Issue #286, opened 2021-12-10
(tracer): Lambda -> SQS -> Lambda support


flochaz:
## Description of the feature request

**Problem statement**

SQS currently don't propagate properly traces when chaining Lambda. See https://github.com/aws/aws-xray-sdk-node/issues/208 for more details.

**Summary of the feature**

1. Implement a mechanism to detect when event coming from SQS and contains XRay trace id.
2. Rebuild segments to be part of the main trace

**Code examples**

I did a first implem here : https://github.com/flochaz/x-ray-lambda-sqs-lambda/blob/main/lib/lambda/consumer-function/index.js

**Benefits for you and the wider AWS community**

The issue on xray sdk has been open for 2 years, people are getting really frustrated, 84 👍  and tens of comments ....

**Describe alternatives you've considered**

Wait for service team to fix it :)

dreamorosi:
Closing this as it currently won't be fixed.

We'll reopen the issue in case things change on the service side.

PR #364, opened 2021-12-29
feat(tracer): Add capability to continue traces comming from SQS triggers

flochaz:
## 🚧 WIP 🚧  :
* Missing proper unit and e2e tests assertions
* Need to evaluate the API design : 
  * is continueSQSRecordTrace and the way to invoke it good enough ? 
  * Should we store toCloseSegments on our side and provide a function to close them ?
* can we really can't do better in term of implementation ??? 
  * might worth to contact a xray sdk node developer 
  * check in deeper details what is done here https://github.com/Sam-Martin/aws-xray-lambda-segment-shim

This PR is a first draft trying to give a solution to the Xray node sdk issue https://github.com/aws/aws-xray-sdk-node/issues/208 where traces should be continued when lambdaare triggered from SQS: `Producer -> SQS -> Consumer Lambda -> ...` (Today it creates 2 traces: `Producer -> SQS` and `Lambda -> ...`).

Here are the findings while implementing this:

* aws-xray-sdk-node has a specific behavior for lambda based on `process.env.LAMBDA_TASK_ROOT` (https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/aws-xray.js#L381)
* Even if active tracing is disabled on the consumer lambda a trace is created for the Consumer Lambda
* If XRay permission is not granted to consumer then no traces are created.
* We can force env var _X_AMZN_TRACE_ID but 
  * need to be set before import of the xray-sdk-core
  * it does not prevent the extra trace to be created with a different traceID
  * It does not help with the multi message use case

### How to verify this change

```
cd packages/tracing
export DISABLE_TEARDOWN=true
npx jest --group=e2e/tracer/sqs
```

Check Xray for traces
Screenshot 2021-12-29 at 10 18 44

Or Cloudwatch Service Map
Screenshot 2021-12-29 at 10 20 07
Screenshot 2021-12-29 at 10 20 24

### Related issues, RFCs

#286

### PR status

## Checklist

- [ ] [My changes meet the tenets criteria](https://awslabs.github.io/aws-lambda-powertools-python/#tenets)
- [ ] The *code coverage* hasn't decreased
- [ ] The PR title follows the [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/semantic.yml#L2)

### Breaking change checklist

---


packages/tracing/tests/unit/ProviderService.test.ts
```diff
@@ -18,229 +33,296 @@ jest.mock('aws-xray-sdk-core', () => ({
...
+          SenderId: 'AROAT26JIZQWSOCCOUNE5:sqsProducer',
+          ApproximateFirstReceiveTimestamp: '1640763398127',
+        },
+        messageAttributes: {},
+        md5OfBody: 'bbdc5fdb8be7251f5c910905db994bab',
+        eventSource: 'aws:sqs',
+        eventSourceARN:
+          'arn:aws:sqs:eu-west-1:123456789012:queue',
+        awsRegion: 'eu-west-1',
+      };
```
simonireilly, line 163: ref

flochaz:
Waiting for X-Ray Service team strategy. Closing for now ...

Issue #419, opened 2022-01-06
Feature request: reduce npm_modules size 


saragerion:
## Bug description
The current bundle size of the Logger library is unnecessarily big. This has an impact on the performance of the Lambda functions (cold starts included).

### Expected Behavior

The node_modules folder has a reasonable size.

### Current Behavior

The node_modules folder is 139M.

### Possible Solution

One of the biggest reasons is the lodash library that is currently a prod dependency, which is unnecessary.
It would also be good to double check whether we publish the NPM packages installing only the prod dependencies and not the dev dependencies.

### Steps to Reproduce

Install the library in your local env, and check the node_modules folder size:

```
du -sh *
139M	node_modules
436K	package-lock.json
4.0K	package.json
```

### Environment

* **Powertools version used**: latest
* **Packaging format (Layers, npm)**: NPM
* **AWS Lambda function runtime:** all
* **Debugging logs:** N/A

### Related issues, RFCs

N/A

saragerion:
Thanks @michaelbrewer for the heads-up!

michaelbrewer:
of coz we should see what size the packaged lambda is and how small we can make it.

dreamorosi:
Good point, we initially removed the whole `lodash` in favor of smaller packages [as mentioned here](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/139#issuecomment-888284501), but somehow it got reintroduced, I think around the time when we had multiple `eslint` configs.

michaelbrewer:
@saragerion there is also a lot of test resources in "commons" that we already have in shared. About 208k worth.

michaelbrewer:
I know this issue closed, but i assume for the production release there is plans to see how to minimize the size of the node_modules?  Like how the aws-sdk-v3 is very lightweight: https://aws.amazon.com/blogs/developer/how-we-halved-the-publish-size-of-modular-aws-sdk-for-javascript-clients/

And for non-typescript users? should @types/aws-lambda always be included?

And maybe tips for the lambda layer:
- https://github.com/ModClean/modclean
- https://github.com/tj/node-prune
- https://tsh.io/blog/reduce-node-modules-for-better-performance/

PR #427, opened 2022-01-06
build(logger): reduce node_modules size

dreamorosi:

As detailed in #419 the size of `node_modules` for logger was unnecessarily big, this PR removes the full `lodash` lib in favor of targeted `lodash.x` depenencies brigning down the `node_modules` size to `1.8M` from `139M`.

Detailed list of changes:
- Removed `lodash` & `@types/lodash`
- Added `lodash.clonedeep`, `lodash.merge`, and `lodash.pickby` & respective `@types/lodash.xxx`
- Updated imports to use new deps
- Updated calls for `merge` in main `Logger` class since it was using the `lodash/fp` version [that swaps the argument order](https://www.measurethat.net/Benchmarks/Show/7660/0/lodash-merge-vs-deepmerge-vs-lodashfp-merge).
- Swapped `npm-shrinkwrap.json` for `packages-lock.json` since it was the last utility to still use it.

### How to verify this change

1. Checkout branch & run unit tests in `logger` to check that there are no regressions.
2. Run `npm pack`
3. Install the packed version in a clean project 
4. Run `du -sh *` to measure `node_moduels` size

(see screenshot below for steps 3 & 4)

Screenshot 2022-01-06 at 20 00 44

### Related issues, RFCs

[#419](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/419)  

### PR status

## Checklist

### Breaking change checklist

N/A

---

### DIFF
--- a/packages/logger/src/Logger.ts
+++ b/packages/logger/src/Logger.ts
@@ -1,7 +1,8 @@
 import type { Context } from 'aws-lambda';
 import { LogFormatterInterface, PowertoolLogFormatter } from './formatter';
 import { LogItem } from './log';
-import { cloneDeep, merge } from 'lodash/fp';
+import cloneDeep from 'lodash.clonedeep';
+import merge from 'lodash.merge';
 import { ConfigServiceInterface, EnvironmentVariablesService } from './config';
 import type {
   Environment,
@@ -69,7 +70,7 @@ class Logger implements ClassThatLogs {
   }
 
   public addPersistentLogAttributes(attributes?: LogAttributes): void {
-    this.persistentLogAttributes = merge(this.getPersistentLogAttributes(), attributes);
+    this.persistentLogAttributes = merge(attributes, this.getPersistentLogAttributes());
   }
 
   public appendKeys(attributes?: LogAttributes): void {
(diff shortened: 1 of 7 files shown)


saragerion (dismissed):
From 139M to 1.8M! Can't wait to have this one released 🚀

flochaz:
well done!!

michaelbrewer:
thanks @dreamorosi

packages/logger/src/Logger.ts
```diff
@@ -148,16 +149,16 @@ class Logger implements ClassThatLogs {
 
   private addToPowertoolLogData(...attributesArray: Array<Partial<PowertoolLogData>>): void {
     attributesArray.forEach((attributes: Partial<PowertoolLogData>) => {
-      this.powertoolLogData = merge(this.getPowertoolLogData(), attributes);
+      this.powertoolLogData = merge(attributes, this.getPowertoolLogData());
```
ijemmy, line 152: (just curious) I understand that the order of param doesn't matter here. What's the reason to swap?

ijemmy (approved):

packages/logger/src/Logger.ts
dreamorosi: See original PR comment:
> Updated calls for merge in main Logger class since it was using the lodash/fp version [that swaps the argument order](https://www.measurethat.net/Benchmarks/Show/7660/0/lodash-merge-vs-deepmerge-vs-lodashfp-merge).

But the TL;DR; is:
- Original version was using `import { merge } from 'lodash/fp';` this version of `lodash` [behaves differently than regular `lodash`](https://dev.to/ifarmgolems/what-is-lodash-fp-even-4ohd).
- As such, using the `import merge from `lodash.merge`;` (that we are using in this PR and is the _non /fp_ version) without changing the args order would make unit tests fail (because the logs objects were merged still recursively but in different order, so with different end result).
- After some digging I found that these two implementations are similar but accept params in different order ([see original link](https://www.measurethat.net/Benchmarks/Show/7660/0/lodash-merge-vs-deepmerge-vs-lodashfp-merge)).
- After swapping the order unit tests were passing again.

With that said, I really never use `lodash` so if anyone spots any issue or I have misunderstood how they behave I'll be more than willing to revert.

ijemmy: Thanks, thi

> With that said, I really never use lodash so if anyone spots any issue or I have misunderstood how they behave I'll be more than willing to revert. 

That makes two of us. I only used it to learn FP many years ago.  
Don't worry. We can pin this on Sara if this breaks :D

dreamorosi:
Need to merge #440 before this can be merged, checks fail because of same linting issue.

dreamorosi:
Ready to be reviewed again, all checks passing now.

flochaz (approved):

Issue #435, opened 2022-01-06
Docs: CDK examples (metrics & tracer) have outdated references


mhboon:
## Bug description

In example-function.MyFunction and example-function.MyFunctionWithDecorator
line 9: const metrics = new Metrics({ namespace: namespace, service: serviceName });
service should be serviceName

In example-function.MyFunctionWithDecorator
typo in line 14 @tracer.captureLambdaHanlder() should be @tracer.captureLambdaHandler()

### Expected Behavior
Try out the examples

### Current Behavior
problems in IDE

### Possible Solution
correct the lines; see summary

### Steps to Reproduce
See summary

1.
2.
3.
4.

### Environment

* **Powertools version used**: 0.02
* **Packaging format (Layers, npm)**:
* **AWS Lambda function runtime:** N/A
* **Debugging logs:**

saragerion:
Thanks for opening the issue, well spotted.

dreamorosi:
@mhboon thanks a lot, we are planning on running tests on our CI for these examples to try to avoid this type of errors in the future, see also #425.

Issue #425, opened 2022-01-06
Maintenance: Write and run tests against examples as part of the CI process


flochaz:
## Description of the feature request

**Problem statement**
Examples are for now only ran manually through manual CDK deploy and manual deployed resource check.

**Summary of the feature**
1. Add E2E tests (following what is done in each packages e2e tests to automate cdk stack deployment).
2. Add a step in CI to run them as part of merge to main.

### Related issues, RFCs

Avoid issue like #423 and #424 

Blocked by: https://github.com/aws/aws-cdk/issues/18211

dreamorosi:
If we do this we'll need to downgrade the examples to cdk version `1.x` right?

flochaz:
yep probably :( that is the reason why I didn't tackled this yet ...

flochaz:
[Current workflow](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/workflows/run-e2e-tests.yml#L48) is giving enough confidence by relying on typescript build to test example don't get broken by new release pkg.

PR #416, opened 2022-01-05
refactor(metrics): rename option property from raiseOnEmptyMetrics to throwOnEmptyMetrics

kozub:
### How to verify this change

Make sure that all existance of "raiseOnEmptyMetrics" were renamed. 

### Related issues, RFCs

[#348](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/348)

### PR status

***Is this ready for review?:*** YES  
***Is it a breaking change?:*** NO

## Checklist

- [x] [My changes meet the tenets criteria](https://awslabs.github.io/aws-lambda-powertools-python/#tenets)
- [x] I have performed a *self-review* of my own code
- [x] I have *commented* my code where necessary, particularly in areas that should be flagged with a TODO, or hard-to-understand areas
- [x] I have made corresponding changes to the *documentation*
- [x] My changes generate *no new warnings*
- [x] The *code coverage* hasn't decreased
- [ ] I have *added tests* that prove my change is effective and works
- [x] New and existing *unit tests pass* locally and in Github Actions
- [x] Any *dependent changes have been merged and published* in downstream module
- [x] The PR title follows the [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/semantic.yml#L2)

### Breaking change checklist

- [ ] I have documented the migration process
- [ ] I have added, implemented necessary warnings (if it can live side by side)

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #439, opened 2022-01-06
fix: updated CDK examples to remove old references & improve comments

dreamorosi:

This PR sets up #425 and aims at:
- Updating references to old parameters/implementations (i.e. #435)
- Improving consistency with examples in docs
- Improving comments in Tracer examples
- Adding full kitchen sink examples that use all three utilities for manual, middy middleware, class decorator based functions

### How to verify this change

~After #416 has been merged,~ checkout the branch, go to `examples/cdk` & run `cdk deploy`.

### Related issues, RFCs

[#435](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/435)  
[#425](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/425)

### PR status

## Checklist

### Breaking change checklist

N/A

---


examples/cdk/lib/example-function.MyFunction.ts
```diff
@@ -38,24 +43,30 @@ export const handler = async (_event: unknown, context: Context): Promise<void>
...
 
+  let res;
   try {
-    throw new Error('test');
-    // Add the response as metadata
+    res = { foo: 'bar' };
+    tracer.addResponseAsMetadata(res, process.env._HANDLER);
   } catch (err) {
     // Add the error as metadata
     subsegment.addError(err as Error, false);
```
ijemmy, line 63: Why do we need try/catch here if there is no throw?

examples/cdk/lib/example-function.MyFunctionWithDecorator.ts
```diff
@@ -36,28 +38,34 @@ export class MyFunctionWithDecorator {
...
     tracer.setSegment(subsegment);
-    // TODO: Add the ColdStart annotation !!! NOT POSSIBLE
-    // tracer.putAnnotation('ColdStart', tracer);
 
+    let res;
     try {
-      throw new Error('test');
-      // Add the response as metadata
+      res = { foo: 'bar' };
     } catch (err) {
```
ijemmy, line 56: Same as above. Why try/catch without throw?

examples/cdk/lib/example-function.MyFunctionWithMiddy.ts
```diff
@@ -1,43 +1,73 @@
...
-    metrics.addMetric('test-metric', MetricUnits.Count, 10);
-    if (_event.throw) {
-      throw new Error('Test error');
-    }
-  }
-}
+  logger.addPersistentLogAttributes({
+    testKey: 'testValue',
+  });
+  logger.debug('This is an DEBUG log'); // Won't show by default
```
ijemmy, line 22: (nit)
```suggestion
  logger.debug('This is an DEBUG log'); // Won't show because we pass logLevel: 'INFO' in the constructor.
```

examples/cdk/lib/example-function.MyFunctionWithMiddy.ts
```diff
@@ -1,43 +1,73 @@
...
-handler.after(async (_request) => {
-  // Won't be flushed since happens after 
-  metrics.addMetric('afterHandlerCalled', MetricUnits.Count, 1);
+  // These traces will be added to the main segment (## index.handler)
+  tracer.putAnnotation('awsRequestId', context.awsRequestId);
+  tracer.putMetadata('eventPayload', event);
 
-});
+  // Create another subsegment & set it as active
+  const handlerSegment = tracer.getSegment(); // This is the custom segment created by Tracer for you (## index.handler)
```
ijemmy, line 45: (nit) In `MyFunction.ts`, the comment is :"// This is the facade segment (the one that is created by AWS Lambda)"

examples/cdk/lib/example-function.Tracer.CaptureErrorDisabled.ts
```diff
@@ -3,11 +3,11 @@ import { Context } from 'aws-lambda';
 import { Events } from '@aws-lambda-powertools/commons';
 import { captureLambdaHandler, Tracer } from '@aws-lambda-powertools/tracer';
 
-// Set environment variable to disable capture response
+// Set environment variable to disable capture response - https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html
 process.env.POWERTOOLS_TRACER_ERROR_RESPONSE = 'false';
 const tracer = new Tracer({ serviceName: 'tracerCaptureErrorDisabledFn' });
 
-// In this example we are using the middleware pattern but you could use also the captureLambdaHandler decorator
+// In this example we are using the middy middleware pattern but you can instrument your functions also with the captureLambdaHandler decorator & manual instrumentation
```
ijemmy, line 10: (nit)
```suggestion
// In this example we are using the Middy middleware pattern but you can instrument your functions also with the captureLambdaHandler decorator & manual instrumentation
```

examples/cdk/lib/example-function.Tracer.PatchAWSSDKv2.ts
```diff
@@ -3,11 +3,15 @@ import { Events } from '@aws-lambda-powertools/commons';
...
-// process.env.POWERTOOLS_SERVICE_NAME = 'tracerManualFn'; // Alternative to setting the service name in the constructor
 const tracer = new Tracer({ serviceName: 'tracerPatchAWSSDKv2Fn' });
-// To patch a specific AWS SDK Client, we pass it to the Tracer that will return an instrumented version of it
+// Alternatively, you can also set the service name using the POWERTOOLS_SERVICE_NAME environment variable
+// Learn more at: https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html
+
+// To patch a specific AWS SDK v2 Client, we pass it to the Tracer that will return an instrumented version of it
 const sts = tracer.captureAWSClient(new STS());
 
+// Here we are showing an example with manual instrumentation but you can do the same also with the captureLambdaHandler Middy Middleware and Class decorator
```
ijemmy, line 13: (minor) I am not aware of this. O.o 

AFAI see, this is not mentioned in the documentation. Can we add it to the list in [this section](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/tracer/#lambda-handler) . Another PR is fine.

ijemmy (changes requested):
The only major one is that we have try/catch without throwing an error.  I am willing approve if that is addressed.

The rest are either minor/nit.

examples/cdk/lib/example-function.MyFunction.ts
dreamorosi: I know in this specific example there's no way that an exception will be thrown in the `try` block, I still added the whole `try/catch/finally` to give an example of the whole function and also to demonstrate how customers can add annotations in the `catch` & close the subsegments in the `finally`.
To sum up, I was intending these examples as a way for customers to copy/paste them and get started writing their own function.

examples/cdk/lib/example-function.MyFunctionWithDecorator.ts
dreamorosi: See [this comment](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/439#discussion_r780100117).

examples/cdk/lib/example-function.MyFunctionWithMiddy.ts
dreamorosi: Yep, that's right. `MyFunction.ts` is an example of manual instrumentation and as such customers need to retrieve the `facade` segment (aka the one created by AWS Lambda) before they can create the one called `## index.handler`, this is needed because customers can't annotate the main Lambda segment.

image
(screenshot above comes [from here](https://docs.aws.amazon.com/lambda/latest/dg/services-xray.html))

In this file instead, since we are using a Middy middleware the retrieval of the `facade` segment + the creation of the custom (`## index.hanlder`) one is done for you, so the first time you run `tracer.getSegment` you're actually getting that one, and not the `facade` one; hence the var name & comment difference.

examples/cdk/lib/example-function.Tracer.PatchAWSSDKv2.ts
dreamorosi: The actual example, in this file, is from L4 to L11 since this example wants to demonstrate how to instrument an AWS SDK v2 client. 

What I wanted to say with this comment is that even though this example uses manual instrumentation, _you don't need_ to use this method specifically over others (Middy middleware, decorator). Essentially, as long as you instrument the AWS SDK client with `tracer.capture*` it'll be traced & appear as custom subsegment on X-Ray no matter how you write your function. 

That's why [in the docs](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/tracer/#patching-aws-sdk-clients) the AWS SDK patching example only has these lines.

I included a function in the example just because otherwise the lines above alone wouldn't run as a Lambda function and the lines from L15 onwards are similar to manual method in the section you linked.

examples/cdk/lib/example-function.MyFunction.ts
dreamorosi: I just re-read the comment, which I misunderstood, sorry, you're right. Adding it!

examples/cdk/lib/example-function.MyFunctionWithDecorator.ts
dreamorosi: This one has the throw at [L57](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/4ac474d6a5e76f00bff4c5d3c597d757c724b113/examples/cdk/lib/example-function.MyFunctionWithDecorator.ts#L57).

examples/cdk/lib/example-function.MyFunction.ts
dreamorosi: Added in [9c2a418](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/439/commits/9c2a4189f9cfc22a24b10e10d68ba2b4108d2117)

michaelbrewer:
As CDK V2 is GA, would it make sense to just start using it from the get go @dreamorosi ?

dreamorosi:
> As CDK V2 is GA, would it make sense to just start using it from the get go @dreamorosi ?

Would love to, but at the moment we are relying on some classes that are not exported in v2, see [this issue I've opened on the CDK repo](https://github.com/aws/aws-cdk/issues/18211).

flochaz (approved):

ijemmy (approved):

examples/cdk/lib/example-function.MyFunction.ts
dreamorosi: Discussed this offline and comments have been addressed.

PR #451, opened 2022-01-10
test(logger): added unit test for URIError

JavierMendozaGomez:

Added unit test for the URIError type to increase the coverage of the log error formatter.

### How to verify this change

Run 
` npm test packages/logger/tests/unit/formatter/PowertoolLogFormatter.test.ts `

### Related issues, RFCs

No RFC's were created as this is not a significant work.

### PR status

## Checklist

- [x] have *added tests* that prove my change is effective and works

### Breaking change checklist

---


package-lock.json
```diff
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "aws-lambda-powertools-typescript",
```
dreamorosi, line 8: Could you please revert this? Since the change is in the unit tests it shouldn't affect the lock file.

dreamorosi (changes requested):

dreamorosi:
Hello @JavierMendozaGomez thanks for opening the PR. 

Just to make sure we are all on the same page I have a couple of questions:
* Was the coverage below 100%?
* Could you elaborate a bit on what kind of case this test is trying to cover?
* Why this type of error deserves special treatment?

Thank you in advance!

package-lock.json
JavierMendozaGomez: Hello @dreamorosi  yes, for sure. This was automatically added when I execute npm install , I'm using the npm version 7.24.2.  
Previously I was using the version 6 of npm and it broke completely the npm lock as you can see in the attached screenshot. It would be good to add the check of having at least npm v7  in the contributors section. 
Screenshot 2022-01-10 at 6 16 41 PM

JavierMendozaGomez:
Hello @dreamorosi . Sorry for my confused description of the PR. 
I was going through your unit test and I saw that you were making unit test to check the format of the error for each one of the possible errors. You had the ReferenceError, AssertionError, RangeError, SyntaxError, TypeError and I realised that you were missing the URIError on your tests.  
The jest coverage  was at 100 per cent before the addition. What I mean was increase in the coverage of the format of the log of the errors adding the URIError.

package-lock.json
dreamorosi: Thanks, yes we are working on standardising the guidelines about which node/npm versions are recommended so that this can be avoided. In the meantime thanks for reverting it.

dreamorosi (approved):

flochaz (approved):

Issue #494, opened 2022-01-21
Maintenance: release process not updating package-lock.json for examples after bumping deps


flochaz:
## Bug description

package.json not matching package-lock.json after bump

### Expected Behavior

lerna version should bump package.json and package-lock.json

### Current Behavior

Build failing https://github.com/awslabs/aws-lambda-powertools-typescript/runs/4895012434?check_suite_focus=true

### Steps to Reproduce

1.
2.
3.
4.

### Environment

* **Powertools version used**:
* **Packaging format (Layers, npm)**:
* **AWS Lambda function runtime:**
* **Debugging logs:**

dreamorosi:
@flochaz is this still happening? Just wanted to make sure the issue is not stale.

PR #498, opened 2022-01-21
chore(examples): fix version bumping of examples

flochaz:

Currently when a version are bumped, examples cdk package got his dependencies bumped as well to match latest package that are going to be released.
1. This can't be control due to this https://github.com/lerna/lerna/issues/2326 bug.
2. Bumping dependencies can make example code buggy since it won't be updated to go around potential breaking changes ...

This PR leverage [versionning life cycle](https://github.com/lerna/lerna/tree/6cb8ab2d4af7ce25c812e8fb05cd04650105705f/commands/version#lifecycle-scripts) to force versioning of deps and workaround https://github.com/lerna/lerna/issues/2326 .

NB: Examples will always be behind if we don't update manually the dependencies.

### Alternative

Remove `examples/cdk` from lerna.json and add lint, build and test specific scripts to either root `package.json` or in github actions and husky scripts ...

### How to verify this change

1. Bump version `npx lerna version --force-publish=\* --no-git-tag-version --no-push  --conventional-commits`
2. Check that examples package.json and lock for no change on deps

Build is passing https://github.com/awslabs/aws-lambda-powertools-typescript/runs/4898015010 so package-lock.json is fixed.

Next bump then run e2e tests will demonstrate the fix

### Related issues, RFCs

#494

### PR status

## Checklist

---

### DIFF
--- a/examples/cdk/package.json
+++ b/examples/cdk/package.json
@@ -11,7 +11,7 @@
     "package": "echo 'Not applicable'",
     "test:unit": "npm run build && jest",
     "test:e2e": "echo 'To be implemented ...'",
-    "version": "npm install @aws-lambda-powertools/commons@0.4.0 @aws-lambda-powertools/logger@0.4.0 @aws-lambda-powertools/tracer@0.4.0 @aws-lambda-powertools/metrics@0.4.0",
+    "version": "npm install @aws-lambda-powertools/commons@0.4.0 @aws-lambda-powertools/logger@0.4.0 @aws-lambda-powertools/tracer@0.4.0 @aws-lambda-powertools/metrics@0.4.0 && git add package.json",
     "cdk": "cdk"
   },
   "devDependencies": {
--- a/examples/cdk/package.json
+++ b/examples/cdk/package.json
@@ -11,6 +11,7 @@
     "package": "echo 'Not applicable'",
     "test:unit": "npm run build && jest",
     "test:e2e": "echo 'To be implemented ...'",
+    "version": "npm install @aws-lambda-powertools/commons@0.4.0 @aws-lambda-powertools/logger@0.4.0 @aws-lambda-powertools/tracer@0.4.0 @aws-lambda-powertools/metrics@0.4.0",
     "cdk": "cdk"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


examples/cdk/package.json
```diff
@@ -11,6 +11,7 @@
     "package": "echo 'Not applicable'",
     "test:unit": "npm run build && jest",
     "test:e2e": "echo 'To be implemented ...'",
+    "version": "npm install @aws-lambda-powertools/commons@0.4.0 @aws-lambda-powertools/logger@0.4.0 @aws-lambda-powertools/tracer@0.4.0 @aws-lambda-powertools/metrics@0.4.0",
```
dreamorosi, line 14: Why are we setting a specific version? And who is going to call `npm run version`?

flochaz: Examples are written against a certain version of our package. By definition its a version that has already been pushed to NPM.

flochaz: lerna version is calling version

dreamorosi (approved):

ijemmy (commented):
@flochaz  

Feel really awkward about this. But I also can't think of a better fix and want to put my thoughts before approval.

## `^0.4.0` and `0.4.0`
My main concern is that we now have `^0.4.0` (with caret) and will be override with `0.4.0` only when running `lerna`. The effect are:

1. If we release `0.4.1`, this `version` will be reverted back to 0.4.0 in `lerna` run. Any patching we did won't be applied here. We need to remember to update this every time we bump a patch version.
1. Most contributors will not know that `version` target is run automatically by `lerna`. 

## Should we keeping example up-to-date through checklist?
Users who try this example with `npm install` will use the latest patched version. This is good. However, users may copy the code from example into their project with the latest version of library.  
   * If the latest is just patching (0.4.1) or  minor (0.5.0), it should be fine as it is backward compatible.
   * If the latest is major (1.0.x), the example can break.

To avoid this, we have to either add an updating example as part of Definition of Done checklist. This is manual now.
   
## The alternative (leaving it as it is) is worse option
Alternatively, we  can leave it as it is. We risk having broken examples for users.  The users may or may not report back to us. So we might have broken examples for a very long time before any detection. 

Choosing between two choice, I prefer to have this change in than leaving it as it is.

ijemmy (approved):
This is blocking merge. Let's go with this for now. Will discuss with the team on adding a checklist in PR so that we don't miss updating this.

Issue #518, opened 2022-01-27
Bug: cold start flag is set to true in subsequent invocations in Logger


ijemmy:
## Bug description

When `logger` [capture context information](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/logger/#capturing-lambda-context-info), it adds `cold_start` key with boolean value into each log.

I call the same Lambda twice. The `cold_start` value should be `false` after the first call. However, it is still `true` in other calls.

![image](https://user-images.githubusercontent.com/61999/151420677-d8d6d094-78aa-4599-95a6-046a38deb76b.png)

### Expected Behavior
`cold_start` value should be `false` in subsequent calls.

### Current Behavior
`cold_start` value is still `true` in subsequent calls.

### Steps to Reproduce
1. This is the Lambda function code I use.

```typescript
import { injectLambdaContext, Logger } from "../../src";
import { APIGatewayProxyEvent } from "aws-lambda";
import middy from "@middy/core"

const persistentKey = process.env.PERSISTENT_KEY_IN_LOG;
const persistentValue = process.env.PERSISTENT_VALUE_IN_LOG;

const logger = new Logger({
  persistentLogAttributes: {
    [persistentKey]: persistentValue,
  }
});

const testFunction = async (event: APIGatewayProxyEvent) => {

  // 1. Context data appears in log
  // 2. LOG_LEVEL works as expected
  // 3. persistentLogAttributes prop works 
  logger.debug("##### This should not appear");
  logger.info("This is an INFO log with some context and persistent key");

  // 4. Can add additional key value in the log method
  logger.info("This is an INFO log with some context", {
    additionalKey: 'additionalValue'
  });

  // 5. Can log error object
  try {
    throw new Error('you cannot prevent this');
  }catch(e){
    logger.error("There was an error", e);
  }

  return formatJSONResponse({
    message: `E2E testing Lambda function`,
    event,
  });
}

const formatJSONResponse = (response: Record<string, any>) => {
  return {
    statusCode: 200,
    body: JSON.stringify(response)
  }
}

export const handler = middy(testFunction)
  .use(injectLambdaContext(logger));
```

2. I deploy and trigger the Lambda function twice with this code
```typescript

    const sdkProvider = await SdkProvider.withAwsCliCompatibleDefaults({
      profile: process.env.AWS_PROFILE,
    });
    const cloudFormation = new CloudFormationDeployments({ sdkProvider });

    // WHEN lambda function is deployed
[...]

PR #535, opened 2022-02-06
fix(logger): fix cold start heuristic and logger attribute merging

okoskine:

- logger attribute (context & custom persisted attribute) merges obey the expected priority, i.e. newly added values overwrite older ones
- logger attribute merges no longer mutate the user-provided attribute objects
- cold start heuristic is now based on the request id

### How to verify this change
Newly added unit tests in Logger.test.ts

### Related issues, RFCs

[#518](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/518)  

### PR status

## Checklist

### Breaking change checklist

---


okoskine:
Hi, I edited the PR not to include the cold start changes. I initially included these in a single PR since the `coldStart` value was also affected by the merge problems described here. I think the introduced unit tests now more easily show the intention as there are fewer changes. However, here's a quick explanation:

The [merge](https://lodash.com/docs/4.17.15#merge)  function from lodash mutates the first parameter, so a call of the form

`activeValues = merge(newValues, activeValues)`

actually mutates the `newValues` object by adding the old `activeValues` contents and then binds that object also to the name `activeValues`.

The apparent intention of these usages is captured with the form

`merge(activeValues, newValues)`

which only mutates the `activeValues` object and overrides the old values with the new ones instead of the other way round.

The actual problem I encountered was that `awsRequestId` was not updated to the correct value on later invocations but stayed constant (the request id of the first (cold start) invocation). The same problem existed for `appendKeys` / `addPersistentLogAttributes`.

dreamorosi:
@okoskine thanks a lot for the explanation and for accepting to exclude the cold start changes.

I'll review the changes & unit tests tomorrow morning and provide a review for the PR.

dreamorosi (approved):

packages/logger/src/Logger.ts
```diff
@@ -174,7 +174,7 @@ class Logger implements ClassThatLogs {
    * @returns {void}
    */
   public addPersistentLogAttributes(attributes?: LogAttributes): void {
-    this.persistentLogAttributes = merge(attributes, this.getPersistentLogAttributes());
+    merge(this.persistentLogAttributes, attributes);
```
saragerion, line 177: Well spotted!

saragerion (approved):

dreamorosi (approved):

Issue #484, opened 2022-01-18
Maintenance: move redundant code and functionalities in shared commons package


saragerion:
## Description of the feature request
The execution of this issue should be divided into 3 steps:

1. Vet the business logic of Logger, Tracer and Metrics utilities to identify code de-duplication opportunities. Opportunities found so far: cold-start business logic, fetching of environment variables. 
2. Add that logic in commons.
3. Remove duplicated logic from the single utilities, in favour of the logic from the commons package.

**Problem statement**

Right now there is a number of functionalities that are repeated throughout the 3 different utilities, all with their own business logic and separated unit tests. 
This is not optimal because:
- The same feature should have the same behaviour across all utilities
- This unnecessarily increases the cost of maintenance
- Code duplication inflates the libraries's size unnecessarily, with impact over the Lambda package size and consequently cold-starts 

**Summary of the feature**

See above.

**Code examples**

**Benefits for you and the wider AWS community**

**Describe alternatives you've considered**

**Additional context**

See here:
https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/logger/src/config/EnvironmentVariablesService.ts
https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/logger/src/Logger.ts#L92

### Related issues, RFCs

#165

N/A

dreamorosi:
An internal RFC for centralising the cold start logic in the `commons` package has been created & reviewed by the team. Will be working on a PR in the coming days.

dreamorosi:
First batch of changes regarding the cold start heuristic have been implemented in `commons` and released in `v0.6.0`, adoption in other utilities will be released with `v0.7.0` in the coming days.

ijemmy:
We'll focus on moving the EnvironmentVariable helper class and close this ticket for Production Release.

saragerion:
Update about this issue: we addressed successfully the cold start logic separation. 
We will think about better separations of concerns and abstractions when we will address this issue after GA:
https://github.com/awslabs/aws-lambda-powertools-typescript/issues/846
If necessary we will open another more scoped issue when we will address this unit of work.
Closing this one.

dreamorosi:
This issue was a dependency of #165.

For the time being I have added #165 in the issue text of #846 as it has a fair bit of context & research on how the environment variables are being used. 

[...]

PR #605, opened 2022-02-28
build(dev-deps): consolidate all common dev deps in workspaces

dreamorosi:

In continuing the effort to reduce redundant code in the repo, and after the consolidation of local & CI environments around `npm@8.x` and subsequent full adoption of `workspaces`; this PR _pulls-up_ **all the shared dev dependencies** from the package-specific `package.json` files to the one in the root.

With the previous PRs the dependencies were being already installed in the main `node_modules` directory so this change only affects their placement. This will also help in the long term to keep more consistent versioning across shared `devDependencies` and also reduce Dependabot's diff sizes.

### How to verify this change

See successful checks under this PR and the [e2e successful execution](https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/1911460449) for this branch.

### Related issues, RFCs

[#484](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/484)

### PR status

## Checklist

### Breaking change checklist

N/A

---


flochaz (dismissed):

packages/commons/package.json
```diff
@@ -59,4 +44,4 @@
   "dependencies": {
     "@types/aws-lambda": "^8.10.72"
```
saragerion, line 45: This can be moved as a dev-dependency too right?

packages/tracing/package.json
```diff
@@ -30,29 +30,8 @@
...
-    "eslint": "^8.3.0",
-    "eslint-import-resolver-node": "^0.3.6",
-    "eslint-import-resolver-typescript": "^2.5.0",
-    "eslint-plugin-import": "^2.25.3",
-    "jest": "^27.5.1",
-    "jest-runner-groups": "^2.1.0",
-    "ts-jest": "^27.1.3",
-    "ts-node": "^10.0.0",
-    "typescript": "^4.1.3"
+    "@middy/core": "^2.5.6"
```
saragerion, line 34: Question: can this be moved in the root package.json? same as for metrics and logger

dreamorosi: Yes it can, in this case I deliberately decided to leave it because it's the only one that historically was a runtime dependency, but on second thought it might belong with the others as I find it unlikely that customers who install the package will dig into their `node_modules` to check which version to install.
I'll move it, in the future we should consider making explicit which version we support in the docs maybe, what do you think?

packages/commons/package.json
dreamorosi: I need to check this one, I assumed it had to stay but maybe it can be removed. I'll test and get back to you

packages/tracing/package.json
saragerion: I agree

packages/commons/package.json
dreamorosi: Ok I've tested this and you were right, I removed it from this file and added it to the main `package.json` as `devDependency` in the last commit.

saragerion: That's great to hear, and it means the library's prod bundle decreased :)

packages/commons/package.json
```diff
@@ -56,7 +41,5 @@
   "bugs": {
     "url": "https://github.com/awslabs/aws-lambda-powertools-typescript/issues"
   },
-  "dependencies": {
-    "@types/aws-lambda": "^8.10.72"
-  }
-}
+  "dependencies": {}
```
saragerion, line 44: Nice!

saragerion (approved):
Looks good! Nicely done 🥳

ijemmy (approved):
LGTM

Issue #644, opened 2022-03-08
Maintenance: avoid fixed-time waiting in our E2E testings


ijemmy:
## Bug description
Currently, our E2E tests are very flaky (fail about 40-50% of the time) due to fixed waiting time. This happens in both metrics and tracers E2E tests.

For example, `tracer.test.ts` waits for 2 minutes before finishing the `beforeAll()` function. 

![image](https://user-images.githubusercontent.com/61999/157275005-0804bb21-5bf3-4856-b00a-7e6c74c2e5b1.png)

Sometimes, the trace come back later than two minutes. The `test()` blocks will fail. 
 
### Expected Behavior
The tests should not be flaky. 

### Current Behavior
The tests are flaky. If the traces/metrics appear after the fixed wait time. 

### Possible Solution
1. **Implement retry** The tests should never wait for fixed amount of time. Instead, it should poll for trace or metrics and retry until it gets expected number of traces.  This could be implemented with [promise-retry library](https://www.npmjs.com/package/promise-retry). See example code in "Reference" section below

2. **Introduce uuid** However, we cannot simply count the number of traces or metrics because there might be another test running concurrently (e.g. one for Node14, another for Node12) . The traces/metrics found may be from another test running. Thus, we need to introduce a uuid in metadata and filter base on that.  You can pass UUID via environment variable like in [this e2e test in Logger](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/8aab29900c5fac8eb625eb747acbc23ceac8f6ba/packages/logger/tests/e2e/sampleRate.decorator.test.ts#L55) 

If you need refactoring, take note of Logger's e2e tests and try to structure it similar to that. 

### Steps to Reproduce
Run E2E tests a few times. You can use this script:

```sh
repeat 20 {jest --group=e2e/PACKAGE; sleep 0.5}
```

### Environment

* **Powertools version used**:0.7.0
* **Packaging format (Layers, npm)**: N/A
* **AWS Lambda function runtime:** 12 and 14
* **Debugging logs:** N/A

### Related issues, RFCs
N/A

## Reference

Here's an example implementation on fetching log stream with `promise-retry` library. It keep fetching until the log group has the `atLeast` number of log stream. You can adapt this with traces/metrics. (but need to check uuid)

```typescript
export const fetchStreamsUntil = async (logGroupName: string, atLeast:number): Promise<string[]> => {
  const retryOption = {
    retries: 6,
    factor: 1.5,
[...]

PR #654, opened 2022-03-11
fix(tracer, metrics): use polling instead of fixed wait in e2e tests

awsdb:

This PR changes the way the E2E tests for **metrics** and **tracer** perform their checks.
Instead of waiting for a fixed amount of time (and hoping that the expected data will be available after the wait time elapses), the tests will now actively and periodically poll the corresponding services (CloudWatch metrics, X-Ray traces). When the expected number of result items arrives, the tests will perform their validation as usual. In case there will be not enough (or too many) items, the polling will stop after a timeout, so the tests will fail as intended.

The implementation relies on polling using unique IDs as keys. For metrics, the namespaces already represent such unique IDs in form of UUIDs. For traces, instead of introducing additional UUID annotations, it turned out to be easier to use unique Lambda function names for each test run.

This PR adds a new dev dependency: the [promise-retry](https://www.npmjs.com/package/promise-retry) package (and its corresponding type declaration package).

### How to verify this change

Run the E2E tests for **tracer** and **metrics** multiple times:

```bash
cd ./packages/tracing/ && for i in {1..10}; do npm run test:e2e; done && cd ..
cd ./packages/metrics/ && for i in {1..10}; do npm run test:e2e; done && cd ..
```
Explore the test results.

**NOTE:** The tests might fail for a different (legitimate) reason, so a red test result doesn't necessarily mean an issue with this PR. If any test fails, compare the actual stored trace or metric data with the expected values using AWS CLI or the Management Console. If the values differ, the test failure is caused not by the way the tests gather the data, so those failures must be investigated separately.

### Related issues, RFCs

#644

### PR status

## Checklist

### Breaking change checklist

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4781,6 +4781,21 @@
       "integrity": "sha512-ReVR2rLTV1kvtlWFyuot+d1pkpG2Fw/XKE3PDAdj57rbM97ttSp9JZ2UsP+2EHTylra9cUf6JA7tGwW1INzUrA==",
       "dev": true
     },
+    "node_modules/@types/promise-retry": {
+      "version": "1.1.3",
+      "resolved": "https://registry.npmjs.org/@types/promise-retry/-/promise-retry-1.1.3.tgz",
+      "integrity": "sha512-LxIlEpEX6frE3co3vCO2EUJfHIta1IOmhDlcAsR4GMMv9hev1iTI9VwberVGkePJAuLZs5rMucrV8CziCfuJMw==",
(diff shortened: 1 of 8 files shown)


flochaz (approved):
Nice !

dreamorosi (approved):
Hi @awsdb thanks a lot for the PR, just ran the e2e tests a couple of times and the duration of the tracer one went down more than 50%!

dreamorosi (approved):
Great one, thanks a lot.

Issue #497, opened 2022-01-21
Maintenance: make e2e test stack name unique


flochaz:
## Description of the feature request

**Problem statement**
As of today Stack names in e2e tests are static (see https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/metrics/tests/e2e/decorator.test.ts#L23 for instance) which prevent us from running multiple instances of e2e against different branches.

**Summary of the feature**
Add a random unique id, a branch name or a commit id to stack name in all e2e.

**Code examples**
```typescript
const stack = new Stack(integTestApp, `MetricsE2EDecoratorStack-${process.env.GITHUB_REF_NAME ? process.env.GITHUB_REF_NAME : 'unknown'}`);
```

List of available env var in github action: https://docs.github.com/en/actions/learn-github-actions/environment-variables

**Benefits for you and the wider AWS community**
Validate PR quickly

ijemmy:
@flochaz  I'm going to refactor metrics & tracer e2e tests to be like logger (using uuid). I can take over this if you're ok.

Issue #445, opened 2022-01-08
RFC: Testing Factories for AWS Data Objects


simonireilly:
>ref https://github.com/awslabs/aws-lambda-powertools-typescript/issues/26for strucutre of my RFC

## Description of the proposal

There should be a consistent way to produce test fixtures for AWS lambda environments

----

## Name of the core utility (and consequently the folders that contain the libraries):

- `packages/factories`

## Justification

When it comes to making effective use of AWS in unit and integration testing, developers will create data objects which have the interfaces of the 'aws-lambda' types package.

This is true, even within this project;

- Here, for testing the SQS event format is compatible with a function interface: https://github.com/awslabs/aws-lambda-powertools-typescript/pull/364/files#r780652925
- Here, where a collection of untyped JSON is used for testing fixtures: https://github.com/awslabs/aws-lambda-powertools-typescript/tree/main/tests/resources/events/aws

These untyped objects become a cause for churn, and often because a single source of truth is not present, they can misdiagnose behaviors through simple omission of fields.

### Goals

- Provide users of the nodejs lambda platform effective utilities for testing lambda handler
- Provide consistent type-safe factories for AWS data structures and events
- Support version of event formats to test utilities against different versions

## Proposed API

There is an example here that utilizes `faker` and `fishery` to create type compliant factories that can be deep merged to allow for standard testing objects:

- https://github.com/simonireilly/aws-cdk-utils/tree/main/packages/factories

### Installation:

```bash
yarn add --dev @aws-lambda-powertools/factories
npm i -d @aws-lambda-powertools/factories
```

### Usage

Consider the following test that is written in #364 

```ts

test('when called, it set the correct segment', () => {
  // Prepare
  const provider: ProviderService = new ProviderService();
  const record: SQSRecord = {
    messageId: 'fd95260b-1600-4028-b252-590cfcc9fe6d',
    receiptHandle: 'test',
    body: 'Information about current NY Times fiction bestseller for week of 12/11/2016.',
    attributes: {
      ApproximateReceiveCount: '1',
      AWSTraceHeader: 'Root=1-61cc1005-53ff3b575736e3c74eae6bfb;Parent=1f57c53badf96998;Sampled=1',
      SentTimestamp: '1640763398126',
      SenderId: 'AROAT26JIZQWSOCCOUNE5:sqsProducer',
      ApproximateFirstReceiveTimestamp: '1640763398127',
    },
    messageAttributes: {},
[...]

PR #678, opened 2022-03-23
test(common,logger,metrics): run e2e tests for metrics in both Node runtimes

ijemmy:

Changes in this PR.
- refactor metrics e2e tests to make stack and resource name unique (so that they can be run in parallel)
- refactor metrics e2e tests to be consistent with logger and move common code to `commons`
- run e2e tests on both Node12 and Node14 runtimes

### How to verify this change

Run `npm run lerna-test:e2e` at the root.  

Every test should pass and metrics tests should be on both node14 and node12

### Related issues, RFCs
#497 

### PR status

## Checklist

### Breaking change checklist

---


packages/commons/src/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from './utils/lambda';
 export * from './Utility';
+export * from './tests/e2e';
```
dreamorosi, line 3: If we export it here, does it mean the module gets included in the published package? If that's the case maybe we don't want that

packages/commons/jest.config.js
```diff
@@ -19,6 +19,12 @@ module.exports = {
   'testEnvironment': 'node',
   'coveragePathIgnorePatterns': [
     '/node_modules/',
+    /** 
+    * e2eUtils.ts contain helper methods that simplify several calls to CDK and SDK interface.
```
dreamorosi, line 23:
```suggestion
    * e2eUtils.ts contains helper methods that simplify several calls to CDK and SDK interface.
```

packages/commons/src/tests/e2e/e2eUtils.ts
```diff
@@ -1,19 +1,24 @@
+/* istanbul ignore file */
```
dreamorosi, line 1: The reason why we ignore this file from coverage is the one you mention in the comment at `packages/commons/jest.config.js`? 
If I understand it correctly we are saying that it's not worth it to write unit tests for the files because by running the integration tests we are already "testing" it.

The only minor issue I see is that at the moment we are not enforcing e2e tests on merge, but I see your point and agree on the low value of writing the unit tests for this.

packages/commons/src/index.ts
ijemmy: Good point. I just saw Florian's PR that he import directly (e.g. `../../../commons/tests/utils/cdk-cli`) . This makes more sense as the code is internal.

packages/commons/src/tests/e2e/e2eUtils.ts
ijemmy: If forgot to remove this, the ignore in `jest.config.js` is enough.

ijemmy: But yes, that's why I didn't write unit tests for the file.  If I don't ignore it, it'll drop our coverage percentage from 100%.

dreamorosi (commented):
Thanks a lot for the PR, I like the direction you're giving to the integration tests. It'll make them much easier to maintain moving forward.

My only concern relates to exporting the e2e utilities in the `commons package`. This package is depended and bundled by all the other utilities and as such it will be included in the deployment package of each function that uses Powertools. 

While I think there's definitely value in providing test utilities (see #445), I don't know if it should be in `commons` and in general, if it's code that should be deployed in Lambda.

What's your take?

ijemmy:
> Thanks a lot for the PR, I like the direction you're giving to the integration tests. It'll make them much easier to maintain moving forward.
> 
> My only concern relates to exporting the e2e utilities in the `commons package`. This package is depended and bundled by all the other utilities and as such it will be included in the deployment package of each function that uses Powertools.
> 
> While I think there's definitely value in providing test utilities (see #445), I don't know if it should be in `commons` and in general, if it's code that should be deployed in Lambda.
> 
> What's your take?

I totally agree with you. We'll unnecessarily increase package size. I'll change this. 
Found that I modify quite a few files with #676 . I'll wait until that one is merge first, resolve conflict, and change this.

packages/metrics/package.json
```diff
@@ -13,7 +13,9 @@
     "commit": "commit",
     "test": "npm run test:unit",
     "test:unit": "jest --group=unit --detectOpenHandles --coverage --verbose",
-    "test:e2e": "jest --group=e2e",
+    "test:e2e:nodejs12x": "RUNTIME=nodejs12x jest --group=e2e/logger",
```
flochaz, line 16: wrong copy/paste. And I don't think you need to specify the module after e2e since it's a module specific package.json so it won't see otheres e2e tests anyway

packages/metrics/package.json
```diff
@@ -13,7 +13,9 @@
...
+    "test:e2e:nodejs14x": "RUNTIME=nodejs14x jest --group=e2e/logger",
```
flochaz, line 17: same

flochaz (changes requested):

packages/metrics/package.json
ijemmy: Good catch!

packages/metrics/package.json
ijemmy: Good catch!

packages/metrics/package.json
ijemmy: Good catch!

packages/metrics/package.json
ijemmy: Good catch!

flochaz (approved):

dreamorosi (approved):

Issue #672, opened 2022-03-18
Maintenance: ignore available updates for CDK v2 and reduce frequency of new PRs


saragerion:
## Description of the feature request

Context - Dependabot
We use dependabot for version updates:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates

Our current dependabot configuration can be found here:
https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/dependabot.yml
Our team merges or rejects PRs automatically opened by the dependabot around once a week (usually on Fridays). 

We need to update our current dependabot configuration to reduce the number of PRs open at once, especially the ones we don't need. 

**Problem statement**

Problem 1:
As stated in our [Contributing guidelines](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/CONTRIBUTING.md), we use [AWS CDK v1](https://docs.aws.amazon.com/cdk/v1/guide/getting_started.html) library and not v2 due to [this limitation](https://github.com/aws/aws-cdk/issues/18211).

Right now we receive a lot of PRs requests opened by the dependabot, to update CDK dependencies to the v2 versions. This adds a lot of noise and requires the continuous action of closing/ignoring those PRs.

Problem: 2
Since we merge dependabot PRs once a week but the dependabot PRs are created daily, this results in a lot of PRs being opened every day that add a lot of noise and pollution to our PR list.

**Summary of the feature**

[] Reduce the frequency of PRs from daily to weekly (ideally on a thursday)
[] Ignore version updates for dependecies related to CDK v2.

**Code examples**

Github docs: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#specifying-dependencies-and-versions-to-ignore

**Benefits for you and the wider AWS community**

Maintainers more focused and less distracted. 

**Describe alternatives you've considered**

N/A

**Additional context**

N?A

### Related issues, RFCs

https://github.com/aws/aws-cdk/issues/18211

dreamorosi:
Just an additional nuance to complement the CDK topic: we are using CDK v1 in the integration tests of the utilities (i.e. under `packages/logger/tests/e2e/`), this is rolled up into the `package.json` & `package-lock.json` in the root of the project ([here](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/package.json#L50-L53)).

[...]

PR #688, opened 2022-03-24
build(docs): updated docs dependencies

dreamorosi:
## Description of your changes

`on-merge-to-main` action is failing ([#140](https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/2034859221), [#141](https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/2034856950)) due to a Python dependency issue.

Attempting to fix it by updating the dependencies related to the documentation to the latest version, one of which (`mkdocs-material` - which is the main one) had a [release yesterday](https://github.com/squidfunk/mkdocs-material/releases).

### How to verify this change

Docs should get built successfully.

### Related issues, RFCs

N/A

### PR status

***Is this ready for review?:*** YES  
***Is it a breaking change?:*** NO

## Checklist

- [x] [My changes meet the tenets criteria](https://awslabs.github.io/aws-lambda-powertools-python/#tenets)
- [x] I have performed a *self-review* of my own code
- [x] My changes generate *no new warnings*
- [x] The *code coverage* hasn't decreased
- [x] The PR title follows the [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/semantic.yml#L2)

### Breaking change checklist

N/A

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #766, opened 2022-04-14
chore: updated depdendabot configs

dreamorosi:

As discussed in #672 we would like to reduce the noise from Dependabot while still being up to date with the dependency ecosystem. This PR introduces the following changes:
- Set Dependabot to run weekly, on Fridays, at 5AM (`Europe/Amsterdam` tz)
- Set PR limit to 20 **
- Add Dependabot checks for Python dependencies in `docs/requirements.txt`, this was previously ignored.

** I propose to set this number to 20 (default is 5) so to increase the chance that Dependabot will open most/all the PR at once. This will allow the maintainers to have a better idea of the amount of PR to handle on the day (Friday). The current behaviour of 5 at the time made it so that Dependabot would keep opening PRs every time one was merged causing frustration. Read more about this setting [here](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit).

### How to verify this change

TBD

### Related issues, RFCs

[#672](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/672)  

### PR status

## Checklist

### Breaking change checklist

N/A

---

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,14 +1,32 @@
 version: 2
 updates:
 
+  # Maintain dependencies for Python
+  - package-ecosystem: "pip"
+    directory: "/docs"
+    schedule:
+      interval: "weekly"
+      day: "friday"
+      time: "05:00"
+      timezone: "Europe/Amsterdam"
+    open-pull-requests-limit: 20
+
   # Maintain dependencies for GitHub Actions
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
       interval: "weekly"
+      day: "friday"
+      time: "05:00"
+      timezone: "Europe/Amsterdam"
+    open-pull-requests-limit: 20
 
   # Maintain dependencies for npm
   - package-ecosystem: "npm"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
+      day: "friday"
+      time: "05:00"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,14 +1,32 @@
...
 
+  # Maintain dependencies for Python
+  - package-ecosystem: "pip"
+    directory: "/docs"
+    schedule:
+      interval: "weekly"
+      day: "friday"
+      time: "05:00"
+      timezone: "Europe/Amsterdam"
+    open-pull-requests-limit: 20
```
ijemmy, line 12: I have to say that this check scares me. It has been a while... :D

ijemmy (approved):
Thanks for doing this.

.github/dependabot.yml
dreamorosi: Yes, but luckily we don't have that many dependencies ([only 3](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/chore/set_dependabot_weekly/docs/requirements.txt)) and we have been keeping them relatively up to date ([see latest PR](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/688)). But let's see...

saragerion (approved):
LGTM

Issue #275, opened 2021-12-08
Feature request: tracer feature improvements


dreamorosi:
## Description of the feature request

**Problem statement**
While an initial implementation of the `Tracer` core package has been done there are a handful of features that are missing when compared with the Python version. Some of these features were implemented recently while others, due to the language difference, need to be implemented differently and for this reason where left out in the initial implementation.

**Summary of the feature**
* [x] Cold start annotation present for every invocation ([#851](https://github.com/awslabs/aws-lambda-powertools-python/pull/851/files)). At the moment `Tracer` annotates only `ColdStart = true` during the first invocation.
* [x] Service name is added as an annotation ([#861](https://github.com/awslabs/aws-lambda-powertools-python/pull/861)). At the moment `Tracer` uses the service name only as default namespace when adding metadata to traces.
* [x] Automatic traces for HTTP requests ([Python version captures all libraries](https://awslabs.github.io/aws-lambda-powertools-python/latest/core/tracer/#tracing-aiohttp-requests)). At the moment `Tracer` ([supports this](https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-nodejs-httpclients.html)) only via explicit opt-in using the escape hatch mechanism.
* [ ] ~Reusing tracer across your code ([Python version implements Singleton pattern](https://awslabs.github.io/aws-lambda-powertools-python/latest/core/tracer/#reusing-tracer-across-your-code)). At the moment every `new Tracer()` returns a completely new instance, JS/TS could implement it [like so](https://stackoverflow.com/questions/1479319/simplest-cleanest-way-to-implement-a-singleton-in-javascript).~
⬆️ We are opting to not introduce this feature in order to not force this pattern on every customer. See [this comment](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/767#issuecomment-1104995642).

**Code examples**
N/A

**Benefits for you and the wider AWS community**
Same behaviour as other Powertools libraries & better DX

**Describe alternatives you've considered**
N/A

**Additional context**

PR #674, opened 2022-03-18
docs(examples): add example for AWS SAM

bpauwels:
## Description of your changes
Added an example for AWS SAM

### How to verify this change
see new SAM example, use SAM CLI to deploy it

### Related issues, RFCs

https://github.com/awslabs/aws-lambda-powertools-typescript/issues/606

### PR status

***Is this ready for review?:*** YES  
***Is it a breaking change?:*** NO

## Checklist

- [x] [My changes meet the tenets criteria](https://awslabs.github.io/aws-lambda-powertools-python/#tenets)
- [x] I have performed a *self-review* of my own code
- [x] I have *commented* my code where necessary, particularly in areas that should be flagged with a TODO, or hard-to-understand areas
- [x] I have made corresponding changes to the *documentation*
- [x] I have made corresponding changes to the *examples*
- [x] My changes generate *no new warnings*
- [x] The *code coverage* hasn't decreased
- [ ] I have *added tests* that prove my change is effective and works
- [x] New and existing *unit tests pass* locally and in Github Actions
- [x] Any *dependent changes have been merged and published* in downstream module
- [x] The PR title follows the [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/semantic.yml#L2)

### Breaking change checklist
no breaking changes

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #767, opened 2022-04-14
docs(tracer): Allow to reuse Tracer across your code

dreamorosi:

When using Tracer customers might want to trace parts of code that are distributed across files/modules. At the moment, in order to maintain the correct relationship between segments, customers would have had to either pass the `Tracer` instance or a parent segment around, which is tedious and leads to tighter coupling.

This PR makes Tracer reusable by returning the a previously initialised instance, if one exists. With this change customers can:
```ts
// index.ts
import collectPayment from './payment'; 

const tracer = new Tracer();

export const handler = async (event: unknown) => {
    const segment = tracer.getSegment(); // This is the facade segment (the one that is created by AWS Lambda)
    // Create subsegment for the function & set it as active
    const subsegment = segment.addNewSubsegment(`## ${process.env._HANDLER}`);
    tracer.setSegment(subsegment);

    await collectPayment(event.chargeId);

    ....
};
```
and then do:
```ts
// payment.ts

const tracer = new Tracer();

const collectPayment = async (chargeId: string) => {
  // process charge
 tracer.putAnnotation('chargeId', chargeId); // This annotation will be done on the `## index.handler` segment 
};

export default collectPayment;
```
or:
```ts
// payment.ts

const tracer = new Tracer();

const collectPayment = async (chargeId: string) => {
  const mainSegment = tracer.getSegment();
  const subsegment = mainSegment.addNewSubsegment('### collectPayment');
  tracer.setSegment(subsegment);

  // This annotation will be done on the `### collectPayment` segment 
  tracer.putAnnotation('chargeId', chargeId);

   // Process charge

  subsegment.close();
  tracer.setSegment(mainSegment);
};

export default collectPayment;
```

### How to verify this change

See existing & new test cases passing.

See screenshot of segments generated by second example above:
image

See video of how the docs look after the changes:

https://user-images.githubusercontent.com/7353869/163380017-b0b457ee-afcc-48e4-b2eb-bb0d375ace25.mp4

### Related issues, RFCs

[#275](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/275)  

### PR status

## Checklist

### Breaking change checklist

N/A

---

### DIFF
--- a/docs/core/tracer.md
+++ b/docs/core/tracer.md
@@ -407,6 +407,85 @@ Use **`POWERTOOLS_TRACER_CAPTURE_ERROR=false`** environment variable to instruct
 
     1. You might **return sensitive** information from exceptions, stack traces you might not control
 
+### Reusing Tracer across your code
+
+Tracer keeps a copy of its configuration after the first initialization. This is useful for scenarios where you want to use Tracer in more than one location across your code base.
+
+=== "index.ts"
+
+    ```typescript hl_lines="5"
+    import { Tracer } from '@aws-lambda-powertools/tracer';
+    import middy from '@middy/core';
+    import collectPayment from './payment';
+
+    const tracer = new Tracer({ serviceName: 'serverlessAirline' }); // (1)
+
+    export const handler = middy(async (event: any, _context: any): Promise<void> => {
+        await collectPayment(event.chargeId);
+    }).use(captureLambdaHandler(tracer));
+    ```
+
+    1.  Since this is the first time Tracer is initialized, the settings are reused across all other Tracer instances
+=== "payment.ts"
+
+    ```typescript hl_lines="3"
+    import { Tracer } from '@aws-lambda-powertools/tracer';
+
+    const tracer = new Tracer(); // (1)
+
+    const collectPayment = async (chargeId: string) => {
+        const mainSegment = tracer.getSegment(); // (2)
+        const subsegment = mainSegment.addNewSubsegment('### collectPayment');
+        tracer.setSegment(subsegment);
+
+        // This annotation will be done on the `### collectPayment` segment 
+        tracer.putAnnotation('chargeId', chargeId);
+
+        /* ... */
+
+        subsegment.close();
+        tracer.setSegment(mainSegment);
+    };
+
+    export default collectPayment;
+    ```
+    
+    1.  You don't need to pass any parameters here, this instance will have the same configuration as the first Tracer instance
+    2.  This is the main subsegment called `### index.handler` that was created by the `captureLambdaHandler` middleware
+=== "Example Raw X-Ray Trace excerpt"
+
+    ```json hl_lines="3 17 20 24"
+    {
+        "id": "22883fbc730e3a0b",
+        "name": "## index.handler",
+        "start_time": 1647956168.22749,
+        "end_time": 1647956169.0679862,
+        "annotations": {
+            "ColdStart": true
+        },
+        "metadata": {
+            "default": {
+                "index.handler response": {
+                    "body": "{\"message\":\"Payment collected for chargeId: 1234\"}",
+                    "statusCode": 200
+                }
+            }
+        },
+        "subsegments": [
+            {
+                "id": "24542b252e5c0756",
+                "name": "collectPayment",
+                "start_time": 1647956169.22749,
+                "end_time": 1647956169.32749,
+                "annotations": {
+                    "chargeId": "1234"
+                }
+            }
+        ]
+    }
+    ```
+
+
 ### Escape hatch mechanism
 
 You can use `tracer.provider` attribute to access all methods provided by the [AWS X-Ray SDK](https://docs.aws.amazon.com/xray-sdk-for-nodejs/latest/reference/AWSXRay.html).
--- a/docs/core/tracer.md
+++ b/docs/core/tracer.md
@@ -418,14 +418,14 @@ Tracer keeps a copy of its configuration after the first initialization. This is
     import middy from '@middy/core';
     import collectPayment from './payment';
 
-    const tracer = new Tracer({ serviceName: 'serverlessAirline' }); // (1)
+    const tracer = new Tracer(); // (1)
 
     export const handler = middy(async (event: any, _context: any): Promise<void> => {
         await collectPayment(event.chargeId);
     }).use(captureLambdaHandler(tracer));
     ```
 
-    1.  Since this is the first time Tracer is initialized, the settings are reused across all other Tracer instances
+    1.  Because of the way ESM modules work, in order to set up Tracer, you'll need to pass the configurations as [environment variables](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/#environment-variables).
 === "payment.ts"
 
     ```typescript hl_lines="3"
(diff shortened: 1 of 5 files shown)


humanzz:
I wonder if this singleton pattern should be done for all 3 utilities; tracer, logger and metrics. In larger projects, with many files and modules, this would be really useful I believe.

dreamorosi:
> I wonder if this singleton pattern should be done for all 3 utilities; tracer, logger and metrics. In larger projects, with many files and modules, this would be really useful I believe.

It's something we can consider. Could you please open an issue requesting the feature? We'll then prioritize accordingly.

humanzz:
> > I wonder if this singleton pattern should be done for all 3 utilities; tracer, logger and metrics. In larger projects, with many files and modules, this would be really useful I believe.
> 
> It's something we can consider. Could you please open an issue requesting the feature? We'll then prioritize accordingly.

https://github.com/awslabs/aws-lambda-powertools-typescript/issues/777

docs/core/tracer.md
```diff
@@ -407,6 +407,85 @@ Use **`POWERTOOLS_TRACER_CAPTURE_ERROR=false`** environment variable to instruct
...
+    import collectPayment from './payment';
+
+    const tracer = new Tracer(); // (1)
+
+    export const handler = middy(async (event: any, _context: any): Promise<void> => {
+        await collectPayment(event.chargeId);
+    }).use(captureLambdaHandler(tracer));
+    ```
+
+    1.  Because of the way ESM modules work, in order to set up Tracer, you'll need to pass the configurations as [environment variables](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/#environment-variables).
```
ijemmy, line 428: This line sounds like I cannot pass any parameter to Tracer object.  I can only use environment variable.

That isn't the intention, is it?

dreamorosi: That's exactly what I'm hinting at. I was going to raise this during the next meeting but essentially what happens is the following:

Suppose we have a file `handler.ts` which is the main entry point for Lambda:
```ts
import payment from './payment.ts`

const tracer = new Tracer({ ... some parameters });

export handler = () => {
   payment()
}
```
and then another file `payment.ts`:
```ts
import payment from './payment.ts`

const tracer = new Tracer({ ... some OTHER parameters });

const payment = () => {}
```

What will happen in this case is that customers **can pass parameters** to either of the two instances, but while you'd expect the ones in `handler.ts` to be picked up, the instance in `payment.ts` is instead the one that is initialised first and so these settings are picked up and reused.

This happens, to the best of my knowledge, because the module is imported (and so executed) before and even if the order of the imports is changed (i.e. I initialise the "main" tracer before `import payment...`), when using something like `esbuild` the imports are still placed at the top.

With that in mind we have three options:
1. Let customers pass the settings when initialising, but explain this behaviour. This would be in line with the current behaviour that other examples and use cases have but now customers would have to handle the cognitive load of keeping track of which instance is initialised first.
2. Specify that when reusing the tracer across file, the only way to pass configurations is via environment variables. This would mean a different behaviour but since the configurations are in the environment whichever instance is initialised first will always get the correct settings.
3. Push back this feature until we find a better option (if it exists). While evaluating this last option, we should also consider that this same issue will likely affect the other utilities.

Happy to discuss this and definitely open to ideas/suggestions/alternatives.

dreamorosi: cc @saragerion @flochaz

dreamorosi:
Following a discussion with the other maintainers we have decided to scrap this feature in order to avoid forcing a pattern on customers and also in light of the behaviour [detailed here](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/767#discussion_r854129149).

In Javascript 

For example, customers could create a module (even in a Lambda Layer) called `powertools.ts`, that imports the utilities and instantiates the modules with common configurations:
```ts

const awsLambdaPowertoolsVersion = '0.6.0';

const defaultValues = {
  awsAccountId: process.env.AWS_ACCOUNT_ID || 'N/A',
  environment:  process.env.ENVIRONMENT || 'N/A'
};

const logger = new Logger({
  persistentLogAttributes: {
    ...defaultValues,
    logger: {
      name: '@aws-lambda-powertools/logger',
      version: awsLambdaPowertoolsVersion,
    }
  },
});

const metrics = new Metrics({
  defaultDimensions: defaultValues
});

const tracer = new Tracer();

export {
  logger,
  metrics,
  tracer
};
```

This module can then be imported and used in other places and files:
```ts
// payment.ts
import { tracer, logger } from './powertools.ts';

const collectPayment = async (chargeId: string) => {
  const mainSegment = tracer.getSegment();
  const subsegment = mainSegment.addNewSubsegment('### collectPayment');
  tracer.setSegment(subsegment);

  // This annotation will be done on the `### collectPayment` segment 
  tracer.putAnnotation('chargeId', chargeId);

   // Process charge

   // This log will have the same config & persistent keys as the others
   logger.info('charge processed', { chargeId: chargeId });

  subsegment.close();
  tracer.setSegment(mainSegment);
};

export default collectPayment;
```
and
```ts
import { captureLambdaHandler } from 'from '@aws-lambda-powertools/tracer';
import { injectLambdaContext } from 'from '@aws-lambda-powertools/logger';
import { tracer, logger } from './powertools.ts';
import collectPayment from './payment.ts';

export const handler = middy(async (event: unknown) => {
    const segment = tracer.getSegment(); // This is the facade segment (the one that is created by AWS Lambda)
    // Create subsegment for the function & set it as active
    const subsegment = segment.addNewSubsegment(`## ${process.env._HANDLER}`);
    tracer.setSegment(subsegment);

    await collectPayment(event.chargeId);

    ....
}).use(captureLambdaHandler(tracer)).use(injectLambdaContext(logger));
```

This PR will now focus on documenting this pattern in the documentation with the aim of providing prescriptive guidance about this use case.

docs/core/tracer.md
dreamorosi: Resolving this conversation in light of the decision detailed [here](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/767#issuecomment-1104995642).

ijemmy:
> Following a discussion with the other maintainers we have decided to scrap this feature in order to avoid forcing a pattern on customers and also in light of the behaviour [detailed here](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/767#discussion_r854129149).
>  ....
> 
> This PR will now focus on documenting this pattern in the documentation with the aim of providing prescriptive guidance about this use case.

After reading #777 , two options came to my mind:
1. Should we also update the `examples/cdk` to use this pattern? 
2. Should we have a "Recommended practices" section with this topic? This is applicable for all utilities. Then we could have a link for each utility doc to this page.

dreamorosi:
> > Following a discussion with the other maintainers we have decided to scrap this feature in order to avoid forcing a pattern on customers and also in light of the behaviour [detailed here](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/767#discussion_r854129149).
> > ....
> > This PR will now focus on documenting this pattern in the documentation with the aim of providing prescriptive guidance about this use case.
> 
> After reading #777 , two options came to my mind:
> 
> 1. Should we also update the `examples/cdk` to use this pattern?
> 2. Should we have a "Recommended practices" section with this topic? This is applicable for all utilities. Then we could have a link for each utility doc to this page.

1. Yes, I think we should but I would wait till at least #674 is merged. At that point we'll have to refactor the CDK examples anyways so it'll be a good moment.
2. I like the idea and I think we should open a separate issue referencing #777 and this PR to track it. I don't think it's critical for our current milestone so if we all agree I would propose to:
- scrap and close this PR
- close #275 since this is not a feature we'll implement (it's a pattern) and all others have been already merged
- keep the soon-to-be-opened issue about documenting this in a "Recommended practices" section & prioritise it after RC/GA.

dreamorosi:
We are going to be closing this PR as detailed [here](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/767#issuecomment-1105163367).

We will, after RC/GA release, work on a "Recommended/Best Practices" section in the docs that will showcase this pattern for all modules. We will be tracking this effort in #777.

Issue #773, opened 2022-04-14
Bug: xray_trace_id is not populated in log lines


humanzz:
## Bug description

According to [documentation](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/logger/#standard-structured-keys), the `xray_trace_id` should be emitted in all log lines if the request is traced. However, I'm not seeing any setting of that field in my logs despite the `REPORT` log line showing the request is traced.

### Expected Behavior

All log lines to have `xray_trace_id` populated if the request is traced (as confirmed by `REPORT` line).

### Current Behavior

No log lines showing the `xray_trace_id`.

### Possible Solution

I've read a little bit in the code in [Logger.ts](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/logger/src/Logger.ts) and I suspect that [xRayTraceId](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/logger/src/Logger.ts#L632) is set only once at `Logger` construction time.

The traceId is read via an environment variable that'll change for each different request. As such, I think a possible solution is to leverage `addContext` to also set that value when calling [this.addToPowertoolLogData](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/logger/src/Logger.ts#L166)

The actual behaviour I've seen is that `xray_trace_id` is not being written to the logs. If my understanding is correct, I believe this issue can be manifested in 1 of 2 ways
1. The one observed i.e. lack of `xray_trace_id` which is likely to happen if the first request is not traced (i.e. cold start request). In this case all subsequent requests will not have the field
2. Wrong `xray_trace_id` (not observed) if the first request is traced and the environment variables is read and the field is set in Logger's state but subsequent requests will not update that value.

### Steps to Reproduce

For a handler looking like

```
export const handler = middy(vanillaHandler)
  .use(captureLambdaHandler(tracer))
  .use(logMetrics(metrics))
  .use(injectLambdaContext(logger)),
```

1. Invoke the function multiple times
3. Inspect the log lines looking for `xray_trace_id` in the different requests (especially those where `REPORT` line indicates request is traced)

`xray_trace_id` is missing

### Environment

* **Powertools version used**: 0.8.0
* **Packaging format (Layers, npm)**: npm
* **AWS Lambda function runtime:** Node.js 14.x
* **Debugging logs:**

### Related issues, RFCs

N/A

dreamorosi:
[...]

PR #776, opened 2022-04-15
fix(logger): add xray_trace_id to every log

dreamorosi:

As reported by user @humanzz in #773 the Logger utility was not including the `xray_trace_id` key in logs. This field is [listed in the documentation](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/logger/#standard-structured-keys) as a standard key that should appear in each log.

The user reporting the issue was helpful in detailing the fact that the `_X_AMZN_TRACE_ID` environment variable used to retrieve the value changes at every invocation while Logger was instead attempting to read this value when instantiated (see `constructor` -> `setOptions` -> `setPowertoolLogData`).

This PR changes this behaviour and appends the `xray_trace_id` key when the log entry is created and populated instead (`createAndPopulateLogItem` function). This behaviour [is consistent with the implementation](https://github.com/awslabs/aws-lambda-powertools-python/issues/137#issuecomment-683454720) also used in the Python version of Powertools.

As additional notes:
- This was not caught in unit tests because the environment variable is set by `packages/logger/tests/helpers/populateEnvironmentVariables.ts` which is imported before tests are run. So when running the tests the `xray_trace_id` would always be present.
- This was not caught in the e2e/integration tests either because no assertion was made on the presence of this key in the logs (this PR adds a test case to check this).
- The documentation erroneously stated that this key should be set only when Active Tracing is enabled. This is not true as the `_X_AMZN_TRACE_ID ` environment variable [is always set in the Lambda execution environment](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-retrieve), regardless of whether tracing is enabled or not. This means the key will always be present in each log. AWS Lambda Powertools for Python makes the same statement and I have tested the latest version available of the library on a Lambda function with Active Tracing disabled and the key is also always present in each log.

### How to verify this change

- See unit test results in this PR
- See result of e2e tests here (<- TBA)
- See example of CloudWatch logs generated after this change (below)
image

### Related issues, RFCs

[#773](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/773)

### PR status

## Checklist

### Breaking change checklist

N/A

---


dreamorosi:
See e2e test result here (passing): https://github.com/awslabs/aws-lambda-powertools-typescript/runs/6041778790?check_suite_focus=true

docs/core/logger.md
```diff
@@ -100,7 +100,7 @@ Key | Example | Note
 **sampling_rate**: `float` |  `0.1` | When enabled, it prints all the logs of a percentage of invocations, e.g. 10%
 **service**: `string` | `serverlessAirline` | A unique name identifier of the service this Lambda function belongs to, by default `service_undefined`
 **timestamp**: `string` | `2011-10-05T14:48:00.000Z` | Timestamp string in simplified extended ISO format (ISO 8601)
-**xray_trace_id**: `string` | `1-5759e988-bd862e3fe1be46a994272793` | When [tracing is enabled](https://docs.aws.amazon.com/lambda/latest/dg/services-xray.html){target="_blank"}, it shows X-Ray Trace ID
+**xray_trace_id**: `string` | `1-5759e988-bd862e3fe1be46a994272793` | X-Ray Trace ID, when [tracing is enabled](https://docs.aws.amazon.com/lambda/latest/dg/services-xray.html){target="_blank"} it's used to send data to X-Ray
```
ijemmy, line 103: From the code, it seems like we always send it (and it alway exist as in PR description).  It might be better clarified here to avoid confusion.

```suggestion
**xray_trace_id**: `string` | `1-5759e988-bd862e3fe1be46a994272793` | X-Ray Trace ID. This value is always presented in Lambda environment, whether [tracing is enabled](https://docs.aws.amazon.com/lambda/latest/dg/services-xray.html){target="_blank"} or not. Logger will always log this value.
```

ijemmy (approved):

flochaz (approved):

Issue #825, opened 2022-04-29
Maintenance: have descriptive logs in the integration tests


dreamorosi:
## Description of the feature request

**Problem statement**
Now that we are running more integration tests (multiple runtimes for each utility) and in parallel it would be beneficial to have logs that explicitly indicate which utility they belong to. This would help troubleshoot faster in case of issues.

**Summary of the feature**
Currently this is how the log prefix are:
![image](https://user-images.githubusercontent.com/7353869/165973199-70315f69-f7b4-453f-8667-4fed7e2aff1a.png)

Ideally we could have something like:
```
[logger:e2e:nodejs14x]
[tracer:e2e:nodejs14x]
[metrics:e2e:nodejs12x]
```

**Code examples**
N/A

**Benefits for you and the wider AWS community**
N/A - this change concerns mainly maintainers.

**Describe alternatives you've considered**
None / scroll through the logs 

**Additional context**
N/A

### Related issues, RFCs

N/A

ijemmy:
GitHub matrix strategy may be a better approach. It will fix this issue and be more efficient. I'm looking into that.

PR #828, opened 2022-05-09
test(all): switch to use GitHub strategy matrix and fix flaky tests

ijemmy:

The e2e tests have been flaky due to following reasons:
1. Some tests (mostly Logger) time out. **Cause:** Refactoring tracing increase the number of concurrent deployment. This slows down the overall run and we are more likely to hit timeout. (GitHub runner has only 2 cores)
2. Tracer tests sometimes fail from not having a particular field (`error`). **Causes:** We removed the explicit wait on Tracer during refactoring. We now rely on polling until we receive the expected number of traces. However, not all subsegments are available in the traces yet. 

This PR addresses the issues by:
1. Increases timeout
2. Use matrix strategies to run 6 combinations in 6 different runners ( 2 Lambda runtimes x 3 packages)
3. Add a logic on polling to retry if the subsegments are not available yet.

### How to verify this change
You can trigger the e2e test workflow manually.

I've run this 5 times without any failure so far. > https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/2295223762. 

### Related issues, RFCs
#825  <-- the log will be descriptive with Matrix strategy

### PR status

## Checklist

### Breaking change checklist

---


packages/commons/tests/utils/e2eUtils.ts
```diff
@@ -57,7 +57,7 @@ export const createStackWithLambdaFunction = (params: StackWithLambdaFunctionOpt
 };
 
 export const generateUniqueName = (name_prefix: string, uuid: string, runtime: string, testName: string): string => 
-  `${name_prefix}-${runtime}-${testName}-${uuid}`.substring(0, 64);
+  `${name_prefix}-${runtime}-${uuid.substring(0,5)}-${testName}`.substring(0, 64);
```
ijemmy, line 60: When test name is really long, the uuid is truncated on some resources. This results in name clashing when the same test cases are run at the same time.

dreamorosi (dismissed):
Already looks good and passing!

As only note I'd recommend removing `concurrently` from the `package-lock.json`. Since we are no longer using it we can avoid having to maintain future updates for it.

.github/workflows/run-e2e-tests.yml
```diff
@@ -2,48 +2,63 @@ name: run-e2e-tests
...
+          npm run test
+  e2e-tests:
     runs-on: ubuntu-latest
     permissions:
       id-token: write # needed to interact with GitHub's OIDC Token endpoint.
       contents: read
+    strategy:
+      matrix:
+        version: [12, 14]
+        package: [logger, metrics, tracing]
```
saragerion, line 42: Outside of the scope of this PR, but we should rename the tracer folder to "tracer" for consistency. Created an issue:
https://github.com/awslabs/aws-lambda-powertools-typescript/issues/829

packages/logger/tests/e2e/constants.ts
```diff
@@ -1,5 +1,6 @@
 export const RESOURCE_NAME_PREFIX = 'Logger-E2E';
-export const TEST_CASE_TIMEOUT = 20_000; // 20 seconds
-export const SETUP_TIMEOUT = 300_000; // 300 seconds
-export const TEARDOWN_TIMEOUT = 200_000; 
+export const ONE_MINUTE = 60 * 1000;
+export const TEST_CASE_TIMEOUT = ONE_MINUTE;
+export const SETUP_TIMEOUT = 5 * ONE_MINUTE;
+export const TEARDOWN_TIMEOUT = 5 * ONE_MINUTE;
```
saragerion, line 5: I like this syntax. More readable!

packages/tracing/tests/helpers/tracesUtils.ts
```diff
@@ -14,11 +14,16 @@ import {
   expectedCustomResponseValue, 
   expectedCustomErrorMessage,
 } from '../e2e/constants';
+import { FunctionSegmentNotDefinedError } from './FunctionSegmentNotDefinedError';
```
saragerion, line 17: Question: what do you mean here with segment not defined?

packages/tracing/tests/helpers/tracesUtils.ts
```diff
@@ -14,11 +14,16 @@ import {
...
 interface ParsedDocument {
   name: string
   id: string
   start_time: number
   end_time: number
+  // This flag may be set if the segment hasn't been fully processed
+  // The trace may have already appeared in the `getTraceSummaries` response 
+  // but a segment may still be in_progress
+  in_progress?: boolean
```
saragerion, line 26: By looking at the documentation, also the `end_time` key might be not defined, correct? 

> end_time – number that is the time the segment was closed. For example, 1480615200.090 or 1.480615200090E9. Specify either an end_time or in_progress.

https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html#api-segmentdocuments-fields

packages/tracing/tests/helpers/tracesUtils.ts
```diff
@@ -117,6 +122,28 @@ const getTraces = async (xrayClient: XRay, startTime: Date, resourceArn: string,
...
+        invocationSubsegment = getInvocationSubsegment(trace);
+      } catch (error) {
+        if (error instanceof FunctionSegmentNotDefinedError){
+          retry(new Error(`There is no Function subsegment (AWS::Lambda::Function) yet. Retry.`));
+        } else {
+          throw error;
+        }
+      }
+
+      retryFlag = retryFlag || (!!invocationSubsegment.in_progress);
```
saragerion, line 141: Question: essentially here we are looking at the `in_progress` key. If it's set, we retry. Correct?

saragerion (commented):
Thanks for the PR! I left some comments/questions but already looking good :-)

dreamorosi (approved):

saragerion (approved):

packages/tracing/tests/helpers/tracesUtils.ts
ijemmy: If X-Ray hasn't fully processed all segments, some of them may be missing.
In our case, we may try to get the Function (AWS::Lambda::Function) segment too early. I use this custom error to distinguish the type of error form others.

packages/tracing/tests/helpers/tracesUtils.ts
ijemmy: Yes. In my experiments, `end_time` field doesn't exist.  It should be optional.

image

packages/tracing/tests/helpers/tracesUtils.ts
saragerion: I see! thanks for clarifying

.github/workflows/run-e2e-tests.yml
dreamorosi: #854

PR #881, opened 2022-05-16
fix(commons) prune issue - fix #664

Muthuveerappanv:
## Importing Common Index in Logger Folder raises Module Error #664

* Import Commons Utility from the commons using `import { Utility } from '@aws-lambda-powertools/commons/lib/Utility';` instead of directly from the package's index
* The index exports contain `tests/resources` references. Pruning tools removes `tests` folder within the code during build 
* Hence importing the Utility directly from its inner package

### Related issues, RFCs

[#664](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/664)  

### PR status

***Is this ready for review?:*** NO  
***Is it a breaking change?:*** NO

## Checklist

- [x] [My changes meet the tenets criteria](https://awslabs.github.io/aws-lambda-powertools-python/#tenets)
- [x] I have performed a *self-review* of my own code
- [x] I have *commented* my code where necessary, particularly in areas that should be flagged with a TODO, or hard-to-understand areas
- [x] I have made corresponding changes to the *documentation*
- [x] I have made corresponding changes to the *examples*
- [x] My changes generate *no new warnings*
- [x] The *code coverage* hasn't decreased
- [x] I have *added tests* that prove my change is effective and works
- [x] New and existing *unit tests pass* locally and in Github Actions
- [x] Any *dependent changes have been merged and published* in downstream module
- [ ] The PR title follows the [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/semantic.yml#L2)

### Breaking change checklist

- [ ] I have documented the migration process
- [ ] I have added, implemented necessary warnings (if it can live side by side)

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #882, opened 2022-05-18
fix(commons): rename tests subfolder to samples to avoid being deleted by tools such as node-prune

flochaz:

Current project structure is having a folder called `tests` under `commons/src` package and was hosting samples context and events to use when  calling a lambda function.

Customers using tools like [node-prune](https://github.com/tj/node-prune) where having issue (#881) because the tool was deleting the folder. A fix would have been to simply use the `-exclude` option of this tool but here we propose to simply rename the folder to `samples`.

### How to verify this change

1. prep the env
   ```sh
   npm ci --foreground-scripts
   npm run lerna-package
   cd examples/cdk
   npm ci
   npm install ../../packages/**/dist/aws-lambda-powertools-*
   npm run test
   ```
1. use node-prune
   ```
   node-prune -exclude '*.d.ts' -excludes 'assets'
   ```
1. test again, they should still pass
   ```sh
   npm run test
   ```

### Related issues, RFCs

#881 

### PR status

## Checklist

### Breaking change checklist

- [x] I have documented the migration process
- [x] I have added, implemented necessary warnings (if it can live side by side)

---


dreamorosi:
Is that any other way in which we can avoid including all these 50+ files in the deployment package?

Also note that there is another community PR #881 that attempts to fix this issue, we should at least address it before moving forward with this. And generally speaking, what do you think of that way of fixing it @flochaz?

flochaz:
I think we should do both : import Utility as suggested in #881 and change name because indeed `tests` is a confusing name. After the question regarding the need of this samples ... I am not sure ...

ijemmy (approved):
Even without `node-prune` issue, I do agree that `tests` is misleading here and `samples` is a better name.

Issue #880, opened 2022-05-16
Bug: regression - not possible to log arbitrary objects


dreamorosi:
## Bug description

After the refactoring introduced in #614 the Logger utility only allows to log values that are of type `Error` or `string`.

As such it's not possible anymore to log an arbitrary object, which causes unneeded friction since the point of logger is to promote structured logging.

As a trivial example, I might want to log the `event` of my function:
```ts
logger.debug('Lambda invoked', {
  event: { event }
});
```

This causes compile errors when using TypeScript (see below for details).

### Expected Behavior

1. Types allow passing strings, `Error`, and arbitrary objects
2. Logger parses/structures each type correctly (i.e. strings are appended to a key & not broken down like it used [to happen here](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/565#issue-1145667907), `Error` objects are broken down correctly, other types of objects are appended)

### Current Behavior

```
Argument of type '[{ details: { event: APIGatewayRequestAuthorizerEvent; }; }]' is not assignable to parameter of type 'LogItemExtraInput'.
  Type '[{ details: { event: APIGatewayRequestAuthorizerEvent; }; }]' is not assignable to type '[string | Error]'.
    Type '{ details: { event: APIGatewayRequestAuthorizerEvent; }; }' is not assignable to type 'string | Error'.
      Object literal may only specify known properties, and 'details' does not exist in type 'Error'.ts(2345)
```

### Possible Solution

N/A

### Steps to Reproduce

```ts
import { APIGatewayRequestAuthorizerEvent } from 'aws-lambda';
import { Logger } from '@aws-lambda-powertools/logger';

const logger = new Logger({ logLevel: 'DEBUG' });

export const handler = async (event:  APIGatewayRequestAuthorizerEvent) => {
  
  logger.debug('Lambda invoked', {
    details: { event }
  });

  return;
};
```

<img width="775" alt="image" src="https://user-images.githubusercontent.com/7353869/168583811-1a5d7f3f-8cf6-4e57-a360-6b5570ae1ddb.png">

### Environment

* **Powertools version used**: v0.9.0
* **Packaging format (Layers, npm)**: npm
* **AWS Lambda function runtime:** N/A
* **Debugging logs:** N/A

### Related issues, RFCs

#565 
#614

awsdb:
I'd like to take this one! @dreamorosi, please assign it to me.

PR #614, opened 2022-03-02
fix(logger): fix handling of additional log keys

awsdb:
## Description of your changes

`Logger` accepts a second parameter that allows to append additional keys and values to a single log item.
Currently, there are 5 use cases that are supported for this:

```ts
// 1) object
logger.info('This is the message value', { extra: 'parameter' });

// 2) object, object, ...
logger.info('This is the message value', { parameterOne: 'foo' }, { parameterTwo: 'bar' });

// 3) Object (that contains the message key)
logger.info( { message: 'This is the message value', extra: 'parameter' });

// 4) error
logger.info('This is the message value', new Error('Something happened!') );

// 5) error with custom key
logger.info('This is the message value', { myCustomErrorKey: new Error('Something happened!') });
```

However, the current implementation allows passing simple strings, arrays, and any other `unknown` objects (single or multiple) as a second (third, fourth, ...) parameter, causing unwanted side effects.

This PR addresses this issue by aligning the implementation to other AWS Lambda Powertools flavors like e.g. Python.
So, one more use cases is allowed:

```ts
// 6) simple string
logger.info('This is the message value', 'Additional string value');
```

Additionally:
- passing multiple objects is restricted to only `{ key: 'value' }` (key-value) kind of ones
- in general, passing arrays and other `unknown` objects is not supported anymore (**BREAKING**)
- passing multiple strings and `Error`s is not supported anymore (**BREAKING**)

### How to verify this change

See the unit tests.
Check the updated documentation (section "Appending additional log keys and values to a single log item").
Try to pass unsupported objects, e.g.:
```ts
logger.info('foo', 'bar', 'baz');
logger.info('foo', ['bar', 'baz']);
```

### Related issues, RFCs

[#565](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/565)  

### PR status

***Is this ready for review?:*** YES  
***Is it a breaking change?:*** YES

## Checklist

- [x] [My changes meet the tenets criteria](https://awslabs.github.io/aws-lambda-powertools-python/#tenets)
- [x] I have performed a *self-review* of my own code
- [x] I have *commented* my code where necessary, particularly in areas that should be flagged with a TODO, or hard-to-understand areas
- [x] I have made corresponding changes to the *documentation*
- [x] I have made corresponding changes to the *examples*
- [x] My changes generate *no new warnings*
- [x] The *code coverage* hasn't decreased
- [x] I have *added tests* that prove my change is effective and works
[...]

Issue #565, opened 2022-02-21
bug(logger): standardize extra input logging 


dreamorosi:
## Bug description

As described in [the docs](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/logger/#appending-additional-log-keys-and-values-to-a-single-log-item) `Logger` accepts a second parameter that allows to append additional keys and values in a single log item.

This works when passing an object, for example `logger.info('my-message', { data: 'my-value' }); ` results in:
```json
{
  "cold_start": true,
  "function_arn": "arn:aws:lambda:eu-west-1:123456789101:function:myFunction",
  "function_memory_size": 128,
  "function_name": "AwsLambdaPowertoolsDemoSt-HelloWorldFunctionB2AB6E-sYstm6IvDAEB",
  "function_request_id": "b5fe0e01-91f5-43f2-8ef9-d572df88b61c",
  "level": "INFO",
  "message": "my-message",
  "service": "service_undefined",
  "timestamp": "2022-02-21T11:15:18.493Z",
  "data": "my-value"
}
```

But when passing directly a string, like: `logger.info('my-message', 'my-value');` it results in this:
```json
{
  "0": "m",
  "1": "y",
  "2": "-",
  "3": "v",
  "4": "a",
  "5": "l",
  "6": "u",
  "7": "e",
  "cold_start": false,
  "function_arn": "arn:aws:lambda:eu-west-1:123456789101:function:myFunction",
  "function_memory_size": 128,
  "function_name": "AwsLambdaPowertoolsDemoSt-HelloWorldFunctionB2AB6E-sYstm6IvDAEB",
  "function_request_id": "a0c5ab8f-c0f0-40fa-8c75-347020065880",
  "level": "INFO",
  "message": "my-message",
  "service": "service_undefined",
  "timestamp": "2022-02-21T11:08:19.611Z"
}
```

Likewise, when passing an array, `logger.info('my-message', ['my-value', 'my-value2']);` it results in:
```json
{
  "0": "my-value",
  "1": "my-value2",
  "cold_start": true,
  "function_arn": "arn:aws:lambda:eu-west-1:123456789101:function:myFunction",
  "function_memory_size": 128,
  "function_name": "AwsLambdaPowertoolsDemoSt-HelloWorldFunctionB2AB6E-sYstm6IvDAEB",
  "function_request_id": "b5fe0e01-91f5-43f2-8ef9-d572df88b61c",
  "level": "INFO",
[...]

PR #883, opened 2022-05-18
fix(logger): enable logging of arbitrary objects

awsdb:

This PR relaxes the restrictions introduced by #614. In #565, we decided to restrict the additional logging data objects in their shape. It turned out, that this restriction prevents from logging of arbitrary objects, which is inconvenient.  
In this PR, the type restrictions are removed, so it is now possible to specify any object as additional payload when calling the logger methods directly.

### How to verify this change

- Run unit tests
- Run E2E tests
- Review docs and examples
- Try to log an arbitrary object by providing it as a property of a string-indexed object, e.g.

```typescript

const logger = new Logger({ logLevel: 'DEBUG' });

  logger.debug('Lambda invoked', { details: { event } });
};
```

No compile time errors should be generated for the above example. The code should generate a valid log entry when executed.

### Related issues, RFCs

[#880](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/880)  

### PR status

## Checklist

### Breaking change checklist

---

### DIFF
--- a/docs/core/logger.md
+++ b/docs/core/logger.md
@@ -286,7 +286,7 @@ If you already have an object containing a `message` key and an additional prope
 
     const logger = new Logger();
     
-    export const handler = async (_event: any, _context: any): Promise<unknown> => {
+    export const handler = async (event: any, _context: any): Promise<unknown> => {
     
         const myImportantVariable = {
             foo: 'bar'
@@ -301,7 +301,7 @@ If you already have an object containing a `message` key and an additional prope
         logger.info('This is a log with 3 extra objects',
             { data: myImportantVariable },
             { correlationIds: { myCustomCorrelationId: 'foo-bar-baz' } },
-            { lambdaEvent: _event }
+            { lambdaEvent: event }
         );
 
         // Simply pass a string for logging additional data
--- a/docs/core/logger.md
+++ b/docs/core/logger.md
@@ -298,9 +298,10 @@ If you already have an object containing a `message` key and an additional prope
         logger.info('This is a log with an extra variable', { data: myImportantVariable });
         
         // You can also pass multiple parameters
-        logger.info('This is a log with 2 extra variables',
+        logger.info('This is a log with 3 extra objects',
             { data: myImportantVariable },
-            { correlationIds: { myCustomCorrelationId: 'foo-bar-baz' } }
+            { correlationIds: { myCustomCorrelationId: 'foo-bar-baz' } },
+            { lambdaEvent: _event }
         );
 
         // Simply pass a string for logging additional data
--- a/docs/core/logger.md
+++ b/docs/core/logger.md
@@ -273,15 +273,17 @@ You can append additional persistent keys and values in the logs generated durin
 
 !!! tip "Logger will automatically ignore any key with an `undefined` value"
 
-### Appending additional log keys and values to a single log item
+### Appending additional data to a single log item
 
-You can append additional keys and values in a single log item passing them as parameters.
-Pass a string for logging it with default key name `extra`. Alternatively, pass one or multiple objects with custom keys.
-If you already have an object containing a `message` key and an additional property, you can pass this object directly.
+You can append additional data to a single log item by passing objects as additional parameters.
+
+* Pass a simple string for logging it with default key name `extra`
+* Pass one or multiple objects containing arbitrary data to be logged. Each data object should be placed in an enclosing object as a single property value, you can name this property as you need: `{ myData: arbitraryObjectToLog }`
+* If you already have an object containing a `message` key and an additional property, you can pass this object directly
 
 === "handler.ts"
 
-    ```typescript hl_lines="14 18-19 23 31"
+    ```typescript hl_lines="14 18-20 24 32"
     import { Logger } from '@aws-lambda-powertools/logger';
 
     const logger = new Logger();
@@ -292,12 +294,12 @@ If you already have an object containing a `message` key and an additional prope
             foo: 'bar'
         };
         
-        // Pass additional keys and values in single log items
+        // Log additional data in single log items
         
         // As second parameter
         logger.info('This is a log with an extra variable', { data: myImportantVariable });
         
-        // You can also pass multiple parameters
+        // You can also pass multiple parameters containing arbitrary objects
         logger.info('This is a log with 3 extra objects',
             { data: myImportantVariable },
             { correlationIds: { myCustomCorrelationId: 'foo-bar-baz' } },
@@ -323,14 +325,14 @@ If you already have an object containing a `message` key and an additional prope
     ```
 === "Example CloudWatch Logs excerpt"
 
-    ```json hl_lines="7 15-16 24 32"
+    ```json hl_lines="7 15-21 28 37"
     {
         "level": "INFO",
         "message": "This is a log with an extra variable",
         "service": "serverlessAirline",
         "timestamp": "2021-12-12T22:06:17.463Z",
         "xray_trace_id": "abcdef123456abcdef123456abcdef123456",
-        "data": { foo: "bar" }
+        "data": { "foo": "bar" }
     }
     {
         "level": "INFO",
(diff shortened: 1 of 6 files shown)


awsdb:
@dreamorosi Ready for review

dreamorosi (commented):
Looking good, I'm going to run the integration tests on this branch: https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/2347203386

flochaz:
Can you add such object log in the e2e tests as well ?

dreamorosi:
Run with the latest changes:

https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/2351346144

packages/logger/tests/unit/Logger.test.ts
```diff
@@ -280,6 +280,9 @@ describe('Class: Logger', () => {
           });
           const consoleSpy = jest.spyOn(logger['console'], methodOfLogger).mockImplementation();
 
+          interface ArbitraryObject { value: 'CUSTOM' | 'USER_DEFINED' }
+          const arbitraryObject: ArbitraryObject = { value: 'CUSTOM' };
```
saragerion, line 284: Improvement: can use test cases where the value type is not a string but a nested object with different custom types? Example: Lambda events types from the `aws-lambda` library?

docs/core/logger.md
```diff
@@ -298,9 +298,10 @@ If you already have an object containing a `message` key and an additional prope
         logger.info('This is a log with an extra variable', { data: myImportantVariable });
         
         // You can also pass multiple parameters
-        logger.info('This is a log with 2 extra variables',
+        logger.info('This is a log with 3 extra objects',
```
saragerion, line 301: Question: why did you reintroduced the possibility to add > 2 parameters?

packages/logger/src/types/Log.ts
```diff
@@ -9,7 +9,7 @@ type LogLevelThresholds = {
   [key in LogLevel]: number;
 };
 
-type LogAttributeValue = string | number | boolean | null | undefined | LogAttributeValue[] | { [key: string]: LogAttributeValue } | Error;
+type LogAttributeValue = unknown;
```
saragerion, line 12: Question: what was the reason why you removed these more granular types?

packages/logger/tests/e2e/basicFeatures.middy.test.FunctionCode.ts
```diff
@@ -34,6 +36,14 @@ const testFunction = async (event: APIGatewayProxyEvent, context: Context): Prom
     logger.error(ERROR_MSG, e as Error);
   }
 
+  // Test feature 7: Logging an arbitrary object
+  const obj: APIGatewayProxyResult = {
+    body: ARBITRARY_OBJECT_DATA,
```
saragerion, line 41: Improvement: by looking at this PR, the type of the body value will be a string or undefined. 
One of the goals of this PR would be allow for logging of nested objects. I suggest we use that scenario instead.

saragerion (commented):
Thanks for opening this PR! 
I have some questions about the choice of making the types more loose than they were before. If there are some struggle with make the types work, I am happy to help out myself with those and improve the existing types, rather than using `unknown`. 

Also I would like us to cover a more diverse scenarios that include nested objects next to the one were the object value is a string.

dreamorosi:
E2E test is failing ([link](https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/2351346144)), it seems for a missing module (so before running the actual tests).

We need to investigate further.

packages/logger/tests/unit/Logger.test.ts
awsdb: Yes, we can - this is what I implemented in the E2E test. Do we also want this in the unit tests?
BTW: the type of `value` here is not `string`, but a literal type - which was not possible to use prior to this PR.

docs/core/logger.md
awsdb: It was actually always possible, both prior of #614 and after that. Is there a requirement to limit the number to two? In [your comment](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/565#issuecomment-1048639733) in #565 you give an example where providing more than two objects should be possible.

packages/logger/src/types/Log.ts
awsdb: The reason is exactly this PR: relaxing the type restrictions. I could have added `| unknown` at the end of the union type, but that makes little sense IMHO since `unknown` just includes all the others.

packages/logger/tests/unit/Logger.test.ts
saragerion: > Yes, we can - this is what I implemented in the E2E test. Do we also want this in the unit tests?

Yes, thanks.

> BTW: the type of value here is not string, but a literal type - which was not possible to use prior to this PR.

This is what I am referring to: 
const arbitraryObject: ArbitraryObject = { value: 'CUSTOM' };

docs/core/logger.md
saragerion: I see, yes that's how I implemented it originally. No requirement or action needed. 
As I was reviewing this PR I saw this change, which confused me on why this was added for the current scope.

awsdb:
@dreamorosi I haven't changed anything in the modules. In fact, we already had a similar issue with my [first PR](#614). The issue was caused by a faulty commit on `main` I based my branch on. Should I rebase my branch on the latest `main` again?

dreamorosi:
> @dreamorosi I haven't changed anything in the modules. In fact, we already had a similar issue with my [first PR](#614). The issue was caused by a faulty commit on `main` I based my branch on. Should I rebase my branch on the latest `main` again?

Yes, please, let's try to rebase. 
And yes, I see that there are no changes in the modules.

packages/logger/tests/unit/Logger.test.ts
awsdb: Please take a look on the interface definition. The type of the `value` property in the object `arbitraryObject` is not `string`, it's a literal type `'CUSTOM' | 'USER_DEFINED'`. You cannot assign `value: 'SOME_OTHER'` here. And as such, this literal type cannot be expressed by anything else but `unknown` unless we exactly know it.

awsdb:
@dreamorosi I double checked and can confirm that my branch is based on the latest `main`. In my fork, the tests are passing on my branch, please see [the test run](https://github.com/awsdb/aws-lambda-powertools-typescript/actions/runs/2351625881). Could you please check if something is going wrong on checkout/merge?

packages/logger/src/types/Log.ts
dreamorosi: I don't like to be pushy when I can't propose a solution (like now), but is there any way that we can explore using [Generics](https://www.typescriptlang.org/docs/handbook/2/generics.html) in a similar way that we do for the Tracer [here](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/tracer/src/types/Tracer.ts#L32)?

In this case I'm afraid this is a bit above my typing skills so I can't suggest the actual type, but in that case we could at least try restricting types that we never want (i.e. function references, etc.).

Also for transparency, when we talked offline I know I said `unknown` could work. It was because at the time I had no better idea; now I think generics could work(?)

awsdb:
@saragerion Thanks for your comments! I replied directly in the threads. I also updated the tests to operate on complex and nested objects of various types.

To your main question: why `unknown`? I have not found any other way to allow logging of objects typed via arbitrary interfaces.  Since we can't control how users define their interfaces (neither how AWS SDK does this), we would not be able to come up with a type definition that covers all those interfaces. A possible solution would be an empty interface, but it would actually mean `unknown` as well. Researching generics as suggested by @dreamorosi hasn't worked out either. If there's a trick in TypeScript I'm not aware of, please let me know!

`unknown' is very permissive, that is true. However, if we want to allow logging of _arbitrary_ objects as stated in the issue, this is probably the way to go. Users could define some types we haven't thought of.

The [latest E2E test run](https://github.com/awsdb/aws-lambda-powertools-typescript/actions/runs/2357889963) is passing.

saragerion (approved):

Issue #482, opened 2022-01-18
Feature request: feature parity with Python Powertools logger


saragerion:
## Description of the feature request

As addressed in [this comment](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/293#issuecomment-992005781), we should revisit the current features of the Logger to make sure there is feature parity.

Some of the mentioned items are not listed as they are captured by another ticket.

- [x] Logging incoming event functionality
- [x] Custom keys are persisted across warm invocations
- [ ] `location` key in log structure
- [x] Removing log keys (#901)
- [x] Clearing all state (#902)
- [x] ~~Concept of logger name~~
- [x] ~~Correlation ID will be captured in #129 #139~~ 

**Problem statement**

The Logger utility has a number of functionalities shipped in 2021 that are currently not in TS.  

**Summary of the feature**

**For each functionality above:**
- [x] Evaluate if the feature should be ported (does it make sense for the TS ecosystem?) or can't be ported because of language's different behaviour

If it can and should be ported:

- Business logic
- Unit tests
- E2E tests
- Examples
- JsDocs (when applicable)
- Documentation

For this reason it's recommended that whoever picks this issue should create one separated PR for each checklist item.

**Code examples**

N/A

**Benefits for you and the wider AWS community**

Feature parity with the Python Powertool.

**Describe alternatives you've considered**

N/A

**Additional context**

N/A

### Related issues, RFCs
N/A

saragerion:
Note: the location logic will be discussed further and captured in another PR, as agreed with @flochaz and @dreamorosi

PR #1004, opened 2022-06-22
Feat(logger): log event functionality

saragerion:

This PR introduces a new functionality: logging Lambda events. Developers can opt in for this behaviour by one of the following methods:
- passing the parameter `logEvent` set to `true` to the `injectLambdaContext` middleware
- passing the parameter `logEvent` set to `true` to the `injectLambdaContext` decorator
- setting the env var `POWERTOOLS_LOGGER_LOG_EVENT` to through while using the injectLambdaContext middleware/decorator

Note that this functionality is only available when using a middleware or decorator. 

The Lambda invocation will then print one log item with the following key-value pairs:
- `logLevel` INFO
- `message` key with value `Lambda invocation event`
- `event` key with the current Lambda invocation event as value

Logging a Lambda event is very typical use of a logger in AWS Lambda. This logic aims to simplify and reduce the code that developers need to maintain. 

### How to verify this change

See E2E tests results.

To see the docs:

```
npm run docs-website-build-run
```

carbon (4)

### Related issues, RFCs

[#482](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/482)  

### PR status

## Checklist

### Breaking change checklist

---


packages/logger/tests/e2e/basicFeatures.middy.test.FunctionCode.ts
```diff
@@ -33,7 +34,7 @@ const testFunction = async (event: LambdaEvent, context: Context): Promise<{requ
   const specialValue = event.invocation;
   if (specialValue === 0) {
     logger.appendKeys({
-      specialKey: specialValue
+      [PERSISTENT_KEY_FIRST_INVOCATION_ONLY]: specialValue
```
saragerion, line 37: House keeping

docs/core/logger.md
```diff
@@ -14,9 +14,9 @@ Logger provides an opinionated logger with output structured as JSON.
 
 ## Key features
 
-* Capture key fields from Lambda context, cold start and structures logging output as JSON
-* Log Lambda context when instructed (disabled by default)
-* Log sampling prints all logs for a percentage of invocations (disabled by default)
+* Capture key fields from Lambda context, cold start and structure logging output as JSON
+* Log Lambda event when instructed (disabled by default)
```
dreamorosi, line 18: Should this be `event and context`?

saragerion: I think that's captured in the point before, so it would be redundant?

dreamorosi: True, sorry, I was reading only the diff.

Ignore me.

package.json
```diff
@@ -29,8 +29,10 @@
     "lerna-preversion": "lerna exec -- npm run lint",
     "lerna-version": "lerna exec -- npm run format && git add -A src",
     "postversion": "git push && git push --tags",
+    "docs-website-build-run": "npm run docs-buildDockerImage && npm run docs-runLocalDocker",
```
saragerion, line 32: Added this for laziness: 1 command instead of 2 
:>

flochaz (approved):
LGTM

package.json
dreamorosi: That's cool, especially for first-time usage.

If you are a returning contributor and you already have the image built you can just run the second command and you'll still see the changes made to the docs source.

CONTRIBUTING.md
```diff
@@ -238,7 +238,11 @@ You can build and start the API reference website by running these two commands
 
 #### Docs website
 
-You can build and start a local docs website by running these two commands:
+You can build and start a local docs website by running:  
+
+`npm run docs-website-build-run`
```
dreamorosi, line 243: would be cool in the future (not this PR) to have an additional combo command to run the docs & the API docs at the same time

saragerion:
E2E tests:
https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/2549417761

CONTRIBUTING.md
dreamorosi: Sorry, I meant something like (name is just an example):

```
"docs-run-docs-and-api": "npm run docs-runLocalDocker && npm run docs-api-build-run"
```

This command would allow you to run the documentation + the API docs so you can browse between both. Right now when editing locally, if you start from the docs and click on "API Docs" on the left navigation menu it gives you a 404 because there's nothing on the `/api` path.

But again, we can do it at some point in the future!

dreamorosi (approved):
Thanks for this PR, looking forward to merge it!

A couple of minor notes, which I will capture in a separate issue for later:
- I'd like to see a corresponding method for manual mode like `logger.logEvent(event);` - this is purely syntactic sugar as you can easily do `logger.debug('event', { event });` but still.
- Since in the docs we call out that this method is for debugging only, I'd expect the log level to be `DEBUG` instead of `INFO`.
- In the packages/logger/src/middy.ts` docstring of the `injectLambdaContext` middleware we should document the new flag/new behaviour so that it appears also in the API docs.
- In the new sub-section of the docs where we show this method, it'd be nice to have a log excerpt. I was not sure which key the event would use in the JSON log (`event` duh).

PR #1010, opened 2022-06-28
feat(tracer): auto disable when running inside amplify mock

### LINKED ISSUE #1007: Feature request: disable automatically when running inside fn mocked with Amplify CLI

**Problem statement**

Amplify CLI offers a workflow that allows developers to invoke locally a Lambda function during development, this is done by running a command like `amplify mock function <function_name> --event <path_to_event_json>` - [source](https://docs.amplify.aws/cli/usage/mock/#function-mocking-setup).

When not running inside the actual AWS Lambda execution environment the `Tracer` should be disabled by default. This is needed because we cannot know if the X-Ray daemon is running, and in case it's not indeed running the `aws-xray-sdk` for Node.js will throw an error.

We support this behaviour already for local invocations made with SAM CLI, as discussed [here](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/tracer/#testing-your-code), but we don't support this with Amplify CLI.

Ideally, Tracer should be able to detect it's being invoked inside an Amplify-mocked environment and disable itself.

**Summary of the feature**

During initialisation of the `Tracer` instance, check for the presence of an Amplify CLI-specific environment variable (like [this one](https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-util-mock/src/utils/lambda/populate-lambda-mock-env-vars.ts#L46)) and if present, disable tracing for said instance.

**Code examples**

[This is the logic](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/tracer/src/Tracer.ts#L613-L615) that checks for the presence of a SAM CLI-specific environment variable and disables the Tracer if present. The new feature should follow the same exact logic but for Amplify CLI.

Being able to locally mock functions using Amplify CLI and without any code change.

Manually disabling Tracer via environment variable and/or not using Amplify CLI to mock functions that use the tracer.

**Additional context**

N/A

### Related issues, RFCs

N/A

dreamorosi:

As discussed in #1007, the Amplify CLI allows developers to run Lambda functions locally during development. When not running inside the AWS Lambda execution environment the `xray-sdk`, and by extension `Tracer`, will not be able to work properly as there isn't any X-Ray Daemon to connect to and send traces. Unless explicitly accounted for, the default behaviour of the SDK is to throw an exception.

This is a problem as it requires developers using Tracer to make changes to their code only to be able to test locally. Similarly to what we already do with SAM CLI, in this PR I introduced some logic to leverage a special value set by the Amplify CLI in the environment variables ([details here](https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-util-mock/src/utils/lambda/populate-lambda-mock-env-vars.ts#L46)) to detect that Tracer is indeed running inside a process started by the Amplify CLI and automatically disable itself to avoid attempting to send traces and segments.

When merged this PR will close #1007.

### How to verify this change

Check the newly added test cases as well as the result of the e2e tests (link in the comments below).

### Related issues, RFCs

#1007

### PR status

## Checklist

### Breaking change checklist

N/A

---

### DIFF
--- a/docs/core/tracer.md
+++ b/docs/core/tracer.md
@@ -65,7 +65,7 @@ For a **complete list** of supported environment variables, refer to [this secti
 
 #### Example using AWS Serverless Application Model (SAM)
 
-The `Tracer` utility is instantiated outside of the Lambda handler. In doing this, the same instance can be used across multiple invocations inside the same execution environment. This allows `Metrics` to be aware of things like whether or not a given invocation had a cold start or not.
+The `Tracer` utility is instantiated outside of the Lambda handler. In doing this, the same instance can be used across multiple invocations inside the same execution environment. This allows `Tracer` to be aware of things like whether or not a given invocation had a cold start or not.
 
 === "handler.ts"
 
--- a/packages/tracer/src/Tracer.ts
+++ b/packages/tracer/src/Tracer.ts
@@ -597,6 +597,14 @@ class Tracer extends Utility implements TracerInterface {
   private getEnvVarsService(): EnvironmentVariablesService {
     return this.envVarsService;
   }
+
+  /**
+   * Determine if we are running inside an Amplify CLI process.
+   * Used internally during initialization.
+   */
+  private isAmplifyCli(): boolean {
+    return this.getEnvVarsService().getAwsExecutionEnv() === 'AWS_Lambda_amplify-mock';
+  }
   
   /**
    * Determine if we are running in a Lambda execution environment.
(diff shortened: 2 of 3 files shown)


docs/core/tracer.md
```diff
@@ -65,7 +65,7 @@ For a **complete list** of supported environment variables, refer to [this secti
 
 #### Example using AWS Serverless Application Model (SAM)
 
-The `Tracer` utility is instantiated outside of the Lambda handler. In doing this, the same instance can be used across multiple invocations inside the same execution environment. This allows `Metrics` to be aware of things like whether or not a given invocation had a cold start or not.
+The `Tracer` utility is instantiated outside of the Lambda handler. In doing this, the same instance can be used across multiple invocations inside the same execution environment. This allows `Tracer` to be aware of things like whether or not a given invocation had a cold start or not.
```
dreamorosi, line 68: Housekeeping

dreamorosi:
E2E tests passing:
https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/2575872038

flochaz (approved):

PR #1015, opened 2022-07-08
fix(logger): POWERTOOLS_LOGGER_LOG_EVENT precedence is respected

dreamorosi:

This PR aims at fixing a bug that was identified by @saragerion in the logger. In [v0.11.0](https://github.com/awslabs/aws-lambda-powertools-typescript/releases/tag/v0.11.0) we released a feature that allows to log the event parameter of a function handler. This is an opt-in feature that can be enabled by users by either passing a `logEvent` parameter equal to true in the middleware/decorator or by setting a `POWERTOOLS_LOGGER_LOG_EVENT` environment variable also equal to true. In the current iteration of the utility only the first method works.

I was able to reproduce the issue in my own tests.

There were two separate issues that contributed to this bug:

When deciding on whether to log the event or not a function named [`logEventIfEnabled`](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/logger/src/Logger.ts#L318) is called. This function takes in account - and gives precedence to - the parameter passed in the middleware/decorator but also checks the value of the [`logEvent`](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/logger/src/Logger.ts#L122) property that is part of the current `Logger` class. This property is set to `false` by default (as documented). That same property should be set to `true` upon class instantiation and as a part of the [`setOptions`](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/12d4c93011c279d77154338708d94e23b23cd3bf/packages/logger/src/Logger.ts#L701) lifecycle if the `POWERTOOLS_LOGGER_LOG_EVENT` env var is present & with a _truthy_ value, but this never happened.

The second issue, perhaps minor or intended (would appreciate feedback if that was the case) was that both middleware and decorator called the `logEventIfEnabled` function mentioned above only when an `options` parameter is passed to the middleware/decorator. For example:

Even fixing the logic as described previously, the code below still doesn't work because since the `injectLambdaContext` doesn't have a second parameter (`options`), the Logger class never checks if the log event feature was enabled via env var:
```ts
process.env.POWERTOOLS_LOGGER_LOG_EVENT = 'true';

    .use(injectLambdaContext(logger));
```

To make it work, developers would have had to pass an empty object (or an object with other configs like `clearState`):
```ts
process.env.POWERTOOLS_LOGGER_LOG_EVENT = 'true';

    .use(injectLambdaContext(logger, { }));

// OR
    .use(injectLambdaContext(logger, { clearState: true }));
```

The same applies to the decorator usage.

Since we cannot rely on the presence of other configs (i.e. `clearState`) we shouldn't force devs to pass an empty object for the env var config to be evaluated.

### How to verify this change

See additional unit test cases that were added.

### Related issues, RFCs

N/A

### PR status

## Checklist

### Breaking change checklist

N/A

---

### DIFF
--- a/packages/logger/src/Logger.ts
+++ b/packages/logger/src/Logger.ts
@@ -292,8 +292,11 @@ class Logger extends Utility implements ClassThatLogs {
         }
 
         this.addContext(context);
-        const logEvent = options ? options.hasOwnProperty('logEvent') ? options.logEvent : undefined : undefined;
-        this.logEventIfEnabled(event, logEvent);
+        let shouldLogEvent = undefined;
+        if ( options && options.hasOwnProperty('logEvent') ) {
+          shouldLogEvent = options.logEvent;
+        }
+        this.logEventIfEnabled(event, shouldLogEvent);
 
         /* eslint-disable  @typescript-eslint/no-non-null-assertion */
         const result = originalMethod!.apply(target, [ event, context, callback ]);
--- a/packages/logger/src/Logger.ts
+++ b/packages/logger/src/Logger.ts
@@ -637,6 +637,19 @@ class Logger extends Utility implements ClassThatLogs {
   }
 
   /**
+   * If the log event feature is enabled via env variable, it sets a property that tracks whether
+   * the event passed to the Lambda function handler should be logged or not.
+   *
+   * @private
+   * @returns {void}
+   */
+  private setLogEvent(): void {
+    if (this.getEnvVarsService().getLogEvent()) {
+      this.logEvent = true;
+    }
+  }
+
+  /**
    * It sets the log formatter instance, in charge of giving a custom format
    * to the structured logs
    *
@@ -716,7 +729,8 @@ class Logger extends Utility implements ClassThatLogs {
     this.setLogsSampled();
     this.setLogFormatter(logFormatter);
     this.setPowertoolLogData(serviceName, environment);
-
+    this.setLogEvent();
+    
     this.addPersistentLogAttributes(persistentLogAttributes);
 
     return this;
--- a/packages/logger/src/Logger.ts
+++ b/packages/logger/src/Logger.ts
@@ -292,9 +292,8 @@ class Logger extends Utility implements ClassThatLogs {
         }
 
         this.addContext(context);
-        if (options) {
-          this.logEventIfEnabled(event, options.logEvent);
-        }
+        const logEvent = options ? options.hasOwnProperty('logEvent') ? options.logEvent : undefined : undefined;
+        this.logEventIfEnabled(event, logEvent);
 
         /* eslint-disable  @typescript-eslint/no-non-null-assertion */
         const result = originalMethod!.apply(target, [ event, context, callback ]);
--- a/packages/logger/src/middleware/middy.ts
+++ b/packages/logger/src/middleware/middy.ts
@@ -38,8 +38,12 @@ const injectLambdaContext = (target: Logger | Logger[], options?: HandlerOptions
       if (options && options.clearState === true) {
         persistentAttributes.push({ ...logger.getPersistentLogAttributes() });
       }
-      const logEvent = options ? options.hasOwnProperty('logEvent') ? options.logEvent : undefined : undefined;
-      logger.logEventIfEnabled(request.event, logEvent);
+
+      let shouldLogEvent = undefined;
+      if ( options && options.hasOwnProperty('logEvent') ) {
+        shouldLogEvent = options.logEvent;
+      }
+      logger.logEventIfEnabled(request.event, shouldLogEvent);
     });
   };
 
--- a/packages/logger/src/middleware/middy.ts
+++ b/packages/logger/src/middleware/middy.ts
@@ -38,9 +38,8 @@ const injectLambdaContext = (target: Logger | Logger[], options?: HandlerOptions
       if (options && options.clearState === true) {
         persistentAttributes.push({ ...logger.getPersistentLogAttributes() });
       }
-      if (options) {
-        logger.logEventIfEnabled(request.event, options.logEvent);
-      }
+      const logEvent = options ? options.hasOwnProperty('logEvent') ? options.logEvent : undefined : undefined;
+      logger.logEventIfEnabled(request.event, logEvent);
     });
   };
 
--- a/packages/logger/tests/e2e/logEventEnvVarSetting.middy.test.FunctionCode.ts
+++ b/packages/logger/tests/e2e/logEventEnvVarSetting.middy.test.FunctionCode.ts
@@ -1,0 +1,16 @@
+import { injectLambdaContext, Logger } from '../../src';
+import { Context } from 'aws-lambda';
+import middy from '@middy/core';
+
+type LambdaEvent = {
+  invocation: number
+};
+
+const logger = new Logger();
+
+const testFunction = async (event: LambdaEvent, context: Context): Promise<{requestId: string}> => ({
+  requestId: context.awsRequestId,
+});
+
+export const handler = middy(testFunction)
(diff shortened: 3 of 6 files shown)


packages/logger/src/Logger.ts
```diff
@@ -292,9 +292,8 @@ class Logger extends Utility implements ClassThatLogs {
         }
 
         this.addContext(context);
-        if (options) {
-          this.logEventIfEnabled(event, options.logEvent);
-        }
+        const logEvent = options ? options.hasOwnProperty('logEvent') ? options.logEvent : undefined : undefined;
```
flochaz, line 295:
```suggestion
        const logEvent = options?. logEvent;
```

packages/logger/src/middleware/middy.ts
```diff
@@ -38,9 +38,8 @@ const injectLambdaContext = (target: Logger | Logger[], options?: HandlerOptions
       if (options && options.clearState === true) {
         persistentAttributes.push({ ...logger.getPersistentLogAttributes() });
       }
-      if (options) {
-        logger.logEventIfEnabled(request.event, options.logEvent);
-      }
+      const logEvent = options ? options.hasOwnProperty('logEvent') ? options.logEvent : undefined : undefined;
```
flochaz, line 41:
```suggestion
      const logEvent = options?.logEvent;
```

flochaz (commented):
lgtm, minor suggestion , can be ignored

packages/logger/src/middleware/middy.ts
dreamorosi: @flochaz: I haven't purposefully used that notation because we must support Node.js 12.x and that would break it, see this other PR from @ijemmy where he had to remove them all to make the build pass: https://github.com/awslabs/aws-lambda-powertools-typescript/pull/925

flochaz (approved):

packages/logger/src/middleware/middy.ts
bilalq: Is Node 12 support actually needed? Active support for it ended over a year and a half ago and even extended security support for it ended 2 months ago. [It's now officially past EOL](https://endoflife.date/nodejs).

That said, I don't have insight into how many Node12 Lambdas are still running, so maybe supporting existing functions makes it worthwhile.

dreamorosi: Hello @bilalq, yes. 

For the time being we are committed to support Node.js 12.x since AWS Lambda still supports that runtime. 

So far, and aside from these minor changes in some lines, it hasn't been a blocker for us, so we still see value in allowing those customers who might have not migrated to newer and actively maintained runtimes to use the Powertools utilities.

dreamorosi:
@saragerion: a comment on the e2e test I added like you asked.

Before this PR we had 3 stacks for e2e tests & Logger:
- middy (many features being tested, including the `logEvent` via middleware config)
- manual (child logger)
- decorator (sample rate).

Each stack defines one single Lambda function with some env vars and some params in the constructor / middleware.

In order to test this case the function would need the following configs:
- `LOG_LEVEL=INFO` or more verbose
- `logEvent` param not present in the middleware/decorator
- `POWERTOOLS_LOGGER_LOG_EVENT` env var set on the function

I have considered adding this test case to the middleware one but this is not possible because the function passes the `logEvent=true` param in the middleware.

Likewise I have also considered adding it to the decorator one but this is also not compatible as that function/stack sets log level equal to `ERROR` to test sampling and the log event feature logs using `INFO`, so the log would never appear.

In order to add this I had to add a new and dedicated stack + function.

Considering that the feature in itself is already being tested in the e2e (via middleware param) and that this PR introduces unit test cases that specifically target this behaviour (env variable but not middleware/decorator param), are we okay with having a new stack / function that makes the e2e tests longer?

This is not a big issue but I wanted to point this out & document it as it kind of breaks the neat organization of the test cases we had in the logger.

packages/logger/src/Logger.ts
```diff
@@ -292,9 +292,8 @@ class Logger extends Utility implements ClassThatLogs {
...
+        this.logEventIfEnabled(event, logEvent);
```
ijemmy, line 296: (nit) I find this difficult to read. Will this be better?

```
const shouldLogEvent = undefined;
if( options && options.hasOwnProperty('logEvent') ) {
  shouldLogEvent = options.logEvent;
}
this.logEventIfEnabled(event, shouldLogEvent);
```

Or we can wait until we drop Node12 support and simply use `options?.logEvent`?

packages/logger/tests/e2e/logEventEnvVarSetting.middy.test.FunctionCode.ts
```diff
@@ -0,0 +1,16 @@
+import { injectLambdaContext, Logger } from '../../src';
```
ijemmy, line 1: (nit) missing Copyright header?

packages/logger/src/middleware/middy.ts
ijemmy, line 41: (nit) The same as above.

packages/logger/tests/e2e/logEventEnvVarSetting.middy.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    // Invoke the function twice (one for cold start, another for warm start)
+    invocationLogs = await invokeFunction(functionName, 2, 'SEQUENTIAL');
+
+    console.log('logGroupName', logGroupName);
+
+  }, SETUP_TIMEOUT);
+
+  describe('Log event', () => {
+
+    it('should log the event on the first invocation', async () => {
```
ijemmy, line 79: (Question) Why only first invocation? not both of them?

packages/logger/tests/unit/middleware/middy.test.ts
```diff
@@ -153,6 +153,64 @@ describe('Middy middleware', () => {
       });
 
     });
+
+  });
+
+  describe('Feature: clear state', () => {
+
+    test('when enabled, the persistent log attributes added in the handler are removed after the handler\'s code is executed', async () => {
```
ijemmy, line 161: (Questions) should they be removed?  From the code line 195-206, they are not removed but persisted through out handler executions.

packages/logger/tests/unit/middleware/middy.test.ts
```diff
@@ -300,34 +356,69 @@ describe('Middy middleware', () => {
...
-      const persistentAttribsAfterInvocation = { ...logger.getPersistentLogAttributes() };
+      await handler(event, context, () => console.log('Lambda invoked!'));
 
       // Assess
-      expect(persistentAttribs).toEqual({
-        foo: 'bar',
-        biz: 'baz'
-      });
-      expect(persistentAttribsAfterInvocation).toEqual(persistentAttribs);
+      expect(consoleSpy).toBeCalledTimes(1);
```
ijemmy, line 421: Should we also assert here that the event is not really logged in the `consoleSpy`?

ijemmy (approved):

packages/logger/src/Logger.ts
dreamorosi: Changed to your suggestion, only difference is that I made `shouldLogEvent` a `let` instead of `const`

packages/logger/src/middleware/middy.ts
dreamorosi: Changed

packages/logger/tests/unit/middleware/middy.test.ts
dreamorosi: Added an assertion on the log content to verify that the event is not present.

ijemmy (approved):

saragerion (approved):

Issue #991, opened 2022-06-21
Maintenance: package size report failing


dreamorosi:
## Bug description

The newly merged "Package size report" step in our CI/CD is currently failing for all executions, [see example here](https://github.com/awslabs/aws-lambda-powertools-typescript/runs/6975736080?check_suite_focus=true).

### Expected Behavior

Step should be succeeding and report the size.

### Current Behavior

It fails

### Possible Solution

N/A

### Steps to Reproduce

N/A

### Environment

* **Powertools version used**:
* **Packaging format (Layers, npm)**:
* **AWS Lambda function runtime:**
* **Debugging logs:**

### Related issues, RFCs

#878

flochaz:
Dependabot only have Read-only permission (https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/) . Need to investigate what is the best to give only permission to comment (and read https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

dreamorosi:
Okay, so this issue would happen only for dependant's PRs?

For PR that come from real users would it work fine? And what about from forks?

flochaz:
PR from fork won't work neither. 
Changing to `pull_request_target` with an explicit PR checkout would most likely work but is opening vulnerability (https://securitylab.github.com/research/github-actions-preventing-pwn-requests/). Risk can be mitigated with what we already have (need of approval before workflow run for forked PR) or add label check (see https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ for more details on that).

But the recommended approach would be to split in 2 flows : one to build and capture all the actions outputs in a dedicated workflow triggered `on pull_request` and have a chain workflow to add comment with write permission ...

dreamorosi:
Oh I see what you mean, check [this PR that I was working a year ago](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/169/files) that used [data artifacts](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts) from one workflow and then used them to do something.

Thanks for looking into this!

flochaz:
> Oh I see what you mean, check [this PR that I was working a year ago](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/169/files) that used [data artifacts](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts) from one workflow and then used them to do something.
[...]

PR #1026, opened 2022-07-26
fix: wrong scope in captureMethod

ratscrew:
using the captureMethod decorator the "this" is no longer the original obj so changing to the target fixes this

change the binding of the decorated method back to the target so the below code will work:

```
export class Lambda implements LambdaInterface {
   @tracer.captureMethod()
   myMethod(){

      //other method is undefined as "this" is no longer the Lambda class
      this.otherMethod()
   }

   otherMethod(){}
}
```

### Related issues, RFCs

#1028 

### PR status

## Checklist

### Breaking change checklist

---

### DIFF
--- a/packages/tracer/src/Tracer.ts
+++ b/packages/tracer/src/Tracer.ts
@@ -424,7 +424,7 @@ class Tracer extends Utility implements TracerInterface {
         return this.provider.captureAsyncFunc(`### ${originalMethod.name}`, async subsegment => {          
           let result;
           try {
-            result = await originalMethod.apply(this, [...args]);
+            result = await originalMethod.apply(target, [...args]);
             this.addResponseAsMetadata(result, originalMethod.name);
           } catch (error) {
             this.addErrorAsMetadata(error as Error); 
--- a/packages/tracer/src/Tracer.ts
+++ b/packages/tracer/src/Tracer.ts
@@ -415,20 +415,20 @@ class Tracer extends Utility implements TracerInterface {
       // The descriptor.value is the method this decorator decorates, it cannot be undefined.
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
       const originalMethod = descriptor.value!;
-
-      descriptor.value = ((...args: unknown[]) => {
+      
+      descriptor.value = (...args: unknown[]) => {
         if (!this.isTracingEnabled()) {
           return originalMethod.apply(target, [...args]);
         }
 
-        return this.provider.captureAsyncFunc(`### ${originalMethod.name}`, async subsegment => {
-
-          let result: unknown;
+        return this.provider.captureAsyncFunc(`### ${originalMethod.name}`, async subsegment => {          
+          let result;
           try {
-            result = await originalMethod.apply(target, [...args]);
+            result = await originalMethod.apply(this, [...args]);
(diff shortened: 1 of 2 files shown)


dreamorosi:
Hi @ratscrew thank you for taking the time to open a PR to this repo.

Before investing time in reviewing the PR, could you please open an issue to report the bug and provide also reproduction steps / example of this bug? We'd like to understand more about what's happening.

Then, after we have all agreed that it's in fact a bug, please review & edit the PR body and make sure to fill/check all the relevant sections.

ratscrew:
@dreamorosi thanks for your quick response i have added in the bug issue.  

thanks

dreamorosi:
Thank you for opening the issue and providing more info, as I mentioned in the issue I was able to reproduce the bug.

The fix that you are proposing seems like it would fix the issue. I was wondering if you could also please add a couple of tests to make sure it's the case and also to avoid future regressions.

For the unit tests you can use this one as a reference: https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/tracer/tests/unit/Tracer.test.ts#L892-L934 and add one in that same file & test group. Ideally we should have a passing test similar to the one in your issue.

packages/tracer/src/Tracer.ts
```diff
@@ -415,20 +415,20 @@ class Tracer extends Utility implements TracerInterface {
       // The descriptor.value is the method this decorator decorates, it cannot be undefined.
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
       const originalMethod = descriptor.value!;
-      
-      descriptor.value = (...args: unknown[]) => {
+
+      descriptor.value = ((...args: unknown[]) => {
         if (!this.isTracingEnabled()) {
-          return originalMethod.apply(target, [...args]);
+          return originalMethod.apply(target, [ ...args ]);
```
dreamorosi, line 421: This is mostly a styling comment, but could you please revert all the changes to this file except the one that contains the fix? (here I'm referring to the tabs/spaces, extra parenthesis, extra spaces, extra lines.

It's a bit difficult to see the actual diff & these would not be in line with the rest of the codebase.

dreamorosi (changes requested):

dreamorosi:
Thank you @ratscrew, we are really close, just some minor styling comments in one file - thanks for adding the unit test.

Looking forward to merge this!

packages/tracer/src/Tracer.ts
ratscrew: Sorry did not see your comment at first, i have fixed this and re-committed

dreamorosi:
It seems that the linting checks are not passing:

Screenshot 2022-07-28 at 18 24 28

According to the style we are using methods inside classes must be ordered alphabetically by name.

dreamorosi (approved):
Thank you @ratscrew for the contribution and for taking the time to address the review comments, looks good to me.

I'm going to request the review of a second maintainer as we have a 2-approvals model before merging.

Note to the next maintainer: the checks are failing due to #991 but **the tests, lint, etc. are OK**

dreamorosi:
image
image
image


Ran integration tests manually, they are all passing

saragerion (approved):

PR #1031, opened 2022-07-29
chore(build): broke up pr workflow & measure package size

dreamorosi:

This PR proposes changes to the workflow that is run every time a new PR is created & synchronised (aka new commit is pushed). It also spins off in a separate file the workflow that measures the size of the packages as well as changing is trigger from even based to on-demand.

### On PR code update

The workflow that runs every time new code is added to a PR was running completely in sequential mode before this PR, aside from the matrix feature that was used to select the runtime. I have applied the same pattern also to the examples and break up the two sections in two different jobs that can now run in parallel:
image

I also have removed completely all the `lerna`-related commands in favor of the `npm workspaces` related ones. This allows us to select the packages we want to run the commands in. 

Initially I used a more terse version:
```sh
npm run lint -ws
```
but then I moved to a slightly longer, but more explicit one, that will allow us to exclude packages that are work in progress until they are ready, while still being part of the workspace:
```sh
npm run lint -w packages/logger -w packages/tracer -w packages/metrics
```

Each job is also now caching the content of the `node_modules` folders (both main and ones in `examples/*`) between different executions. The cache uses a combination of environment variables, node version, and hash of the `package-lock.json` as key so that if environment, version, or dependencies change the cache is rebuilt.

Here's an [example of cache hit](https://github.com/dreamorosi/aws-lambda-powertools-typescript/actions/runs/2770252600).

### Measure packages size

This action was previously running after the one described in the section above. This meant each new commit was causing the action to run and leave a comment. After several weeks some of us (me included) have felt like this was generating a bit too much noise in the comments section.

This PR moves the action in its own workflow and allows maintainers to run it on demand by specifying a PR number. The action will use the PR number to get the info related to `head` and `base` and then, after having computed the changes and sizes, it will leave a comment only when manually triggered.

This PR is dependant on [this other PR](https://github.com/flochaz/pkg-size-action/pull/1) made in the action's repo.

Like in the previous workflow, I have also removed the `lerna` related commands and used the `npm workspaces` ones. This has the added benefit of avoiding running unnecessary commands in the `examples/*` folders.

Screenshot 2022-07-29 at 17 06 33

Finally, this new version of the action and workflow, once merged **will allow to safely run the checks also for PRs that come from external contributors**.

### How to verify this change

See example of the `On PR code update` workflow run in a fork here:
https://github.com/dreamorosi/aws-lambda-powertools-typescript/actions/runs/2724186137

Also see an example of the comment left on a PR opened by an external user on a fork (of a fork):
https://github.com/dreamorosi/aws-lambda-powertools-typescript/pull/3

### Related issues, RFCs

**Issue number:** #991 

### PR status

## Checklist

### Breaking change checklist

N/A

---

### DIFF
--- a/.github/workflows/measure-packages-size.yml
+++ b/.github/workflows/measure-packages-size.yml
@@ -31,7 +31,7 @@ jobs:
         with:
           ref: ${{ steps.extract_PR_details.outputs.headSHA }}
       - name: Packages size report
-        uses: flochaz/pkg-size-action@v1.2.15
+        uses: flochaz/pkg-size-action@v2.0.0
         with:
           build-command: mkdir dist && npm run package -w packages/logger -w packages/tracer -w packages/metrics -w packages/commons && npm run package-bundle -w packages/logger -w packages/tracer -w packages/metrics -w packages/commons && bash -c "mv ./packages/*/dist/* dist/" && ls dist
           dist-directory: /dist
--- a/.github/workflows/measure-packages-size.yml
+++ b/.github/workflows/measure-packages-size.yml
@@ -14,12 +14,26 @@ jobs:
       NODE_ENV: dev
       PR_NUMBER: ${{ inputs.prNumber }}
     steps:
+      # Since we are manually triggering the workflow the previous checkout has the main branch. In order to checkout the branch/code of the PR
+      # we need first to use the PR number to retrieve the PR SHA number. This means we need three steps to: checkout the repo,
+      # run a custom script to get the SHA, and then finally checkout the PR branch
       - name: Checkout Repo
         uses: actions/checkout@v3
+      - name: Extract PR details
+        id: extract_PR_details
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const script = require('.github/scripts/get_pr_info.js');
+            await script({github, context, core});
+      - name: Checkout PR code
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ steps.extract_PR_details.outputs.headSHA }}
       - name: Packages size report
         uses: flochaz/pkg-size-action@v1.2.15
         with:
-          build-command: mkdir dist && npm run package -w packages/logger -w packages/tracer -w packages/metrics && npm run package-bundle -w packages/logger -w packages/tracer -w packages/metrics && bash -c "mv ./packages/*/dist/* dist/" && ls dist
+          build-command: mkdir dist && npm run package -w packages/logger -w packages/tracer -w packages/metrics -w packages/commons && npm run package-bundle -w packages/logger -w packages/tracer -w packages/metrics -w packages/commons && bash -c "mv ./packages/*/dist/* dist/" && ls dist
           dist-directory: /dist
           pr-number: ${{ inputs.prNumber }}
         env:
--- a/.github/workflows/measure-packages-size.yml
+++ b/.github/workflows/measure-packages-size.yml
@@ -1,0 +1,26 @@
+name: Measure packages size
+
+on:
+  workflow_dispatch:
+    inputs:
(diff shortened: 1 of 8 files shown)


examples/sam/package.json
```diff
@@ -18,6 +18,7 @@
   },
   "devDependencies": {
     "@types/aws-lambda": "^8.10.86",
+    "@types/jest": "^27.5.2",
```
dreamorosi, line 21: This change is needed because we are now running the tests outside of the `lerna` scope. For this reason the dependency (which is present in both the `examples/cdk` and the main package) is also needed here now.

.github/workflows/measure-packages-size.yml
```diff
@@ -0,0 +1,26 @@
...
+  measure-utils-sizes:
+    runs-on: ubuntu-latest
+    env:
+      NODE_ENV: dev
+      PR_NUMBER: ${{ inputs.prNumber }}
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v3
+      - name: Packages size report
+        uses: flochaz/pkg-size-action@v1.2.15
```
dreamorosi, line 20: TODO: update this version once https://github.com/flochaz/pkg-size-action/pull/1 is merged & a new version has been released

flochaz (dismissed):

.github/workflows/pr_lint_and_test.yml
```diff
@@ -12,63 +13,74 @@ jobs:
...
+        env:
+          cache-name: cache-node-modules
+        with:
+          path: "./node_modules"
+          # Use the combo between node version, name, and SHA-256 hash of the lock file as cache key so that
+          # if one of them changes the cache is invalidated/discarded
+          key: ${{ matrix.version }}-build-${{ env.cache-name }}-${{ hashFiles('./package-lock.json') }}
+      - name: Install dependencies
+        # We can skip the install if there was a cache hit
+        if: steps.cache-node-modules.outputs.cache-hit != 'true'
```
ijemmy, line 37: ❤️ this.

ijemmy: Question:  I understand that `env.cache-name` will always be the same right?  Any advantage of introducing this env var (and put it in the key)?

.github/workflows/pr_lint_and_test.yml
```diff
@@ -12,63 +13,74 @@ jobs:
...
-          build-command: mkdir dist && npm run lerna-package && npm run lerna-package-bundle && bash -c "mv ./packages/*/dist/* dist/" && ls dist
-          dist-directory: /dist
+          node-version: 16
+          cache: "npm"
+      - name: Cache node modules
+        id: cache-node-modules
+        uses: actions/cache@v3
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          cache-name: cache-node-modules
```
ijemmy, line 75: The same question as above. What's the use of this env var?

ijemmy (approved):
Got a question around environment variable for cache key.

Apart from that, everything looks good to me.

.github/workflows/pr_lint_and_test.yml
dreamorosi: You are right, it won't change ever. 

I have lifted this from a documentation example and they were using it but now that you mention it I agree it's redundant.

Happy to make another commit and remove it

.github/workflows/pr_lint_and_test.yml
dreamorosi: Removed it in the latest commit, thanks for spotting it!

ijemmy (approved):
LGTM

flochaz (approved):

PR #1045, opened 2022-08-09
fix(logger): fix clearstate bug when lambda handler throws

saragerion:
### Related issues, RFCs

**Issue number:** 

### PR status

## Checklist

- [ ] [My changes meet the tenets criteria](https://awslabs.github.io/aws-lambda-powertools-python/#tenets)
- [ ] The *code coverage* hasn't decreased
- [ ] The PR title follows the [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/semantic.yml#L2)

### Breaking change checklist

---

### DIFF
--- a/packages/logger/src/Logger.ts
+++ b/packages/logger/src/Logger.ts
@@ -275,8 +275,6 @@ class Logger extends Utility implements ClassThatLogs {
    * @returns {HandlerMethodDecorator}
    */
   public injectLambdaContext(options?: HandlerOptions): HandlerMethodDecorator {
-    const isClearStateEnabled = options && options.clearState === true;
-
     return (target, _propertyKey, descriptor) => {
       /**
        * The descriptor.value is the method this decorator decorates, it cannot be undefined.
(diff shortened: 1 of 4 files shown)


saragerion:
E2E tests:
https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/2828668395

packages/logger/tests/unit/Logger.test.ts
```diff
@@ -982,6 +982,55 @@ describe('Class: Logger', () => {
...
+          throw new Error('Unexpected error occurred!');
+        }
+      }
+
+      const persistentAttribs = { ...logger.getPersistentLogAttributes() };
+
+      // Act
+      try {
+        await new LambdaFunction().handler({ user_id: '123456' }, dummyContext, () => console.log('Lambda invoked!'));
+      } catch (error) {
```
dreamorosi, line 1020: This is likely a minor thing, but I think we should consider asserting on the fact that the function actually threw an error.

Written this way, if the handler doesn't throw but the persistent attribs are OK the test would still pass I think.

You could wrap it with an `expect( () => ... ).toThrow(..)` so that if the handler doesn't throw it'll make the test fail.

This extra assertion would help us detect regressions around us catching error that we didn't mean to, like in #1016

dreamorosi (dismissed):
Thank you for the fix! Left one comment in the unit tests.

packages/logger/tests/unit/middleware/middy.test.ts
```diff
@@ -10,6 +10,7 @@ import { Logger } from './../../../src';
 import middy from '@middy/core';
 import { PowertoolLogFormatter } from '../../../src/formatter';
 import { Console } from 'console';
+import { context as dummyContext } from '../../../../../tests/resources/contexts/hello-world';
```
dreamorosi, line 13: I think that unless you're doing something specific with the context in each test case you could use this one instead of redeclaring it in each test, alternatively just remove the line.

dreamorosi (approved):

dreamorosi (approved):
LGTM, thanks for addressing the comments!

packages/logger/src/Logger.ts
```diff
@@ -286,30 +284,43 @@ class Logger extends Utility implements ClassThatLogs {
...
-          initialPersistentAttributes = { ...this.getPersistentLogAttributes() };
-        }
-
-        this.addContext(context);
-        let shouldLogEvent = undefined;
-        if ( options && options.hasOwnProperty('logEvent') ) {
-          shouldLogEvent = options.logEvent;
-        }
-        this.logEventIfEnabled(event, shouldLogEvent);
+        const initialPersistentAttributes = Logger.injectLambdaContextBefore(this, event, context, options);
```
ijemmy, line 287: (optional) I find it hard to understand why `injectLambdaContextBefore()` return its initial state. It seems to do two things at once. 

Would it be easier to understand if we move the return logic out of the function? Something like this.

```suggestion
let initialPersistentAttributes = {};
if (options && options.clearState === true) {
  initialPersistentAttributes = logger.getPersistentLogAttributes();
}

Logger.injectLambdaContextBefore(/*...*/); // This method return nothing. 
```

packages/logger/src/middleware/middy.ts
```diff
@@ -30,34 +30,29 @@ import { HandlerOptions, LogAttributes } from '../types';
 const injectLambdaContext = (target: Logger | Logger[], options?: HandlerOptions): middy.MiddlewareObj => {
 
   const loggers = target instanceof Array ? target : [target];
-  const persistentAttributes: LogAttributes[] = [];
+  const persistentAttributes: (LogAttributes | undefined)[] = [];
 
   const injectLambdaContextBefore = async (request: middy.Request): Promise<void> => {
     loggers.forEach((logger: Logger) => {
-      logger.addContext(request.context);
+      const initialPersistentAttributes = Logger.injectLambdaContextBefore(logger, request.event, request.context, options);
```
ijemmy, line 37: (Optional) the same as above.

Q: I assume that decorator + manual has already been covered by the change in `Logger.ts`?

ijemmy (approved):
LGTM. 

I have one suggestion but it's subjective opinion on readability.

packages/logger/src/middleware/middy.ts
saragerion: Yep, this is the reason i created the static methods injectLambdaContextBefore and injectLambdaContextAfterAndOnError which I use with middy and the decorator: because I wanted to make sure the behaviour is the same across different instrumentation strategies.

saragerion:
E2E tests passing: https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/2832787138

dreamorosi (approved):

ijemmy (approved):

Issue #1046, opened 2022-08-10
Maintenance: e2e tests failing


dreamorosi:
## Bug description

E2E tests are failing after the changes related to #826.

There’s a dependency issue in the [job section that tests the layers](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/workflows/run-e2e-tests.yml#L74-L102), and as far as I can tell it’s because it’s requiring something from commons but we are installing the dependencies only in `[layer-publisher](https://github.com/awslabs/aws-lambda-powertools-typescript/tree/main/layer-publisher)`.

### Expected Behavior

E2E tests passing.

### Current Behavior

See [here](https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/2832997617).

### Possible Solution

I remember advocating for this usage during review, now I'm wondering if we should use the published version or the local one. If the local one then we'll have to install `commons` in that job.

### Steps to Reproduce

N/A

### Environment

* **Powertools version used**: `main`
* **Packaging format (Layers, npm)**: Layers
* **AWS Lambda function runtime:** N/A
* **Debugging logs:** N/A

### Related issues, RFCs

#826

Issue #826, opened 2022-05-02
Feature request: AWS Lambda Layers support


saragerion:
## Description of the feature request

**Problem statement**

Currently AWS Lambda Powertools for Python supports usage via a Lambda Layer. The TypeScript Powertools do not. 

**Summary of the feature**

[Lambda Layer](https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html) is a .zip file archive that can contain additional code, pre-packaged dependencies, data, or configuration files. Layers promote code sharing and separation of responsibilities so that you can iterate faster on writing business logic.

This issue would include Lambda Powertools Lambda Layer using [AWS Lambda Console](https://docs.aws.amazon.com/lambda/latest/dg/invocation-layers.html#invocation-layers-using), or your preferred deployment framework.

When using Layers, you can add Lambda Powertools as a dev dependency (or as part of your virtual env) to not impact the development process.

Additionally, Lambda Layers allow us to understand who uses this library in a non-intrusive way. This helps us justify and gain future investments for other Lambda Powertools languages.

**Code examples**

https://awslabs.github.io/aws-lambda-powertools-python/latest/#install

**Benefits for you and the wider AWS community**

A different way to install and leverage the library.

**Describe alternatives you've considered**

N/A

**Additional context**

N/A

### Related issues, RFCs

N/A

orozcoadrian:
Lambda layers can be architecture-specific, if they contain native code. Assuming there's no native code in this library, it will be compatible with both `x86` and `arm64`. It'd be nice to make that explicit in the documentation.

Reference:
https://aws.amazon.com/blogs/compute/migrating-aws-lambda-functions-to-arm-based-aws-graviton2-processors/#:~:text=Functions%20without%20architecture%2Dspecific%20dependencies%20or%20binaries

For illustration, here's a lambda layer that only works with `x86`, but there's no mention of it in the docs: https://serverlessrepo.aws.amazon.com/applications/us-east-1/145266761615/image-magick-lambda-layer

idm-ryou:
Thank you for your great work. 
It would be helpful to have a table on the website that maps the npm package version to the Lambda Layer version.

dreamorosi:
> Thank you for your great work.
> It would be helpful to have a table in the documentation that maps the npm package version to the Lambda Layer version.

Hi @idm-ryou, thanks for the feedback. While looking at the docs you should be able to see the package version at the top of the page:
[...]

PR #1047, opened 2022-08-10
Fix(build): e2e workflow failing

dreamorosi:

As described in #1046 there is a small bug in the newly introduced layer job that is now part (#826) of the e2e tests.

The issue was caused by the fact that the e2e tests for this part of the codebase required code in the `packages/commons` folder. The dependencies for that packages were not being installed causing the run to fail.

After fixing the issue I then found out two more issues:
- The workflow uses the matrix strategy to run the e2e tests on multiple NodeJS versions. There was a resource in the tested stack (an SSM Param) for which the name was not unique making 2/3 versions of the matrix fail automatically. I have added a suffix with the NodeJS version being tested to make the name unique.
- The workflow deploys a Lambda function that is used to test the correctness of the Lambda Layer being tested. That function imports the Powertools utilities. Since the utils are expected to be in the layer, they are not installed (`npm i`) as part of the setup process. This was however causing the `esbuild` process to fail because they were not marked as [external](https://docs.aws.amazon.com/cdk/api/v1/docs/aws-lambda-nodejs-readme.html#externals). Marking a dependency as external tells CDK & esbuild to not bundle said dependency since it's expected to be present in the Lambda environment.

This PR aims at fixing all three issues mentioned above.

### How to verify this change

See this cool badge for the e2e test results for this branch:
[run-e2e-tests](https://github.com/awslabs/aws-lambda-powertools-typescript/actions/workflows/run-e2e-tests.yml)

### Related issues, RFCs

**Issue number:** #1046 

### PR status

## Checklist

### Breaking change checklist

N/A

---

### DIFF
--- a/.github/workflows/run-e2e-tests.yml
+++ b/.github/workflows/run-e2e-tests.yml
@@ -102,7 +102,7 @@ jobs:
           path: "./node_modules"
           # Use the combo between node version, name, and SHA-256 hash of the lock file as cache key so that
           # if one of them changes the cache is invalidated/discarded
-          key: 16-cache-utils-node-modules-${{ hashFiles('./package-lock.json') }}
+          key: ${{ matrix.version }}-cache-utils-node-modules-${{ hashFiles('./package-lock.json') }}
       - name: Install dependencies
         # We can skip the install if there was a cache hit
         if: steps.cache-node-modules.outputs.cache-hit != 'true'
--- a/.github/workflows/run-e2e-tests.yml
+++ b/.github/workflows/run-e2e-tests.yml
@@ -95,6 +95,26 @@ jobs:
         with:
           role-to-assume: ${{ secrets.AWS_ROLE_ARN_TO_ASSUME }}
           aws-region: eu-west-1
+      - name: Cache node modules for commons
+        id: cache-node-modules
+        uses: actions/cache@v3
+        with:
+          path: "./node_modules"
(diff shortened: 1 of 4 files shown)


flochaz (approved):
YOU ROCK!

.github/workflows/run-e2e-tests.yml
```diff
@@ -81,20 +81,40 @@ jobs:
       matrix:
         version: [12, 14, 16]
     steps:
-      - name: "Checkout"
+      - name: Checkout
         uses: actions/checkout@v3
-      - name: "Use NodeJS 14"
+      - name: Use NodeJS
```
saragerion, line 86: Thanks for the house keeping!

.github/workflows/run-e2e-tests.yml
```diff
@@ -81,20 +81,40 @@ jobs:
...
           role-to-assume: ${{ secrets.AWS_ROLE_ARN_TO_ASSUME }}
           aws-region: eu-west-1
+      - name: Cache node modules for commons
+        id: cache-node-modules
+        uses: actions/cache@v3
+        with:
+          path: "./node_modules"
+          # Use the combo between node version, name, and SHA-256 hash of the lock file as cache key so that
+          # if one of them changes the cache is invalidated/discarded
+          key: ${{ matrix.version }}-cache-utils-node-modules-${{ hashFiles('./package-lock.json') }}
```
saragerion, line 105: Nice

saragerion (approved):

Issue #1038, opened 2022-08-03
Docs: document CI/CD behavior / implementation


dreamorosi:
## Description of the improvement

**Summary of the proposal**
With the number of `workflows` and `scripts` growing in our CI/CD processes (based on GitHub Actions) there's a need of documenting the relationships between each workflow/script and also the different triggers.

The proposal is to create a `.github/workflows/README.md` file and have a short explanation + diagram of each major workflow, similar to what done in [this comment](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/1023#issue-1315269180) of #1023.

**How, where did you look for information**
N/A

**Missing or unclear documentation**
`.github/workflows/README.md`

**Improvement**
Creating the doc would help all existing maintainers to have context on how the workflows. This would also be useful to onboard additional maintainers / collaborators in the future.

### Related existing documentation

N/A

### Related issues, RFCs

#1023

dreamorosi:
Most of the automation has now been documented in the Processes section of the documentation as well as the [Maintainers handbook](https://docs.powertools.aws.dev/lambda/typescript/latest/maintainers/). Additionally, each workflow has now a mini documentation at the top of the file that explains how it works and its purpose.

We also have a couple dedicated issues that should be addressed during the next few iterations:
- #2044
- #1981

With this in mind, I'll be closing this issue as completed.

PR #1052, opened 2022-08-11
fix(layers): release process + remove duplicate code

dreamorosi:

This PR tackles two separate but related issues: #1047 and #1053. All changes are related to the workflows that run as a result of a PR, a merge, or a release.

In an effort to try to minimise issues related to workflows this PR tries to remove duplicate code across workflows by creating a reusable and separate workflow around two flows:
- Running unit tests on all utils, examples, and layer-publisher
- Publishing docs (both main docs & API) 

These generalised workflows can now be called by another workflow so that the amount of repeated code is reduced.

For example, before this change, the same code for running the unit tests was repeated in the workflows that run:
- On merge
- On a PR update
- On a release

Every time a change was needed or a new directory was added us maintainers had to go in each file and replicate the changes. This was needlessly time consuming and error prone, especially this last point is proved by the uptick of issues generated by the introduction of a new folder (`layer-publisher`).

With this change, the logic needed to run the tests has been extracted in a separate workflow that is now found at `.github/workflows/reusable-run-unit-tests.yml`. This workflow can be called by another workflow this way:
```yml
run-unit-tests:
    uses: ./.github/workflows/reusable-run-unit-tests.yml
```
This way, the three workflows mentioned above will have only these two lines, and any change required to the logic will happen in a single place.

The steps needed to publish the docs also followed a similar pattern. The logic was replicated across two workflows that run:
- On merge
- On a release

In this case, the steps had slight differences related to the fact that we want to publish the docs under two different paths based on the stage. 

This PR creates a reusable workflow called `.github/workflows/reusable-publish-docs.yml` that accepts some inputs that will help the workflow to publish on the right path based on the values passed.

Note: this PR also sneaks in a minor wording change in the `PR_TEMPLATE` file that was causing some workflows that are in charge of labelling issues to fail.

### How to verify this change

I have ran the various the workflows in a fork but **please note that some of them fail at later stages because the token needed to publish to NPM is not present in the fork**.

Example of "Make Release" workflow:
image

Example of "On PR Merge" workflow - when a PR is still open:
image

Example of "On PR Merge" workflow - when a PR is actually merged: 
image

To see an example of the workflow that runs when a PR is updated, see the checks below.

### Related issues, RFCs

**Issue number:** #1047 
**Issue number:** #1053

### PR status

## Checklist

### Breaking change checklist

N/A

---


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -21,8 +21,8 @@
 
 ### Related issues, RFCs
 
-<!--- Add here the number to the Github Issue or RFC that is related to this PR. -->
-<!-- **Issue number:** #123 -->
```
dreamorosi, line 25: This line was interpreted by the workflow as an actual issue number. The workflow was then trying to assign a label to it but since it doesn't exist it would fail.

flochaz (approved):
Love it !

.github/workflows/reusable-publish-docs.yml
```diff
@@ -0,0 +1,104 @@
...
+    secrets:
+      token:
+        required: true
+concurrency:
+  group: on-release-publish
+
+jobs:
+  publish-docs:
+    # see https://github.com/awslabs/aws-lambda-powertools-python/issues/1349
+    if: inputs.workflow_origin == 'dreamorosi/aws-lambda-powertools-typescript'
```
ijemmy, line 26: Forget to update this to `awslabs`?

ijemmy (changes requested):
Please double check the value in `inputs.workflow_origin`

(Optional)
Would it be possible to put reusable workflows in a `reusable` directory?

It might also worth going through this in our maintainer sync so everyone is aware, how we structure them + naming convention.

.github/workflows/reusable-publish-docs.yml
dreamorosi:
```suggestion
    if: inputs.workflow_origin == 'awslabs/aws-lambda-powertools-typescript'
```

dreamorosi:
> It might also worth going through this in our maintainer sync so everyone is aware, how we structure them + naming convention.

Yes, I'm still planning to document all this in written form #1038

dreamorosi:
> (Optional) Would it be possible to put reusable workflows in a `reusable` directory?

I'd need to test this (and I will), but according to [the docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#about-yaml-syntax-for-workflows) (sentence after first paragraph - doesn't have a `/*`) & [this SO answer](https://stackoverflow.com/a/62289855/2304421), GitHub expects the workflow to be only in the `.github/workflows` directory and subdirectories might not work.

ijemmy (approved):

dreamorosi:
Merging this as Florian already approved the previous iteration and the only difference is a word in the repo name of a workflow

dreamorosi:
@ijemmy I have tested your suggestion in a fork, by defining a dummy workflow at top level that calls another in a subfolder:
image
it throws this error when pushed preventing to start the workshop:
image

PR #1055, opened 2022-08-12
fix(tracer): use instance as `this` in capture decorators

misterjoshua:
## Description of your changes

This PR fixes bugs in the `tracer.captureMethod()` and `tracer.captureLambdaHandler()` decorators. These decorators provide the class **prototype** as the `this` scope and not the **instance** of the class. While the prototype scope allows member functions to be run, it does not allow the customer to access *instance* member variables, as these instance member variables are not part of the class prototype.

Shout out to @dreamorosi for assistance in building an acceptable proof of concept in Slack.

### How to verify this change

I have added two new unit tests which reproduce the bug. You may run them to verify the change.

### Related issues, RFCs

This is related to #1028.  #1026 previously attempted to fix this and got it us part of the way to a fix.

**Issue number:** #1056

### PR status

***Is this ready for review?:*** YES
***Is it a breaking change?:*** NO

## Checklist

- [x] [My changes meet the tenets criteria](https://awslabs.github.io/aws-lambda-powertools-python/#tenets)
- [x] I have performed a *self-review* of my own code
- [x] I have *commented* my code where necessary, particularly in areas that should be flagged with a TODO, or hard-to-understand areas
- [ ] I have made corresponding changes to the *documentation*
- [ ] I have made corresponding changes to the *examples*
- [x] My changes generate *no new warnings*
- [x] The *code coverage* hasn't decreased
- [x] I have *added tests* that prove my change is effective and works
- [x] New and existing *unit tests pass* locally and in Github Actions
- [ ] Any *dependent changes have been merged and published* in downstream module
- [x] The PR title follows the [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/semantic.yml#L2)

### Breaking change checklist

- [ ] I have documented the migration process
- [ ] I have added, implemented necessary warnings (if it can live side by side)

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1059, opened 2022-08-14
Fix(metrics): decorated class methods cannot access `this`

### LINKED ISSUE #1057: Bug: decorated class methods cannot access `this` in Metrics
## Bug description

A long form description of this issue is available in #1056. This is an issue that affects all three Powertools utilities because they all share similar implementation for the decorators.

---

When using Powertools in with a Class and when decorating a method, the decorated method does not have access to the correct value of `this` and so it's not able to call other methods nor read attributes from its same class.

This is due to the fact that the decorator implementation uses an `() => ()` arrow function instead of a regular `function () {}` one. This mixes the value of `this` which then causes the wrong value to be passed.

Additionally there's a missing detail in the documentation of the Class instrumentation that prevents compounds the issue described above.

### Expected Behavior

Decorated methods have access to the correct value of `this` and is able to call methods or access attributes/properties.

### Current Behavior

See "Steps to Reproduce" section.

### Possible Solution

Apply the same changes done to Tracer in #1055 

### Steps to Reproduce

```ts
import {
  LambdaInterface,
  SyncHandler,
  AsyncHandler,
} from "@aws-lambda-powertools/commons";
import { Metrics } from "@aws-lambda-powertools/metrics";

const metrics = new Metrics({ serviceName: "serverlessAirline" });

  greeting: string;

  constructor(message: string) {
    this.greeting = message;
  }

  public getGreeting(): string {
    /* ... */
    return `Hello ${this.greeting}`;
  }

  @metrics.logMetrics({ captureColdStartMetric: true })
  public handler(_event: any, _context: any, _callback: any): void {
    console.log(this.getGreeting());
  }
}

export const myFunction = new Lambda("World");
```

The value logged from `console.log(this.getGreeting());` should be `Hello World` but instead is `Hello undefined`:
Screenshot 2022-08-13 at 22 27 25

### Environment

* **Powertools version used**: 1.1.0
* **Packaging format (Layers, npm)**: all
* **AWS Lambda function runtime:** all
* **Debugging logs:** N/A

### Related issues, RFCs

#1056

dreamorosi:

As described in #1057 the decorator implementation was affected by an issue that prevented the decorated method/class to access other class members as the value of `this` was passed incorrectly.

This PR aims at fixing the issue, creating an unit test case, making minimal changes to the function in the e2e test to verify the change, and update docstrings/docs.

A long form explanation of the issue can be found at #1055, when merged this PR will close #1057. 

### How to verify this change

See checks below the PR, also see the results of the [e2e test run](https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/2855116627).

### Related issues, RFCs

**Issue number:** #1057 

### PR status

## Checklist

### Breaking change checklist

N/A

---

### DIFF
--- a/docs/core/metrics.md
+++ b/docs/core/metrics.md
@@ -277,7 +277,7 @@ You can add default dimensions to your metrics by passing them as parameters in 
         }
     }
 
-    export const handlerClass = new Lambda();
+    const handlerClass = new Lambda();
     export const handler = handlerClass.handler.bind(handlerClass); // (1)
     ```
 
@@ -375,7 +375,7 @@ The `logMetrics` decorator of the metrics utility can be used when your Lambda h
         }
     }
 
-    export const handlerClass = new Lambda();
+    const handlerClass = new Lambda();
     export const handler = handlerClass.handler.bind(handlerClass); // (1)
     ```
 
@@ -639,7 +639,7 @@ CloudWatch EMF uses the same dimensions across all your metrics. Use `singleMetr
         }
     }
 
-    export const handlerClass = new Lambda();
+    const handlerClass = new Lambda();
     export const handler = handlerClass.handler.bind(handlerClass); // (1)
     ```
 
--- a/docs/core/metrics.md
+++ b/docs/core/metrics.md
@@ -269,14 +269,19 @@ You can add default dimensions to your metrics by passing them as parameters in 
     const metrics = new Metrics({ namespace: 'serverlessAirline', serviceName: 'orders' });
     const DEFAULT_DIMENSIONS = { 'environment': 'prod', 'foo': 'bar' };
 
-    export class MyFunction implements LambdaInterface {
+    export class Lambda implements LambdaInterface {
         // Decorate your handler class method
         @metrics.logMetrics({ defaultDimensions: DEFAULT_DIMENSIONS })
         public async handler(_event: any, _context: any): Promise<void> {
             metrics.addMetric('successfulBooking', MetricUnits.Count, 1);
         }
     }
+
+    export const handlerClass = new Lambda();
+    export const handler = handlerClass.handler.bind(handlerClass); // (1)
     ```
+
+    1. Binding your handler method allows your handler to access `this` within the class methods.
 
 If you'd like to remove them at some point, you can use the `clearDefaultDimensions` method.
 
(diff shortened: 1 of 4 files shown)


dreamorosi:
Note to reviewers: don't merge this until #1054 is merged.

packages/metrics/tests/e2e/basicFeatures.decorator.test.functionCode.ts
```diff
@@ -1,6 +1,6 @@
 import { Metrics, MetricUnits } from '../../src';
 import { Context } from 'aws-lambda';
-import { LambdaInterface } from '../../examples/utils/lambda/LambdaInterface';
+import { LambdaInterface } from '@aws-lambda-powertools/commons';
```
ijemmy, line 3: Won't this use the published version instead of local version?

ijemmy (changes requested):
Please check the comment.  Want to double check if that is intentional.

Also, similar to #1055, we should also update the `examples/cdk/src/*.Decorator.ts` files. They still don't use .bind() method. This could be done in another PR. But I prefer to have it now or we will forget about this.

packages/metrics/tests/e2e/basicFeatures.decorator.test.functionCode.ts
dreamorosi: As far as I understand, given that we are inside the NPM workspace & it uses the only lock file that is present at the root, then no, it should get the local one (see screenshot below):

dreamorosi: image

ijemmy (approved):

docs/core/metrics.md
```diff
@@ -362,15 +367,20 @@ The `logMetrics` decorator of the metrics utility can be used when your Lambda h
...
-    export class MyFunction implements LambdaInterface {
+    class Lambda implements LambdaInterface {
 
         @metrics.logMetrics()
         public async handler(_event: any, _context: any): Promise<void> {
             metrics.addMetric('successfulBooking', MetricUnits.Count, 1);
         }
     }
+
+    export const handlerClass = new Lambda();
```
saragerion, line 378: Same as above

saragerion: https://github.com/awslabs/aws-lambda-powertools-typescript/pull/1059/files#r947884023

docs/core/metrics.md
```diff
@@ -629,6 +639,8 @@ CloudWatch EMF uses the same dimensions across all your metrics. Use `singleMetr
         }
     }
 
-    export const myFunction = new Lambda();
-    export const handler = myFunction.handler;
+    export const handlerClass = new Lambda();
```
saragerion, line 642: Same as above: https://github.com/awslabs/aws-lambda-powertools-typescript/pull/1059/files#r947884023

packages/metrics/tests/unit/Metrics.test.ts
```diff
@@ -619,6 +619,33 @@ describe('Class: Metrics', () => {
...
+        }
+
+        private dummyMethod(): void {
+          metrics.addMetric('test_name', MetricUnits.Seconds, 1);
+        }
+      }
+      await new LambdaFunction().handler(dummyEvent, dummyContext.helloworldContext, () => console.log('Lambda invoked!'));
+
+      // Assess
+      expect(console.log).toBeCalledTimes(1);
```
saragerion, line 647: Can you check that it's the metric output?

saragerion (commented):
Thanks a lot Andrea!

packages/metrics/tests/unit/Metrics.test.ts
dreamorosi: Done! See 3 lines below

saragerion (approved):

ijemmy (approved):

PR #1060, opened 2022-08-14
fix(logger): decorated class methods cannot access `this`

### LINKED ISSUE #1058: Bug: decorated class methods cannot access `this` in Logger
## Bug description

A long form description of this issue is available in #1056. This is an issue that affects all three Powertools utilities because they all share similar implementation for the decorators.

---

When using Powertools in with a Class and when decorating a method, the decorated method does not have access to the correct value of `this` and so it's not able to call other methods nor read attributes from its same class.

This is due to the fact that the decorator implementation uses an `() => ()` arrow function instead of a regular `function () {}` one. This mixes the value of `this` which then causes the wrong value to be passed.

Additionally there's a missing detail in the documentation of the Class instrumentation that prevents compounds the issue described above.

### Expected Behavior

Decorated methods have access to the correct value of `this` and is able to call methods or access attributes/properties.

### Current Behavior

See "Steps to Reproduce" section.

### Possible Solution

Apply the same changes done to Tracer in #1055 

### Steps to Reproduce

```ts
import {
  LambdaInterface,
  SyncHandler,
  AsyncHandler,
} from "@aws-lambda-powertools/commons";

const logger = new Logger({ serviceName: "serverlessAirline" });

  greeting: string;

  constructor(message: string) {
    this.greeting = message;
  }

  public getGreeting(): string {
    /* ... */
    return `Hello ${this.greeting}`;
  }

  @logger.injectLambdaContext()
  public handler(_event: any, _context: any, _callback: any): void {
    console.log(this.getGreeting());
  }
}

export const myFunction = new Lambda("World");
```

The value logged from `console.log(this.getGreeting());` should be `Hello World` but instead is `Hello undefined`:
image

### Environment

* **Powertools version used**: 1.1.0
* **Packaging format (Layers, npm)**: all
* **AWS Lambda function runtime:** all
* **Debugging logs:** N/A

### Related issues, RFCs

#1056

dreamorosi:

As described in #1058 the decorator implementation was affected by an issue that prevented the decorated method/class to access other class members as the value of `this` was passed incorrectly.

This PR aims at fixing the issue, creating an unit test case, making minimal changes to the function in the e2e test to verify the change, and update docstrings/docs.

A long form explanation of the issue can be found at #1055, when merged this PR will close #1058. 

### How to verify this change

See checks below the PR, also see the results of the [e2e test run](https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/2855192482).

### Related issues, RFCs

**Issue number:** #1058  

### PR status

## Checklist

### Breaking change checklist

N/A

---

### DIFF
--- a/packages/logger/tests/e2e/sampleRate.decorator.test.FunctionCode.ts
+++ b/packages/logger/tests/e2e/sampleRate.decorator.test.FunctionCode.ts
@@ -2,27 +2,39 @@ import { Logger } from '../../src';
 import { APIGatewayProxyEvent, Context } from 'aws-lambda';
 import { LambdaInterface } from '@aws-lambda-powertools/commons';
 
-const SAMPLE_RATE = parseFloat(process.env.SAMPLE_RATE);
-const LOG_MSG = process.env.LOG_MSG;
+const SAMPLE_RATE = parseFloat(process.env.SAMPLE_RATE || '0.1');
+const LOG_MSG = process.env.LOG_MSG || 'Hello World';
 
 const logger = new Logger({
   sampleRateValue: SAMPLE_RATE,
 });
 
 class Lambda implements LambdaInterface {
+  private readonly logMsg: string;
+
+  public constructor() {
+    this.logMsg = LOG_MSG;
+  }
+
   // Decorate your handler class method
   @logger.injectLambdaContext()
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
   public async handler(event: APIGatewayProxyEvent, context: Context): Promise<{requestId: string}> {
-    logger.debug(LOG_MSG);
-    logger.info(LOG_MSG);
-    logger.warn(LOG_MSG);
-    logger.error(LOG_MSG);
-
+    this.dummyMethod();
+    
     return {
       requestId: context.awsRequestId,
     };
   }
+
+  private dummyMethod() : void {
+    logger.debug(this.logMsg);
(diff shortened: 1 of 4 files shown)


dreamorosi:
Note to reviewers: don't merge this until #1054 is merged.

packages/logger/tests/e2e/sampleRate.decorator.test.FunctionCode.ts
```diff
@@ -2,27 +2,39 @@ import { Logger } from '../../src';
...
-    logger.error(LOG_MSG);
-
+    this.dummyMethod();
+    
     return {
       requestId: context.awsRequestId,
     };
   }
+
+  private dummyMethod() : void {
```
ijemmy, line 31: (optional) Calling this `dummyMethod` is a bit misleading.  What about `printLogInAllLevels`?

The reason is that these logs are used for assertion later so they are important.

ijemmy (commented):
One small nit.

Also, similar to #1055, we should also update the `examples/cdk/src/*.Decorator.ts` files. They still don't use .bind() method. We can have this in another PR if you prefer (then I'll approve this).

ijemmy (approved):
Approved.

Discussed with @dreamorosi , we'll go with this to push the fix first.

The example will be updated in a separated PR.

ijemmy (dismissed):

saragerion (approved):

ijemmy (approved):

Issue #1063, opened 2022-08-18
Docs: incorrect version number in the docs + API docs returning 404


dreamorosi:
## Bug description

As per title of the issue, after releasing the [`v1.1.1`](https://github.com/awslabs/aws-lambda-powertools-typescript/releases/tag/v1.1.1) version, the action that publishes the documentation [completed successfully](https://github.com/awslabs/aws-lambda-powertools-typescript/runs/7900433814?check_suite_focus=true) but the version number that corresponds to `latest` is incorrect (`v1.1.0`) and the API docs that should be [published here](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/api/), return 404.

Currently investigating, will provide more info once I finish triaging.

----

**[EDIT 1: 7:51PM CEST]** After a first assessment I have found why the API docs return a 404. In #1052 I mistakenly removed [this step from the workflow](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/a0f1db2be96bb27d2ec01eb07b05991e3bdf19c9/.github/workflows/on-release-prod.yml#L105-L111) which causes the API docs to not be published to `latest` but only under the version number - i.e. [if you browse v1.1.0](https://awslabs.github.io/aws-lambda-powertools-typescript/1.1.0/api/) they are there.

There is however another issue related to the version issue that I'm still investigating. Even though we just released v1.1.1, the latest version in the docs is v1.1.0. The version number [is read from one](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/workflows/reusable-publish-docs.yml#L62) of the `package.json` files, and in the last run the version that was read was in fact `v1.1.0` (see screenshot below & run [here](https://github.com/awslabs/aws-lambda-powertools-typescript/runs/7900433814?check_suite_focus=true)) even though that action ran after the versioning had already happened.

![image](https://user-images.githubusercontent.com/7353869/185462290-9ecc255e-7a43-49c1-a96a-08519c62b546.png)

----

**[EDIT 2: 3:58PM CEST]** I have found the "issue" which is not an issue but a specific behavior of GitHub actions. I have opened a PR and more details are available there, after merging the issue should be resolved.

### Expected Behavior

Published docs should reflect latest version & API docs should be available.

### Current Behavior

Published docs reflect wrong version & API docs are not available (404).

### Possible Solution

N/A

### Steps to Reproduce

[...]

PR #1066, opened 2022-08-19
fix(docs): docs published with incorrect version number + api docs missing after release

dreamorosi:

As described in #1063 there were two issues related to the docs publishing after a release. Both issues were due to oversights on my part.

### Issue 1 - API docs are missing

The first one was that I removed a step from the `reusable-publish-docs` ([link](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/workflows/reusable-publish-docs.yml)) and this step was the one that published the API docs under the correct version number in the docs. You can verify this by checking the `/latest` version of the docs, which works (while the `v1.1.0` doesn't). Adding back the step fixes the docs missing issue.

### Issue 2 - Incorrect version number

The second one was instead due to a behavior of GitHub Actions that I misunderstood.

For context: we have 2 workflows: `make-release` ([link](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/workflows/make-release.yml)) and `reusable-publish-docs` ([link](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/workflows/reusable-publish-docs.yml)). 

The first one is triggered manually, the second is called by the former as the last step (we also use it elsewhere, hence it’s `reusable-` prefix).

During the `make-release` workflow we perform some actions, [one of these](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/workflows/make-release.yml#L48) is to increment the version of our packages, commit, & push a tag.

**When `make-release` calls `reusable-publish-docs` I was expecting the latter to checkout the code that contains the latest changes & tag, so that it can build the docs using the correct version number**.

What happens instead is that when checking out the actions/checkout action detects the latest tag (L517 of “Checkout code step” [here](https://github.com/awslabs/aws-lambda-powertools-typescript/runs/7900433814?check_suite_focus=true#step:2:522)) but then the code in the repo & the tag used are the previous & same as the one in the make-release context.

Below a diagram that visually represents what described so far:
IMG_0009

Reading the [docs on reusing workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) though, there are some info that I was missing. Below two key excerpts:
> A workflow that uses another workflow is referred to as a "caller" workflow. The reusable workflow is a "called" workflow. One caller workflow can use multiple called workflows.

and

> For example, if the called workflow uses actions/checkout, the action checks out the contents of the repository that hosts the caller workflow, not the called workflow.

This means that even though the called workflow (`reusable-publish-docs`) was correctly detecting the presence of a new tag, the context used and the the version of the repo checked out, were the same as the `make-release` one, which was in a point in time before the release & consequent version bump.

There are two solutions that come to mind here:
1. Have the caller workflow export the commit SHA of the post-version state of the repo, pass it to the called workflow, and have the called workflow to use said SHA to checkout the correct version of the workflow. This approach would be laborious and would make the two workflow tightly coupled, reducing the _reusability_ of the called workflow.
2. Change the trigger of the called workflow and execute it in response to an [actual release event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release). This is an event that that is triggered **after a maintainer publishes a release**, which means that the docs will be synced to the GitHub release instead of the the NPM publish. This is the approach that I am proposing in this PR which also removes both issues described above since it'll get always the latest tag in the event.

Finally, I'm also adding a way to **manually trigger the publish docs workflow** as a break-glass solution that will allow us to manually trigger a docs publish if any of the other triggers fails.

### How to verify this change

Review the code & see these workflow runs in a fork:
- I run `make-release` - [result here](https://github.com/dreamorosi/aws-lambda-powertools-typescript/actions/runs/2889145596) - (only difference here is that I'm not publishing to NPM, otherwise the workflow is the same)
- I then [make a release](https://github.com/dreamorosi/aws-lambda-powertools-typescript/releases/tag/v1.2.17) on the GH repo (manually) - (This is a release in the release section, not a workflow execution)
- This triggers the `reusable-publish-docs` workflow - result [here](https://github.com/dreamorosi/aws-lambda-powertools-typescript/actions/runs/2889171377)
- Check that the docs correspond to the released version - [link](https://dreamorosi.github.io/aws-lambda-powertools-typescript/latest/) (I'll delete this after the PR is merged to avoid confusion)

Additionally, I have also run the publish docs workflow manually, [here's the result](https://github.com/dreamorosi/aws-lambda-powertools-typescript/actions/runs/2889878925).

### Related issues, RFCs

**Issue number:** #1063 

### PR status

## Checklist

### Breaking change checklist

N/A

---

### DIFF
--- a/.github/workflows/on-merge-to-main.yml
+++ b/.github/workflows/on-merge-to-main.yml
@@ -18,6 +18,8 @@ jobs:
     secrets:
       token: ${{ secrets.GITHUB_TOKEN }}
   run-unit-tests:
+    needs: get_pr_details
+    if: ${{ needs.get_pr_details.outputs.prIsMerged == 'true' }}
     uses: ./.github/workflows/reusable-run-unit-tests.yml
   publish:
     needs:
--- a/.github/workflows/publish-docs-on-release.yml
+++ b/.github/workflows/publish-docs-on-release.yml
@@ -1,0 +1,17 @@
+name: Publish docs on release
+
+on:
+  # Triggered manually
+  workflow_dispatch: {}
+  # Or triggered as result of a release
+  release:
+    types: [released]
+
+jobs:
+  run-unit-tests:
+    uses: ./.github/workflows/reusable-publish-docs.yml
+    with:
+      workflow_origin: ${{ github.event.repository.full_name }}
+      isRelease: "true"
+    secrets:
+      token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/publish-docs-on-release.yml
+++ b/.github/workflows/publish-docs-on-release.yml
@@ -13,7 +13,7 @@ on:
     types: [released]
 
 jobs:
-  run-unit-tests:
+  publish-docs:
     uses: ./.github/workflows/reusable-publish-docs.yml
     with:
       workflow_origin: ${{ github.event.repository.full_name }}
--- a/.github/workflows/publish-docs-on-release.yml
+++ b/.github/workflows/publish-docs-on-release.yml
@@ -2,7 +2,12 @@ name: Publish docs on release
 
 on:
   # Triggered manually
-  workflow_dispatch: {}
+  workflow_dispatch:
+    inputs:
+      versionNumber:
+        required: true
+        type: string
+        description: "If running this manually please insert a version number that corresponds to the latest published in the GitHub releases (i.e. v1.1.1)"
   # Or triggered as result of a release
   release:
     types: [released]
@@ -13,5 +18,6 @@ jobs:
     with:
       workflow_origin: ${{ github.event.repository.full_name }}
       isRelease: "true"
+      versionNumber: ${{ inputs.versionNumber }}
     secrets:
       token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/reusable-publish-docs.yml
+++ b/.github/workflows/reusable-publish-docs.yml
@@ -21,7 +21,7 @@ on:
 jobs:
   publish-docs:
     # see https://github.com/awslabs/aws-lambda-powertools-python/issues/1349
-    if: inputs.workflow_origin == 'awslabs/aws-lambda-powertools-typescript'
+    if: ${{ inputs.workflow_origin == 'awslabs/aws-lambda-powertools-typescript' }}
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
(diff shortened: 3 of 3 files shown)


.github/workflows/publish-docs-on-release.yml
```diff
@@ -0,0 +1,23 @@
...
+      versionNumber:
+        required: true
+        type: string
+        description: "If running this manually please insert a version number that corresponds to the latest published in the GitHub releases (i.e. v1.1.1)"
+  # Or triggered as result of a release
+  release:
+    types: [released]
+
+jobs:
+  run-unit-tests:
```
ijemmy, line 16: Forgot to change job name?

.github/workflows/on-merge-to-main.yml
```diff
@@ -18,6 +18,8 @@ jobs:
     secrets:
       token: ${{ secrets.GITHUB_TOKEN }}
   run-unit-tests:
+    needs: get_pr_details
+    if: ${{ needs.get_pr_details.outputs.prIsMerged == 'true' }}
```
ijemmy, line 22: Q: why do we need `prIsMerged` to be true? I thought this workflow is triggerred only when PR merged.

ijemmy (commented):
@dreamorosi   Thanks for the detailed description about the issues!

Let me double check my understanding. 

`Reusable Publish docs` workflow can be triggered from 3 scenarios:
1. (GitHub) Release  (`publish-docs-on-releases.yml`)
2. Manual trigger (from `publish-docs-on-releases.yml`)
3. Merge to main (line 26 in `on-merge-to-main.yml`)

Is that correct? 

If that is the case, can the issue 2 (checking out the wrong version) also happens in merge to main scenario?

.github/workflows/on-merge-to-main.yml
dreamorosi: This workflow is actually triggered by [`record_pr.yml`](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/workflows/record_pr.yml#L5), which in turn is triggered every time a PR is opened, updated (title or description change), or closed. This is a design that we copied from the Powertools Python repo.

The reason why we don't want to trigger directly is that this workflow is somewhat privileged and we don't want to allow forks to access it. So we trigger the workflow that exports the PR details for most events, and then dispatch the other downstream workflows. Then those workflows can decide what to do based on the content/status of the PR.

.github/workflows/publish-docs-on-release.yml
dreamorosi:
```suggestion
  publish-docs:
```

dreamorosi:
> Is that correct?

Yes, your understanding is correct.

> If that is the case, can the issue 2 (checking out the wrong version) also happens in merge to main scenario?

No, because when we merge on `main` we don't care about the version since we publish the docs always to `dev`. The docs are published under a specific version only when we make a release.

ijemmy (approved):

saragerion (approved):

Issue #1044, opened 2022-08-04
Feature request: Increase maximum number of metrics dimension from 10 to 30


ijemmy:
## Summary
Increase maximum number of metrics dimension from 10 to 30

## Bug description
CloudWatch Metrics has increase maximum number of dimensions from 10 to 30 [ref](https://aws.amazon.com/about-aws/whats-new/2022/08/amazon-cloudwatch-metrics-increases-throughput/).  

Currently, `metrics` utility throws an error when user try to add more than 10 dimension. This should be updated to 30. 

Note: Powertools for Python is having the same change in [this issue](https://github.com/awslabs/aws-lambda-powertools-python/issues/1413).  Note that we **DO NOT** consider this as breaking change. (See that issue discussion on this)

### Expected Behavior
Clients should be able to add up to 30 dimensions before throwing an error.

### Current Behavior
Adding more than 10 dimensions will throw an error.

### Possible Solution
You can find the relevant code to update in this [search](https://github.com/awslabs/aws-lambda-powertools-typescript/search?q=MAX_DIMENSION_COUNT).  

Required changes are:
1. Update `MAX_DIMENSION_COUNT` variable from `9` to `29`
2. Update related unit tests 

### Steps to Reproduce
Call `metrics.addDimension()` more then 10 times. See [this test case](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/014c5bd7d5c807064af8f04c16d297a8fe3bc0d9/packages/metrics/tests/unit/Metrics.test.ts#L83) for example.

### Environment

* **Powertools version used**: all
* **Packaging format (Layers, npm)**: npm
* **AWS Lambda function runtime:** all runtimes
* **Debugging logs:** n/a

### Related issues, RFCs
* [Issue with the same change in Python](https://github.com/awslabs/aws-lambda-powertools-python/issues/1413)

sthuber90:
Happy to take over this task

dreamorosi:
Hi @sthuber90 thanks for commenting under the issue. 

Based on the description above the change should involve both the `Metric` class & unit tests, not sure yet about integration tests. Do you need any additional info/support in defining the scope of the feature?

I'll assign the issue to you and I'll be happy to help with reviewing the PR once it's opened.

sthuber90:
Thank you very much. I will try to work out a minimal PR and look into whether integration tests make sense for this feature. All in all, it really seems like a good first issue and easy to start with :)

Issue #163, opened 2021-08-06
Maintenance: metrics unit testing strategy


alan-churley:
## Description of the feature request

**Problem statement**
Metrics unit tests currently are not structured the same as the Logger module, 

In the logger only public methods are tested, following the recommendations here: https://github.com/goldbergyoni/javascript-testing-best-practices/#-%EF%B8%8F-14-stick-to-black-box-testing-test-only-public-methods

Opening a ticket to track discussions and work on improving this,

### Related issues, RFCs
https://github.com/awslabs/aws-lambda-powertools-typescript/pull/102

dreamorosi:
The issue is still current so contributions are welcome.

This is a good issue for newcomers as it allows to understand how the utility works in depth, as well as for those who are looking at sharpening their Jest skills.

If anyone is interested in picking this up, please leave a comment here, then discuss your findings / proposed changes before opening a PR.

arnabrahman:
@dreamorosi Interested, I will look into this & share my findings.

dreamorosi:
Hi @arnabrahman, thank you for your interest in this issue ✨!

Looking forward to review / collaborate on your proposal and findings in the issue.

If you have any question while you look into it, please let me know both here or on [our Discord](https://discord.gg/B8zZKbbyET).

arnabrahman:
Hello @dreamorosi, sorry for replying late to this. 

I think this particular issue was first mentioned [here](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/102/files#r679956229) by @saragerion . After looking at both the unit tests for `Logger`/`Tracer`, my understanding of what she meant by the issue is, in the `Logger`/`Tracer` package all the tests are written for public functions rather than testing a particular feature of the package. But in `Metrics`, here we are following a pattern of testing a particular behavior/feature of `Metrics` rather than taking a single function & testing every case of that function. 

So, if we want to follow a similar pattern/structure for `Metrics` similar to `Logger`/`Tracer`, we would take a function & test every case of that function. For example, for `addMetric` function, we want to test

-  the behavior when a new metric is added
-  the behavior when an old metric is added
-  the behavior when adding a single metric (for single metric, it will publish StoredMetrics)

We can take this approach & apply it to every public function of `Metrics`. 

These are my findings. I might be wrong about this, so let me know if i am in the right direction or not. Thanks 🙂

dreamorosi:
[...]

PR #1072, opened 2022-08-23
feat(metrics):  increase maximum dimensions to 29

sthuber90:

Amazon added support to up 30 dimensions per metric. In this PR we increase this limit

### How to verify this change

Review the code changes and the results of the automated builds.

### Related issues, RFCs

**Issue number:**  #1044 

### PR status

## Checklist

### Breaking change checklist

N/A

---

### DIFF
--- a/packages/metrics/tests/unit/Metrics.test.ts
+++ b/packages/metrics/tests/unit/Metrics.test.ts
@@ -13,7 +13,7 @@ import { Metrics, MetricUnits } from '../../src/';
 import { populateEnvironmentVariables } from '../helpers';
 
 const MAX_METRICS_SIZE = 100;
-const MAX_DIMENSION_COUNT = 9;
+const MAX_DIMENSION_COUNT = 29;
 const DEFAULT_NAMESPACE = 'default_namespace';
 
 const consoleSpy = jest.spyOn(console, 'log').mockImplementation();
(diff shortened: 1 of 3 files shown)


packages/metrics/tests/unit/Metrics.test.ts
```diff
@@ -459,6 +459,7 @@ describe('Class: Metrics', () => {
     });
 
     test('Should throw an error if the same metric name is added again with a different unit', ()=> {
+      expect.assertions(1);
```
dreamorosi, line 462: Could you please move this assertion at the end of the test? This file specifically is a bit of an outlier, but generally speaking we try to do 1/ Setup, 2/ Action, 3/ Assess

dreamorosi: I just saw that the assertion is used at the beginning in other test cases of this file, we will be changing this in the context of #163 to make them more uniform to the other two utils.

You can choose to ignore the comment or address it, it won't be a blocker for this PR.

dreamorosi (approved):
Thank you for opening the PR.

I'm going to ask for a review to another maintainer as we have a 2 approvals model for each PR.

Note for the next maintainer: I have run e2e tests and the result was successful:
image

packages/metrics/tests/unit/Metrics.test.ts
sthuber90: As you already said, I added it at the beginning just as it's the case in other tests. If I hadn't seen this, I would have used the `fail('reason')` before the `catch` block but that's something to be adressed in the issue that you've referenced

ijemmy (approved):
LGTM, thanks for your contribution @sthuber90   :)

PR #1087, opened 2022-09-12
fix(logger): wait for decorated method return before clearing out state

### LINKED ISSUE #1085: Bug: Decorator clears states too early
## Bug description

Looks like decorator clears states too early on async/await codes.

https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/logger/#clearing-all-state
```ts
import { LambdaInterface } from "@aws-lambda-powertools/commons";
import { setTimeout } from "timers";

const logger = new Logger({
  logLevel: "DEBUG",
  persistentLogAttributes: {
    foo: "bar",
    biz: "baz",
  },
});

  // Enable the clear state flag
  @logger.injectLambdaContext({ clearState: true })
  public async handler(event: any, _context: any): Promise<void> {
    logger.appendKeys({ temporaryState: true });
    logger.debug("This is a DEBUG log");
    await this.doSomethingAsync();
    logger.debug("This is a DEBUG log");
  }

  private async doSomethingAsync(): Promise<void> {
    return new Promise((resolve) => setTimeout(resolve, 5 * 1000));
  }
}

const myFunction = new Lambda();
```

### Expected Behavior

States are not cleared until handler execution is complete.

#### CloudWatch Logs

```jsonc
{
    "cold_start": true,
    "function_arn": "arn:aws:lambda:us-east-1:000000000000:function:function_foo",
    "function_memory_size": 128,
    "function_name": "function_foo",
    "function_request_id": "4577ef0f-1e74-47b9-b8e6-40d45e736089",
    "level": "DEBUG",
    "message": "This is a DEBUG log",
    "timestamp": "2022-09-09T05:03:32.625Z",
    "xray_trace_id": "1-631ac924-7290afce25c885af5a324679",
    "foo": "bar",
    "biz": "baz",
    "temporaryState": true
}

{
    "cold_start": true,
    "function_arn": "arn:aws:lambda:us-east-1:000000000000:function:function_foo",
    "function_memory_size": 128,
    "function_name": "function_foo",
    "function_request_id": "4577ef0f-1e74-47b9-b8e6-40d45e736089",
    "level": "DEBUG",
    "message": "This is a DEBUG log",
    "timestamp": "2022-09-09T05:03:37.670Z",
    "xray_trace_id": "1-631ac924-7290afce25c885af5a324679",
    "foo": "bar",
    "biz": "baz",
    "temporaryState": true // <==
}
```

### Current Behavior

State are cleared before the handler's execution is complete.

#### CloudWatch Logs
```jsonc
{
    "cold_start": true,
    "function_arn": "arn:aws:lambda:us-east-1:000000000000:function:function_foo",
    "function_memory_size": 128,
    "function_name": "function_foo",
    "function_request_id": "4577ef0f-1e74-47b9-b8e6-40d45e736089",
    "level": "DEBUG",
    "message": "This is a DEBUG log",
    "timestamp": "2022-09-09T05:03:32.625Z",
    "xray_trace_id": "1-631ac924-7290afce25c885af5a324679",
    "foo": "bar",
    "biz": "baz",
    "temporaryState": true
}

{
    "cold_start": true,
    "function_arn": "arn:aws:lambda:us-east-1:000000000000:function:function_foo",
    "function_memory_size": 128,
    "function_name": "function_foo",
    "function_request_id": "4577ef0f-1e74-47b9-b8e6-40d45e736089",
    "level": "DEBUG",
    "message": "This is a DEBUG log",
    "timestamp": "2022-09-09T05:03:37.670Z",
    "xray_trace_id": "1-631ac924-7290afce25c885af5a324679",
    "foo": "bar",
    "biz": "baz"
    // <== !!!
}
```

### Possible Solution
`await` handler and then clear states?
https://github.com/awslabs/aws-lambda-powertools-typescript/blob/34d3a766acccce9c8292aeb3795d95d2aeea683a/packages/logger/src/Logger.ts#L299-L305

### Steps to Reproduce

1. deploy lambda using AWS CDK.

```ts
new lambdaNodeJs.NodejsFunction(
  stack,
  "MyLambdaFunction",
  {
    entry: "src/lambda/powertools-bug.ts",
    memorySize: 128,
    runtime: lambda.Runtime.NODEJS_16_X,
    architecture: lambda.Architecture.ARM_64,
    timeout: cdk.Duration.minutes(1),
    retryAttempts: 0,
    tracing: lambda.Tracing.ACTIVE,
    bundling: { keepNames: true },
  }
);
```

### Environment

* **Powertools version used**: 1.2.1
* **Packaging format (Layers, npm)**: npm
* **AWS Lambda function runtime:**: Node.js 16.x (arm)
* **Debugging logs:**

### Related issues, RFCs
- https://github.com/awslabs/aws-lambda-powertools-typescript/issues/1027
- https://github.com/awslabs/aws-lambda-powertools-typescript/pull/902

dreamorosi:

As reported in #1085 the injectLambdaContext decorator, when used on async methods, was not awaiting the method to return before performing other clean-up/post-exec logic. 

This PR introduces an `async/await` in the decorator logic so that the decorated method is always awaited. This should appears to fix #1085.

**Note to reviewers/readers:**
It should be noted that the decorator will now always `await` the decorated method regardless of it being an async function or not. Based on what I was able to find online awaiting a non-promise should be a no-op and should not have a performance impact. If this is not the case please point this out during review and we'll try to implement a conditional logic. At the moment I went with always awaiting for the reason above and also because this is how the decorator is implemented in `Tracer`.

### How to verify this change

See results of checks below the PR, or check the result of the integration tests here:
https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/3032214494

Optionally, run `npm pack -w packages/logger` & test the change in the repo mentioned in the comment.

### Related issues, RFCs

**Issue number:** #1085 

### PR status

## Checklist

### Breaking change checklist

N/A

---

### DIFF
--- a/packages/logger/src/Logger.ts
+++ b/packages/logger/src/Logger.ts
@@ -285,7 +285,7 @@ class Logger extends Utility implements ClassThatLogs {
       const loggerRef = this;
       // Use a function() {} instead of an () => {} arrow function so that we can
       // access `myClass` as `this` in a decorated `myClass.myMethod()`.
-      descriptor.value = (function (this: Handler, event, context, callback) {
+      descriptor.value = (async function (this: Handler, event, context, callback) {
 
         let initialPersistentAttributes = {};
         if (options && options.clearState === true) {
@@ -297,7 +297,7 @@ class Logger extends Utility implements ClassThatLogs {
         /* eslint-disable  @typescript-eslint/no-non-null-assertion */
         let result: unknown;
         try {
-          result = originalMethod!.apply(this, [ event, context, callback ]);
+          result = await originalMethod!.apply(this, [ event, context, callback ]);
         } catch (error) {
           throw error;
         } finally {
--- a/packages/logger/src/Logger.ts
+++ b/packages/logger/src/Logger.ts
@@ -275,7 +275,7 @@ class Logger extends Utility implements ClassThatLogs {
    * @returns {HandlerMethodDecorator}
    */
   public injectLambdaContext(options?: HandlerOptions): HandlerMethodDecorator {
-    return (target, _propertyKey, descriptor) => {
+    return (_target, _propertyKey, descriptor) => {
       /**
        * The descriptor.value is the method this decorator decorates, it cannot be undefined.
        */
--- a/packages/logger/tests/unit/Logger.test.ts
+++ b/packages/logger/tests/unit/Logger.test.ts
@@ -1166,6 +1166,48 @@ describe('Class: Logger', () => {
 
     });
 
+    test('when used as decorator on an async method, the method is awaited correctly', async () => {
+
+      // Prepare
+      const injectLambdaContextAfterOrOnErrorMock = jest.fn().mockReturnValue('worked');
+      // Temporarily override the cleanup static method so that we can "spy" on it.
+      // This method is always called after the handler has returned in the decorator
+      // implementation.
+      Logger.injectLambdaContextAfterOrOnError = injectLambdaContextAfterOrOnErrorMock;
+      const logger = new Logger({
+        logLevel: 'DEBUG',
+      });
+      const consoleSpy = jest.spyOn(logger['console'], 'info').mockImplementation();
+      class LambdaFunction implements LambdaInterface {
+        @logger.injectLambdaContext()
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        public async handler<TResult>(_event: unknown, _context: Context, _callback: Callback<TResult>): void | Promise<TResult> {
+          await this.dummyMethod();
+          logger.info('This is a DEBUG log');
+
+          return;
+        }
+
+        private async dummyMethod(): Promise<void> {
+          return;
+        }
+      }
+
+      // Act
+      const lambda = new LambdaFunction();
+      const handler = lambda.handler.bind(lambda);
+      await handler({}, dummyContext, () => console.log('Lambda invoked!'));
+
+      // Assess
+      expect(consoleSpy).toBeCalledTimes(1);
+      // Here we assert that the logger.info method is called before the cleanup function that should awlays 
+      // be called after the handler has returned. If logger.info is called after it means the decorator is
+      // NOT awaiting the handler which would cause the test to fail.
+      expect(consoleSpy.mock.invocationCallOrder[0]).toBeLessThan(injectLambdaContextAfterOrOnErrorMock.mock.invocationCallOrder[0]);
+
+    });
+
   });
 
   describe('Method: refreshSampleRateCalculation', () => {


ijemmy (commented):
Nice finding. How did you figure out this is where the bug is?

`await` will wrap non-promise item in promise. It does change execution order. But that's the correct behavior in this case. (See Jonas Wilms' [answer](https://stackoverflow.com/questions/55262996/does-awaiting-a-non-promise-have-any-detectable-effect))


A few questions:
1. Is it possible to add unit test to cover this case? 
2. I see the same issue in `Tracer.ts` 's `captureMethod` and `captureLambdaHandler`. Could we also fix that in this PR or log an issue for this (could be a good first issue) if not urgent.

flochaz (approved):

packages/logger/src/Logger.ts
```diff
@@ -285,7 +285,7 @@ class Logger extends Utility implements ClassThatLogs {
       const loggerRef = this;
       // Use a function() {} instead of an () => {} arrow function so that we can
       // access `myClass` as `this` in a decorated `myClass.myMethod()`.
-      descriptor.value = (function (this: Handler, event, context, callback) {
+      descriptor.value = (async function (this: Handler, event, context, callback) {
```
flochaz, line 288: could we add a unit test to demonstrate the fix ?

flochaz (changes requested):

packages/logger/src/Logger.ts
dreamorosi: Done in the latest commit

dreamorosi:
Hello, I have added a unit test case for this feature. 

I have detailed this in the comments inside the tests, but in order to test the fix I am asserting the call order between the cleanup function of the decorator (that should always be called after `handler` returns) and something inside the handler that is placed after an `await`. The correct order in this case should be: 1/ whatever is inside `handler`, 2/ the cleanup function.

This relies on an implementation detail of the decorator so it might not be the best solution formally speaking, but the only other alternative that I was able to test on was adding an artificial delay of at least 5000ms to the awaited method, so I went with this one.

If anyone has a better idea I'm open to adopt it.

dreamorosi:
@ijemmy I have opened an issue for Tracer like you suggested: #1091 

Since I'm planning on using the same method to test the fix in the unit tests I'll wait till this PR is approved before starting to work on that PR.

flochaz (approved):

packages/logger/src/Logger.ts
```diff
@@ -275,7 +275,7 @@ class Logger extends Utility implements ClassThatLogs {
    * @returns {HandlerMethodDecorator}
    */
   public injectLambdaContext(options?: HandlerOptions): HandlerMethodDecorator {
-    return (target, _propertyKey, descriptor) => {
+    return (_target, _propertyKey, descriptor) => {
```
dreamorosi, line 278: Housekeeping. The `target` param is never used so I'm prefixing it with underscore to tell the IDE that this is intended.

saragerion (approved):

Issue #165, opened 2021-08-10
Maintenance: `undefined` environment variable returning empty string from `EnvironmentVariablesService`


alan-churley:
## Description of the feature request

**Problem statement**
Currently the EnvironmentVariablesService returns an empty string if an env var is not set, rather than undefined.

Ticket to discuss the pros and cons of this approach and consider changing the logic to use undefined

**Summary of the feature**

**Code examples**

**Benefits for you and the wider AWS community**

**Describe alternatives you've considered**

**Additional context**

### Related issues, RFCs

This issue should be addressed after:
#484

dreamorosi:
Would like to pick up again the discussion about this topic to see if we want to change the current behaviour or not. In case we decide to change I'm available to pick up the task.

The outcome of this discussion should be an agreement on whether `EnvironmentVariablesService.get()` should return an empty string (i.e. `''`) or `undefined` when a given environment variable is not set.

## Context
The example below is from [`packages/logger/src/config/EnvironmentVariablesService.ts`](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/logger/src/config/EnvironmentVariablesService.ts#L16), but a similar implementation is found in `Tracer` and `Metrics`:
```ts
/**
 * Class EnvironmentVariablesService
 *
 * This class is used to return environment variables that are available in the runtime of
 * the current Lambda invocation.
 * These variables can be a mix of runtime environment variables set by AWS and
 * variables that can be set by the developer additionally.
 *
 * @class
 * @extends {ConfigService}
 * @see https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime
 * @see https://awslabs.github.io/aws-lambda-powertools-typescript/latest/#environment-variables
 */
class EnvironmentVariablesService extends ConfigService {

  // Reserved environment variables
  private xRayTraceIdVariable = '_X_AMZN_TRACE_ID';

  /**
   * It returns the value of an environment variable that has given name.
   *
   * @param {string} name
   * @returns {string}
   */
  public get(name: string): string {
    return process.env[name]?.trim() || '';
  }

  /**
   * It returns the value of the AWS_REGION environment variable.
   *
   * @returns {string}
   */
  public getAwsRegion(): string {
    return this.get(this.awsRegionVariable);
  }

  /**
   * It returns the value of the POWERTOOLS_SERVICE_NAME environment variable.
   *
   * @returns {string}
   */
[...]

PR #1123, opened 2022-10-17
feat(all): moved EnvService to commons + exposed getXrayTraceId in tracer

### LINKED ISSUE #1070: Feature request: access to the root trace id in Tracer

**Problem statement**
When handling application errors, I want to embed the XRay Trace ID in the payload & error message returned by my API so that customers can cite the trace ID in support requests, and so our development team can easily look up the trace from a customer service request via AWS XRay's web console.

**Summary of the feature**
I'd like the tracer to provide the root trace id so that I can access it in my (otherwise) unhandled exceptions handler when I construct a user-facing error message.

**Code examples**

```ts
const rootTraceId = tracer.getRootTraceId();

// Example of returning an error with Express.js
res.status(500).json({
  error: `Internal Error - Please contact support and quote the following id: ${rootTraceId}`,
  // Include the rootTraceId in the response so we can show a "contact support" button that
  // takes the customer to a customer service form with the trace as additional context.
  rootTraceId, 
});
```

Makes it easier for the wider community to expose trace IDs to their customers so that developers can look up the conditions and contexts of the errors reported by their customers.

For now, we've implemented this by parsing out the `Root=` traceid from `process.env['_X_AMZN_TRACE_ID']`, but our preference would be to depend on Powertools for this.

```ts
export function getTraceId(): string | undefined {
  const traceId = process.env['_X_AMZN_TRACE_ID'];
  if (!traceId) return;

  return parseTraceId(traceId);
}

export function parseTraceId(traceId: string): string | undefined {
  if (traceId.includes('Root=')) {
    // Example:
    // Root=1-6303d71b-63c75c564e336c9c2cffc1d5;Parent=58e97d4cc1ed1447;Sampled=1
    const [rootTraceId] = traceId
      .split(';')
      .filter((segment) => segment.startsWith('Root='))
      .map((segment) => {
        const [_, rootTraceId] = segment.split('=');
        return rootTraceId;
      });

    if (rootTraceId) {
      // 1-6303d71b-63c75c564e336c9c2cffc1d5
      return rootTraceId;
    }
  }

  return traceId;
}
```

**Additional context**
None

### Related issues, RFCs
None

dreamorosi:

In #1070 customers have requested the addition of a new method in Tracer that allows to access the AWS X-Ray Root Trace ID. This is useful in those situations in which customers want to use this id as correlation id in the responses returned by a function or manually pass it to downstream processes.

The logic to retrieve and parse this ID was already present in Logger ([here](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/logger/src/Logger.ts#L559-L563)). For that utility we use the method to append the `xray_trace_id` to log entries [in some cases](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/logger/#standard-structured-keys).

During the discussion of the feature it became clear that this could be a good moment to also extract this logic to the `commons` package continuing the work started in #484, and potentially unlocking/facilitating #165.

#### Before this change
image

#### After this change
image

This issue also introduces a new public method in Tracer called `getRootXrayTraceId` that allows customers to access the AWS X-Ray Root Trace ID within a Lambda handler.

Fixes #1070 

### How to verify this change

Check the unit tests results, as well as package size report, and e2e tests.

### Related issues, RFCs

**Issue number:** #1070

### PR status

## Checklist

### Breaking change checklist

N/A

---


dreamorosi:
E2E tests run: https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/3271574957

docs/core/tracer.md
```diff
@@ -484,6 +484,38 @@ Use **`POWERTOOLS_TRACER_CAPTURE_ERROR=false`** environment variable to instruct
 
     1. You might **return sensitive** information from exceptions, stack traces you might not control
 
+### Access AWS X-Ray Root Trace ID
+
+Tracer exposes a `getRootXrayTraceId()` method that allows you to retrieve the [AWS X-Ray Root Trace ID](https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-traces) corresponds to the current function execution.
+
+!!! info "This is commonly useful in two scenarios"
+
+    1. By including the root trace id in your response consumers can use it to correlate requests
```
misterjoshua, line 493:
```suggestion
    1. By including the root trace id in your response consumers, can use it to correlate requests
```

packages/logger/src/config/EnvironmentVariablesService.ts
```diff
@@ -123,7 +120,7 @@ class EnvironmentVariablesService extends ConfigService {
    * @returns {string}
    */
   public getXrayTraceId(): string {
-    return this.get(this.xRayTraceIdVariable);
+    return super.getXrayTraceId();
```
misterjoshua, line 123: Given that this class extends commons' `EnvironmentVariablesService` and that the commons superclass marks `getXrayTraceId()` as public, do we need to define a `getXrayTraceId()` method here?

packages/tracer/src/config/EnvironmentVariablesService.ts
```diff
@@ -40,7 +44,7 @@ class EnvironmentVariablesService extends ConfigService {
   }
 
   public getXrayTraceId(): string {
-    return this.get(this.xRayTraceIdVariable);
+    return super.getXrayTraceId();
```
misterjoshua, line 47: Same as for `EnvironmentVariablesService` in logger, are we sure we need this?

misterjoshua (commented):
This PR looks pretty good to me. For now, I have just a few minor comments.

packages/logger/src/config/EnvironmentVariablesService.ts
dreamorosi: Very good spot @misterjoshua, I missed this. 

I have also removed the extra `getServiceName` one from all.

packages/tracer/src/config/EnvironmentVariablesService.ts
dreamorosi: Thanks, removed, see here https://github.com/awslabs/aws-lambda-powertools-typescript/pull/1123#discussion_r999231243

misterjoshua (dismissed):
This all looks good to me!

saragerion (dismissed):
LGTM!

packages/commons/tests/unit/config/EnvironmentVariablesService.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+      // Assess
+      expect(value).toEqual('shopping-cart-api');
+    });
+
+  });
+
+  describe('Method: getXrayTraceId', () => {
+
+    test('It returns the value of the environment variable _X_AMZN_TRACE_ID', () => {
```
flochaz, line 73: can we add better test for this logic: 
1. if empty ?
2. if real (see the split works)

flochaz (approved):
LGTM, minor update on UT would b great

flochaz (approved):

PR #1107, opened 2022-09-29
fix: version command refactor

saragerion:
## Description of your changes

This PR removes all linting related commands from the NPM versioning hooks and makes sure it's done in the CI/CD pipelines actions. 

### Related issues, RFCs

**Issue number:** 

### PR status

***Is this ready for review?:*** NO  
***Is it a breaking change?:*** NO

## Checklist

- [ ] [My changes meet the tenets criteria](https://awslabs.github.io/aws-lambda-powertools-python/#tenets)
- [ ] I have performed a *self-review* of my own code
- [ ] I have *commented* my code where necessary, particularly in areas that should be flagged with a TODO, or hard-to-understand areas
- [ ] I have made corresponding changes to the *documentation*
- [ ] I have made corresponding changes to the *examples*
- [ ] My changes generate *no new warnings*
- [ ] The *code coverage* hasn't decreased
- [ ] I have *added tests* that prove my change is effective and works
- [ ] New and existing *unit tests pass* locally and in Github Actions
- [ ] Any *dependent changes have been merged and published* in downstream module
- [ ] The PR title follows the [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/semantic.yml#L2)

### Breaking change checklist

- [ ] I have documented the migration process
- [ ] I have added, implemented necessary warnings (if it can live side by side)

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1125, opened 2022-10-18
chore(all): removed unused npm scripts + moved linting in PR checks

### LINKED ISSUE #1106: Maintenance: run tests on examples only in CI and not on pre-push locally

**Problem statement**

At the moment we are running unit tests for all examples every time we are pushing code as a part of the `pre-push` hooks. This is a relic of a time (a year ago) during which we were using the monorepo/linked version of the utilities in the examples.

Since a long time we have moved the examples to use the published version. This means that even when one of the utilities changes, testing the examples is redundant as they won't be using the changed version.

This issue proposes to remove the unit test run for examples from the `pre-push` commit. Examples will be tested only as a part of the CI checks in a PR and potentially only when there are changes in the files under `examples/*`.

**Summary of the feature**
Remove the examples from the `pre-commit`, `pre-push` hooks.

**Code examples**
N/A

With this change the DX for maintainers will improve as it removes the need of using `--no-verify` when wanting a faster feedback look.

N/A

**Additional context**
N/A

### Related issues, RFCs

N/A

dreamorosi:

This PR supersedes #1107 by replicating most of the changes introduced in that PR as well as removing all `lerna` related commands (except `lerna version` & `lerna publish` in the Make Release workflow) in favor of npm workspace-based ones.

Just like #1107, this PR:
- Changes the `husky` scripts to:
  - Run `npm run lint-fix -ws` on `pre-commit`
  - Run `npm t -ws` on `pre-push` which runs unit tests on all utilities & not on examples & layer publisher (read on)
- Updates the workflow that is run on PR updates to check linting, if non-formatted code is pushed to a PR the automated checks will fail
- Removed all extra `lint`, `lint-fix`, etc. around versioning. The changes from the point above should prevent non-formatted code to be merged in the first place.
- Removes all `lerna-*` commands from the main `package.json` file & update the only two instances in which they were referenced in the CONTRIBUTING doc.
- Adds `.eslintrc.js` and `tsconfig.es.json` files, as well as the `--resolve-plugins-relative-to .` flag to `npm run lint` so to `examples/*` and `layer-publisher` to allow for local linting. This is needed because they're not part of the npm workspace so they expect their own configs and not the ones from the monorepo while in CI/CD.
- Update GH CodeSpaces & GitPod devfiles so that the changes above are reflected.

Fixes #1106

### How to verify this change

See automated checks 

### Related issues, RFCs

**Issue number:**  #1106

### PR status

## Checklist

### Breaking change checklist

N/A

---


CONTRIBUTING.md
```diff
@@ -156,9 +155,8 @@ See `metrics/tests/e2e/basicFeatures.decorator.test.ts` as an example.
 
 To run e2e tests, you can either use:
 
-* npm task `lerna-test:e2e` (`npm run lerna-test:e2e`) in root folder
-* npm task `test:e2e` (`npm run test:e2e`) in module folder (for example: `packages/metrics`) to run the module specific ones
-* jest directly `npx jest --group=e2e` in module folder. (You can run selective tests by restricting the group to the one you want. For instance `npx jest --group=e2e/metrics/decorator`)
```
ijemmy, line 161: Question:

Is there a way to pass `--group` to Jest via `npm test:e2e -w packages/metrics`?  
I use this a lot to run only a specific set of E2E tests. If there is no option, I will prefer to keep this bullet if there is no alternative.

CONTRIBUTING.md
```diff
@@ -126,9 +126,8 @@ As mentioned before, tests are split into groups thanks to [jest-runner-groups](
 
 To run unit tests, you can either use:
 
-* npm task `lerna-test:unit` (`npm run lerna-test:unit`) in root folder to run them all
-* npm task `test:unit` (`npm run test:unit`) in module folder (for example: `packages/metrics`) to run the module specific ones
-* jest directly `npx jest --group=unit` in module folder to run the module specific ones (You can run selective tests by restricting the group to the one you want. For instance `npx jest --group=unit/metrics/class`)
+* `npm test -ws` while in the root folder to run them all
```
ijemmy, line 129: Shall we make an update to line 91-92? 

```md
We use npm's [workspace feature](https://docs.npmjs.com/cli/v8/using-npm/workspaces) to manage multiple packages in a single repository. In addition, we use [lerna](https://github.com/lerna/lerna#about) to release packages to npm (via `lerna publish`).

For contribution, you can use only npm workspace. It is not necessary to learn lerna to contribute.
```

.gitpod.yml
```diff
@@ -5,22 +5,13 @@ tasks:
...
       # Install AWS SAM CLI
       wget https://github.com/aws/aws-sam-cli/releases/latest/download/aws-sam-cli-linux-x86_64.zip
       unzip aws-sam-cli-linux-x86_64.zip -d sam-installation
       sudo ./sam-installation/install
       rm -rf sam-installation aws-sam-cli-linux-*
-      cd ../../
-      # Setup husky hooks
-      npm run init-environment
+      # Install repo dependencies
+      npm run setup-local
```
ijemmy, line 14: Very nice. :)

ijemmy (changes requested):
Thanks a lot for cleaning this up.

Just one question and a change on `CONTRIBUTING.md`

CONTRIBUTING.md
dreamorosi: The command above (the one with `npx`) will still work if run in the module folder (like specified by the guide).

Generally speaking, this `npm run xxx -w packages/logger -- --group=e2e/metrics/decorator` forwards all the flags after the `--` to the original scoped command.

In this specific instance, since both `test:unit` & `test:e2e` already have the `--group` flag applied (see image below):
image

this won't work.

We have three options here:
- Restore the line above (requires to be run within the module folder)
- Add a generic `jest` commad/script to each `pacakges/*/package.json` so that you can pass arbitrary flags and do something like this:
image
- Create a more scoped command/script in each `pacakges/*/package.json` called `test:group` so that you can do the same as above.

Regardless of which one we choose I'm happy to add the line back to the document.

flochaz (approved):
Great work !

CONTRIBUTING.md
ijemmy: I'm inclined to use a generic jest command. But let's do that in another PR. 
Let's keep the change as it is.

ijemmy (approved):

Issue #751, opened 2022-04-11
Feature request: pretty printing logs in local development


krivachy:
## Description of the feature request

Have the ability to pretty print log statements when running lambdas locally. This would make local development nicer.

**Problem statement**

The logger module always prints out JSON which is required for CloudWatch, but not a nice experience when running a lambda locally. It is hard to parse the JSON that is printed in a single line.

There is no way to override this via `Logger` options.

**Summary of the feature**

To solve this, I would propose an overridable `logPrinter` function that would default to JSON printing. The function would only be invoked if the log level will be outputted. The function would also be invoked after `logFormatter` has run.

This would allow users of the library to conditionally override the printer when running a lambda locally.

**Code examples**

The interface could look like this:

```typescript
interface LogPrintterInterface {
    /**
     * it serializes the log attributes to a string to be printed out.
     *
     * @param {LogAttributesWithMessage} attributes
     * @returns {string}
     */
    printAttributes(attributes: LogAttributesWithMessage): string;
}
```

Usage could look like this:

```typescript
const localPrinter = {
    printAttributes: (attributes: LogAttributesWithMessage) => `${attributes.logLevel} ${attributes.message}`
};
const log = new Logger({
  logPrinter: process.env.IS_LOCAL ? localPrinter : undefined
});
```

**Benefits for you and the wider AWS community**

This overridable config would allow users of the AWS community to develop various output formats, for example ones that colorize the console output.

**Describe alternatives you've considered**

Overriding the actually outputting to stdout, but most users of the library would just use the default `console.log` so overriding printing seemed like a better option.

**Additional context**

`pino` is a library that has pluggable prettification via [`pino-pretty`](https://github.com/pinojs/pino-pretty):

```typescript
const pino = require('pino')
const pretty = require('pino-pretty')
const logger = pino(pretty())

logger.info('hi')
```

Which then outputs logs that look like this:

![](https://github.com/pinojs/pino-pretty/raw/master/demo.png)

dreamorosi:
Hi @krivachy, thanks for opening an issue and proposing the feature.

I can see where you are coming from and what is the use case you're suggesting. I haven't formed an opinion on the topic yet though; the comments that follow are aimed at having an open discussion to evaluate options.

[...]

Issue #1030, opened 2022-07-28
Feature request: disable logs while testing with jest


IslamWahid:
## Description of the feature request

**Problem statement**

While running jest I would like to suppress the logs on the console to make it readable and clean.

jest CLI is offering [`--silent`](https://jestjs.io/docs/cli#--silent) option to prevent tests from printing messages through the console, but it doesn't seem to work with the logger.

**Summary of the feature**

Suggestions:
- the logger is silenced when running in jest with `--silent`
- the logger is disabled when running in the test environment or when `NODE_ENV` is a `test`
- an option on the constructor to disable the logger

**Code examples**

```shell
jest --silent
```
**Benefits for you and the wider AWS community**

make the console readable and clean while running tests.

**Describe alternatives you've considered**

I have created a `__mocks__/@aws-lambda-powertools/logger.ts` with this setup

```typescript
export const Logger = jest.fn().mockImplementation(() => {
  return {
    debug: () => jest.fn(),
    info: () => jest.fn(),
    error: () => jest.fn(),
  };
});
```
and this solves the issue.

**Additional context**

dreamorosi:
Thank you @IslamWahid for opening the issue. I'm going to report part of the content of our discussion on the Slack community channel.

Based on what I could gather, when the `--silent` flag is present Jest patches the `console` [with a custom version](https://github.com/facebook/jest/blob/24e0472ac41ea03cdd89ffaea8c5f410ecf6054f/packages/jest-runner/src/runTest.ts#L134).

On our end, we also don't use the global `console` because of what described in #748. This most probably impedes Jest to suppress the logs coming from Powertools Logger.

Reproduction example (`lib/repro.ts`):
```ts
import { Logger } from "@aws-lambda-powertools/logger";

const logger = new Logger();

export const handler = async (event: any, _context: any) => {
  logger.info("Hello from Lambda");

  return {
    statusCode: 200,
    body: JSON.stringify({
      message: "Hello World",
    }),
  };
};
```
And corresponding test (`test/repro.test.ts`):
```ts
import { handler } from "../lib/repro";
import { ContextExamples } from "@aws-lambda-powertools/commons";

describe("MyUnitTest", () => {
  test("Lambda invoked successfully", async () => {
    const testEvent = { test: "test" };
    await handler(testEvent, ContextExamples.helloworldContext);
  });
});
```
Does in fact generate logs even when using the `--silent` flag:
```
 npm run test            

[...]

PR #1141, opened 2022-11-03
feature:(logger): pretty printing logs in local and non-prod environment

shdq:

Changes increase JSON.stringify indentation to 4 if the POWERTOOLS_DEV env variable has a truthy value.

Feature described and discussed in #751 

### Related issues, RFCs
RFC https://github.com/awslabs/aws-lambda-powertools/issues/86

**Issue number:**  #751 

### PR status

## Checklist

### Breaking change checklist

---

### DIFF
--- a/packages/logger/src/Logger.ts
+++ b/packages/logger/src/Logger.ts
@@ -6,6 +6,7 @@ import { LogItem } from './log';
 import cloneDeep from 'lodash.clonedeep';
 import merge from 'lodash.merge';
 import { ConfigServiceInterface, EnvironmentVariablesService } from './config';
+import { LogJsonIndent } from './types';
 import type {
   ClassThatLogs,
   Environment,
@@ -569,7 +570,10 @@ class Logger extends Utility implements ClassThatLogs {
 
     const consoleMethod = logLevel.toLowerCase() as keyof ClassThatLogs;
 
-    this.console[consoleMethod](JSON.stringify(log.getAttributes(), this.removeCircularDependencies()));
+    const isDevMode = this.getEnvVarsService().getDevMode();
+    const INDENTATION: LogJsonIndent = isDevMode ? LogJsonIndent.PRETTY : LogJsonIndent.COMPACT;
+
+    this.console[consoleMethod](JSON.stringify(log.getAttributes(), this.removeCircularDependencies(), INDENTATION));
   }
 
   /**
--- a/packages/logger/tests/unit/config/EnvironmentVariablesService.test.ts
+++ b/packages/logger/tests/unit/config/EnvironmentVariablesService.test.ts
@@ -220,4 +220,112 @@ describe('Class: EnvironmentVariablesService', () => {
 
   });
 
+  describe('Method: getDevMode', () => {
+
+    test('It returns true if the environment variable POWERTOOLS_DEV is "true"', () => {
+
+      // Prepare
+      process.env.POWERTOOLS_DEV = 'true';
+      const service = new EnvironmentVariablesService();
+
+      // Act
+      const value = service.getDevMode();
+
+      // Assess
+      expect(value).toEqual(true);
+    });
+
+    test('It returns true if the environment variable POWERTOOLS_DEV is "TRUE"', () => {
+
+      // Prepare
+      process.env.POWERTOOLS_DEV = 'TRUE';
+      const service = new EnvironmentVariablesService();
+
+      // Act
+      const value = service.getDevMode();
+
+      // Assess
+      expect(value).toEqual(true);
+    });
+
+    test('It returns true if the environment variable POWERTOOLS_DEV is "1"', () => {
+
+      // Prepare
+      process.env.POWERTOOLS_DEV = '1';
+      const service = new EnvironmentVariablesService();
+
+      // Act
+      const value = service.getDevMode();
+
+      // Assess
+      expect(value).toEqual(true);
+    });
+
+    test('It returns false if the environment variable POWERTOOLS_DEV is "false"', () => {
+
+      // Prepare
+      process.env.POWERTOOLS_DEV = 'false';
+      const service = new EnvironmentVariablesService();
+
+      // Act
+      const value = service.getDevMode();
+
+      // Assess
+      expect(value).toEqual(false);
+    });
+
+    test('It returns false if the environment variable POWERTOOLS_DEV is "FALSE"', () => {
+
+      // Prepare
+      process.env.POWERTOOLS_DEV = 'FALSE';
+      const service = new EnvironmentVariablesService();
+
+      // Act
+      const value = service.getDevMode();
+
+      // Assess
+      expect(value).toEqual(false);
+    });
+
+    test('It returns false if the environment variable POWERTOOLS_DEV is "0"', () => {
+
+      // Prepare
+      process.env.POWERTOOLS_DEV = '0';
+      const service = new EnvironmentVariablesService();
+
+      // Act
+      const value = service.getDevMode();
+
+      // Assess
+      expect(value).toEqual(false);
+    });
+
+    test('It returns false if the environment variable POWERTOOLS_DEV is NOT set', () => {
+
+      // Prepare
+      process.env.POWERTOOLS_DEV = 'somethingsilly';
+      const service = new EnvironmentVariablesService();
+
+      // Act
+      const value = service.getDevMode();
+
+      // Assess
+      expect(value).toEqual(false);
+    });
+
+    test('It returns false if the environment variable POWERTOOLS_DEV is "somethingsilly"', () => {
+
+      // Prepare
+      process.env.POWERTOOLS_DEV = 'somethingsilly';
+      const service = new EnvironmentVariablesService();
+
+      // Act
+      const value = service.getDevMode();
+
+      // Assess
+      expect(value).toEqual(false);
+    });
+
+  });
+
 });
--- a/packages/logger/src/config/ConfigServiceInterface.ts
+++ b/packages/logger/src/config/ConfigServiceInterface.ts
@@ -23,6 +23,13 @@ interface ConfigServiceInterface {
   getCurrentEnvironment(): string
 
   /**
+   * It returns the value of the POWERTOOLS_DEV environment variable.
+   *
+   * @returns {boolean}
+   */
+  getDevMode(): boolean
+
+  /**
    * It returns the value of the POWERTOOLS_LOGGER_LOG_EVENT environment variable.
    *
    * @returns {boolean}
--- a/packages/logger/src/config/EnvironmentVariablesService.ts
+++ b/packages/logger/src/config/EnvironmentVariablesService.ts
@@ -20,6 +20,7 @@ class EnvironmentVariablesService extends CommonEnvironmentVariablesService impl
   // Reserved environment variables
   private awsRegionVariable = 'AWS_REGION';
   private currentEnvironmentVariable = 'ENVIRONMENT';
+  private devModeVariable = 'POWERTOOLS_DEV';
   private functionNameVariable = 'AWS_LAMBDA_FUNCTION_NAME';
   private functionVersionVariable = 'AWS_LAMBDA_FUNCTION_VERSION';
   private logEventVariable = 'POWERTOOLS_LOGGER_LOG_EVENT';
@@ -43,6 +44,17 @@ class EnvironmentVariablesService extends CommonEnvironmentVariablesService impl
    */
   public getCurrentEnvironment(): string {
     return this.get(this.currentEnvironmentVariable);
+  }
+
+  /**
+   * It returns the value of the POWERTOOLS_DEV environment variable.
+   *
+   * @returns {boolean}
+   */
+  public getDevMode(): boolean {
+    const value = this.get(this.devModeVariable);
+
+    return value.toLowerCase() === 'true' || value === '1';
   }
 
   /**
--- a/packages/logger/src/types/Logger.ts
+++ b/packages/logger/src/types/Logger.ts
@@ -74,3 +74,8 @@ export {
   ConstructorOptions,
   HandlerOptions
 };
+
+export const enum LogJsonIndent {
+  PRETTY = 4,
+  COMPACT = 0,
+}
--- a/packages/logger/tests/unit/Logger.test.ts
+++ b/packages/logger/tests/unit/Logger.test.ts
@@ -11,7 +11,7 @@ import * as dummyEvent from '../../../../tests/resources/events/custom/hello-wor
 import { createLogger, Logger } from '../../src';
 import { EnvironmentVariablesService } from '../../src/config';
 import { PowertoolLogFormatter } from '../../src/formatter';
-import { ClassThatLogs } from '../../src/types';
+import { ClassThatLogs, LogJsonIndent } from '../../src/types';
 import { Context, Handler } from 'aws-lambda';
 import { Console } from 'console';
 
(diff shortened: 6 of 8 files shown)


packages/logger/src/Logger.ts
```diff
@@ -569,7 +570,10 @@ class Logger extends Utility implements ClassThatLogs {
 
     const consoleMethod = logLevel.toLowerCase() as keyof ClassThatLogs;
 
-    this.console[consoleMethod](JSON.stringify(log.getAttributes(), this.removeCircularDependencies()));
+    const isDevMode = this.getEnvVarsService().getDevMode();
```
dreamorosi, line 573: If implemented here this and the following line will be evaluated every time a log is about to be emitted.

Given that this is a configuration that has to be explicitly set by the developer and it lives in the environment, it's unlikely to change at runtime.

If this assumption is correct, what do you think about implementing this during class initialization in the [`setOptions` method](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/8b85db14e3cd4efbf6ec77472d97a63e036bdc2e/packages/logger/src/Logger.ts#L715) and store the indentation value as private class property (i.e. `private logIndentation: number`)?

This way the evaluation would be done once, during the init phase, and then whenever logging the value would simply be accessed.

What do you think?

dreamorosi (commented):
Hi @shdq thanks a lot for the code contribution, this is looking promising :) 

I have left a comment around the placement/evaluation order of the `isDevMode`/`INDENTATION` values in `packages/logger/src/Logger.ts`. I would like to hear your opinion on the idea.

Another point that I would like to bring on the table is whether or not the changes currently applied under `packages/logger/src/config/*` should stay there or they should be placed under `packages/commons/src/config`.

On one hand, at least for now, the only utility to use this variable & this logic from the `EnvironmentVariablesService` would be Logger. So having the logic under this package makes sense.

On the other hand, the original RFC talks about the `POWERTOOLS_DEV` variable influencing the behavior of multiple utilities, some of which don't exist today in this project. 

One could argue that in order to be future proof, the logic could be placed today in the `EnvironmentVariablesService` that is present in commons ([link](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/8b85db14e3cd4efbf6ec77472d97a63e036bdc2e/packages/commons/src/config/EnvironmentVariablesService.ts)) and that is extended by all other utilities.

Having the logic already in the commons package would allow to easily integrate the feature to other utilities in the future, however this might be a case of premature optimization since we don't know if/when these features will land and until then users who use only the Metrics or Tracer utilities will be having extra unused code as dependency.

Would be keen to hear your opinion on the topic.

shdq:
>  `packages/logger/src/config` vs `packages/commons/src/config`

I would stick to the RFC since there is a decision made that `POWERTOOLS_DEV` will influence multiple utilities, so I think the best place for it is in `packages/commons/src/config`. I don't see it as premature optimization since it is a shared setting, but the trade-off is the timeframe of adding support of `POWERTOOLS_DEV` into other packages.

When I wrote the plan in the issue I mentioned that I didn't find a "common" place. My fault that I didn't sync and see the recent change #1123 that you moved the config to `commons`. So I would have reconsidered the decision.

packages/logger/src/Logger.ts
shdq: 100% agree. It is always better to do "heavy" things once during the initialization phase than on every call. Thank you, I will make changes.

dreamorosi:
Hi @shdq thank you for the changes, I'll wait for your confirmation before reviewing again.

shdq:
@dreamorosi hello! If we stick to `packages/logger/src/config` it is ready to review.

dreamorosi:
Okay, let's keep it as it is now and if we add similar features to other utilities we'll move this logic in `commons` then.

I've already done a first pass of review and I see that the tests are passing. Before marking it as approved I'd like to test it locally to see how it would look from an user perspective. I'll try to do that by Monday end of morning (CEST).

In the meantime I'll also ask for a second review from other maintainers.

Thank you!

dreamorosi:
I have built the package from this branch & tested the following dummy lambda code:
```ts
import { Logger, injectLambdaContext } from "@aws-lambda-powertools/logger";

const logger = new Logger({ serviceName: "someName" });

  logger.info("test", { details: "foo" });

  try {
    throw new Error("error");
  } catch (err) {
    logger.error("some error", err as Error);
  }
}).use(injectLambdaContext(logger));
```

Then run the file with `export POWERTOOLS_DEV=true && node lib/test.mjs`, below the indented output:
<details>

```json
{
    "cold_start": true,
    "function_memory_size": null,
    "level": "INFO",
    "message": "test",
    "service": "someName",
    "timestamp": "2022-11-07T11:03:23.627Z",
    "details": "foo"
}
{
    "cold_start": true,
    "function_memory_size": null,
    "level": "ERROR",
    "message": "some error",
    "service": "someName",
    "timestamp": "2022-11-07T11:03:23.629Z",
    "error": {
        "name": "Error",
        "location": "file:///Users/aamorosi/Codes/powertools-playground/node_modules/@middy/core/index.js:85",
        "message": "error",
        "stack": "Error: error\n    at file:///Users/aamorosi/Codes/powertools-playground/lib/test.mjs:10:11\n    at runRequest (file:///Users/aamorosi/Codes/powertools-playground/node_modules/@middy/core/index.js:85:17)"
    }
}
```

</details>

docs/core/logger.md
```diff
@@ -104,6 +104,9 @@ Your Logger will include the following keys to your structured logging (default
 | **xray_trace_id**: `string` | `1-5759e988-bd862e3fe1be46a994272793`                                                                            | X-Ray Trace ID. This value is always presented in Lambda environment, whether [tracing is enabled](https://docs.aws.amazon.com/lambda/latest/dg/services-xray.html){target="_blank"} or not. Logger will always log this value. |
 | **error**: `Object`         | `{ name: "Error", location: "/my-project/handler.ts:18", message: "Unexpected error #1", stack: "[stacktrace]"}` | Optional - An object containing information about the Error passed to the logger                                                                                                                                                |
 
+???+ info
+    When `POWERTOOLS_DEV` environment variable is present and set to `"true"`, Logger will pretty-print log messages for easier readability. It is for debugging purposes and is not supposed to use in production environment.
```
dreamorosi, line 108:
```suggestion
    When `POWERTOOLS_DEV` environment variable is present and set to `"true"`, Logger will pretty-print log messages for easier readability. We recommend to use this setting only when debugging on local environments.
```

dreamorosi (approved):
Hi @shdq I'm ready to approve the PR, however I have left one minor comment in the docs section. Apologies for not flagging this earlier, I noticed how it reads only after rendering the docs locally.

dreamorosi (dismissed):

shdq:
Hey @dreamorosi! Thank you for the fast feedback loop and help with this PR. I enjoyed working on it. If you need help with other issues or maybe have a particular one in mind – let me know.

dreamorosi:
> Hey @dreamorosi! Thank you for the fast feedback loop and help with this PR. I enjoyed working on it. If you need help with other issues or maybe have a particular one in mind – let me know.

Glad to hear, I also had a very pleasant experience :smile: 

In case you're interested, I have two in mind:
- I would appreciate your review on this PR I made #1144 (context is in the linked issue)
- And then, after your PR is merged, there's this other issue that is somewhat related: #1030 (context is in [this message](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/1030#issuecomment-1305704449) / discussion)

As usual feel free to decline and/or choose different issues if the above don't interest you 😊 


PS: not sure if you're interested, but we have also a chat-based channel on Discord ([join here](https://discord.gg/B8zZKbbyET)), in case you want to discuss and you think it's easier feel free to post in the #typescript channel

docs/core/logger.md
```diff
@@ -104,6 +104,9 @@ Your Logger will include the following keys to your structured logging (default
 | **xray_trace_id**: `string` | `1-5759e988-bd862e3fe1be46a994272793`                                                                            | X-Ray Trace ID. This value is always presented in Lambda environment, whether [tracing is enabled](https://docs.aws.amazon.com/lambda/latest/dg/services-xray.html){target="_blank"} or not. Logger will always log this value. |
 | **error**: `Object`         | `{ name: "Error", location: "/my-project/handler.ts:18", message: "Unexpected error #1", stack: "[stacktrace]"}` | Optional - An object containing information about the Error passed to the logger                                                                                                                                                |
 
+???+ info
+    When `POWERTOOLS_DEV` environment variable is present and set to `"true"`, Logger will pretty-print log messages for easier readability. We recommend to use this setting only when debugging on local environments.
```
ijemmy, line 108: Can we be explicit in all possibilities here to avoid issue around "truthy" value (e.g. `"false"` (string) is `true`)? 

```suggestion
    When `POWERTOOLS_DEV` environment variable is present and set to `"true"` or `"1"`, Logger will pretty-print log messages for easier readability. We recommend to use this setting only when debugging on local environments.
```

Given that the default value is `0` (in `docs/index.md`), I would be passing `1` here.

ijemmy (changes requested):
@dreamorosi  I'm surprised that they use

flochaz (approved):
Ok after @ijemmy requested change

flochaz (approved):

flochaz:
e2e are failing :/ https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/3418100576/jobs/5690304692

dreamorosi:
> e2e are failing :/ https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/3418100576/jobs/5690304692

We need to take a look at them, they're failing in `Metrics` which were not changed in this PR

dreamorosi:
We have a PR open #1146 that should fix the issue highlighted by the e2e tests.

dreamorosi:
Hi @shdq, after discussing this with the other maintainers I think we should tweak the function that resolves the value of the `POWERTOOLS_DEV` function so that it's closer to the implementation that the Python's version of the library has:
```py
value = value.lower()
if value in ("1", "y", "yes", "t", "true", "on"):
    return True
if value in ("0", "n", "no", "f", "false", "off"):
    return False
```

Basically we should accept more values for `true` & check other values for `false`.

Also, we should update the value in the docs to say `false`, I have [opened a PR](https://github.com/awslabs/aws-lambda-powertools-python/pull/1695) in the Python's repo so that it's also changed to `false`.

Thanks for your understanding!

ijemmy:
@shdq  Sorry about this last minute after you have addressed everything. 

I was the one who raised this discrepancy to Python team. It sounds wrong to me that `POWERTOOLS_DEV` use `0` instead of `false`. This could lead to configuration error for some users who aren't familiar with truthy evaluation.  So we decided to change this.

I am happy to make an adjustment + add test cases myself if you don't have time for this. Just let me know.

shdq:
Hello, @ijemmy I'll make the changes!

I'm actually looking at the python implementation of the function that resolves the value and in the situation, when the env var value is "somethingsilly" it returns `None`.

Here we have a test for [this](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/103a3a19ee0a2089738864ce2a93df1c2eb44791/packages/logger/tests/unit/config/EnvironmentVariablesService.test.ts#L174) 

@dreamorosi @ijemmy what do you think is I implement it as a switch statement with `default: false` ?

shdq:
Another question:

There is the same check for the `POWERTOOLS_LOGGER_LOG_EVENT` variable. Should it also be truthy? In the docs, it says `Whether to log or not the incoming event when using the decorator or middleware. Supported values are: true, or false, disabled by default`. But it also checks for `1`.

And there was a function `isValueTrue`, that was removed after [this refactoring](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/1123/commits/8d3fa287a25b3385188d8f404259c5b704c66290#). What do you think, maybe it makes sense to make a common function for checking for `true` and use it for both variables? Or across all packages. How it is checking in `Tracer` and `Metrics`?

dreamorosi:
> There is the same check for the `POWERTOOLS_LOGGER_LOG_EVENT` variable. Should it also be truthy? In the docs, it says `Whether to log or not the incoming event when using the decorator or middleware. Supported values are: true, or false, disabled by default`. But it also checks for `1`.

Possibly yes, since this new function is going to be implemented (at least for now) in `packages/logger/src/config/EnvironmentVariablesService.ts` it makes sense to do the same check there.

> And there was a function `isValueTrue`, that was removed after [this refactoring](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/1123/commits/8d3fa287a25b3385188d8f404259c5b704c66290#).

I opted to remove that function because at the time that was the only usage.

> What do you think, maybe it makes sense to make a common function for checking for `true` and use it for both variables? Or across all packages. How it is checking in `Tracer` and `Metrics`?

At the moment they're not really checking for truth values but exactly and only `true`, however I agree that there's room for improvement and extend the same checks to these other utilities.

For now I would keep the change scoped to this PR. Later if you want, in a separate PR after this one, we could move up the logic in the `packages/commons` as well as potentially address #165 

Let's discuss this again once this is merged.

shdq:
@dreamorosi sorry for being dense, but I want to make sure that I got it right. I need to implement a function `isValueTrue` in `packages/logger/src/config/EnvironmentVariablesService.ts` and apply it to both `POWERTOOLS_DEV` and `POWERTOOLS_LOGGER_LOG_EVENT` checks. Correct?

Am I right, that we treat any other than `"1", "y", "yes", "t", "true", "on"` values as `false`? So it is not needed to explicitly check for falsy values.

Speaking of implementation details, I have two options:

```
switch (value) {
  case '1':
  case 'y':
  case 'yes':
  case 't':
  case 'true':
  case 'on':
    return true;
  default:
    return false;
}
```
Another option is to put it in an array of truthy values:
```
const truthyValues: string[] = ["1", "y", "yes", "t", "true", "on"];
return truthyValues.includes(value); // possible to use indexOf(), but it doesn't matter since it is going to be transpilled.
```
I like it, it is short and concise, but it uses an array. What do you think?

Thank you for your time and help!

shdq:
I hope I didn't mess up with the previous push. And I can't find the commit to the docs with the default `false` value.

Should we add tests for `isValueTrue`?

If any other changes are needed I'm happy to make them!

dreamorosi:
> I hope I didn't mess up with the previous push. And I can't find the commit to the docs with the default `false` value.

Don't worry, it happens! As far as I can see the diff looks okay and anyway we'll squash before merging so the single commits don't matter as much.

> Should we add tests for `isValueTrue`?

Yes please, would be great to have them so we can detect regressions in the future. Maybe you could reduce the ones for `getDevMode` and convert some of them to test `isValueTrue` instead.

For `getDevMode` I would test only `true`, `false`, not set, and `somethingsilly`. The others would go to `isValueTrue`. For the latter I don't think there's no need to test every single combination: i.e. it's okay to test `TRUE` and `false` without the need of having to write tests for `TRUE`, `true`, `FALSE`, and `False` - we can assume that the `String.toLowerCase()` from the standard lib works.

Just make sure each string (case-insensitive) is represented once to test we support it.

> If any other changes are needed I'm happy to make them!

As far as I can see, and from my side, that would be all 🙂

packages/logger/tests/unit/config/EnvironmentVariablesService.test.ts
```diff
@@ -119,105 +119,149 @@ describe('Class: EnvironmentVariablesService', () => {
...
+      [ 'TRUE', true ],
+      [ 'on', true ],
+      [ '', false ],
+      [ 'false', false ],
+      [ 'fasle', false ],
+      [ 'somethingsilly', false ],
+      [ '0', false ]
+    ];
+
+    test.each(valuesToTest)('It takes string "%s" and returns %s', (input, output) => {
```
dreamorosi, line 256: Neat, I like it

dreamorosi (approved):
Looks good to me, thank you for the attention to details 💯

dreamorosi:
E2E tests are passing: https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/3437044442

Issue #1142, opened 2022-11-05
Documentation Improvement: inconsistency Tracer.ts example code - Tracer Website Docs


niko-achilles:
### Expected Behaviour

consistency of  generated Docs for `Tracer` -> `Lambda Handler` -> `Manual` 
Links: 
- https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/tracer/#lambda-handler
- https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/docs/core/tracer.md

with

`Tracer.ts` code example snippet  in IDE editor
Link:  https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/tracer/src/Tracer.ts#L80

### Current Behaviour

Using my IDE and reading the documentation on how to use the Tracer i read 
that i should additionally `import { Segment } from 'aws-xray-sdk-core';`
See line: https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/tracer/src/Tracer.ts#L80

This is inconsistent with the docs generated on the Website:
See `Tracer` -> `Lambda Handler` -> `Manual` 

Links:
- https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/tracer/#lambda-handler
- https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/docs/core/tracer.md

### Code snippet

```JavaScript
see expected and current behavior description
```

### Possible Solution

see expected behavior

### Steps to Reproduce

import Tracer in IDE and follow the example without switching to online Docs of Powertools.

### AWS Lambda Powertools for TypeScript version

latest

### AWS Lambda function runtime

16.x

### Packaging format used

Lambda Layers

dreamorosi:
Hi, thank you for opening this issue. We'll be checking the documentation and the docstrings to make sure they match and any update will land in the next release.

However this is not an issue that affects runtime behavior so I've removed the bug categorization and marked this as documentation improvement.

niko-achilles:
ok, i saw that you will look at things for maintenance and found the time to write about this issue, so that the website docs and code example with IDE support is the same.

dreamorosi:
I have reviewed the examples and opened a PR #1144, if you check the PR and notice any other issue I have missed please leave a comment in the PR.

I'll report back once it's merged.

dreamorosi:
Improvements to the docs released in `v1.4.1`.

PR #1144, opened 2022-11-07
docs(all): update & revise docstring examples

dreamorosi:

Following the report in #1142 a community member highlighted the fact that some examples in the docstrings of the packages (mainly Tracer) were inconsistent with the ones in the published documentation.

In response to the issue I went through the examples and checked that they were appropriate and consistent, this PR contains the changes related to this review.

Changes:
- Added middleware example to Metrics
- Standardized order of usages in main docstring: middleware, decorator, manual
- Cleaned up imports
- Standardized usage of single quotes 

### How to verify this change

Check diff.

### Related issues, RFCs

**Issue number:** #1142 

### PR status

## Checklist

### Breaking change checklist

N/A

---

### DIFF
--- a/packages/metrics/src/Metrics.ts
+++ b/packages/metrics/src/Metrics.ts
@@ -30,11 +30,33 @@ const DEFAULT_NAMESPACE = 'default_namespace';
  *   * Context manager to create a one off metric with a different dimension
  *
  * ## Usage
+ * 
+ * ### Functions usage with middleware
+ * 
+ * Using this middleware on your handler function will automatically flush metrics after the function returns or throws an error.
+ * Additionally, you can configure the middleware to easily:
+ * * ensure that at least one metric is emitted before you flush them
+ * * capture a `ColdStart` a metric
+ * * set default dimensions for all your metrics
+ * 
+ * @example
+ * ```typescript
+ * import { Metrics, logMetrics } from '@aws-lambda-powertools/metrics';
+ * import middy from '@middy/core';
+ * 
+ * const metrics = new Metrics({ namespace: 'serverlessAirline', serviceName: 'orders' });
+ * 
+ * const lambdaHandler = async (_event: any, _context: any) => {
+ *   ...
+ * };
+ * 
+ * export const handler = middy(lambdaHandler).use(logMetrics(metrics));
+ * ```
  *
  * ### Object oriented way with decorator
  *
  * If you are used to TypeScript Class usage to encapsulate your Lambda handler you can leverage the [@metrics.logMetrics()](./_aws_lambda_powertools_metrics.Metrics.html#logMetrics) decorator to automatically:
- *   * create cold start metric
+ *   * capture a `ColdStart` metric
  *   * flush buffered metrics
  *   * throw on empty metrics
  *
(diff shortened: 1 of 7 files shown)


packages/metrics/src/Metrics.ts
```diff
@@ -30,35 +30,57 @@ const DEFAULT_NAMESPACE = 'default_namespace';
...
  *
- *   // FYI: Decorator might not render properly in VSCode mouse over due to https://github.com/microsoft/TypeScript/issues/39371 and might show as *@metrics* instead of `@metrics.logMetrics`
+ *   // FYI: Decorator might not render properly in VSCode mouse over due to https://github.com/microsoft/TypeScript/issues/47679 and might show as *@metrics* instead of `@metrics.logMetrics`
  *
  *   @metrics.logMetrics({ captureColdStartMetric: true, throwOnEmptyMetrics: true })
- *   public handler(_event: any, _context: Context, _callback: Callback<any>): void | Promise<void> {
- *    // ...
- *    metrics.addMetric('test-metric', MetricUnits.Count, 10);
- *    // ...
+ *   public handler(_event: any, _context: any): Promise<void> {
```
flochaz, line 76: why do we push of `any` type for Context ? I'd prefer to keep `Context` type

packages/metrics/src/Metrics.ts
```diff
@@ -164,11 +186,10 @@ class Metrics extends Utility implements MetricsInterface {
    *
    * ```typescript
    * import { Metrics } from '@aws-lambda-powertools/metrics';
-   * import { Context } from 'aws-lambda';
    *
-   * const metrics = new Metrics({ namespace:'serverlessAirline', serviceName:'orders' });
+   * const metrics = new Metrics({ namespace: 'serverlessAirline', serviceName: 'orders' });
    *
-   * export const handler = async (event: any, context: Context): Promise<void> => {
+   * export const handler = async (event: any, _context: any): Promise<void> => {
```
flochaz, line 192: Why removing Context type ?

packages/metrics/src/middleware/middy.ts
```diff
@@ -2,6 +2,33 @@ import type { Metrics } from '../Metrics';
 import type middy from '@middy/core';
 import type { ExtraOptions } from '../types';
 
+/**
+ * A middy middleware automating capture of metadata and annotations on segments or subsegments for a Lambda Handler.
```
flochaz, line 6: hmmm seems mixing tracing  and logging middleware

flochaz (changes requested):

packages/metrics/src/Metrics.ts
dreamorosi: I hear you. Let me explain why, but I'm open to put it back.

I removed it here because these were the only examples in the whole repo that have it. Even the published docs don't have it.

Removing it here was a way of standardizing the examples. 

Now, we could go two ways:
- put the context type in all examples and docstrings
- leave as it is, but in the docs put a tip/note about `@types/aws-lambda` and explain it can be installed as `devDependency`

Ultimately, since we don't bundle the dependency, having it here in the IDE tooltip could lead to one of two outcomes:
- you copy/write the content of the tooltip and it doesn't work because you don't have the dep installed
- you might think the dependency is mandatory to use Powertools

What do you think?

flochaz: Make sense. Indeed the copy paste that does not work is really frustrating. deal . approving

flochaz (dismissed):

ijemmy (approved):

dreamorosi:
Since I have addressed the comments from @flochaz I'm going to merge this PR now instead of waiting for him to click the "approve" button.

Issue #1159, opened 2022-11-11
Maintenance: add Idempotency utility to unit tests


dreamorosi:
### Summary

Given that #1110 is now merged it's now a good time to add the new utility to the unit testing ran at different points of the dev lifecycle.

Additionally, there were some inconsistencies in the `package.lock` dependencies that prevented some other CI processes from completing successfully ([example of failed run](https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/3444715921)).

### Why is this needed?

To increase confidence in the code to be merged and to prevent regressions.

### Which area does this relate to?

Tests, Idempotency

### Solution

See linked PR 

### Acknowledgment

- [X] This request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/#tenets)
- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [Java](https://github.com/awslabs/aws-lambda-powertools-java/)

dreamorosi:
Closing this in favor of #1100

PR #1160, opened 2022-11-11
tests(idempotency): add utility to workspace unit tests and CI

dreamorosi:

As described in #1159, this PR makes it so that the existing unit tests for the Idempotency utility are now run as part of the PR checks as well as on merge & pre-push hooks. 

Additionally, this PR introduces the following changes:
- Minor styling changes to package structure and exports
- Introduction of one additional type
- Minor refactor of function that creates the idempotencyKey prefix for DynamoDB storage
- Rename DynamoDbPersistenceLayer to DynamoDBPersistenceLayer (to follow product name)
- Added missing unit test cases for implemented logic to reach 100% coverage
- Temporarily excluded `makeFunctionIdempotent` from test coverage (to be reintroduced once implemented)

### How to verify this change

Check the results of the automated checks below this PR.

### Related issues, RFCs

**Issue number:** #1159

### PR status

## Checklist

### Breaking change checklist

---


packages/idempotency/src/persistence/PersistenceLayer.ts
```diff
@@ -10,22 +11,30 @@ abstract class PersistenceLayer implements PersistenceLayerInterface {
...
-    this.functionName = this.getEnvVarsService().getLambdaFunctionName() + '.' + functionName;
+
+  /**
+   * Configures the persistence layer by passing the name of the idempotent function. This will be used
+   * in the prefix of the idempotency key
+   * 
+   * @param {PersistenceLayerConfigureOptions} options - configuration object for the persistence layer
+   */
+  public configure(options?: PersistenceLayerConfigureOptions): void {
+    if (options?.functionName && options.functionName.trim() !== '') this.idempotencyKeyPrefix = `${this.idempotencyKeyPrefix}.${options.functionName}`;
```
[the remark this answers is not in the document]
ijemmy: (Question) 
Now that I review this function again. It seems like it uses both two lambda names (one from environment variable  + another from options object).   

I looked at Python (see the image below), it is doing the same thing.  Do you know the root cause? (If not, don't worry about it, I'll check with Python team myself)

image

packages/idempotency/tests/helpers/populateEnvironmentVariables.ts
```diff
@@ -1,2 +1,7 @@
 // Reserved variables
-process.env.AWS_REGION = 'us-east-1';
+process.env._X_AMZN_TRACE_ID = '1-abcdef12-3456abcdef123456abcdef12';
+process.env.AWS_LAMBDA_FUNCTION_NAME = 'my-lambda-function';
+process.env.AWS_EXECUTION_ENV = 'nodejs16.x';
+process.env.AWS_LAMBDA_FUNCTION_MEMORY_SIZE = '128';
+process.env.AWS_REGION = 'eu-west-1';
+process.env._HANDLER = 'index.handler';
```
ijemmy, line 7: (Question) Do we really need all of these env vars?  Wouldn't only `AWS_LAMBDA_FUNCTION_NAME` be enough?

packages/idempotency/tests/unit/persistence/DynamoDbPersistenceLayer.test.ts
```diff
@@ -29,31 +34,77 @@ describe('Class: DynamoDbPersistenceLayer', () => {
...
+        statusAttr: testDynamoDBPersistenceLayerOptions.statusAttr,
+        expiryAttr: testDynamoDBPersistenceLayerOptions.expiryAttr,
+        inProgressExpiryAttr: testDynamoDBPersistenceLayerOptions.inProgressExpiryAttr,
+        dataAttr: testDynamoDBPersistenceLayerOptions.dataAttr
+      }));
+
+    });
+
+  });
+
```
ijemmy, line 89: All member variables in this class are private. Should we break the abstraction here?

image

The private value are indirectly tested in the `_putRecord()` and other protected fields. Those test cases would detect an issue if the private values are incorrect.

packages/idempotency/src/persistence/PersistenceLayer.ts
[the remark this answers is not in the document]
dreamorosi: Yes, I was very confused by this as well and during my first refactor (not pushed) I actually removed this, but then reintroduced it.

From the way I understand it, the first one (that comes from the environment in our implementation) is the name of the Lambda resource as it appears in AWS Lambda.
The second one instead is a name that can be set and refers to the function being made idempotent in your code.

Imagine that you have a function called `MyLambda` like this:
```ts
const processPayment = async () => {
  // ... do stuff
};

  const result = makeIdempotentFn(await processPayment());
};
```

In this case, the idempotency key would be `MyLambda.processPayment#${hash}`.

If my understanding is correct, we cannot rely on the resource name only because there might be multiple functions in the runtime code that customers want to make idempotent.

What do you think?

packages/idempotency/tests/unit/persistence/DynamoDbPersistenceLayer.test.ts
dreamorosi: I see your point.

The reason why I introduced these tests is primarily to reach 100% coverage.

If we remove this test, coverage drops:
image

and the reason, I think, is because we have conditional statements in the `constructor`:
```ts
this.tableName = constructorOptions.tableName;
this.keyAttr = constructorOptions.keyAttr ?? 'id';
this.statusAttr = constructorOptions.statusAttr ?? 'status';
this.expiryAttr = constructorOptions.expiryAttr ?? 'expiration';
this.inProgressExpiryAttr = constructorOptions.inProgressExpiryAttr ?? 'in_progress_expiry_attr';
this.dataAttr = constructorOptions.dataAttr ?? 'data';
```

Since in all other test cases, including `_putRecord` we are always using the defaults, we never testing the left-side of the `??` (except for `tableName`).

Now, I could have an additional test to one of the other methods (i.e. `_putRecord`) and pass some values when initializing the class, however I saw that test as really testing the constructor, and so I added its own case.

If there's another way I'm happy to change it though, I have to admit I went this way also because this is how I've done it in Tracer.

ijemmy (approved):
I think it will be beneficial to have someone from who's working on Idempotency feature review this as it might affect them. I'll talk with them.

dreamorosi:
I will be merging this PR so that it can be included in this week's release.

In case of any comment please feel free to either leave one here or open an issue, we'll address any change in a new PR.

PR #748, opened 2022-04-08
fix(logger): change logging to use stdout

dreamorosi:
## Description of your changes

As reported by user @olanb7 on the community Slack and in #747 before this PR Logger would output logs like this: 
```
2022-04-08T10:58:44.811Z	b6511ee9-4873-404e-b60c-a23cb45d3ff2	INFO {"cold_start":false,"function_arn":"arn:aws:lambda:eu-west-2:12345:function:test","function_memory_size":1024,"function_name":"test","function_request_id":"b6511ee9-4873-404e-b60c-a23cb45d3ff2","level":"INFO","message":"test message","service":"test","timestamp":"2022-04-08T10:58:44.811Z"}
```

These logs, although containing a structured JSON, are not JSON as they are always prefixed by `2022-04-08T10:58:44.811Z	b6511ee9-4873-404e-b60c-a23cb45d3ff2	INFO`. 

This also confirmed [by a line in the Lambda docs](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-logging.html#node-logging-output) (emphasis mine):
> The Node.js runtime logs the START, END, and REPORT lines for each invocation. **It adds a timestamp, request ID, and log level to each entry logged by the function**.

This behaviour works well with CloudWatch but the presence of these prefixes was an issue for customers wanting to use these structured logs elsewhere (both 3rd party services and AWS Managed Grafana).

From this PR onwards Logger will be using its own internal and private version of the [`Console` class](https://nodejs.org/api/console.html) that only outputs the structured JSON log without any prefix:
```
{"cold_start":false,"function_arn":"arn:aws:lambda:eu-west-2:12345:function:test","function_memory_size":1024,"function_name":"test","function_request_id":"b6511ee9-4873-404e-b60c-a23cb45d3ff2","level":"INFO","message":"test message","service":"test","timestamp":"2022-04-08T10:58:44.811Z"}
```

Aside from 3-lines change in the Logger class, all the rest of the changes in this PR are related to the unit and integration tests.

### How to verify this change

See the results of checks inside this PR and e2e tests (to be run).

Also see screenshot below:
image

### Related issues, RFCs

[#747](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/747)

### PR status

***Is this ready for review?:*** YES  
***Is it a breaking change?:*** NO

## Checklist

- [x] [My changes meet the tenets criteria](https://awslabs.github.io/aws-lambda-powertools-python/#tenets)
- [x] I have performed a *self-review* of my own code
[...]

PR #1165, opened 2022-11-14
feat(logger): disable logs while testing with jest --silent in dev env

shdq:

It conditionally initializes console property as an instance of the internal version of Console() class (PR #748) or as the global node console if the `POWERTOOLS_DEV' env variable is set and has a truthy value.

### Related issues, RFCs

**Issue number:** 
#1030 

### PR status

## Checklist

### Breaking change checklist

---

### DIFF
--- a/docs/core/logger.md
+++ b/docs/core/logger.md
@@ -975,3 +975,7 @@ This is a Jest sample that provides the minimum information necessary for Logger
 ### Suppress logs with Jest
 
 When unit testing your code with [Jest](https://jestjs.io) you can use the `POWERTOOLS_DEV` environment variable in conjunction with the Jest `--silent` CLI option to suppress logs from Logger.
+
+```bash title="Disabling logs while testing with Jest"
+export POWERTOOLS_DEV=true && npx jest --silent
+```
--- a/docs/core/logger.md
+++ b/docs/core/logger.md
@@ -974,4 +974,4 @@ This is a Jest sample that provides the minimum information necessary for Logger
 
 ### Suppress logs with Jest
 
-When unit testing your code with the `POWERTOOLS_DEV` environment variable enabled, you could run Jest with the `--silent` CLI option to make the console readable and clean. It prevents printing messages through the console from tests and logs as well.
+When unit testing your code with [Jest](https://jestjs.io) you can use the `POWERTOOLS_DEV` environment variable in conjunction with the Jest `--silent` CLI option to suppress logs from Logger.
--- a/docs/core/logger.md
+++ b/docs/core/logger.md
@@ -971,3 +971,7 @@ This is a Jest sample that provides the minimum information necessary for Logger
 
 !!! tip
     If you don't want to declare your own dummy Lambda Context, you can use [`ContextExamples.helloworldContext`](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/commons/src/samples/resources/contexts/hello-world.ts#L3-L16) from [`@aws-lambda-powertools/commons`](https://www.npmjs.com/package/@aws-lambda-powertools/commons).
+
+### Suppress logs with Jest
+
+When unit testing your code with the `POWERTOOLS_DEV` environment variable enabled, you could run Jest with the `--silent` CLI option to make the console readable and clean. It prevents printing messages through the console from tests and logs as well.
--- a/packages/logger/src/Logger.ts
+++ b/packages/logger/src/Logger.ts
@@ -782,7 +782,7 @@ class Logger extends Utility implements ClassThatLogs {
     } = options;
 
     this.setEnvVarsService();
-    // order is important, it uses EngVarsService()
+    // order is important, it uses EnvVarsService()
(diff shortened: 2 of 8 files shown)


packages/logger/src/Logger.ts
```diff
@@ -766,6 +782,9 @@ class Logger extends Utility implements ClassThatLogs {
     } = options;
 
     this.setEnvVarsService();
+    // order is important, it uses EngVarsService()
+    this.initConsole();
```
dreamorosi, line 786: Should we call this `setConsole` like the other methods? Mostly for consistency but also because we are not always initializing. Thoughts?

shdq: I agree, naming is one of two hard things in CS 😀

dreamorosi:
Also I think it could be interesting to surface this new behavior in the docs, specifically in [this section](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/logger/#testing-your-code).

We could add a new sub section under _Testing your code_

docs/core/logger.md
```diff
@@ -971,3 +971,7 @@ This is a Jest sample that provides the minimum information necessary for Logger
 
 !!! tip
     If you don't want to declare your own dummy Lambda Context, you can use [`ContextExamples.helloworldContext`](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/commons/src/samples/resources/contexts/hello-world.ts#L3-L16) from [`@aws-lambda-powertools/commons`](https://www.npmjs.com/package/@aws-lambda-powertools/commons).
+
+### Suppress logs with Jest
+
+When unit testing your code with the `POWERTOOLS_DEV` environment variable enabled, you could run Jest with the `--silent` CLI option to make the console readable and clean. It prevents printing messages through the console from tests and logs as well.
```
dreamorosi, line 977:
```suggestion
When unit testing your code with [Jest](https://jestjs.io) you can use the `POWERTOOLS_DEV` environment variable in conjunction with the Jest `--silent` CLI option to suppress logs from Logger.
```

dreamorosi: And then let's put a CLI example of how to run a test with the env variable & ``--silent flag, kind of like this: https://awslabs.github.io/aws-lambda-powertools-python/2.2.0/core/logger/#pytest-live-log-feature

shdq:
What about testing? Is there a way to unit test this "silent" behavior? I only came up with the idea to somehow check what type of `console` object we use with `POWERTOOL_DEV` on and off.

dreamorosi:
> What about testing? Is there a way to unit test this "silent" behavior? I only came up with the idea to somehow check what type of `console` object we use with `POWERTOOL_DEV` on and off.

I don't think we should test Jest's behavior. From my point of view we just need to test that when `POWERTOOL_DEV` is enabled or disabled the value of `Logger.console` is a new `Console` object or the global one.

shdq:
> From my point of view we just need to test that when `POWERTOOL_DEV` is enabled or disabled the value of `Logger.console` is a new `Console` object or the global one.

This is what I meant actually.

I tested the global node object when `POWERTOOLS_DEV` is set. But for the other option, since instances of a class are not the same I presumed the opposite:

```
expect(logger).not.toEqual({
  ...logger,
  console: console,
});
```

@dreamorosi  what do you think?

dreamorosi (commented):
Thank you @shdq!

Tomorrow morning I will test the code locally (same process as the one in the issue) & see how it renders in the documentation - then if it's all good I'll approve & we can merge :)

packages/logger/src/Logger.ts
```diff
@@ -766,6 +782,8 @@ class Logger extends Utility implements ClassThatLogs {
     } = options;
 
     this.setEnvVarsService();
+    // order is important, it uses EngVarsService()
```
dreamorosi, line 785:
```suggestion
    // order is important, it uses EnvVarsService()
```

dreamorosi: Why is the order important here?

dreamorosi (commented):
Hi @shdq, I have reviewed the code and it looks good. I also have tested this version in a separate codebase & verified that the `--silent- flag is indeed respected.

I'm going to run the integration tests on this branch now.

In the meanwhile I have left a comment/question about one of the sections of your changes.

packages/logger/src/Logger.ts
shdq: I wanted to make the order of methods alphabetically sorted, like the linting rule for methods. But `setConsole` uses `setEnvVarsService`, so it won't work.

dreamorosi: Oh I see, makes sense! Thanks.

Let's correct the typo in the comment (see suggestion) and we can resolve this conversation.

dreamorosi (approved):
Integration tests are passing: https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/3469565442

@shdq thank you so much for all the hard work here!

ijemmy (approved):

PR #1168, opened 2022-11-16
feat(parameters): added `BaseProvider` class

dreamorosi:

This PR introduces a new package (`parameters`) in the npm workspace as well as the `BaseProvider` for the new utility and its unit tests.

The changes related to the new workspace package include:
- New `pacakges/parameters` folder with `package.json`
- TSConfig files
- Eslint file
- Jest test harness, and jestconfig file
- README file

The changes related to the `BaseProvider` include the content of the `src` and `tests` directories:
- New `BaseProvider` class
- New types and exceptions to support the class
- Unit tests for `BaseProvider`

> **Note**
> I have left a few comments/questions that I would appreciate reviewers to engage with during the review

Please don't take in account the content of the `README.md` file nor the docstrings (or lack thereof) for the various features, both items will be addressed in their own PRs.

### How to verify this change

See unit tests added as part of this PR.

### Related issues, RFCs

**Issue number:** #1173 

### PR status

## Checklist

### Breaking change checklist

---


packages/parameters/src/BaseProvider.ts
```diff
@@ -0,0 +1,214 @@
...
+    this.ttl = timeNow.setSeconds(timeNow.getSeconds() + maxAge);
+  }
+
+  public isExpired(): boolean {
+    return this.ttl < Date.now();
+  }
+}
+
+abstract class BaseProvider implements BaseProviderInterface {
+  public store: Map<string, ExpirableValue> = new Map;
```
dreamorosi, line 48: Question: Should this be `public` or `private`?

packages/parameters/src/BaseProvider.ts
```diff
@@ -0,0 +1,214 @@
...
+    if (value && configs.transform) {
+      value = transformValue(value, configs.transform, true);
+    }
+
+    if (value) {
+      this.addToCache(key, value, configs.maxAge);
+    }
+
+    // TODO: revisit return type once providers are implemented, it might be missing binary when not transformed
+    return value;
```
dreamorosi, line 104: As mentioned in the comment here, I stayed generic with types because at this stage I'm not yet sure what each provider/AWS SDK (from `_get`) will return.

I have opened #1172 to track this work.

packages/parameters/src/BaseProvider.ts
```diff
@@ -0,0 +1,214 @@
...
+
+    if (Object.keys(values) && configs.transform) {
+      values = transformValues(values, configs.transform, configs.throwOnTransformError);
+    }
+
+    if (Array.from(Object.keys(values)).length !== 0) {
+      this.addToCache(key, values, configs.maxAge);
+    }
+
+    // TODO: revisit return type once providers are implemented, it might be missing something
```
dreamorosi, line 132: I have opened #1172 to track this work.

packages/parameters/src/BaseProvider.ts
```diff
@@ -0,0 +1,214 @@
...
+
+      return value.isExpired();
+    }
+    
+    return true;
+  }
+
+}
+
+// TODO: revisit `value` type once we are clearer on the types returned by the various SDKs
```
dreamorosi, line 170: I have opened #1172 to track this work.

packages/parameters/src/BaseProvider.ts
```diff
@@ -0,0 +1,214 @@
...
+    ) {
+      return JSON.parse(value) as Record<string, unknown>;
+    } else if (
+      (normalizedTransform === TRANSFORM_METHOD_BINARY ||
+      (normalizedTransform === 'auto' && key.toLowerCase().endsWith(`.${TRANSFORM_METHOD_BINARY}`))) &&
+      typeof value === 'string'
+    ) {
+      return new TextDecoder('utf-8').decode(fromBase64(value));
+    } else {
+      // TODO: revisit this type once we are clearer on types returned by SDKs
```
dreamorosi, line 187: I have opened #1172 to track this work.

packages/parameters/src/BaseProvider.ts
```diff
@@ -0,0 +1,214 @@
...
+   * 
+   * It returns true if the key is expired or not present in the cache.
+   * 
+   * @param {string} key - Stringified representation of the key to retrieve
+   */
+  private hasKeyExpiredInCache(key: string): boolean {
+    const value = this.store.get(key);
+    if (value) {
+      if (value.isExpired()) {
+        this.store.delete(key);
```
dreamorosi, line 157: **As far as I can tell**, the Python implementation **doesn't** clear expired items from the cache.

Can you think of any case in which doing that might become an issue?

packages/parameters/src/BaseProvider.ts
ijemmy: I'll go with `private` first and broaden later if needed.

packages/parameters/src/BaseProvider.ts
```diff
@@ -0,0 +1,216 @@
...
+  public isExpired(): boolean {
+    return this.ttl < Date.now();
+  }
+}
+
+abstract class BaseProvider implements BaseProviderInterface {
+  public store: Map<string, ExpirableValue> = new Map;
+
+  public constructor () {
+    this.store = new Map();
```
ijemmy, line 51: (non-blocking) Just double check. This is redundant with the initialize in line 48. Is it intentional?

If it is (or it can complicate the type if we leave it possibly having `undefined` value), feel free to resolve this comment.

packages/parameters/src/BaseProvider.ts
```diff
@@ -0,0 +1,214 @@
...
+  public maxAge: number = DEFAULT_MAX_AGE_SECS;
+  public sdkOptions?: unknown;
+  public transform?: TransformOptions;
+
+  public constructor(options: GetOptionsInterface = {}) {
+    Object.assign(this, options);
+  }
+}
+
+class GetMultipleOptions implements GetMultipleOptionsInterface {
```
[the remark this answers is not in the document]
ijemmy: Given this file's size, I would say so. But it is not critical.

packages/parameters/src/BaseProvider.ts
```diff
@@ -0,0 +1,214 @@
...
+   * 
+   * Since the number of supported transform is small and the probability that a given parameter will always be used in a specific transform,
+   * this should be an acceptable tradeoff.
+   * 
+   * @param {string} name - Parameter name
+   * @param {GetOptionsInterface} options - Options to configure maximum age, trasformation, AWS SDK options, or force fetch
+   */
+  public async get(name: string, options?: GetOptionsInterface): Promise<undefined | string | Record<string, unknown>> {
+    const configs = new GetOptions(options);
+    const key = [ name, configs.transform ].toString();
```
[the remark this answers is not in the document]
ijemmy: As long as the test cases are rigorous enough, this should be ok. I cannot see how we can run into having a duplicated here, can we?

packages/parameters/src/BaseProvider.ts
ijemmy: No. We just trade time for space here. The operation is O(1) so I don't think there is any issue to delete it now. 

The problem is more on doing more than one thing in a single method. It is not obvious to me that `hasKeyExpiredInCache()` will prune the `this.store`.  The new maintainer may use this function and didn't know that he's pruned the cache. Thus, I'm incline to remove this.

If we want to prune expired items, I would expect a public/protected method to do this explicitly. But that will cause O(n) and I am not sure if it's useful.  Map has O(1) access so we are only saving the memory consumption here, which I expect it to be very small gain.

Thus, I'm incline to remove this and keep the expired item.

ijemmy: That said. I don't think that it's bad enough to block this PR. So I would give an approval. You can decide in later commit if you want to remove this.

ijemmy (dismissed):
One comment on prunning expire key

packages/parameters/src/BaseProvider.ts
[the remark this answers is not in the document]
dreamorosi: Addressed in f9259eb9688525cbc26dfa0f4554369280d69504

packages/parameters/src/BaseProvider.ts
dreamorosi: If I use `private` I'm not sure how to test around it. So I went with `private` and extended the class in the tests, similar to what done in the Idempotency utility.

If you have any better way to do it that also allows a pure `private` let me know.

packages/parameters/src/BaseProvider.ts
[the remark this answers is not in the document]
dreamorosi: Agree.

I wasn't worried about duplicates as much as special characters and such.

packages/parameters/src/BaseProvider.ts
dreamorosi: Good points, I didn't think about the operation costs and I agree that the memory savings are marginal.

I have removed it.

ijemmy (approved):

packages/parameters/src/BaseProvider.ts
ijemmy: That should be good for now. I would expect that the tests cover all `privates` variable from the `public` methods.

packages/parameters/src/types/BaseProvider.ts
```diff
@@ -0,0 +1,34 @@
...
+  transform?: TransformOptions
+}
+
+interface GetMultipleOptionsInterface {
+  maxAge?: number
+  forceFetch?: boolean
+  sdkOptions?: unknown
+  transform?: string
+  throwOnTransformError?: boolean
+}
```
shdq, line 16: Just to reduce boilerplate:
```
interface GetMultipleOptionsInterface extends GetOptionsInterface {
    throwOnTransformError?: boolean
}
```

dreamorosi: Good point, please feel free to address this in your PR for `AppConfigprovider`

Issue #1163, opened 2022-11-13
Bug: child logger with overwritten options clear all other options 


dreamorosi:
### Expected Behaviour

When creating a child logger (see "Code snippet" section) and passing some options the child logger should retain all the parent properties/options except the ones explicitly overwritten. 

### Current Behaviour

The child logger won't have all the same options that the parent has whenever the user creates the logger while also passing some options to overwrite. For example:
```ts
const logger = new Logger({
  serviceName: "MyService",
  sampleRateValue: 0.01,
});
const auditLogger = logger.createChild({
  sampleRateValue: 1,
});
```

From the example above, I would expect `auditLogger` to have the same `serviceName` as `logger` and only differ in `sampleRateValue` - however as you can see from the logs in the "Debugging logs" section, `serviceName` is undefined in `auditLogger`.

### Code snippet

```JavaScript
import { Logger, injectLambdaContext } from "@aws-lambda-powertools/logger";
import middy from "@middy/core";

const logger = new Logger({
  serviceName: "MyService",
  sampleRateValue: 0.01,
});
const auditLogger = logger.createChild({
  sampleRateValue: 1,
});

export const handler = middy(async () => {
  logger.info("I log sometimes");
  auditLogger.info("I always log");
})
  .use(injectLambdaContext(logger))
  .use(injectLambdaContext(auditLogger));
```

### Possible Solution

Rework the [`createChild` function in Logger](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/664c4da08ac668ff1311dcc7fc1fc866f9ac06ff/packages/logger/src/Logger.ts#L208) to account for passed options and add unit tests to confirm fix & prevent future regressions.

### Steps to Reproduce

Deploy a function using the code shared in the "Code snippet" section, then execute it, finally observe the logs (see example in Debugging logs section).

### AWS Lambda Powertools for TypeScript version

latest

### AWS Lambda function runtime

16.x

### Packaging format used

Npm

### Debugging logs

```Shell
{
    "cold_start": false,
    "function_arn": "arn:aws:lambda:eu-west-1:536254204126:function:lambda-powertools-playground-fn",
    "function_memory_size": 128,
    "function_name": "lambda-powertools-playground-fn",
    "function_request_id": "6bb4c509-31d0-4dc7-9e16-f08f22a70b59",
    "level": "INFO",
    "message": "I log sometimes",
    "sampling_rate": 0.01,
    "service": "MyService",
[...]

PR #1153, opened 2022-11-10
docs: update environment variable tables for utilities

ConnorKirk:
## Description of your changes

This PR updates the environment variable documentation for the main page, and three utilities.
It also sets the default service name to `service_undefined`. Previously, the Tracer and Metrics utilities did not set the service name

* [x] Update `docs/index.md`
* [x] Update Tracer table
* [x] Update Metric table
* [x] Update Logger table
* [x] Set default Service Name in Utility class
* [x] Update Tracer class to use default service name
* [x] Update Metrics class to use default service name
* [x] Extract `isValidServiceName` to Utility class
* [x] Update unit tests

### Related issues, RFCs

**Issue number:** #724

### PR status

***Is this ready for review?:*** Yes  
***Is it a breaking change?:*** No

## Checklist

- [x] [My changes meet the tenets criteria](https://awslabs.github.io/aws-lambda-powertools-python/#tenets)
- [x] I have performed a *self-review* of my own code
- [x] I have *commented* my code where necessary, particularly in areas that should be flagged with a TODO, or hard-to-understand areas
- [x] I have made corresponding changes to the *documentation*
- [ ] I have made corresponding changes to the *examples*
- [x] My changes generate *no new warnings*
- [x] The *code coverage* hasn't decreased
- [ ] I have *added tests* that prove my change is effective and works
- [ ] New and existing *unit tests pass* locally and in Github Actions
- [ ] Any *dependent changes have been merged and published* in downstream module
- [x] The PR title follows the [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/semantic.yml#L2)

### Breaking change checklist

- [ ] I have documented the migration process
- [ ] I have added, implemented necessary warnings (if it can live side by side)

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1178, opened 2022-11-18
fix(logger): merge child logger options correctly

shdq:

Initial options (if any) now store in the `initLogger` class property when an instance of Logger creates. When a child creates out of the parent, `createChild` uses the parents' initial options to create a new instance of Logger.

- `lodash.clonedeep` dependency is no longer needed

### Related issues, RFCs

**Issue number:** #1163 

### PR status

## Checklist

### Breaking change checklist

---


shdq:
Hey, @dreamorosi I have some questions to discuss. Could you help me figure this out, please?

- what about `addContext`? Should it be included in the child?
- some tests use createLogger helper, some just new Logger()? It is needed?
- it is possible to make `initOption` optional property and check for empty options objects.
- how to remove a dependency of the package in monorepo? using `lerna exec ‘nom remove lodash.clonedeep' --scope=logger` or manually cleanup?

packages/logger/src/Logger.ts
```diff
@@ -151,7 +152,7 @@ class Logger extends Utility implements ClassThatLogs {
    */
   public constructor(options: ConstructorOptions = {}) {
     super();
-
+    this.initOptions = options;
```
dreamorosi, line 155: I would move this inside `setOptions`

packages/logger/src/Logger.ts
```diff
@@ -205,7 +206,7 @@ class Logger extends Utility implements ClassThatLogs {
    * @returns {Logger}
    */
   public createChild(options: ConstructorOptions = {}): Logger {
-    return cloneDeep(this).setOptions(options);
+    return new Logger(merge({}, this.initOptions, options));
```
dreamorosi, line 209: Given that in the `constructor`, `options` already defaults to `{}` (when no options are passed), is the first `{}` in this `merge` needed?

dreamorosi:
> Hey, @dreamorosi I have some questions to discuss. Could you help me figure this out, please?

Definitely, and thank you for the PR 👍.

> * what about `addContext`? Should it be included in the child?

I would say yes, just like all other attributes [set by other means](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/logger/#appending-persistent-additional-log-keys-and-values).

If later on there's demand to have this behavior customizable we'll add an option to `createChild`, but let's cross that bridge when we get there (if ever).

> * some tests use `createLogger` helper, some just new Logger()? It is needed?

Go with `new Logger()`, that was a helper function that we introduced when initially creating the tests but I wouldn't mind removing it at some point in the future.

> * it is possible to make `initOption` optional property and check for empty options objects.

If I understand the point correctly, and read it in the context of your current implementation, I think it should stay as not optional and default to `{}` since `options` defaults to that in the constructor when no options are passed. Correct me if I misread the question.

> * how to remove a dependency of the package in monorepo? using `lerna exec ‘nom remove lodash.clonedeep' --scope=logger` or manually cleanup?

The command would be `npm remove lodash.clonedeep @types/lodash.clonedeep -w packages/logger` (from the project's root).

While we still have `lerna` in the repo, it is used only to version & publish the packages to npm ([here](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/workflows/make-release.yml#L48-L49)). For all other operations we use [npm workspaces](https://docs.npmjs.com/cli/v8/using-npm/workspaces?v=true).

packages/logger/src/Logger.ts
shdq: `merge` mutate the first argument, so `initOptions` of the parent will be mixed with the child's options. If we create a second child from the parent after that, it will have both parent and first-child options.

packages/logger/src/Logger.ts
shdq: You are right. I'll created a method like `setInitOptions(options)`, initialize `initOptions` inside it, and put it in `setOptions` like other things. This method will be helpful, to update `initOptions` with attributes that can be added to an existing Logger instance.

shdq:
> I would say yes, just like all other attributes [set by other means](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/logger/#appending-persistent-additional-log-keys-and-values).

Ok, I'll make `addContext`, `appendKeys`, and `addPersistentLogAttributes` update `initOptions` inside `addToPowertoolLogData` and add some tests for it next.

> If I understand the point correctly, and read it in the context of your current implementation, I think it should stay as not optional and default to {} since options defaults to that in the constructor when no options are passed. Correct me if I misread the question.

I thought not to include property in instances initialized without options or with an empty object. I tested it and decided, that not having an empty object in some instances of the Logger class is not worth the changes needed.

One other thing not connected to this PR: here #1153 `defaultServiceName` was moved to `Utility`, and it imports from `@aws-lambda-powertools/commons`, because of this in every package TS shows me errors:
```
Property 'getDefaultServiceName' does not exist on type 'Logger'.ts(2339)
```
If I import `Utility` directly from `src`, it disappears. Is it because these changes are not released in `commons`?

Thanks for the fast responses!

packages/logger/src/Logger.ts
dreamorosi: Okay, thanks for explaining it. I'm not very familiar with `lodash` so I assumed it worked differently. 

Let's go with your implementation!

dreamorosi:
> Ok, I'll make `addContext`, `appendKeys`, and `addPersistentLogAttributes` update `initOptions` inside `addToPowertoolLogData` and add some tests for it next.

Nice, sounds sensible.

> I thought not to include property in instances initialized without options or with an empty object. I tested it and decided, that not having an empty object in some instances of the Logger class is not worth the changes needed.

Agreed.

> One other thing not connected to this PR: here #1153 `defaultServiceName` was moved to `Utility`, and it imports from `@aws-lambda-powertools/commons`, because of this in every package TS shows me errors:
> 
> ```
> Property 'getDefaultServiceName' does not exist on type 'Logger'.ts(2339)
> ```
> 
> If I import `Utility` directly from `src`, it disappears. Is it because these changes are not released in `commons`?

That's our bad. We should clarify this better in the `CONTRIBUTE` document.

You should be able to fix this by running `npm run build -w packages/commons` - this will rebuild the package which will now have the changes reflected.

From the way that `npm workspaces` work, the package is resolved locally, so if you were to peak inside `package-lock.json` you should see that the `@aws-lambda-powertools/commons` dependency is linked to the `packages/commons` folder (or at least it should be). Now, since this is a TS project, and also most probably because we haven't found a better way, the package actually links to the `dist` folder inside `packages/commons` and not whatever is in `packages/commons/src`. So when there are changes in `commons` from another PR, or if you make changes to it, you'll have to rebuild it before these changes are available in any other of the packages.

Hope this clarifies it!

shdq:
Hey @dreamorosi.

I found a way to get rid of `initOptions` too. It seems to work smoothly. The best code is no code, right?

It's ready to review. If more tests are needed, let me know.

packages/logger/src/Logger.ts
```diff
@@ -205,7 +203,14 @@ class Logger extends Utility implements ClassThatLogs {
    * @returns {Logger}
    */
   public createChild(options: ConstructorOptions = {}): Logger {
-    return cloneDeep(this).setOptions(options);
+    const parentsPowertoolsLogData = this.getPowertoolLogData();
+    const childLogger = new Logger(merge({}, parentsPowertoolsLogData, options));
+    const parentsPersistentLogAttributes = this.getPersistentLogAttributes();
+    childLogger.addPersistentLogAttributes(parentsPersistentLogAttributes);
+    const childPowertoolsLogData = childLogger.getPowertoolLogData();
+    childLogger.addToPowertoolLogData(merge({}, parentsPowertoolsLogData, childPowertoolsLogData));
```
dreamorosi, line 211: @shdq: question: why do we merge and use the `parentsPowertoolsLogData` twice? Maybe worth documenting with a comment (inside the method, so we don't expose it in the docstring)

dreamorosi:
On Thursday I'm planning on making a new release, I'm mentioning this just in case you'd like this PR to be included - if not it'll be guaranteed to be in the next one in ~2 weeks

shdq:
> On Thursday I'm planning on making a new release, I'm mentioning this just in case you'd like this PR to be included - if not it'll be guaranteed to be in the next one in ~2 weeks

I think this PR is finished and will be in the next release 🤞

dreamorosi (approved):
Thank you @shdq for working on this, appreciate the effort as usual.

Integration tests were passing: https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/3533218456

Issue #1177, opened 2022-11-17
Feature request: implement `AppConfigProvider`


dreamorosi:
### Summary

The Parameters utility should allow customers to retrieve one parameter stored in a AppConfig.

This work includes the following tasks:
- [x] Write provider + types
- [x] Unit tests (100% coverage)

### Why is this needed?

1/ So that customers are able to retrieve parameters from AppConfig, 2/ to reach feature parity with Powertools for Python.

### Which area does this relate to?

Parameters

### Acknowledgment

- [X] This request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/#tenets)
- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [Java](https://github.com/awslabs/aws-lambda-powertools-java/)

shdq:
Hello, @dreamorosi. I'm ready to help, but the scope of this PR is bigger than usual, and I don't know where to start. My plan is:
- use [implementation](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/utilities/parameters/appconfig.py) in the python repository, as an example. We are porting it to TypeScript, right?
- dive deep into both #1187 and #1168 PR's (I glanced at them and I already have a basic understanding)
- I will use [AppConfig client from AWS SDK](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-appconfig/index.html) to get actual parameters from service.

Correct me if I'm wrong with my mental model about the `Parameters` milestone. So the idea is to provide users of powertools ways to get parameters from the range of AWS products, that can store settings, such as `AppConfig`. But what parameters are that? I still miss some parts of the whole picture. Could you explain in general how it works or share a link to further reading, please?

Speaking of implementation, I need to add my provider class that will extend `BaseProvider` class and implement two methods `_get` – for one parameter and `_getMultiple` using `AppConfig client` from SDK. 

Thank you for the guidance as always!

dreamorosi:
Hi @shdq thank you for the bias for action, I'm happy to support you throughout the process.

> use [implementation](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/utilities/parameters/appconfig.py) in the python repository, as an example. We are porting it to TypeScript, right?

[...]

PR #1200, opened 2022-12-21
feat(parameters): AppConfigProvider

shdq:

Related issue #1177 

### Related issues, RFCs

**Issue number:** #1177

### PR status

## Checklist

### Breaking change checklist

---


packages/parameters/src/AppConfigProvider.ts
```diff
@@ -0,0 +1,111 @@
...
+    return value;
+  }
+
+  /**
+   * Retrieving multiple parameter values is not supported with AWS App Config Provider.
+   *
+   * @throws Not Implemented Error.
+   */
+  protected async _getMultiple(
+    path: string,
```
dreamorosi, line 89:
```suggestion
   _path: string,
```

packages/parameters/src/AppConfigProvider.ts
```diff
@@ -0,0 +1,111 @@
...
+    });
+    const response = await this.client.send(getConfigurationCommand);
+
+    //  Should we cache and flush the token after 24 hours?
+    //  NextPollConfigurationToken expires in 24 hours after the last request and causes `BadRequestException`
+    this.token = response.NextPollConfigurationToken;
+
+    const configuration = response.Configuration;
+
+    // should we convert Uint8Array to string?
```
dreamorosi, line 76: So, keeping in mind that this method is called by `BaseProvider.get`, I think this method should return the value as-is and leave the caller decide whether to apply transforms or not based on the arguments passed by the user.

Thoughts?

packages/parameters/src/AppConfigProvider.ts
```diff
@@ -0,0 +1,111 @@
...
+      this.token = session.InitialConfigurationToken;
+    }
+
+    const getConfigurationCommand = new GetLatestConfigurationCommand({
+      ConfigurationToken: this.token,
+    });
+    const response = await this.client.send(getConfigurationCommand);
+
+    //  Should we cache and flush the token after 24 hours?
+    //  NextPollConfigurationToken expires in 24 hours after the last request and causes `BadRequestException`
```
dreamorosi, line 71: Is the token scoped to a specific configuration or can I retrieve multiple configurations with the same token?

I.e.: If I do `.get('configA')` and then `.get('configB')`, do they use the same token or do they need two different tokens?

Before making a decision on caching I'd like to understand this. By default we are already caching the response (in `BaseProvider`) but I can see how customers might want to hit AppConfig more times (i.e. retrieving multiple configs or refreshing the same config).

I'm not excessively concerned about the 24hrs duration given that we are optimizing for Lambda which is ephemeral in most cases. We expect most environments to never reach that lifespan but there could be edge cases in which an environment might last for longer spans of time (perhaps with provisioned concurrency). I would say that having a logic that refreshes the token in case of a `BadRequestException` would be a nice to have but not a must at the moment. I leave it up to you.

shdq: This is a really good question. I opened an [issue](https://github.com/awslabs/aws-lambda-powertools-python/issues/1796) in the python repo about the function logic and going to adjust my implementation.

packages/parameters/src/AppConfigProvider.ts
shdq: I agree. I was confused by the `str` return type of the function. I opened an [issue](https://github.com/awslabs/aws-lambda-powertools-python/issues/1796) in the python repo about the return type

dreamorosi: After implementing the SecretsProvider (in #1206) which can also return `Uint8Array` I'm convinced that we should return the value as-is and then let the user decide on whether it should be transformed or not via the `transform` parameter.

I have updated the `BaseProvider` (#1205 - merged) to support this, you might want to rebase to get the latest from `main`.

Based on the current info I think our implementation is the correct one, but I'm open to get convinced otherwise and change it back if the discussion in the issue you opened requires so.

packages/parameters/src/AppConfigProvider.ts
dreamorosi: Given that the Python's team is not expected to be 100% operational for [a little longer](https://github.com/awslabs/aws-lambda-powertools-python/issues/1784) I would move forward with the current implementation and revisit the implementation when the discussion picks up again.

shdq:
@dreamorosi I need help with types for `AppConfigProvider`. 

TL;DR; `BaseProvider` `get` method  `name` argument and `options?` argument which is optional. But `name` and two more arguments are required to retrieve a configuration.

To retrieve configuration two API two calls are necessary. First to `StartConfigurationSession` three properties are required:

1. `ConfigurationProfileIdentifier` – The configuration profile ID or the configuration profile name. This is `name` argument from the `get` method of `BaseProvider`.
2. `ApplicationIdentifier` – The application ID or the application name.
3. `EnvironmentIdentifier` – The environment ID or the environment name.

When you make the first call to API you get initial token that encapsulates all these fields. Then you make the next API call with the only one initial token to `GetLatestConfiguration` and get the actual configuration and new token for the next call.

I placed the last two properties into `sdkOptions` property of options.

```
interface AppConfigGetOptionsInterface
  extends Omit<GetOptionsInterface, 'sdkOptions'> {
  clientConfig?: AppConfigDataClientConfig
  sdkOptions: {
    application: string
    environment: string
  }
}
```

Right now, `application` and `environment` are required to create an instance of `AppConfigProvider`. I add `Partial<AppConfigGetOptionsInterface>` to `BaseProvider` `get` method or it would break `BaseProvider` tests. As the result, I had to provide `default` values to `application` and `environment` which is not correct.

It also makes sense to "reset" the token, when you receive SDK options to overwrite in `get` method. Or it will make an API call with the old token completely ignoring new options. That's why I opened an [issue](https://github.com/awslabs/aws-lambda-powertools-python/issues/1796) about logic.

Another way is not to require options for a new instance of the class, but raise an error for the first call of the `get` method if it was made without required SKD options for `StartConfigurationSession`.

Sorry for the delay with the changes. We can discuss it in discord to speed up the process.

dreamorosi:
Hey Sergei, thanks for the message and please no need to apologise for the times, you're doing great - no complaints at all here 😄 

Let me pull the code and play with it in my IDE. Before giving an answer I need to better understand it and also wrap my head around the AppConfig flow.

Doing this now, you can expect an answer in the next couple of hours tops.

shdq:
> Doing this now, you can expect an answer in the next couple of hours tops.

Thanks 👍 
At the same time, I'll try to draw a diagram with the flow to help us figure this out.

packages/parameters/src/AppConfigProvider.ts
```diff
@@ -0,0 +1,115 @@
...
+  private environment: string;
+  private latestConfiguration: Uint8Array | undefined;
+  private token: string | undefined;
+
+  /**
+   * It initializes the AppConfigProvider class'.
+   * *
+   * @param {AppConfigGetOptionsInterface} config
+   */
+  public constructor(options: AppConfigGetOptionsInterface) {
```
dreamorosi, line 25: I think the `options` parameter should have its own interface, similar to what [done here for the DynamoDBProvider](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/5144b3b974bc1f988c2d3d0599f6f05a76cc315c/packages/parameters/src/types/DynamoDBProvider.ts#L4-L10).

This interface would look like this:
```ts
interface AppConfigProviderOptions {
  application: string
  environment: string
  clientConfig?: AppConfigDataClientConfig
}
```

This is in line with the Python's implementation, which requires users to specify an `application` and `environment` (both mandatory) and allows for an optional config object that will be passed to the SDK client. This config object will allow users to set things like the AWS Region or pass explicit AWS credentials.

dreamorosi: As an aside, while writing this comment I realized that I will have to adjust the `SSMProvider` to have a similar shape.

packages/parameters/src/AppConfigProvider.ts
```diff
@@ -0,0 +1,115 @@
...
+     **/
+    if (!this.token) {
+      const sessionOptions: StartConfigurationSessionCommandInput = {
+        ConfigurationProfileIdentifier: name,
+        EnvironmentIdentifier: this.application,
+        ApplicationIdentifier: this.environment,
+      };
+
+      if (options?.sdkOptions) {
+        Object.assign(sessionOptions, options.sdkOptions);
```
dreamorosi, line 60: Here we should apply a similar logic to what done in the `DynamoDBProvider` and `SecretsProvider` ([example](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/5144b3b974bc1f988c2d3d0599f6f05a76cc315c/packages/parameters/src/DynamoDBProvider.ts#L45)) and remove from `sdkOptions` some of the parameters that we don't want the user to override, namely: `ConfigurationProfileIdentifier`, `EnvironmentIdentifier`, and `ApplicationIdentifier`, which leaves us only with [`RequiredMinimumPollIntervalInSeconds`](https://docs.aws.amazon.com/appconfig/2019-10-09/APIReference/API_appconfigdata_StartConfigurationSession.html#appconfig-appconfigdata_StartConfigurationSession-request-RequiredMinimumPollIntervalInSeconds).

shdq:
I think this would fix the logic.

appconfig-diagram

packages/parameters/src/AppConfigProvider.ts
dreamorosi: This way we nudge users to configure the application and environment via the class constructor and use the `name` parameter for the configuration identifier. By disallowing these overrides we can streamline our logic.

dreamorosi: And for the sake of completion, the `AppConfigGetOptionsInterface` should look like this:
```ts
interface AppConfigGetOptionsInterface extends GetOptionsInterface {
  sdkOptions?: StartConfigurationSessionCommandInput
}
```

packages/parameters/src/AppConfigProvider.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  /**
+   * Retrieving multiple parameter values is not supported with AWS App Config Provider.
+   *
+   * @throws Not Implemented Error.
+   */
+  protected async _getMultiple(
+    _path: string,
+    _sdkOptions?: Partial<GetLatestConfigurationCommandInput>
```
dreamorosi, line 94: Let's make this `unknown` type

packages/parameters/src/AppConfigProvider.ts
shdq: > There's then a third optional config object that will be passed to the SDK client (note the `AppConfigDataClientConfig` type). This config object will allow users to set things like the AWS Region or pass explicit AWS credentials.

If we extract config from options, how to pass it to the new instance of the class? I asked you about that in [SecretsProvider PR](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/1206/files#diff-ba2b2c24f5637430ac1de6cf8f6de4832ee5c12149e9563de13523ea38c904ecR7).

packages/parameters/src/AppConfigProvider.ts
shdq: > This way we nudge users to configure the application and environment via the class constructor and use the `name` parameter for the configuration identifier. By disallowing these overrides we can streamline our logic.

So if a user wants to get configuration with other `environment` and `application`, they should create a new instance of the `AppConfigProvider`. Correct?

packages/parameters/src/AppConfigProvider.ts
dreamorosi: I don't see any comment at the line you have highlighted

packages/parameters/src/AppConfigProvider.ts
dreamorosi: Yes, that's right.

Looking at the docs for the AppConfig service the two entities are defined as follows:
**Application:**
> In AWS AppConfig, an application is simply an organizational construct like a folder. This organizational construct has a relationship with some unit of executable code. For example, you could create an application called MyMobileApp to organize and manage configuration data for a mobile application installed by your users.

**Environment:**
> An environment is a logical deployment group of AppConfig targets, such as applications in a Beta or Production environment. You can also define environments for application subcomponents such as the Web, Mobile, and Back-end components for your application.

Based on these definitions I think it's safe to assume (and recommend) that users group all their configurations within the same environment & application pair. This has a number of advantages like isolation of both responsibilities and permissions (i.e. this AWS Lambda resource has access to a specific application & environment).

packages/parameters/src/AppConfigProvider.ts
dreamorosi: But aside from the comment I might have missed (sorry if I did):
- `AppConfigProviderOptions` is the interface used for the parameter passed to the Class constructor of a provider.
- `AppConfigGetOptionsInterface` it's the interface used for the `options` parameter of the `get`/`_get` methods

The former has an optional `clientConfig` parameter that allows the user to configure the behavior of the entire AWS SDK client that is instantiated upon class instantiation. The configs passed here will apply to all the API calls made with this client.

The latter instead has an optional `sdkOptions` parameter that allows the user to configure the behavior of a specific API call (made via the SDK).

shdq: > I don't see any comment at the line you have highlighted

My bad, I didn't submit the review. The same for the `SSMProvider` you asked me to review and which I did. I'm not really experienced in reviewing. I think I did it many times and thought why you are not responding to my comments 🤣

packages/parameters/src/AppConfigProvider.ts
shdq: > Based on these definitions I think it's safe to assume (and recommend) that users group all their configurations within the same environment & application pair. This has a number of advantages like isolation of both responsibilities and permissions (i.e. this AWS Lambda resource has access to a specific application & environment).

Make sense and simplifies implementation. Thanks for the clarification.

packages/parameters/src/AppConfigProvider.ts
dreamorosi: Oh I see! 

I think while you're in review mode you have two choices:
- "Post as single comment" (or named something similar) which basically adds the comment immediately and as standalone
- "Add as review" - In this case it'll be marked as "pending" and it won't show up until you actually submit the review (green button top-right of the review page)

Maybe you were doing the second 😁

dreamorosi:
I hope that my comments clarify all the open points we had. I realise that the communication/conversation is a bit all over the place due to the distributed nature of the comments, so if there's any point that we should discuss further, or that I have missed please don't hesitate to let me know.

To sum up:
- `application` and `environment` should be set by the user when instantiating the provider
  - `application` can be set via constructor argument OR via environment variable - if none is set we throw an error.
  - `environment` can be set exclusively via constructor argument and as such it's mandatory
- Each set of `application`, `environment`, and configuration requires its own session token. With the assumption that an instance of `AppConfigProvider` handles exclusively a pair of `application` and `environment` defined during instantiation, we can handle multiple config by keeping track of the configuration and token pairs.

shdq:
> This logic is pretty similar to the one in your diagram, however it would allow us to handle & cache the tokens of multiple configurations at the same time. What do you think?

I like it, I mentioned something like that in the python issue. But I see a potential problem with the latest configuration. Quote from the issue:
> Another option is to use dict to store NextPollConfigurationToken tokens with name as the key. But it increases complexity, cause in this case you should also store last_returned_value for each token. I think flush and starting a new session is the better option.

Let's use your example:
```
const provider = new AppConfigProvider( {
  sdkOptions: {
    application: 'TestApplication',
    environment: 'Lambda',
  }
});

const config = await provider.get('TestFeatureFlag');
```
We got the configuration for `TestFeatureFlag` and saved the token in the map and this.latestConfiguration = "latest TestFeatureFlag configuration".
```
const otherConfig = await provider.get('TestFreeForm');
```
Now the map has two tokens and this.latestConfiguration = "latest TestFreeForm configuration".
Now, lets call the `TestFeatureFlag` again:
```
const config = await provider.get('TestFeatureFlag');
```
We hit the cache and used the token for `TestFeatureFlag` and since we already got the latest configuration API will return an empty response which lead to
```
if (configuration) {
      this.latestConfiguration = configuration;
    }
```
But `latestConfiguration` has the latest TestFreeForm configuration. I assume this behavior, but I can't test it.

> At least based on my tests, if you use a valid token requested for a configuration from the same application/environment pair to retrieve a second configuration, the operation will succeed but return an empty response. This is a little confusing, but I'm sure there are reasons as for why the service was implemented this way although I'll refrain from commenting on it since I don't know much about it.

I think it is intended to reduce end-user costs because AWS charges for each API call and received configuration
>With AWS AppConfig, you pay each time you request configuration data from AWS AppConfig via API calls, as well as each time your requesting target receives configuration data in response to that request.

>A target refers to the host that receives the application configuration (this could be an Amazon EC2 instance, on-premises server, container, AWS Lambda function, mobile app, IoT device, etc.).

>Configuration requests via API Calls cost $0.0000002 per configuration request
>Configurations Received cost $0.0008 per configuration received

If this potential problem with the latest configuration is real, seems like we should store configurations as well as tokens. Or flush the token if it was called with a different `name` and start over with a new session. I think in this case the response won't be empty (didn't test it) if it was made with `InitialToken`.

What do you think?

shdq:
Sorry for the lengthy discussion around that.

Thanks for the help and clarification. I appreciate that. I will adjust the implementation and write more tests for different use cases while we decide on the caching tokens.

dreamorosi:
No, this is a great discussion and it's worth having it I think.

I realised that I might not have been very clear in my previous message, so I have recorded a short screen share where I explain my implementation idea & also show a debugger session to explain what happens at each point, I hope it helps:

https://user-images.githubusercontent.com/7353869/210086160-06573170-c75d-4d6c-ba50-6d939122dd32.mp4

In regards to why this is needed, you are correct, the reason is mainly cost savings but also performances (which in the case of Lambda almost directly translates to additional cost savings since shorter execution = lower cost).

Adding the storage of the tokens would be a second layer of caching. The first one is the storage of the configurations, which as you pointed out, is needed and we are in fact doing.

If you check the implementation of `BaseProvider.get` ([here](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/parameters/src/BaseProvider.ts#L47-L51)) which is the caller of the `_get` method you are implementing, we are in fact checking the cache before hitting the API. This means that we'll call the `_get` method only if the config is not cached or is present in the cache but has expired (as defined by the `maxAge` param - default 5 seconds).

Given this first level of caching, we can assume that a number of retrievals will be covered by the main cache. However the function's environment lasts longer and the cache expires, or the user wants to fetch an updated version (via the `forceFetch` parameter), we need to be able to call the AppConfig API as quickly as possible, and this means that having the token we can do so with one less request.

dreamorosi:
After our conversation on Discord I'd like to amend part of my understanding of how the AppConfig service works and acknowledge that you were correct in your understanding.

Looking at [this section of the docs](https://docs.aws.amazon.com/appconfig/latest/userguide/appconfig-retrieving-the-configuration.html#appconfig-retrieving-the-configuration-example) (which I think you were referring to), there's this statement:
> **Warning**
> The GetLatestConfiguration response includes a Configuration section that shows the configuration data. The Configuration section only appears if the system finds new or updated configuration data. If the system doesn't find new or updated configuration data, then the Configuration data is empty.

This means that regardless of our implementation / caching, if users want to call the API they need to be ready to handle empty responses in their code if the config hasn't changed in the meanwhile.

Additionally, there's another case that we haven't mentioned, but that doesn't affect the implementation: the response returns a `NextPollIntervalInSeconds` which can be tweaked when starting the session via the `RequiredMinimumPollIntervalInSeconds` parameter (minimum 15 seconds). If you poll too quickly the service will throw an error `Request too early`.

In the video I posted above, I mistakenly assumed that the last config was retrieved from the service, while in fact it must have been retrieved from the cache since it should have been empty. I was able to confirm this by re-running the same sample this time using `forceFetch` and confirm that there response was indeed empty since the config hadn't changed.

packages/parameters/src/AppConfigProvider.ts
```diff
@@ -0,0 +1,115 @@
...
+   * Retrieve a parameter value from AWS App config.
+   *
+   * @param {string} name - Name of the configuration
+   * @param {AppConfigGetOptionsInterface} options - SDK options to propagate to `StartConfigurationSession` API call
+   * @returns {Promise<Uint8Array | undefined>}
+   */
+  protected async _get(
+    name: string,
+    options?: AppConfigGetOptionsInterface
+  ): Promise<Uint8Array | undefined> {
```
dreamorosi, line 46: Based on [this message](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/1200#issuecomment-1368015600), the return type should include `string` for the empty string case.

dreamorosi:
After merging #1206 I'm having some issues with the types on the `get` & `getMultiple` methods in the specific providers and in relation to the overloads in the `BaseProvider`.

Not sure how to move ahead yet, but you probably will encounter the same type issues I think.

packages/parameters/src/AppConfigProvider.ts
shdq: In the `GetLatestConfigurationResponse` interface `Configuration` may be `Uint8Array` or `undefined`. I didn't find anywhere in the docs that it can return an empty string.

shdq:
Hey @dreamorosi! Some comments for the PR. If you have any questions, I'm happy to answer.

As I get it from our conversation, a user should handle the cases below:
- the too-early request (`RequiredMinimumPollIntervalInSeconds`)
- empty `GetLatestConfiguration`response when configuration didn't change since the last retrieval (I assume it tracked using tokens)
- `BadRequestException` when the token expires in 24 hours (unlikely with lambda)

So I return it as-is, so all the errors will be caught in `BaseProvider`.

For the utility function, I combined interfaces in `getAppConfigCombinedInterface`, since the function has one argument, but inside I need to provide options for both `new AppConfigProvider(options)` and `DEFAULT_PROVIDERS.appconfig.get(name, options)`

Future thoughts for the next cycles:

- If `forceFetch` option is provided, start a new configuration session and get the latest configuration anyway
- Configuration returns `Uint8Array` and the configuration itself can be in `json` format, so it may require a double transformation

dreamorosi:
> As I get it from our conversation, a user should handle the cases below:
> 
> * the too-early request (`RequiredMinimumPollIntervalInSeconds`)
> * empty `GetLatestConfiguration`response when configuration didn't change since the last retrieval (I assume it tracked using tokens)
> * `BadRequestException` when the token expires in 24 hours (unlikely with lambda)
> 
> So I return it as-is, so all the errors will be caught in `BaseProvider`.


Correct.

> For the utility function, I combined interfaces in `getAppConfigCombinedInterface`, since the function has one argument, but inside I need to provide options for both `new AppConfigProvider(options)` and `DEFAULT_PROVIDERS.appconfig.get(name, options)`

I understand why this is necessary, however I'm on the fence about this.

My main question mark is on whether we should restrict the type to avoid passing `clientConfig` to the utility function, i.e.:
```ts
interface getAppConfigCombinedInterface
   extends Omit<AppConfigProviderOptions, 'clientConfig>,
   AppConfigGetOptionsInterface {}
```

The reason I suggest this is for coherence with the other providers. None of the other providers allows to customize the AWS SDK when using the utility functions. That should be a behavior (and a reason to use) for the class providers.

> Future thoughts for the next cycles:
> * If `forceFetch` option is provided, start a new configuration session and get the latest configuration anyway

That's an interesting proposal, I hadn't thought about that and I can see the value. I am however concerned about implicitly overloading the meaning of the `forceFetch` which across providers would normally mean "1/ skip cache, 2/ fetch **value** from API" while for this provider would mean "1/ skip cache, 2/ discard token, 3/ fetch value from API". 

On the other hand, if someone wanted to have a break-glass solution to forcefully fetch the value no matter the cost / timeout values, there's no way.

I think it's worth discussing this, but I'd like to do so across all versions of Powertools, not just us.

> * Configuration returns `Uint8Array` and the configuration itself can be in `json` format, so it may require a double transformation

Very true, I've been thinking about this a lot as well and I don't love it. However, given that (like the point above) it's a new behavior/feature, I'd like to discuss this with the larger Powertools ecosystem. I'm in the process of writing a RFC which I'll post at a later time, but my current stance is that this is not something that should be there in v1 of the Parameters utility and could come afterwards. 

Basically I'd like to focus our limited resources/cycles on laying out solid foundations and then having people start using the library before adding/perfecting features.

Hope it makes sense.

packages/parameters/src/types/AppConfigProvider.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+/**
+ * Combined options for the getAppConfig utility function.
+ *
+ * @interface getAppConfigCombinedInterface
+ * @extends {AppConfigProviderOptions, AppConfigGetOptionsInterface}
+ */
+interface getAppConfigCombinedInterface
+  extends AppConfigProviderOptions,
```
dreamorosi, line 43: Copying excerpt from [this other comment](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/1200#issuecomment-1373569312).

My main question mark is on whether we should restrict the type to avoid passing `clientConfig` to the utility function, i.e.:
```ts
interface getAppConfigCombinedInterface
   extends Omit<AppConfigProviderOptions, 'clientConfig>,
   AppConfigGetOptionsInterface {}
```

The reason I suggest this is for coherence with the other providers. None of the other providers allows to customize the AWS SDK when using the utility functions. That should be a behavior (and a reason to use) for the class providers.

dreamorosi (commented):
Looking great, I have one last comment/point about the `getAppConfigCombinedInterface` interface that I'd like to hash out.

After that I'm looking forward to merge this asap!

packages/parameters/src/types/AppConfigProvider.ts
dreamorosi: Also can we capitalize the name to `GetAppConfigCombinedInterface` for uniformity?

shdq: > The reason I suggest this is for coherence with the other providers. None of the other providers allows to customize the AWS SDK when using the utility functions. That should be a behavior (and a reason to use) for the class providers.

100% agree. Consistency is important, I'll make the changes.

shdq:
> Basically I'd like to focus our limited resources/cycles on laying out solid foundations and then having people start using the library before adding/perfecting features.

Sure. I just want to let you know and save my thoughts for future discussions.

Issue #1117, opened 2022-10-12
Bug: logging fails if input contains bigint values


samuelsulo:
## Bug (logger): logging fails if input contains bigint values
Since ```JSON.stringify``` is unable to serialize bigint values, the logging function raises an exception if the input contains a bigint value.

### Expected Behavior
The logger should handle bigints to prevent ```JSON.stringify``` from raising an exception.

Log Request:
```typescript
import { Logger } from "@aws-lambda-powertools/logger";

const logger = new Logger();
logger.info('Object with bigint value', { myBigInt: BigInt(1234) })
```
Log Result:
```json
{
  "level": "INFO",
  "message": "Object with bigint value",
  "service": "service_undefined",
  "timestamp": "2022-10-12T22:27:17.067Z",
  "myBigInt": "1234"
}
```

### Current Behavior
The logger does not handle bigint values which causes ```JSON.stringify``` to raise an exception.

Log Request:
```typescript
import { Logger } from "@aws-lambda-powertools/logger";

const logger = new Logger();
logger.info('Object with bigint value', { myBigInt: BigInt(1234) })
```
Raised exception:
```
TypeError: Do not know how to serialize a BigInt
    at JSON.stringify (<anonymous>)
    ...
```

### Possible Solution

Handling bigint values in the ```removeCircularDependencies``` method of the ```Logger``` class should fix the issue.

Before
```typescript
private removeCircularDependencies(): (key: string, value: LogAttributes | Error) => void {
    const references = new WeakSet();

    return (key, value) => {
      let item = value;
      if (item instanceof Error) {
        item = this.getLogFormatter().formatError(item);
      }
      if (typeof item === 'object' && value !== null) {
        if (references.has(item)) {
          return;
        }
        references.add(item);
      }

      return item;
    };
  }
```

After
```typescript
private removeCircularDependencies(): (key: string, value: LogAttributes | Error | bigint) => void {
    const references = new WeakSet();

    return (key, value) => {
      let item = value;
      if (typeof item === 'bigint') {
        return item.toString();
      }
      if (item instanceof Error) {
        item = this.getLogFormatter().formatError(item);
      }
      if (typeof item === 'object' && value !== null) {
        if (references.has(item)) {
          return;
        }
        references.add(item);
      }

      return item;
    };
  }
```

### Environment

* **Powertools version used**: 1.2.1
* **Packaging format (Layers, npm)**: npm
[...]

PR #1201, opened 2022-12-28
fix(logger): logger throws TypeError when log item has BigInt value

shdq:

The replacer function used for `JSON.stringify` converts `BigInt` values to strings.

Added tests and removed external dependency for `removeEmptyKeys()` function.

### Related issues, RFCs

**Issue number:** #1117 

### PR status

## Checklist

### Breaking change checklist

---


packages/logger/src/log/LogItem.ts
```diff
@@ -31,7 +30,14 @@ class LogItem implements LogItemInterface {
...
-    return pickBy(attributes, (value) => value !== undefined && value !== '' && value !== null);
+    const newAttributes: LogAttributes = {};
+    for (const key in attributes) {
+      if (attributes[key] !== undefined && attributes[key] !== '' && attributes[key] !== null) {
+        newAttributes[key] = attributes[key];
+      }
+    }
+    
+    return newAttributes;
   }
```
shdq, line 41: Firstly, I wrote a declarative one-liner:
```
public removeEmptyKeys(attributes: LogAttributes): LogAttributes {
    return Object.fromEntries(Object.entries(attributes).filter(([, value]) => value !== undefined && value !== '' && value !== null));
}
```
But for the sake of readability, I stick to the verbose version.

shdq: I don't know if nested objects could be in the `logItem`, but it wasn't and isn't supported by this function.

dreamorosi: Could you give an example of what do you mean by nested objects?

dreamorosi:
I see that there's no `package-lock.json` in the file diff, if you have done it differently, could you please remove the dependency by running `npm remove lodash.pickby -w packages/logger`?

This should make the necessary updates to the lock file as well as removing it from the `package.json` of the package. Note that in order to successfully be able to run the command you might have to temporarily revert the change to `package.json`, otherwise npm will think the two files are out of sync.

Could you please also remove `@types/lodash.pickby` from `devDependencies` in the same package while at it?

packages/logger/src/log/LogItem.ts
shdq: `logger.info("my message", { value: 42, nested: { value: "nested value", emptyValue: '', undefinedValue: undefined, nullValue: null } });`

dreamorosi:
Also, I see that [the branch you're working on](https://github.com/shdq/aws-lambda-powertools-typescript/tree/1117-bug-logger-cant-serialize-bigint-values) is behind `main`.

Normally this would be fine, but in this case I would strongly advise to rebase/bring it in line with `main`. 

The reason why I say this is that as part of the migration to Node 18 we have updated all the dependencies and added the runtime to the unit & integration tests. For the integration tests to pass we need the CDK version that is currently present on `main` but not in your branch, which [causes the integration tests to fail](https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/3794247482).

packages/logger/src/Logger.ts
```diff
@@ -568,6 +568,37 @@ class Logger extends Utility implements ClassThatLogs {
...
+  /**
+   * When the data added in the log item contains object references or BigInt values,
+   * `JSON.stringify()` can't handle them and instead throws errors:
+   * `TypeError: cyclic object value` or `TypeError: Do not know how to serialize a BigInt`.
+   * To mitigate these issues, this method will find and remove all cyclic references and convert BigInt values to strings.
+   *
+   * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#exceptions
+   * @private
+   */
+  private getReplacer(): (key: string, value: LogAttributes | Error | bigint) => void {
```
dreamorosi, line 580: I like the new name, thanks for making the change

packages/logger/src/log/LogItem.ts
dreamorosi: I think we need to support nested objects though. 

Here's an example of logs emitted by Logger `v1.5.0` (latest):
Screenshot 2022-12-28 at 14 40 48
 
from this code:
```ts
import { Logger, injectLambdaContext } from "@aws-lambda-powertools/logger";

const logger = new Logger({});

  logger.info("my message", {
    value: 42,
    nested: {
      value: "nested value",
      emptyValue: "",
      undefinedValue: undefined,
      nullValue: null,
    },
  });
}).use(injectLambdaContext(logger, { logEvent: true }));
```

this was the event payload passed:
```json
{
  "key1": "value1",
  "key2": "value2",
  "key3": "value3",
  "nested": {
      "some": "key",
      "other": 1
  }
}
```

We need to support nested objects both for the `logEvent` functionality (also present in the unit tests) and arbitrary objects

dreamorosi: Oh after re-reading I realize I might have misunderstood your comment.

Were you referring to supporting the removal of `undefined`, `null`, etc. from nested objects?

shdq: > Oh after re-reading I realize I might have misunderstood your comment.
> 
> Were you referring to supporting the removal of `undefined`, `null`, etc. from nested objects?

yes, it's just go through the first level

dreamorosi: I see, apologies for the misunderstanding!

I think that at the moment we should strive for maintaining the same exact functionality level as the current version without breaking changes. Which based on the test I did above, seems to at least remove `undefined` values.

Then later on, if there's demand we can explore extending the behavior to empty strings and/or `null`.

shdq: `undefined` values are filtered somewhere else 🙃

dreamorosi:
Thank you for syncing the branch & updating the dependencies & lock file.

I can confirm that the integration tests are passing: https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/3794471678

packages/logger/src/log/LogItem.ts
dreamorosi: And I definitely support/agree with the choice of using a more verbose version for the sake of readability.

dreamorosi (approved):
Thanks a lot @shdq for picking this up and for the quick PR.

I appreciate the choices made throughout.

shdq:
> Thanks a lot @shdq for picking this up and for the quick PR.
> 
> I appreciate the choices made throughout.

I'm happy to help! Thanks for the fast feedback loop @dreamorosi!

PR #1202, opened 2022-12-28
feat(parameters): DynamoDBProvider support

dreamorosi:

This PR implements the `DynamoDBProvider` class which will allow the future Parameters utility to support retrieving parameters from DynamoDB.

The `DynamoDBProvider` supports the following retrieval modes:
- `get` this class method allows customers to retrieve a single parameter by name and allows to specify transform, max age.

The provider comes with defaults thanks to which the user only has to specify the table name and the the name of the parameter to retrieve:
```ts
const parameters = new DynamoDBProvider({
  tableName: 'some-table'
});
```
The table named `some-table` expects an item that looks like this:
```json
{ "id": "foo", "value": "bar" }
```
However, if users have items with other type of shapes, like for instance:
```json
{ "key": "foo", "val": "bar" }
```
they can instantiate the provider like so:
```ts
const parameters = new DynamoDBProvider({
  tableName: 'some-table',
  keyAttr: 'key',
  valueAttr: 'val',
});
```

In both cases, they can retrieve the parameter using the `get` method:
```ts
const parameter = parameters.get('foo');
console.log(parameter); // bar
```

- `getMultiple` this class method allows customers to retrieve multiple parameters by path and allows to specify a shared config object that changes the behavior around transform, max age.

Similarly to what described above, the default come also for path-based retrieval:
```ts
const parameters = new DynamoDBProvider({
  tableName: 'some-table'
});
```
The table named `some-table` expects an item that looks like this:
```json
{ "id": "foo", "sk": "a", "value": "value-param-a" }
{ "id": "foo", "sk": "b", "value": "value-param-b" }
```
Which will result, when used, in this:
```ts
const parameterObject = parameters.getMultiple('foo');
for (const [ key, val ] of Object.entries(parameterObject)) {
  console.log(key, val);
}
// a, value-param-a
// b, value-param-b
```
Just like before, if users have items with a different shape, say this:
```json
{ "key": "foo", "sort": "a", "val": "value-param-a" }
{ "key": "foo", "sort": "b", "val": "value-param-b" }
```
They can customize the provider's behavior by instantiating it with different attribute names:
```ts
const parameters = new DynamoDBProvider({
  tableName: 'some-table',
  keyAttr: 'key',
  valueAttr: 'val',
  sortAttr: 'sort',
});
```

Both `get` and `getMultiple` also allow an optional `options` argument that allow to customize the provider's behavior around caching and transform/errors. 

The `options` object also accepts a `sdkOptions` object, which allows the user to pass additional inputs to the SDK client API call (i.e. specify `ConsistentRead`). There are however some SDK options that cannot be passed as they are managed by the provider, these are:
- `TableName`
- `ProjectionExpression`
- `Key` (applies only to `get`)
- `KeyConditionExpression` (applies only to `getMultiple`)
- `ExpressionAttributeValues` (applies only to `getMultiple`)

### How to verify this change

See the newly added unit test cases & optionally compare also the API surface with the implementation [found in Powertools for Python](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/utilities/parameters/dynamodb.py).

### Related issues, RFCs

**Issue number:** #1174

### PR status

## Checklist

### Breaking change checklist

---


kennethwkz-mm (approved):
This is really nice to have

Issue #1172, opened 2022-11-17
Maintenance: Revisit return types of `BaseProvider`


dreamorosi:
### Summary

`BaseProvider` was the first base class implemented for the new Parameters utility.

At the time, it was not yet known what each AWS SDK provider would return. For instance `DynamoDBProvider` might return different types than `SSMProvider`, so the types in `BaseProvider` have been intentionally left generic.

This issue serves to track the task that will have to be carried out after all providers have been implemented.

### Why is this needed?

To enforce strong typing to the values returned by the utility, whenever possible.

### Which area does this relate to?

Parameters

### Acknowledgment

- [X] This request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/#tenets)
- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [Java](https://github.com/awslabs/aws-lambda-powertools-java/)

shdq:
`AppConfig` API returns `Uint8Array`, and the python implementation of the AppConfig provider implicitly converts it to a string. I did [the same](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/1200/files#diff-67776e521a5379e61f3fe11a4d0756663d9701eed5e480060d08dbc986f4cd8fR76-R78) for now, but we have the `transformValue` function in the `BaseProvider`, which I think should do that. Correct me if I'm wrong – the result for the end user is to get raw data (binary) or transformed value to the end type, which is `json` for `AppConfig`.

dreamorosi:
My understanding is that, whenever the user specifies a `transform` when calling `get`, we should apply that transform and return the corresponding type. If no transform is specified, we can return as-is. If a transform has to be applied, then yes, it should be handled by the `transformValue` method from the `BaseProvider`, this applies to all providers.

I don't have 100% confidence on this, but I think your assessment is correct, and the return type on the Python's function might be wrong. Looking at the [docs section for this method](https://awslabs.github.io/aws-lambda-powertools-python/2.5.0/utilities/parameters/#appconfigprovider), the type annotation seems to be `bytes`:
```py
# Retrieve a single secret
value: bytes = appconf_provider.get("my_conf")
```

I think it's worth opening an issue on the Python's repo so that it can be looked at be fixed. If instead we both understood it wrong, then they can clarify what's the expected behavior :)

[...]

PR #1205, opened 2022-12-29
refactor(parameters): BaseProvider support for Uint8Array

dreamorosi:

This PR introduces some changes to the typing in `BaseProvider` so that it's now able to handle, transform, or return `Uint8Array` objects without necessarily coercing them to `string`.

The change came from the fact that AWS Secrets Manager can return both `string` (when the secret is stored as `SecretString`) and pure binaries aka `Uint8Array` (when the secret is stored as `SecretBinary`).

Consumers of the library should be able to both retrieve the latter and have them automatically transformed by the Parameters utility but also have them returned as-is (`Uint8Array`) in case they don't want to apply any transform.

The PR enables the behavior described above.

The PR introduces also some light housekeeping around dependencies (switches from `@aws-sdk/util-base64-node` to `@aws-sdk/util-base64` due to the former having being marked as being on a deprecation path) and the introduction of the `DEFAULT_PROVIDERS` object which is required by the other providers.

### How to verify this change

See status checks under this PR.

### Related issues, RFCs

**Issue number:** #1172

### PR status

## Checklist

### Breaking change checklist

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16471,18 +16471,7 @@
       "version": "1.4.1",
       "license": "MIT-0",
       "dependencies": {
-        "@aws-sdk/util-base64-node": "^3.209.0"
-      }
-    },
-    "packages/parameters/node_modules/@aws-sdk/util-base64-node": {
-      "version": "3.209.0",
-      "license": "Apache-2.0",
-      "dependencies": {
-        "@aws-sdk/util-buffer-from": "3.208.0",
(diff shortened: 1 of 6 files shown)


dreamorosi:
Given that at this time only `DynamoDBProvider`, `SSMProvider`, and `SecretsProvider` return types are fairly clear, we'll leave the linked issue open after merging this PR, as there might be some additional refactoring needed.

Issue #1175, opened 2022-11-17
Feature request: implement `SecretsProvider`


dreamorosi:
### Summary

The Parameters utility should allow customers to retrieve one parameter stored in a Secrets Manager.

This work includes the following tasks:
- [x] Write provider + types
- [x] Unit tests (100% coverage)

### Why is this needed?

1/ So that customers are able to retrieve parameters from Secrets Manager, 2/ to reach feature parity with Powertools for Python.

### Which area does this relate to?

Parameters

### Acknowledgment

- [X] This request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/#tenets)
- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [Java](https://github.com/awslabs/aws-lambda-powertools-java/)

Issue #1043, opened 2022-08-04
Docs: create API docs


dreamorosi:
## Description of the improvement

**Summary of the proposal**
As a part of our documentation we have a [section that focuses on the APIs available](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/api/). The new utility should feature docstrings that will then populate the corresponding pages of this documentation.

**How, where did you look for information**
This unit of work tracks the activities related to having the API docs for the new utility. On an high level, these are:
- Create docstrings for each class / method / function
- Review content accuracy & wording
- Review visually the docs locally

**Missing or unclear documentation**
https://awslabs.github.io/aws-lambda-powertools-typescript/latest/api/modules/_aws_lambda_powertools_parameters.html

**Improvement**
N/A

### Related existing documentation

[N/A](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/api/

### Related issues, RFCs

#846

dreamorosi:
It seems that it's not possible to have multiple entry points and so, in order to have the documentation, we'll have to export all providers from the main `src/index.ts`.

I did some tests both with `esbuild` and `tsc` and as long as using scoped imports like (note ending):
```
import { SSMProvider } from '@aws-lambda-powertools/parameters/ssm';
```

It seems to work without throwing missing dependency errors for the other providers.

PR #1206, opened 2022-12-29
feat(parameters): SecretsProvider support

dreamorosi:

This PR introduces the `SecretsProvider` class which will allow the future Parameters utility to support retrieving secrets from AWS Secrets Manager.

The `SecretsProvider` supports two retrieval modes, both for retrieving one secret at the time. This is in line with the current functionalities supported by the Parameters utility in Powertools for Python.

- `get` this class method allows customers to retrieve a single secret by name and allows to specify aspects like the transformation to apply (none, binary, json, automatic) and max age (aka time to live for the item in the cache).

The class method will be used as such:
```ts
const secrets = new SecretsProvider();

const secret = secrets.get('foo');
console.log(secret); // value as string or binary
```

- `getSecret` utility function that behaves the same as the the method, but doesn't require instantiating a class. 

```ts
const secret = getSecret('foo');
console.log(secret); // value as string or binary
```

In both cases users can configure how long the secret should be cached in the Lambda's execution environment:
```ts
const secretOne = secrets.get('foo', { maxAge: 5000 }); // This will be cached for 5 seconds
// or
const secretTwo = getSecret('foo', { maxAge: 10000 }); // This will be cached for 10 seconds
```

Secrets Manger allows to store secrets both as `SecretString` and `SecretBinary`. Both of them can be base64-encoded.

Using the provider or helper function, users can optionally specify a transform (`json`, `binary`, or `auto`) and SecretsProvider will try to transform the string or binary accordingly. When using `auto`, the name of the secret must have a suffix that specifies the transform to use (i.e. `my-secret.json` will result in a `json` transform). If no transform is specified, the secret will be returned as-is (aka `string` or `Uint8Array`).

Examples:

> **Note**
> The examples below use the `getSecret` notation for brevity, but the are 1:1 equivalent to `SecretsProvider.get(...)`

```ts
// SecretString: foo.json - value: "{ \"hello\": \"world\" }" (stringified version of an object)
const a = getSecret('foo.json');
console.log(a); // "{ \"hello\": \"world\" }" (returned as-is)

const b = getSecret('foo.json', { transform: 'auto' });
console.log(b); // { hello: 'world' } (type object)

const c = getSecret('foo.json', { transform: 'json' });
console.log(c); // { hello: 'world' } (type object)

// SecretString: bar.binary - value: "Yw==" (base64 encoded string - 'hello world')
const a = getSecret('bar.binary');
console.log(a); // "Yw==" (returned as-is)

const b = getSecret('bar.binary', { transform: 'auto' });
console.log(b); // "hello world" (type string - base64 decoded)

const c = getSecret('bar.binary', { transform: 'binary' });
console.log(c); // "hello world" (type string - base64 decoded)

// SecretBinary: baz.binary - value: "Yw==" (base64 encoded string - 'hello world')
const a = getSecret('bar.binary');
console.log(a); // [104, 101, 108, 108, 111,  32, 119, 111, 114, 108, 100] (Uint8Array - returned as-is)

const b = getSecret('bar.binary', { transform: 'auto' });
console.log(b); // "hello world" (type string - base64 decoded)

const c = getSecret('bar.binary', { transform: 'binary' });
console.log(c); // "hello world" (type string - base64 decoded)
```

Finally, users can also customize the behavior of the underlying AWS SDK client calls by passing an optional `sdkOptions` of type `GetSecretValueCommandInput` in the second argument:
```ts
const secret = getSecret('bar.binary', {
  sdkOptions: {
    VersionId: 'test-version',
  }
});
// or
const secrets = new SecretsProvider();

const secret = secrets.get('bar.binary', {
  sdkOptions: {
    VersionId: 'test-version',
  }
});
```

Or, **for class-based usage only**, they can pass their own SDK client config (of type `SecretsManagerClient`) altogether while initialising the instance:
```ts
const secrets = new SecretsProvider({
  clientConfig: { region: 'eu-west-1' }
});
```

### How to verify this change

See the newly added unit test cases & optionally compare also the API surface with the implementation [found in Powertools for Python](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/utilities/parameters/secrets.py).

### Related issues, RFCs

**Issue number:** #1175 

### PR status

## Checklist

### Breaking change checklist

---

### DIFF
--- a/packages/parameters/src/BaseProvider.ts
+++ b/packages/parameters/src/BaseProvider.ts
@@ -39,7 +39,7 @@ abstract class BaseProvider implements BaseProviderInterface {
    * this should be an acceptable tradeoff.
    * 
    * @param {string} name - Parameter name
-   * @param {GetOptionsInterface|DynamoDBGetOptionsInterface} options - Options to configure maximum age, trasformation, AWS SDK options, or force fetch
+   * @param {GetOptionsInterface|SecretsGetOptionsInterface} options - Options to configure maximum age, trasformation, AWS SDK options, or force fetch
    */
   public async get(name: string, options?: SecretsGetOptionsInterface): Promise<undefined | string | Uint8Array | Record<string, unknown>>;
   public async get(name: string, options?: GetOptionsInterface): Promise<undefined | string | Uint8Array | Record<string, unknown>> {
--- a/packages/parameters/src/BaseProvider.ts
+++ b/packages/parameters/src/BaseProvider.ts
@@ -5,6 +5,7 @@ import { ExpirableValue } from './ExpirableValue';
 import { TRANSFORM_METHOD_BINARY, TRANSFORM_METHOD_JSON } from './constants';
 import { GetParameterError, TransformParameterError } from './Exceptions';
 import type { BaseProviderInterface, GetMultipleOptionsInterface, GetOptionsInterface, TransformOptions } from './types';
+import type { SecretsGetOptionsInterface } from './types/SecretsProvider';
 
 // These providers are dinamycally intialized on first use of the helper functions
 const DEFAULT_PROVIDERS: Record<string, BaseProvider> = {};
@@ -38,8 +39,9 @@ abstract class BaseProvider implements BaseProviderInterface {
    * this should be an acceptable tradeoff.
    * 
    * @param {string} name - Parameter name
-   * @param {GetOptionsInterface} options - Options to configure maximum age, trasformation, AWS SDK options, or force fetch
+   * @param {GetOptionsInterface|DynamoDBGetOptionsInterface} options - Options to configure maximum age, trasformation, AWS SDK options, or force fetch
    */
+  public async get(name: string, options?: SecretsGetOptionsInterface): Promise<undefined | string | Uint8Array | Record<string, unknown>>;
   public async get(name: string, options?: GetOptionsInterface): Promise<undefined | string | Uint8Array | Record<string, unknown>> {
     const configs = new GetOptions(options);
     const key = [ name, configs.transform ].toString();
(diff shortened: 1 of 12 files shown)


packages/parameters/src/BaseProvider.ts
```diff
@@ -38,8 +39,9 @@ abstract class BaseProvider implements BaseProviderInterface {
    * this should be an acceptable tradeoff.
    * 
    * @param {string} name - Parameter name
-   * @param {GetOptionsInterface} options - Options to configure maximum age, trasformation, AWS SDK options, or force fetch
+   * @param {GetOptionsInterface|DynamoDBGetOptionsInterface} options - Options to configure maximum age, trasformation, AWS SDK options, or force fetch
```
shdq, line 42: I think you accidentally copied it from `DynamoDBProvider`

dreamorosi:
```suggestion
   * @param {GetOptionsInterface|SecretsGetOptionsInterface} options - Options to configure maximum age, trasformation, AWS SDK options, or force fetch
```

shdq (dismissed):
Great work!

Should we have tests for the transformation in providers or having them in `BaseProvider` is enough?

I would add JSDoc for class and methods with a couple of examples (from the PR description).

It looks good to me.

dreamorosi:
> Great work!

Thank you for the review!

> Should we have tests for the transformation in providers or having them in `BaseProvider` is enough?

I think testing in the `BaseProvider` should be enough for now, we'll add more testing in the integration tests against real end to end retrievals.

> I would add JSDoc for class and methods with a couple of examples (from the PR description).

Definitely, we have a dedicated issue for it (#1043). When doing the planning I put it in a separate one because I thought it'd make more sense to do it once the code is more or less stable. I expect to do this sometime towards the end of the integration tests & before the beta release.

Issue #1080, opened 2022-08-28
Bug: unable to build with tsc when not using @middy/core


rreddypally:
## Bug description

The code doesn't compile using `tsc` when using the Logger module without middy.

### Expected Behavior

Should be able to use Logger without the need to import middy when its not used. This is possible by importing the submodule via `import { Logger } from "@aws-lambda-powertools/logger/lib/Logger";`. If this is the solution maybe the documentation should reflect that.

### Current Behavior

Middy is required atleast as the dev dependency while compiling with `tsc`. This is not a problem when `esbuild` or `ts-node` is used.

### Steps to Reproduce

Steps:
1. npm install and import `Logger` using `import { Logger } from "@aws-lambda-powertools/logger";`
2. Build project using `tsc`
3. Expect this error `Cannot find module '@middy/core' or its corresponding type declarations.`

`package.json`
```
{
  "scripts": {
    "build": "npm i && tsc logger-demo.ts"
  },
  "dependencies": {
    "@aws-lambda-powertools/logger": "1.2.0",
    "@types/aws-lambda": "8.10.101"
    
  },
  "devDependencies": {
    "typescript": "4.7.4"
  }
}
```
`logger-demo`
```
import { Logger } from "@aws-lambda-powertools/logger/lib/Logger";

const logger = new Logger();
logger.info("Testing logger");
```
`tsconfig.json`
```
{
    "compilerOptions": {
        "experimentalDecorators": true,
        "target": "ES2021",
        "module": "CommonJS"
    }
}
```

### Environment

* **Powertools version used**: 1.2.1
* **Packaging format (Layers, npm)**: npm
* **AWS Lambda function runtime:** node16
* **Debugging logs:**n/a

### Related issues, RFCs

https://github.com/awslabs/aws-lambda-powertools-typescript/issues/1068

dreamorosi:
@obiwabrakenobi, not at this time, however we're happy to review & merge any PR coming from the community to fix the issue in a more timely manner.

obiwabrakenobi:
I'll have a look into it

dreamorosi:
Thank you!

ValeryShvyndzikau:
Hi "AWS Lambda Powertools Team", 
We are going to use Tracer, Logger, and Metrics for observability on our project and we are building our codebase with "tsc". Middy is still under consideration, so it would be nice to avoid redundant dependency on the extra library we don't use.
[...]

PR #511, opened 2022-01-25
feat(all): make `@middy/core` optional

dreamorosi:
## Description of your changes

As detailed in #373 `@middy/core` has been bundled in all utilities which means customers who don't use the middleware implementations will have an unnecessary dependency bundled in their Lambda deployment package. This PR aims at correcting that.

Changes:
- Moved `@middy/core` to `devDependency` in all utilities that used it (updated also lock file)
- Changed imports of Middy to include only types
- Updated documentation to specify that Middy must now be installed with `npm i @middy/core`
- Verified that `@middy/core` is a `dependency` in `examples/cdk`

### How to verify this change

See checks execution results + [e2e result here](https://github.com/awslabs/aws-lambda-powertools-typescript/runs/4936612711).

Also ran `npm pack` on one of the utilities and then installed it in a new dummy project (`@middy/core` not installed):
```sh
~/Codes
❯ mkdir powert-est              

~/Codes
❯ cd powert-est 

~/Codes/powert-est
❯ npm init -y
Wrote to /Users/aamorosi/Codes/powert-est/package.json:

{
  "name": "powert-est",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC"
}



~/Codes/powert-est
❯ mv ../aws-lambda-powertools-typescript/packages/tracing/aws-lambda-powertools-tracer-0.4.0.tgz .

~/Codes/powert-est
❯ npm i aws-lambda-powertools-tracer-0.4.0.tgz 

added 15 packages, and audited 16 packages in 7s

found 0 vulnerabilities

~/Codes/powert-est 7s
❯ ls -la node_modules 
total 16
drwxr-xr-x  15 aamorosi  staff   480 Jan 25 13:15 .
drwxr-xr-x   6 aamorosi  staff   192 Jan 25 13:15 ..
drwxr-xr-x   3 aamorosi  staff    96 Jan 25 13:15 .bin
-rw-r--r--   1 aamorosi  staff  5329 Jan 25 13:15 .package-lock.json
drwxr-xr-x   4 aamorosi  staff   128 Jan 25 13:15 @aws-lambda-powertools
drwxr-xr-x   4 aamorosi  staff   128 Jan 25 13:15 @aws-sdk
drwxr-xr-x   5 aamorosi  staff   160 Jan 25 13:15 @types
[...]

PR #1218, opened 2023-01-10
fix(middleware): add @middy/core as dependency

obiwabrakenobi:

As middy is only added as dev dependency any projects using this library with typescript `tsc` fails to compile as is requires middy as dependency. So you need to manually add it to your project. Adding it as dependency to each of the packages installs the required middy package when running `npm install`.

### How to verify this change

The package.json of the respective workspaces have added middy (updated) as dependency and its being removed from the dev dependency.

### Related issues, RFCs

**Issue number:**  #1080

### PR status

## Checklist

### Breaking change checklist

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,6 @@
         "@aws-cdk/cloudformation-diff": "^2.55.0",
         "@aws-cdk/cx-api": "^2.55.0",
         "@commitlint/cli": "^17.3.0",
-        "@middy/core": "^3.6.2",
         "@types/aws-lambda": "^8.10.109",
         "@types/jest": "^29.2.4",
         "@types/node": "^18.11.15",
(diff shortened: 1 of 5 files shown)


dreamorosi:
Hi @obiwabrakenobi, thank you for taking the time to open the PR and propose a solution, however I don't think this is the fix we are looking for.

Making `@middy/core` a dependency means that it will be included and bundled for all users of Powertools, even those who are not actually using Middy. This used to be the case in the past, but Middy was moved to `devDependency` in #511 to avoid this.

A solution for #1080 should account for `@middy/core` being a `devDependency` and allow for `tsc` to correctly bundle when it's not installed.

obiwabrakenobi:
Yes I understand your argument. I found it reasonable to optimize the DX by installing the dependency right away compromising the costs with bundling. As far as I saw middy/core is pretty small. Adding it as dev dependency is further a confusing fact because after installing the build fails.

The only other option I see so far is to create additional packages for the respective middleware’s. If you want I can refactor the packages towards isolated packages for the middy middleware.

dreamorosi:
I will close the PR for now, we can reopen it or create a new one once we have a solution.

PR #1257, opened 2023-01-19
tests(parameters): integration tests for `SSMProvider`

dreamorosi:

This PR introduces the test suite that runs integration tests on the `SSMProvider` which is part of the upcoming Parameters utility. The PR also extends the `ResourceAccessGranter` helper to add compatibility for SSM.

### Test suite

This test suite deploys a CDK stack with a Lambda function and a number of SSM parameters. The function code uses the Parameters utility to retrieve the SSM parameters. It then logs the values to CloudWatch Logs as JSON objects.

Once the stack is deployed, the Lambda function is invoked and the CloudWatch Logs are retrieved. The logs are then parsed and the values are checked against the expected values for each test case.

### The parameters created are:

- Name: param/a - Value: foo
- Name: param/b - Value: bar
- Name: param-encrypted/a - Value: foo-encrypted
- Name: param-encrypted/b - Value: bar-encrypted

These parameters allow to retrieve one or more parameters both by name and by path, as well as mixing encrypted and unencrypted parameters.

#### Test 1
get a single parameter by name with default options

#### Test 2
get a single parameter by name with decrypt

#### Test 3
get multiple parameters by path with default options

#### Test 4
get multiple parameters by path recursively (aka. get all parameters under a path recursively) i.e. given /param, retrieve /param/get/a and /param/get/b (note path depth)

#### Test 5
get multiple parameters by path with decrypt

#### Test 6
get multiple parameters by name with default options

#### Test 7
get multiple parameters by name, some of them encrypted and some not

#### Test 8
get parameter twice with middleware, which counts the number of requests, we check later if we only called SSM API once

#### Test 9
get parameter twice, but force fetch 2nd time, we count number of SDK requests and check that we made two API calls

### How to verify this change

Given that the integration tests are not yet hooked up to the GitHub workflow that runs them on GitHub, at the moment the only way to test this is to checkout the repo locally and then run: npm run test:e2e:nodejs18x -w packages/parameters.

The command above will run the tests on your currently logged in AWS account and use the Node.js 18 runtime.

Below a screenshot of the result as run on my machine:
image

### Related issues, RFCs

**Issue number:** #1240

### PR status

## Checklist

### Breaking change checklist

---

### DIFF
--- a/packages/parameters/src/ssm/SSMProvider.ts
+++ b/packages/parameters/src/ssm/SSMProvider.ts
@@ -147,10 +147,8 @@ class SSMProvider extends BaseProvider {
       ...(options?.sdkOptions || {}),
       Name: name,
     };
-    if (options) {
-      if (options.hasOwnProperty('decrypt'))
-        sdkOptions.WithDecryption = options.decrypt;
-    }
+    sdkOptions.WithDecryption = options?.decrypt !== undefined ?
+      options.decrypt : sdkOptions.WithDecryption;
     const result = await this.client.send(new GetParameterCommand(sdkOptions));
 
     return result.Parameter?.Value;
@@ -167,15 +165,12 @@ class SSMProvider extends BaseProvider {
     const paginationOptions: PaginationConfiguration = {
       client: this.client
     };
-    if (options) {
-      if (options.hasOwnProperty('decrypt'))
-        sdkOptions.WithDecryption = options.decrypt;
-      if (options.hasOwnProperty('recursive'))
-        sdkOptions.Recursive = options.recursive;
-      if (sdkOptions.MaxResults) {
-        paginationOptions.pageSize = sdkOptions.MaxResults;
-      }
-    }
+    sdkOptions.WithDecryption = options?.decrypt !== undefined ?
+      options.decrypt : sdkOptions.WithDecryption;
+    sdkOptions.Recursive = options?.recursive !== undefined ?
+      options.recursive : sdkOptions.Recursive;
+    paginationOptions.pageSize = sdkOptions.MaxResults !== undefined ?
+      sdkOptions.MaxResults : undefined;
     
     const parameters: Record<string, string | undefined> = {};
     for await (const page of paginateGetParametersByPath(paginationOptions, sdkOptions)) {
@@ -388,13 +383,11 @@ class SSMProvider extends BaseProvider {
       const overrides = parameterOptions;
       overrides.transform = overrides.transform || configs.transform;
 
-      if (!overrides.hasOwnProperty('decrypt')) {
-        overrides.decrypt = configs.decrypt;
-      }
-      if (!overrides.hasOwnProperty('maxAge')) {
-        overrides.maxAge = configs.maxAge;
-      }
-
+      overrides.decrypt = overrides.decrypt !== undefined ?
+        overrides.decrypt : configs.decrypt;
+      overrides.maxAge = overrides.maxAge !== undefined ?
+        overrides.maxAge : configs.maxAge;
+      
       if (overrides.decrypt) {
         parametersToDecrypt[parameterName] = overrides;
       } else {
--- a/packages/parameters/src/ssm/SSMProvider.ts
+++ b/packages/parameters/src/ssm/SSMProvider.ts
@@ -144,13 +144,12 @@ class SSMProvider extends BaseProvider {
     options?: SSMGetOptionsInterface
   ): Promise<string | undefined> {
     const sdkOptions: GetParameterCommandInput = {
+      ...(options?.sdkOptions || {}),
       Name: name,
     };
     if (options) {
-      if (options.hasOwnProperty('decrypt')) sdkOptions.WithDecryption = options.decrypt;
-      if (options.hasOwnProperty('sdkOptions')) {
-        Object.assign(sdkOptions, options.sdkOptions);
-      }
+      if (options.hasOwnProperty('decrypt'))
+        sdkOptions.WithDecryption = options.decrypt;
     }
     const result = await this.client.send(new GetParameterCommand(sdkOptions));
 
@@ -162,19 +161,19 @@ class SSMProvider extends BaseProvider {
     options?: SSMGetMultipleOptionsInterface
   ): Promise<Record<string, string | undefined>> {
     const sdkOptions: GetParametersByPathCommandInput = {
+      ...(options?.sdkOptions || {}),
       Path: path,
     };
     const paginationOptions: PaginationConfiguration = {
       client: this.client
     };
     if (options) {
-      if (options.hasOwnProperty('decrypt')) sdkOptions.WithDecryption = options.decrypt;
-      if (options.hasOwnProperty('recursive')) sdkOptions.Recursive = options.recursive;
-      if (options.hasOwnProperty('sdkOptions')) {
-        Object.assign(sdkOptions, options.sdkOptions);
-        if (sdkOptions.MaxResults) {
-          paginationOptions.pageSize = sdkOptions.MaxResults;
-        }
+      if (options.hasOwnProperty('decrypt'))
+        sdkOptions.WithDecryption = options.decrypt;
+      if (options.hasOwnProperty('recursive'))
+        sdkOptions.Recursive = options.recursive;
+      if (sdkOptions.MaxResults) {
+        paginationOptions.pageSize = sdkOptions.MaxResults;
       }
     }
     
--- a/packages/parameters/tests/e2e/ssmProvider.class.test.functionCode.ts
+++ b/packages/parameters/tests/e2e/ssmProvider.class.test.functionCode.ts
@@ -1,134 +1,182 @@
 import { Context } from 'aws-lambda';
-import { SSMProvider } from '../../src/ssm';
+import {
+  SSMProvider,
+} from '../../src/ssm';
+import {
+  SSMGetOptionsInterface,
+  SSMGetMultipleOptionsInterface,
+  SSMGetParametersByNameOptionsInterface
+} from '../../src/types';
 import { TinyLogger } from '../helpers/tinyLogger';
-// # TODO: Uncomment code below once #1222 is fixed
-/*
 import { middleware } from '../helpers/sdkMiddlewareRequestCounter';
 import { SSMClient } from '@aws-sdk/client-ssm';
-*/
+
+// We use a custom logger to log pure JSON objects to stdout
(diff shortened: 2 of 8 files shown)


dreamorosi:
Hi @am29d, review appreciated whenever you have bandwidth

packages/parameters/src/ssm/SSMProvider.ts
```diff
@@ -144,13 +144,12 @@ class SSMProvider extends BaseProvider {
...
+      ...(options?.sdkOptions || {}),
       Name: name,
     };
     if (options) {
-      if (options.hasOwnProperty('decrypt')) sdkOptions.WithDecryption = options.decrypt;
-      if (options.hasOwnProperty('sdkOptions')) {
-        Object.assign(sdkOptions, options.sdkOptions);
-      }
+      if (options.hasOwnProperty('decrypt'))
+        sdkOptions.WithDecryption = options.decrypt;
```
am29d, line 152: Now that we have types, ca we reduce the if blocks to a more simpler version?
```ts
sdk.WithDecryption = options?.decrypt? : options.decrypt : undefined;
```

packages/parameters/src/ssm/SSMProvider.ts
```diff
@@ -162,19 +161,19 @@ class SSMProvider extends BaseProvider {
...
-        Object.assign(sdkOptions, options.sdkOptions);
-        if (sdkOptions.MaxResults) {
-          paginationOptions.pageSize = sdkOptions.MaxResults;
-        }
+      if (options.hasOwnProperty('decrypt'))
+        sdkOptions.WithDecryption = options.decrypt;
+      if (options.hasOwnProperty('recursive'))
+        sdkOptions.Recursive = options.recursive;
+      if (sdkOptions.MaxResults) {
+        paginationOptions.pageSize = sdkOptions.MaxResults;
```
am29d, line 176: Similar to previous typing comment we can assign properties of they are present in options:
```ts
    sdkOptions.WithDecryption = options?.decrypt ? options.decrypt : undefined;
    sdkOptions.Recursive = options?.recursive ? options.recursive : undefined;
    paginationOptions.pageSize = sdkOptions?.MaxResults ? sdkOptions.MaxResults : undefined;
```

packages/parameters/tests/e2e/ssmProvider.class.test.functionCode.ts
```diff
@@ -0,0 +1,183 @@
...
+export const handler = async (_event: unknown, _context: Context): Promise<void> => {
+  // Test 1 - get a single parameter by name with default options
+  await _call_get(paramA, 'get');
+  
+  // Test 2 - get a single parameter by name with decrypt
+  await _call_get(paramEncryptedA, 'get-decrypt', { decrypt: true });
+  
+  // Test 3 - get multiple parameters by path with default options
+  // Get path (/param/get)
+  const parameterPath = paramA.substring(0, paramA.lastIndexOf('/'));
```
am29d, line 119: I see that you want to recycle params in different tests. While the comment above helps to understand the value, it's still difficult to reason, because we use one parameter in different test cases.

I'd suggest so separate the resources and dedicate one specific set of parameter for each test case. This will prevent any unexpected side effects, if maintainers will change `paramA` in the future, for the first test and might break other tests.

packages/parameters/tests/helpers/cdkAspectGrantAccess.ts
```diff
@@ -30,12 +35,27 @@ export class ResourceAccessGranter implements IAspect {
...
+        } else if (
+          resource instanceof Secret
+        ) {
+          resource.grantRead(node);
+        } else if (isStringParameterGeneric(resource)) {          
           resource.grantRead(node);
+          // Grant access also to the path of the parameter
+          node.addToRolePolicy(
+            new PolicyStatement({
+              actions: [
```
am29d, line 51: Is `Effect.ALLOW` not necessary, or implicit, or forgot to add?

packages/parameters/tests/helpers/ssmSecureStringCdk.ts
```diff
@@ -0,0 +1,55 @@
...
+import {
+  SSMClient,
+  PutParameterCommand,
+  DeleteParameterCommand
+} from '@aws-sdk/client-ssm';
+
+const client = new SSMClient({});
+
+/**
+ * Create a new SSM SecureString parameter, overwriting any existing parameter with the same name if it exists.
```
am29d, line 14: I am not sure why we need to update parameter during the test 🤔. Which test cases require that? Or do you simply want to overwrite them in case they were not cleaned up properly before?

am29d (changes requested):
Looks great! Left few minor observations.

packages/parameters/tests/helpers/ssmSecureStringCdk.ts
dreamorosi: CDK doesn't support creating SecureString parameter, so we use a custom resource.

The comment refers only to the fact that in this custom resource handler `event.RequestType === 'Create' || event.RequestType === 'Update'` are handled by the same code path (which creates the parameter regardless of whether it exists or not).

This stack can only be created when the test is run, and it's tore down immediately after. As such, the request should never be `Update` as we don't have a way of updating the stack.

Happy to change the `if/else` statement at L47 and have instead:
```ts
if (event.RequestType === 'Create') {
     await createResource(event);
} else if (event.RequestType === 'Delete') {
     await deleteResource(event);
} else { // This now includes event.RequestType === 'Update'
     console.error('Unknown request type', event);
     throw new Error('Unknown request type');
}
```

What do you think?

packages/parameters/tests/helpers/cdkAspectGrantAccess.ts
dreamorosi: It should be implicit, since it runs successfully & it didn't before me adding this statement.

However I agree it could be made even more explicit so I'll add it.

packages/parameters/tests/e2e/ssmProvider.class.test.functionCode.ts
dreamorosi: Yes, I was undecided about this as I agree it's a good practice for tests to be independent.

However here I was trying to balance this best practice, with the fact that I would want to avoid the tests running forever due to having to create and destroy many resources at each run. 

Thoughts on this?

packages/parameters/src/ssm/SSMProvider.ts
dreamorosi: So I initially thought about this, but I wasn't sure on whether setting these values to `undefined` would overwrite the default values/behavior of the SDK, so I opted for assigning only when the user actually provides a value, otherwise leave them as they are

dreamorosi: I'm going to follow your suggestion.

I have made a quick test with a middleware that logs right before sending the request and thankfully the sdk is smart enough to remove any key that has value `undefined` before sending the request, so it's safe to do like you suggest.

packages/parameters/tests/helpers/ssmSecureStringCdk.ts
am29d: Gotcha, I didn't know about the missing feature. The stance [from CDK team is from 2020](https://github.com/aws/aws-cdk/issues/3520) to not support secrets creation, only consumptions.

packages/parameters/tests/helpers/cdkAspectGrantAccess.ts
am29d: TIL, but adding explicit helps.

packages/parameters/src/ssm/SSMProvider.ts
dreamorosi: I'm not sure this can work.

This is the current implementation:
```ts
const sdkOptions: GetParameterCommandInput = {
  ...(options?.sdkOptions || {}),
  Name: name,
};
if (options) {
  if (options.hasOwnProperty('decrypt'))
    sdkOptions.WithDecryption = options.decrypt;
  }
}
```

In this method, users can influence the behavior around decrypt in two ways:
- By using the `decrypt` option (i.e. `provider.get('something', { decrypt: true });`)
- By passing the `WithDecryption` parameter directly to the SDK (i.e. `provider.get('something', { sdkOptions: { WithDecryption: true }});`)

The former takes precedence over the second.

If we were to modify the statement like suggested, we'd always overwrite the option either with the value of `options.decrypt` or with `undefined`. This would mean never considering the value of `options.sdkOptions` when one is passed.

Now, we could make the ternary operator statement include this case and do something like this:
```ts
sdk.WithDecryption = options?.decrypt? : options.decrypt : options?.sdkOptions?.WithDecryption ? options.sdkOptions.WithDecryption : undefined;
```

but I think this, while shorter in terms of LOC it's less readable. Generally speaking I'm not too concerned about LOC since all this is transpiled to JS before being published.

Thoughts?

packages/parameters/tests/e2e/ssmProvider.class.test.functionCode.ts
am29d: I see your point. I tried to create params per test case it bloats the setup excessively. I am not concerned with the execution time, the results in my tests were very similar, between 165 and 175 seconds. 

Following back to tenets, we should keep it simple like it is right now.

packages/parameters/src/ssm/SSMProvider.ts
dreamorosi: I went with this one:
```ts
sdk.WithDecryption = options?.decrypt !== undefined ? options.decrypt : sdk.WithDecryption
```

this way, if the user passes `options.decrypt === false` we still respect their value. Not passing it would have evaluated to `false`, and thus `undefined`. For most cases the two are equal but in some cases it might need to exactly be `false`, so I made the first comparison stricter.

packages/parameters/tests/helpers/ssmSecureStringCdk.ts
dreamorosi: I ended up making the change in the `if/else`. This way we can detect that the parameter is being updated and make the stack fail. Given it's not an expected behavior I prefer to have a failure in the deployment rather than a silent bug.

Thanks for pointing this out!

dreamorosi:
Thank you for the review & comments Alex, appreciate it.

I've addressed all of them and made changes in line with the decisions we took in each one.

I have also run the e2e tests again:
image

am29d:
LGTM, we are good.

am29d (approved):

Issue #1226, opened 2023-01-11
Maintenance: make `layer-publisher` part of the main npm workspace


dreamorosi:
### Summary

The `layer-publisher` folder contains the utilities and tests used to publish AWS Lambda Layers for Powertools. When this was developed, it was developed in isolation and with its own dependency tree. Now that the component is somewhat stable, and the dependencies of the main npm workspace has been recently updated, we can bring the folder in the npm workspace so that the dependencies are shared.

### Why is this needed?

Because if the two dependency trees are not aligned we risk incurring in the issue described in #1227 every time one of the two trees change. Both trees depend on `aws-cdk-lib` which has a weekly release cadence so, while the issue is now fixed, it's likely it'll happen again.

Additionally, by having the `layer-publisher` in the main npm workspace we can decrease the project's footprint on development & CI hosts (because of sharing dependencies) and also simplify some of the workflow that are now disjointed.

### Which area does this relate to?

Automation, Other

### Acknowledgment

- [X] This request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/#tenets)
- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [Java](https://github.com/awslabs/aws-lambda-powertools-java/)

dreamorosi:
While addressing this issue, also try removing the hardcoded CDK version from [this workflow](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/workflows/publish_layer.yaml#L46).

dreamorosi:
Also remove this other hardcoded version in [this other workflow](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/.github/workflows/reusable_deploy_layer_stack.yml#L68).

PR #1292, opened 2023-02-16
chore(layers): moved `layer-publisher` into npm workspace

dreamorosi:

This PR introduces changes to the files related to publishing public layers that are aimed at:
- Moving the utility into the project's npm workspace
- Reducing technical debt on this area

The first point has already been explained at length in the linked issue (#1226), but the TL;DR; is that by bringing this code into the workspace we can make sure all shared dependencies (which happen to be almost all of them) are version aligned. This is especially desirable given that this code is CDK-heavy and that library infamously doesn't play well with version mismatches.

In regards to the second point, the following high level changes have been made:
- Renamed folder from `layer-publisher` to `layers` (same as Powertools for Python)
- Removed layers from pre-commit / pre-push hooks, all tests are done on the PR. Maintainers can still run the tests locally as needed albeit manually.
- Changed unit test structure to align more closely with the rest of the project. Also changed testing strategy moving away from snapshot testing and towards making assertions on the resources present in the generated stack.
- Changed integration test structure to align more closely with the rest of the project. Also broken down the various test cases to allow maintainers to quickly narrow down the area of code in case of issues.
- Simplified the layer publisher CDK stack by removing all logic related to spawning child processes (now done directly in the test runner)

### How to verify this change

There are three areas that would allow us to definitively say the changes are correct, one of which we can't test until after merging this PR.

The first one is by looking at the checks under this PR. The tests running as a part of this PR include unit tests of the changes. The second one are integration tests, which create a layer similar to how it would be done once pushing to prod and then tests that the file bundled in it are compatible with an actual Lambda function, [here's a successful passing run](https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/4203851247).

The third one would be to actually publish the layers to the production AWS environment. Doing that however would cause the layer version number to increment, which is not desirable. For this reason we have opted to defer this test to the next release date, since the two previous tests already make us confident that any potential issue will be limited to the workflows in charge of publishing the layer, rather than the layer itself.

### Related issues, RFCs

**Issue number:** #1226

### PR status

## Checklist

### Breaking change checklist

---


layers/tests/e2e/layerPublisher.test.ts
```diff
@@ -0,0 +1,147 @@
...
+      functionEntry: path.join(__dirname, lambdaFunctionCodeFile),
+      tracing: Tracing.ACTIVE,
+      environment: {
+        UUID: uuid,
+        POWERTOOLS_PACKAGE_VERSION: powerToolsPackageVersion,
+        POWERTOOLS_SERVICE_NAME: 'LayerPublisherStack',
+      },
+      runtime: runtime,
+      bundling: {
+        externalModules: [
```
am29d, line 84: I am not sure why we need to bundle the modules, we already have the layer, right?

am29d (changes requested):
Looks great! Two small changes:

Need to fix the path to the new file here to `bin/layers.ts` 

https://github.com/awslabs/aws-lambda-powertools-typescript/blob/1e36e45800119810d536a76427f86c24f9a3df35/layers/cdk.json#L2

layers/tests/e2e/layerPublisher.test.ts
dreamorosi: Counterintuitively enough, this [means](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda_nodejs.BundlingOptions.html#externalmodules):
> A list of modules that should be considered as externals (already available in the runtime).

Given they are already in the layer like you said, we don't want `esbuild` to include them in the bundle.

am29d (approved):

layers/tests/e2e/layerPublisher.test.ts
am29d: oh, it's the other way around 😅 

> A list of modules that should be considered as externals (already available in the runtime).

PR #1308, opened 2023-02-17
refactor(tracer): allow streaming threshold configuration

nathanagez:

- Add setStreamingThreshold method to `ProviderServiceInterface.ts` 
- Implement setStreamingThreshold method 
- Add unit-tests

Initially the `setStreamingThreshold` method was not exposed.
Having encountered a Segment too large issue (with a lot of requests to DynamoDB), the only way to bypass the issue was to set the Streaming Threshold to a value of 0. 
This is possible when using `aws-xray-sdk-node` directly but not `aws-lambda-powertools-typescript` because the setStreamingThreshold method was not exposed.

### Related issues, RFCs

**Issue number:** 

### PR status

## Checklist

### Breaking change checklist

---

### DIFF
--- a/docs/core/tracer.md
+++ b/docs/core/tracer.md
@@ -249,6 +249,16 @@ You can opt-out from this feature by setting the **`POWERTOOLS_TRACER_CAPTURE_HT
 
 ## Advanced
 
+### Configure streaming threshold
+
+By default, the SDK is configured to have a threshold of 100 subsegments per segment. This is because the UDP packet maximum size is ~65 kb, and larger segments might trigger the 'Segment too large to send' error.
+
+To remedy this, the SDK automatically sends the completed subsegments to the daemon when the threshold is breached. Additionally, subsegments that complete when over the threshold automatically send themselves. If a subsegment is sent out of band, it is pruned from the segment object. The full segment is reconstructed on the service side. You can change the threshold as needed.
+
+`tracer.setStreamingThreshold(10)`
+
+Subsegments can be marked as in_progress when sent to the daemon. The SDK is telling the service to anticipate the asynchronous subsegment to be received out of band when it has completed. When received, the in_progress subsegment is discarded in favor of the completed subsegment.
+
 ### Disabling response auto-capture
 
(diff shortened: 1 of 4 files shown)


dreamorosi:
Hi Nathan, I appreciate you taking the time to contribute to this project and open a PR, however I'm not inclined to consider merging it.

We decided not to expose this method from the SDK in the provider interface because [the SDK itself already sets the streaming threshold to 0](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/env/aws_lambda.js#L26) whenever running on Lambda. This is done to limit any potential segment loss which could happen due to the ephemeral nature of Lambda. With a streaming threshold always set to 0 the SDK immediately sends out a segment as soon as it's closed.

With this in mind, and considering that we exclusively target Lambda, I don't think its necessary to expose the method at this time.

For future reference, also please note that we like to try following the process described both in the docs ([here](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/roadmap/#process)) and in the [contributing guidelines](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/CONTRIBUTING.md#summary).
According to the process the journey for any change/addition/removal to the project, starts with an issue, continues with a discussion, and then ends in a PR only after having been agreed upon with a maintainer. 

In this way, if a change doesn't fit with the project, the contributor can save time or decide to contribute on another issue/topic.

nathanagez:
Hi @dreamorosi, thank you for your feedback. 

Indeed I just saw the link to the [SDK that sets the value directly to 0](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/env/aws_lambda.js#L26). 

It's quite strange because when we use the tracer wrapper (from the lambda powertools) and `captureAWSv3Client`, when there are huge amounts of calls to DynamoDB when I use we get a `Segment too large` error.
And when we use the SDK itself, it works as expected.

I will open an issue in this case.

dreamorosi:
Yes please, happy to look at the issue regarding the diverging behavior between the SDK and Powertools.

When you open the issue, please include a snippet of code to reproduce the issue if possible. Thanks

dreamorosi:
In light of the discussion happening in #1310 I'm going to close this PR for the time being.

If after discussing we decide that the feature is indeed needed we'll reopen it.

PR #1313, opened 2023-02-20
feat(logger): make loglevel types stricter

### LINKED ISSUE #1309: Feature request: Narrow LogLevel type to not allow arbitrary strings
### Use case

Raising this ticket after the discussion in [#1198](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/1198#issuecomment-1435041684).

Currently the `LogLevel` type allows for any string to be used, but at runtime it [validates the value](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/logger/src/Logger.ts#L745) and if it doesn't match one of the valid log levels (as per `isValidLogLevel()`) it silently ignores the log level passed in and uses the default.

This has two problems:

1. The issue isn't caught at compile time
2. The passed in log level is silently ignored which may not be detected.

Regarding 1. above, the following typo currently compiles. It would be better for the typo to be detected at compile time:

```ts
const logger = new Logger({ logLevel: 'DEBG' });
```

### Solution/User Experience

Narrow the `LogLevel` type definition, removing  the `| string` part.

### Alternative solutions

_No response_

### Acknowledgment

- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [Java](https://github.com/awslabs/aws-lambda-powertools-java/)

dreamorosi:

The Logger constructor accepts an optional `logLevel`  parameter with [default](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/logger/src/Logger.ts#L120) to `INFO`. While the library already has types for all the log levels, and performs a validation of the value passed to `logLevel`, the constructor currently accepts any type of value.

As detailed in #1309, this can lead to issues that go undetected at edit/build time since TypeScript is not being leveraged to validate these inputs. For instance, currently it's possible to pass the following value and Logger will accept the value and silently default to `INFO`:
```ts
const logger = new Logger({ logLevel: 'DEBG' }); // instead of DEBUG
```

This PR, that once merged will close #1309, narrows the types for `logLevel` so that only certain values both upper and lower case variations are accepted. If an invalid value is passed, users will see errors both in their IDE and at compile time.

For the time being I have decided to not include a warning log in case of invalid value, given that we are already narrowing the types. However I'm open to reconsider this decision in a future PR if there's demand.

### How to verify this change

See checks under this PR, as well as the [results of the integration tests run](https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/4222887432).

### Related issues, RFCs

**Issue number:** #1309 

### PR status

## Checklist

### Breaking change checklist

---


packages/logger/src/types/Log.ts
```diff
@@ -3,10 +3,10 @@ type LogLevelInfo = 'INFO';
 type LogLevelWarn = 'WARN';
 type LogLevelError = 'ERROR';
 
-type LogLevel = LogLevelDebug | LogLevelInfo | LogLevelWarn | LogLevelError | string;
+type LogLevel = LogLevelDebug | Lowercase<LogLevelDebug> | LogLevelInfo | Lowercase<LogLevelInfo> | LogLevelWarn | Lowercase<LogLevelWarn> | LogLevelError | Lowercase<LogLevelError>;
```
am29d, line 6: Is it possible to extend the types with `type LogLevelDebug = 'DEBUG' | 'debug';` to simplify it? With that we only need two `toUpperCase()` conversions: in `setLogger` and `shouldPrint`. 

Curious to understand why `Uppercase` and `Lowercase` generics were introduced.

packages/logger/tests/unit/Logger.test.ts
```diff
@@ -1419,7 +1419,7 @@ describe('Class: Logger', () => {
       };
       const childLoggerWithSampleRateEnabled = parentLogger.createChild(optionsWithSampleRateEnabled);
 
-      const optionsWithErrorLogLevel = {
+      const optionsWithErrorLogLevel: ConstructorOptions = {
```
am29d, line 1422: Just a small nitpick, the name for the Logger constructor is too generic, so it's probably better to align it with other modules to `LoggerOptions`.

packages/logger/tests/unit/helpers.test.ts
```diff
@@ -56,7 +56,7 @@ describe('Helper: createLogger function', () => {
     test('when no parameters are set, returns a Logger instance with the correct properties', () => {
 
       // Prepare
-      const loggerOptions = {
+      const loggerOptions: ConstructorOptions = {
```
am29d, line 59: We probably need also tests for lower case configuration of log level.

am29d (changes requested):

packages/logger/tests/unit/Logger.test.ts
dreamorosi: Agreed, we should revisit all the Logger types at some point in the future, for instance [this one](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/logger/src/types/Logger.ts#L7).

However I think it'd be a breaking change, so it's going to be a v2 discussion.

packages/logger/src/types/Log.ts
dreamorosi: I initially started out the same way you suggested. However later on I realized this would clash with how the levels are actually used internally in the Logger. 

The [logic that decides](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/feat/logger/stricter_loglevel_types/packages/logger/src/Logger.ts#L857) on whether a log should be emitted uses numeric values to do so, and these numeric values are [defined here](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/feat/logger/stricter_loglevel_types/packages/logger/src/Logger.ts#L133-L138).

So essentially, to avoid changing the logic too much and have to deal internally with both uppercase & lowercase, I introduced these generics so that Logger is able to accept both upper & lowercase but internally it's all still treated as [if it was uppercase only](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/feat/logger/stricter_loglevel_types/packages/logger/src/types/Log.ts#L8-L10).

packages/logger/tests/unit/Logger.test.ts
dreamorosi: [Added a comment](https://github.com/awslabs/aws-lambda-powertools-typescript/discussions/1269#discussioncomment-5114490) to the v2.0.0 milestone discussion.

packages/logger/tests/unit/helpers.test.ts
dreamorosi: Thank you for pointing this out, by adding the test I found out that I was introducing a bug.

[This commit](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/1313/commits/5488c0af993d299fa3624be459077ad07155e1ca) adds the test and fixes the logic.

packages/logger/src/types/Log.ts
am29d: The `LogLevelThresholds` with `shouldPrint` is the only place were we need to convert to uppercase. Using `setLogger` we can enforce the uppercase type value. Types only help us to handle user input when passing value to Logger constructor. After we created the object, there is no need to convert them during the runtime. Here is an example of `setLogLevel`

```ts
if (this.isValidLogLevel(logLevel)) {
      this.logLevel = logLevel?.toUpperCase() as LogLevel;

      return;
    }
```
After we know that we got a valid `LogLevel` we set it to uppercase. With this we can have

```ts
if (
      this.logLevelThresholds[logLevel.toUpperCase() as Uppercase<LogLevel>] >=
      this.logLevelThresholds[this.getLogLevel().toUpperCase() as Uppercase<LogLevel>]
    ) {
      return true;
    }
``` 

Technically the `logLevel.toUpperCase() as Uppercase<LogLevel>` is unnecessary, because we know it will be uppercase, we set it during `setLogLevel`.

`LogLevelThresholds` type remains:
```ts
type LogLevelThresholds = {
  [key in Uppercase<LogLevel>]: number;
};
```

With this we can remove some type assertions of `Uppercase` and `Lowercase`, but not all of them.

I think my main concern is that Lowercase and Uppercase assertions feels scattered across the codebase. In `Log` type we have lowercase, and in `Logger` it's all uppercase, and then we call `toUppercase()` in many places, and I'd like to avoid them.

am29d (dismissed):

packages/logger/src/types/Log.ts
dreamorosi: I see what you mean and I agree.

During the refactor I mistakenly thought that the instance of `logLevel` was user provided and that as such we should enforce its case. However as you've pointed out this is not the case and we already know it's uppercase.

In [this commit](https://github.com/awslabs/aws-lambda-powertools-typescript/pull/1313/commits/2ed61c677a0c0121f60f890326256aaea3176193) I have fixed the input type of that function and removed the typecast. 

Now the only place where `Lowercase` is used is in the `Log` types, whereas elsewhere in the `Logger` class we use always `Uppercase`. Based on my current understanding of these generics, since `LogLevel` is now mixed, we need to use `Uppercase` to tell TS that we know/want it's uppercase.

am29d (approved):

acdha:
I noticed a regression caused by this improvement where code which used to use the standard environmental variable we use across all of our applications to configure the Logger (e.g. `export const logger = new Logger({logLevel: process.env.APPNAME_LOG_LEVEL})`) now fails to build because there isn't a way to guarantee that the string from the environmental variable is one of the values in the LogLevel union, and none of the exported interfaces make it easy to check that without copy-pasting the values from Log.ts. 

I think it would be useful to have a public method similar to the `isValidLogLevel` method which would allow you to essentially say “get me the `LogLevel` for this string or throw an exception” so you could have a robust way to let linters know that execution will halt if that's not a valid value.

dreamorosi:
Hi @acdha, you should be able to import the `LogLevel` type using `import type { LogLevel } from "@aws-lambda-powertools/logger/types";` (available since v2.0.0).

With this, if you know for sure that the value of your env variable is valid, you should be able to cast the type.

If instead you don't trust/control the value and want to check at runtime, I do agree that at the moment we don't have anything exported to allow you to check easily.

I wouldn't necessarily consider this as a regression, and if anything making the type more strict (which is the content of this PR) goes in the same direction of wanting to make the input safer.

I think exporting the values and possibly adding a method/helper for customers to check this is a valid ask though. Could you please open a feature request so we can consider putting this on the backlog?

acdha:
> If instead you don't trust/control the value and want to check at runtime, I do agree that at the moment we don't have anything exported to allow you to check easily.

I actually do control the inputs completely in our Terraform configuration but the type checker can’t know that. I classed it as a regression since a point release broke our build but am happy to open a feature request tomorrow.

PR #1329, opened 2023-02-25
fix(logger): middleware stores initial persistent attributes correctly

### LINKED ISSUE #1328: Bug: Log state not cleared before logging event in middleware
### Expected Behaviour

I am currently using the `injectLambdaContext` middleware with the middy library to both clear persistent log state and log the invocation event. My lambda handler function puts a request ID on the logger for debugging purposes.

According to the docs, because I've set `clearState` to true, _persistent attributes added inside the handler will **NOT** be cached_. So, my expectation would be that the request_id attribute would never be shared between invocations (assuming each invocation contained a unique request_id).

### Current Behaviour

I am seeing the initial event logged by the middleware including persistent attributes added from a previous invocation, despite `clearState` being set to `true`

### Code snippet

```JavaScript
const logger = new Logger({
    logLevel: 'DEBUG',
    persistentLogAttributes: {
        foo: 'bar'
    }
});

const lambdaHandler = async (event: { headers: {x-request-id: string } }, _context: unknown): Promise<void> => {
    logger.appendKeys({request_id: event.headers['x-request-id']});
    ...
};

    .use(injectLambdaContext(logger, { logEvent: true, clearState: true }));
```

### Possible Solution

It seems like the logEvent step occurs in the "before" middleware, here: https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/logger/src/Logger.ts#L338 whereas the clearState step happens in the "after" middleware step here: https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/logger/src/Logger.ts#L346, with the `initialPersistentAttributes` being used to override any existing persistent attributes set in a previous invocation.

### Steps to Reproduce

1. Host a lambda behind API Gateway.
2. Invoke the lambda twice, each time using a unique `x-request-id` header in the request.
3. Observe via CloudWatch logs that the "Lambda invocation event" log line for the second invocation (assuming the second invocation was warm) has the first `x-request-id` header value in the `request_id` logger attribute.

### AWS Lambda Powertools for TypeScript version

latest

16.x

### Packaging format used

Lambda Layers

### Execution logs

_No response_

dreamorosi:

In #1328 it was reported that the Logger middleware `injectLambdaContext` was not clearing correctly the state and potentially leaking attributes set during an execution in the subsequent one.

Prior to this PR the middleware was improperly storing the initial persistent attributes of each logger instance passed to the middleware. This caused issues when clearing the state after each invocation, since the initial persistent attributes were now unusable.

This PR fixes the issue and adds new unit tests. Once merged this PR will close #1328.

<longer explanation coming soon>

Once merged, this PR closes #1328.

### How to verify this change

See checks under the PR as well as integration [test run here](https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/4272607387). New unit tests were added to demonstrate the fix & avoid future regressions.

### Related issues, RFCs

**Issue number:** #1328

### PR status

## Checklist

### Breaking change checklist

---

### DIFF
--- a/packages/logger/tests/unit/Logger.test.ts
+++ b/packages/logger/tests/unit/Logger.test.ts
@@ -1235,11 +1235,10 @@ describe('Class: Logger', () => {
     test('it awaits the decorated method correctly', async () => {
 
       // Prepare
-      const injectLambdaContextAfterOrOnErrorMock = jest.fn().mockReturnValue('worked');
-      // Temporarily override the cleanup static method so that we can "spy" on it.
-      // This method is always called after the handler has returned in the decorator
-      // implementation.
-      Logger.injectLambdaContextAfterOrOnError = injectLambdaContextAfterOrOnErrorMock;
+      const injectLambdaContextAfterOrOnErrorSpy = jest.spyOn(
+        Logger,
+        'injectLambdaContextAfterOrOnError'
+      );
       const logger = new Logger({
         logLevel: 'DEBUG',
       });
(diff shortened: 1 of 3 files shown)


packages/logger/tests/unit/Logger.test.ts
```diff
@@ -1227,11 +1227,10 @@ describe('Class: Logger', () => {
     test('when used as decorator on an async method, the method is awaited correctly', async () => {
 
       // Prepare
-      const injectLambdaContextAfterOrOnErrorMock = jest.fn().mockReturnValue('worked');
-      // Temporarily override the cleanup static method so that we can "spy" on it.
-      // This method is always called after the handler has returned in the decorator
-      // implementation.
-      Logger.injectLambdaContextAfterOrOnError = injectLambdaContextAfterOrOnErrorMock;
+      const injectLambdaContextAfterOrOnErrorSpy = jest.spyOn(
```
dreamorosi, line 1230: I changed this because I realised that the previous mock was not being cleared correctly, and we really only need a spy.

am29d (approved):
🚢

danrivett:
I believe we also ran into this, but I didn't have time at the time to investigate the root cause. Instead we overrode the `addPersistentLogAttributes()` method to clear out any existing keys matching the ones being added in this call.

This is what we added:

```ts
  // Override addPersistentLogAttributes() and not appendKeys() since the latter delegates to the former
  addPersistentLogAttributes(attributes?: LogAttributes | undefined): void {
    // I've found that attributes don't always get cleared consistently even when `clearState` is set to `true`
    // So we'll clear out any keys we're about to set to make sure
    if (attributes != null) {
      this.removePersistentLogAttributes(Object.keys(attributes));
    }

    super.addPersistentLogAttributes(attributes);
  }
```

We've upgraded to v1.6.0 but haven't yet removed this override, but once we have some time to test, we'll do so and retest as I think this PR should fix the issue we saw. If not, I'll create a separate bug later on.

PR #1347, opened 2023-02-28
feat(logger): add silent log level to suppress the emission of all logs

### LINKED ISSUE #1198: Feature request: Silent mode for all testing environments
### Use case

I'm a huge fan of the logger, thanks for all your hard work on it!! One side affect of it though is it can be super noisy when testing. It's useful to have the POWERTOOLS_DEV env variable, this is helpful. However, the problem with this is that there is no option of _less_ noise, just _silence_.

I often debug with my IDE, but on occasion i'll add a `console.log` and find it useful to see that output. With my previous favourite logger (winston) you could silence within tests and that would stop the winston logger from outputting when explicitly silenced.

### Solution/User Experience

It would be great to be able to set the logger to silent when a certain condition is met, like in a test or dev environment. This could be using `NODE_ENV` or another conditional.

A good solution would be to add an optional parameter of `silent` to the logger on invocation:
```js
const logger = new Logger({
  silent: process.env.NODE_ENV === 'test',
  serviceName: 'AWS-SERVICE',
  persistentLogAttributes: {
    awsRegion: process.env.AWS_REGION || 'N/A',
  },
})
```

This would allow for `console.log` to still make it's way through to the terminal, while reducing the noise from Logger.

### Alternative solutions

_No response_

### Acknowledgment

- [X] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [Java](https://github.com/awslabs/aws-lambda-powertools-java/)

shdq:

This PR introduces a new `logLevel` value `SILENT`. When this level is set, either via constructor or via environment variable, the utility does not emit any log regardless of where it's running.

**Constructor example:**

```typescript
const logger = new Logger({ logLevel: 'SILENT' });
```

**Environment variable example:**

```typescript
process.env.LOG_LEVEL = 'SILENT';
const logger = new Logger();
```

This behavior is desirable for those customers who would like to have the ability to instrument their code with logs, but optionally have the chance to suppress them based on their requirements (see https://github.com/awslabs/aws-lambda-powertools-typescript/issues/1198#issuecomment-1434833994 for example).

Resolves #1198.

### How to verify this change

N/A

### Related issues, RFCs

**Issue number:** #1198 

## Checklist

### Breaking change checklist

---


shdq:
It's also worth mentioning in a `tip` box near the `logLevel` in docs or in [testing your code](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/logger/#testing-your-code) section with a description of particular use case.

dreamorosi:
> It's also worth mentioning in a `tip` box near the `logLevel` in docs or in [testing your code](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/logger/#testing-your-code) section with a description of particular use case.

I agree that we should update the docs but I'm still unsure about where it should go. 

At the moment I'm leaning towards this not being a test-only use case, especially if we keep in mind [the use case](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/1198#issuecomment-1434833994) that influenced the feature being green lit.

If that's okay for you, let's start with the implementation and unit tests, and maybe also adding the new value to [this table](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/logger/#utility-settings). This will give me some more time to think about how to frame this new feature.

packages/logger/tests/unit/Logger.test.ts
```diff
@@ -169,7 +203,7 @@ describe('Class: Logger', () => {
 
           // Prepare
           const logger: Logger = createLogger({
-            logLevel: 'ERROR',
+            logLevel: 'SILENT',
```
shdq, line 206: I changed to the highest log level when it doesn't print to stdout from any method and still prints when the sample rate is 1.

packages/logger/tests/unit/Logger.test.ts
```diff
@@ -159,6 +158,41 @@ describe('Class: Logger', () => {
...
+          const consoleSpy = jest.spyOn(logger['console'], methodOfLogger).mockImplementation();
+
+          // Act
+          logger[methodOfLogger]('foo');
+
+          // Assess
+          expect(consoleSpy).toBeCalledTimes(0);
+        });
+
+        test('when the Logger\'s log level is set through LOG_LEVEL env variable, it DOES print to stdout', () => {
```
shdq, line 176: Added a test that applies to all log methods, when the log level is set through the env var.

shdq:
> This will give me some more time to think about how to frame this new feature.

Take your time and then let me know!

packages/logger/tests/unit/Logger.test.ts
dreamorosi: Could you please elaborate on this one? Not sure I'm following

docs/core/logger.md
```diff
@@ -49,9 +49,9 @@ These settings will be used across all logs emitted:
...
-| **Logging level**       | Sets how verbose Logger should be                                                                                | `LOG_LEVEL`                     | `info`              |`DEBUG`, `INFO`, `WARN`, `ERROR`| `ERROR`            | `logLevel`            |
-| **Log incoming event**  | Whether to log or not the incoming event when using the decorator or middleware.                                 | `POWERTOOLS_LOGGER_LOG_EVENT`   | `false`             | `true`, `false`                | `false`            | `logEvent`            |
-| **Debug log sampling**  | Probability that a Lambda invocation will print all the log items regardless of the log level setting.           | `POWERTOOLS_LOGGER_SAMPLE_RATE` | `0`                 | `0.0` to `1`                   | `0.5`              | `sampleRateValue`     |
+| **Logging level**       | Sets how verbose Logger should be, from the most verbose to the least verbose (no logs)                                                                                | `LOG_LEVEL`                     | `info`              |`DEBUG`, `INFO`, `WARN`, `ERROR`, `SILENT` | `ERROR`            | `logLevel`            |
```
dreamorosi, line 52: nitpick: could you please apply formatting to the table so that the `|` are aligned?

packages/logger/tests/unit/Logger.test.ts
shdq: This test case consists of two tests: testing with and w/o sample rate.

For the first one, it is needed to show that no logs emit with the higher log level. The `error` log level was used to show that it doesn't print to stdout from all methods and is ignored on the `error` method:
```
expect(consoleSpy).toBeCalledTimes(method === 'error' ? 1 : 0)
```
Now we have `silence` as the highest level, so I changed to it.

The second one test that all logs emit with sample rate for every method even with the highest log level.

I pointed out this, because this is a dynamic set of tests, and it wasn't obvious to me what is "higher" level means in the description, and why it uses a particular log level. So I thought it worth mentioning in the review.

docs/core/logger.md
shdq: Sure. I thought there no order in these tables, because they are always looked like a mess in my vscode. I made the vscode window 6-x wider, and I see how it works 🙃

dreamorosi: Thank you for addressing my comment!

Should probably add some proper linting to markdown files at some point, you're right.

In the meantime I use this VSCode extension [that formats on save](https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one).

dreamorosi:
Regarding the docs, I think we should add a new section under "Advanced", and right after "Sampling logs", titled "Silencing logs":
> The `SILENT` log level is a quick and effective way to suppress all log messages without having to modify your code. When this log level is set, all logs, regardless of severity, are suppressed. This is useful in those cases when you want to have your code instrumented to produce logs, but due to some requirement or business decision, you prefer to not emit them. By setting the log level to `SILENT`, either via the `logLevel` constructor option, or via the `LOG_LEVEL` environment variable, you can easily suppress all logs on demand.

Then also add a "Note" box/callout that says:
> Note that using the `SILENT` log level can make it harder to debug and monitor your application, so we recommend to use this log level judiciously.

(Housekeeping) Also, since you're editing that file already, could you please fix these two:
- the typo at [this line](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/docs/core/logger.md?plain=1#L84) - it should be `function's invocation` instead of `function"s invocation` (single quote).
- the line highlights of [this code snippet](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/logger/#using-multiple-logger-instances-across-your-code)

shdq:
You found a good place for this feature in the docs. Good job, Andrea! I paraphrased the text a bit. I hope it's suitable and you like it.

dreamorosi (approved):
Looking great, thanks for the help on this one.

Let's 🚢 it!

PR #1365, opened 2023-03-10
feat(parameters): AppConfigProvider to return the last valid value when the API returns empty value on subsequent calls

### LINKED ISSUE #1363: Feature request: `AppConfigProvider` to return the last valid value when the API returns empty value on subsequent calls
### Use case

When retrieving a configuration profile from AppConfig, customers need to be mindful of the poll interval and to the fact that AppConfig returns empty responses for every subsequent call made after a successful retrieval **and** before the remote config has changed.

To provide a better DX, and to align with the implementation made in Powertools for Python, we should store the last valid configuration and return it instead of returning a response if users call the API again and the main cache was expired.

### Solution/User Experience

Whenever a valid configuration profile is retrieve, store the value in the provider. Then, if the API is polled before the configuration has been refreshed (which would return an empty response), return the stored value instead of the empty response.

### Alternative solutions

_No response_

### Acknowledgment

- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [Java](https://github.com/awslabs/aws-lambda-powertools-java/), and [.NET](https://github.com/awslabs/aws-lambda-powertools-dotnet/)

### Future readers

brnkrygs:

Fixes #1363 

When AppConfig parameters were requested beyond the cache expiration, they were returning empty. This was due to AppConfig `GetLatestConfiguration` returning an empty response if the active session already has the latest value.

This change introduces a local cache inside the `AppConfigProvider` that returns the most recent non-empty value in this scenario, matching the behavior of the [Python PowerTools Repository](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/utilities/parameters/appconfig.py#L121).

### How to verify this change

I added unit and e2e tests that cover this scenario. The simplest way is to develop a Lambda that uses the Parameters library with the AppConfig Provider, then invoke it a series of times, waiting beyond the `maxAge` of the configured provider. Confirm that you get the same return value in cold start, within `maxAge`, and beyond `maxAge` scenarios.

### Related issues, RFCs

**Issue number:** 
#1363

## Checklist

### Breaking change checklist

---

### DIFF
--- a/packages/parameters/src/ExpirableValue.ts
+++ b/packages/parameters/src/ExpirableValue.ts
@@ -4,10 +4,17 @@ class ExpirableValue implements ExpirableValueInterface {
   public ttl: number;
   public value: string | Uint8Array | Record<string, unknown>;
 
+  /**
+   * Creates a new cached value which will expire automatically
+   * @param value Parameter value to be cached
+   * @param maxAge Maximum number of seconds to cache the value for
+   */
   public constructor(value: string | Uint8Array | Record<string, unknown>, maxAge: number) {
     this.value = value;
-    const timeNow = new Date();
-    this.ttl = timeNow.setSeconds(timeNow.getSeconds() + maxAge);
+
+    const maxAgeInMilliseconds = maxAge * 1000;
+    const nowTimestamp = Date.now();
+    this.ttl = nowTimestamp + maxAgeInMilliseconds;
   }
 
   public isExpired(): boolean {
--- a/packages/parameters/src/ExpirableValue.ts
+++ b/packages/parameters/src/ExpirableValue.ts
@@ -4,17 +4,10 @@ class ExpirableValue implements ExpirableValueInterface {
   public ttl: number;
   public value: string | Uint8Array | Record<string, unknown>;
 
-  /**
-   * Creates a new cached value which will expire automatically
-   * @param value Parameter value to be cached
-   * @param maxAge Maximum number of seconds to cache the value for
-   */
   public constructor(value: string | Uint8Array | Record<string, unknown>, maxAge: number) {
     this.value = value;
-
-    const maxAgeInMilliseconds = maxAge * 1000;
-    const nowTimestamp = Date.now();
-    this.ttl = nowTimestamp + maxAgeInMilliseconds;
+    const timeNow = new Date();
+    this.ttl = timeNow.setSeconds(timeNow.getSeconds() + maxAge);
   }
 
   public isExpired(): boolean {
--- a/packages/parameters/src/appconfig/AppConfigProvider.ts
+++ b/packages/parameters/src/appconfig/AppConfigProvider.ts
@@ -13,6 +13,7 @@ import type {
 class AppConfigProvider extends BaseProvider {
   public client: AppConfigDataClient;
   protected configurationTokenStore: Map<string, string> = new Map();
+  protected valueStore: Map<string, Uint8Array> = new Map();
   private application?: string;
   private environment: string;
   
@@ -79,6 +80,10 @@ class AppConfigProvider extends BaseProvider {
      * The new AppConfig APIs require two API calls to return the configuration
      * First we start the session and after that we retrieve the configuration
      * We need to store { name: token } pairs to use in the next execution
+     * We also need to store { name : value } pairs because AppConfig returns
+     * an empty value if the session already has the latest configuration
+     * but, we don't want to return an empty value to our callers.
+     * {@link https://docs.aws.amazon.com/appconfig/latest/userguide/appconfig-retrieving-the-configuration.html}
      **/
     
     if (!this.configurationTokenStore.has(name)) {
@@ -106,14 +111,25 @@ class AppConfigProvider extends BaseProvider {
     });
 
     const response = await this.client.send(getConfigurationCommand);
-    
+
     if (response.NextPollConfigurationToken) {
       this.configurationTokenStore.set(name, response.NextPollConfigurationToken);
     } else {
       this.configurationTokenStore.delete(name);
     }
 
-    return response.Configuration;
+    /** When the response is not empty, stash the result locally before returning
+     * See AppConfig docs: 
+     * {@link https://docs.aws.amazon.com/appconfig/latest/userguide/appconfig-retrieving-the-configuration.html}
+     **/
+    if (response.Configuration !== undefined && response.Configuration?.length > 0 ) {
+      this.valueStore.set(name, response.Configuration);
+
+      return response.Configuration;
+    }
+
+    // Otherwise, use a stashed value
+    return this.valueStore.get(name);
   }
 
   /**
--- a/packages/parameters/tests/e2e/appConfigProvider.class.test.functionCode.ts
+++ b/packages/parameters/tests/e2e/appConfigProvider.class.test.functionCode.ts
@@ -81,8 +81,16 @@ export const handler = async (_event: unknown, _context: Context): Promise<void>
   try {
     providerWithMiddleware.clearCache();
     middleware.counter = 0;
-    await providerWithMiddleware.get(freeFormBase64encodedPlainText);
-    await providerWithMiddleware.get(freeFormBase64encodedPlainText);
+    const result1 = await providerWithMiddleware.get(freeFormBase64encodedPlainText);
+    const result2 = await providerWithMiddleware.get(freeFormBase64encodedPlainText);
+    logger.log({
+      test: 'get-cached-result1',
+      value: result1
+    });
+    logger.log({
+      test: 'get-cached-result2',
+      value: result2
+    });
     logger.log({
       test: 'get-cached',
       value: middleware.counter // should be 1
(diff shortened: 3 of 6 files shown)


packages/parameters/src/appconfig/AppConfigProvider.ts
```diff
@@ -13,6 +13,7 @@ import type {
 class AppConfigProvider extends BaseProvider {
   public client: AppConfigDataClient;
   protected configurationTokenStore: Map<string, string> = new Map();
+  protected valueStore: Map<string, Uint8Array> = new Map();
```
dreamorosi, line 16: I don't think this change is necessary. 

The `AppConfigProvider` extends the `BaseProvider` which is the one that holds the caching logic.

On a high level, the call stack goes like this:
- User runs `provider.get('...');`
- `AppConfigProvider.get()` runs first, and calls `BaseProvider.get()` passing down the same args ([here](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/parameters/src/appconfig/AppConfigProvider.ts#L53))
- Within `BaseProvider.get()`
  - The first action is to check if the value is in the cache & it's not expired, if both are `true`, the value is returned from cache ([here](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/parameters/src/BaseProvider.ts#L47-L51))
  - If there's no cached value, then `AppConfigProvider._get` (this method) is called, and we hit the API ([here](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/parameters/src/BaseProvider.ts#L55))
  - After that, we apply transformations on the returned value (if any), and we add it to cache

With this in mind, I don't think we should add a second cache here. We can consider improving the documentation around this, but if users expect to request a config very often they should handle this with the original `maxAge` setting. This will ensure that the value is returned from cache and won't be empty. Users can also influence the refresh rate from AppConfig by passing a parameter to the API call via `sdkOptions` to refresh more often.
But I don't think we should cache indefinitely and again here.

Also note that this is the same behavior/implementation found in the Python version of this library.

packages/parameters/src/ExpirableValue.ts
```diff
@@ -4,10 +4,17 @@ class ExpirableValue implements ExpirableValueInterface {
...
 
+  /**
+   * Creates a new cached value which will expire automatically
+   * @param value Parameter value to be cached
+   * @param maxAge Maximum number of seconds to cache the value for
+   */
   public constructor(value: string | Uint8Array | Record<string, unknown>, maxAge: number) {
     this.value = value;
-    const timeNow = new Date();
-    this.ttl = timeNow.setSeconds(timeNow.getSeconds() + maxAge);
```
dreamorosi, line 10: Why not simply do:
```ts
this.ttl = timeNow.setSeconds(timeNow.getSeconds() + maxAge * 1000);
```

and minimize the diff?

dreamorosi (changes requested):
Hi Brian, thank you so much for opening the PR and contributing a fix.

I have left a few comments here and there but essentially I don't think we should add an additional layer of caching and the fix could be as simple as converting the `maxAge` input to seconds (`number * 1000`) in the [current logic](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setSeconds#return_value).

brnkrygs:
Hi @dreamorosi, I appreciate the review and feedback.

I can definitely pivot this back to just looking at `maxAge` ttl calculation if you prefer.

I tried the simple fix and still saw unexpected behavior where until `maxAge` expired, I received values from the `AppConfig` Provider, and after it expired I received empty values.

But, maybe this isn't actually unexpected?

You're right that removing the extra cache exposes `AppConfig`-specific behavior to clients of the library. Maybe that _is_ a good thing. 

But as I understood it, part of the purpose of the library is to simplify this exposure, or at least generalize interaction across Parameter providers. If I swap from `DynamoDB` to `AppConfig` for example, I'll have to change client code as well and manage my own secondary cache, because they will act very differently when `maxAge` expires, depending on my `AppConfig` session lifecycle (which I have no visibility to).

Python implements this secondary cache pattern here, which inspired this approach:
https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/utilities/parameters/appconfig.py#L121

packages/parameters/src/ExpirableValue.ts
brnkrygs: I was exploring different algorithms as I tested and this is the last one I landed on before realizing the larger issue. It was a little easier for me to wrap my head around the intent.

[From the docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setSeconds), `Date.setSeconds()` takes a seconds value, not a milliseconds one, so the original implementation might be right.

In my node REPL:

```javascript
> const timeNow = new Date();
undefined
> const maxAge = 20;
undefined
> timeNow.getTime()
1678466472800
> timeNow.setSeconds(timeNow.getSeconds() + maxAge);
1678466492800
> 1678466492800-1678466472800 // New value is 20000ms ahead of old value... 20 seconds
20000
```

packages/parameters/src/appconfig/AppConfigProvider.ts
brnkrygs: I replied to this thread down in a separate comment. I suppose it comes down to how much provider implementation detail we decide to expose to clients.

With that said, I'm _fairly_ sure the Python library is using an intermediate cache (unless I'm misreading!?):
https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/utilities/parameters/appconfig.py#L121

brnkrygs: Either way, if you'd prefer that the `AppConfig` behavior be exposed, this PR can probably be closed. I'll re-test the original `maxAge` algorithm and refocus only there.

dreamorosi:
Thank you for pushing back on this, after re reading your comment I went back to check the implementation in Python and you're indeed right. The discussion that led to implementing that second cache layer comes from these two messages ([here](https://github.com/awslabs/aws-lambda-powertools-python/issues/1506#issuecomment-1256451601) and [here](https://github.com/awslabs/aws-lambda-powertools-python/issues/1506#issuecomment-1260791075)).

I now agree with you that we should do the same. However, if you don't mind I'd prefer to keep the PR scoped to the issue at hand, also and in addition to the fact that we are still unsure whether the initial caching is fixed or not.

----

Once we know that `maxAge` and the cache behavior in the BaseProvider is fixed, I'm more than open to address this other part. I'll open an issue to track this, and we can add this second caching in a second PR (contribution from you is welcome).


Would this work?

dreamorosi:
As discussed on a quick call, we are going to:
- Repurpose the original issue & PR to address this second layer of cache
- We can leave the `ttl` calculation as-is given it's correct
- We can add unit tests around `ExpirableValue` in the same `AppConfigProvider` test file
- We refactor the integration test around this new cache layer to have a single log emitted and to use the following (which should work the same):
```ts
const param1 = provider.get('hello', { maxAge: 0 });
const param2 = provider.get('hello');

// test that param1 === param2, where param2 is from this second cache
```
- Document the new test case in the comments

packages/parameters/src/appconfig/AppConfigProvider.ts
dreamorosi: No, you were right. Let's continue with this change, see this comment for full comment.

Marking this as resolved.

packages/parameters/tests/e2e/appConfigProvider.class.test.functionCode.ts
```diff
@@ -111,4 +119,34 @@ export const handler = async (_event: unknown, _context: Context): Promise<void>
...
+
+  // Test 8
+  // get parameter twice, but wait long enough that cache expires count SDK calls and return values
+  try {
+    providerWithMiddleware.clearCache();
+    middleware.counter = 0;
+    const expiredResult1 = await providerWithMiddleware.get(freeFormBase64encodedPlainText, { maxAge: 5 });
+
+    // Wait
+    await new Promise(resolve => setTimeout(resolve, 6000));
```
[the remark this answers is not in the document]
brnkrygs: Per discussion, will move to `maxAge: 0` to let us avoid the timeout and still express intent.

shdq:
Hey guys!

I see you've covered the second empty response from `AppConfig`. It was one of the two open questions for the current implementation. Another one in my notes is whether we should propagate `forceFetch` and start a new configuration session. Users would expect to get the latest configuration but will get value from the second cache. Should the second cache be pruned in this case?

dreamorosi (commented):
Thank you so much for addressing all the comments.

I think we are very close to getting this in shape to be merged. I have also run the [integration tests on your branch and are passing](https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/4403321820).

I've left a couple of very minor comments and expect this to be the last round of reviews.

packages/parameters/tests/unit/AppConfigProvider.test.ts
```diff
@@ -225,6 +226,51 @@ describe('Class: AppConfigProvider', () => {
       // Act & Assess
       await expect(provider.get(name)).rejects.toThrow();
     });
+
+    test('when session returns an empty configuration on the second call, it returns the last value', async () => {
+
+      client.reset();
```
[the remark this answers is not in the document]
brnkrygs: Sure, that works for me. I added it here to be surgical - the `resolvesOnce` chain was acting weirdly without it. I like the consistency of using an `afterEach` instead.

dreamorosi:
> I see you've covered the second empty response from `AppConfig`. It was one of the two open questions for the current implementation. 

Indeed, this was an open topic/question and I should take ownership of the fact that I misunderstood both the problem and the question originally. Brian was conducting some internal testing before the imminent beta launch and noticed this behavior. So after he compared the implementation with the one in Python he proposed this PR.

> Another one in my notes is whether we should propagate `forceFetch` and start a new configuration session. Users would expect to get the latest configuration but will get value from the second cache. Should the second cache be pruned in this case?

I think it's an interesting proposal. At the moment I'd like us to focus on launching the utility with feature parity and equivalent behavior with its Python corresponding one. After getting enough usage and feedback we can start iterating and proposing new features.

dreamorosi (approved):
Great work Brian, appreciate the time you've put into this.

Let's 🚢 it!

brnkrygs:
> 

Sounds great to me! Happy to help, and appreciate the feedback and patience!

PR #1370, opened 2023-03-13
refactor(tracer): log warning instead of throwing when segment is not found

### LINKED ISSUE #1356: Feature request: Tracer should log warning instead of throwing when segment is not found
### Expected Behaviour

Today, for an unknown reason in one of our Lambdas in Production, the `Tracer` object was unable to retrieve the current `Segment` from X-Ray.

Because the `Segment` was attempted to be retrieved in the middy middleware at the beginning of every request ([here](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/v1.6.0/packages/tracer/src/middleware/middy.ts#L46)) it caused every Lambda request to fail until we forced a cold-start of the Lambda, which resolved our problem.

It caused the request to fail because the `Tracer` [throws an error](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/tracer/src/Tracer.ts#L534-L537) if the `Segment` returned is `undefined`.

Unfortunately our Lambda was kept warm for over an hour before we tracked it down and forced a cold-start. In the meantime every request failed (see execution logs below).

My expected behaviour would be that if a Segment could not be retrieved, that it doesn't cause the whole Lambda invocation to fail, which it currently does when using the `middy` middleware.

Tracing should be done on a best-effort basis, if there is any problem tracing a request it should not cause the main request itself to fail. This particular scenario of getting the `Segment` is one specific case that could be improved to avoid failing the request, but it's quite possible there may be other code-paths that are vulnerable to causing the main request to fail, particularly if instrumented by the [captureLambdaHandler](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/v1.6.0/packages/tracer/src/middleware/middy.ts) middy middleware.

### Current Behaviour

As mentioned above, the `Tracer` currently [throws an error](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/v1.6.0/packages/tracer/src/Tracer.ts#L534-L537) if the `Segment` returned is `undefined`, this causes the middy middleware to fail every request, since `getSegment()` is called in the `captureLambdaHandlerBefore` function ([here](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/v1.6.0/packages/tracer/src/middleware/middy.ts#L53)).

### Code snippet

You could wire in a dummy Tracing provider to return `undefined` from its `getSegment()` method, called [here](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/v1.6.0/packages/tracer/src/Tracer.ts#L534).

### Steps to Reproduce

We don't know yet what caused `aws-xray-sdk-core`'s `getSegment()` ([here](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/v1.6.0/packages/tracer/src/provider/ProviderService.ts#L42)) to return `undefined`, but if you can simulate that, and use the `captureLambdaHandler` middy middleware, you will find that it causes the wrapped request to fail.

For us it caused the request to fail until the function was cold-started since `getSegment()` never started returning a `Segment` object.

### Possible Solution

My suggestion would be to consider updating the [Tracer.getSegment()](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/v1.6.0/packages/tracer/src/Tracer.ts#L530-L540) method to log a warning and return a dummy `SubSegment` if a real segment couldn't be retrieved, similar to how it already does if tracing is disabled.

Something like as follows:

```ts
  public getSegment(): Segment | Subsegment {
    if (!this.isTracingEnabled()) {
      return new Subsegment('## Dummy segment');
    }

    const segment = this.provider.getSegment();

    if (segment == null) { // changed from === undefined to be slightly more robust in case null is returned somehow
      // throw new Error('Failed to get the current sub/segment from the context.');
      console.warn('Failed to get the current sub/segment from the context, so failing over to manually create one to allow the request to proceed.');
      return new Subsegment('## Fallback segment (as unable to retrieve existing Segment)');
    }
 
    return segment;
  }
```

### AWS Lambda Powertools for TypeScript version

1.6.0

18.x

### Packaging format used

Lambda Layers

### Execution logs

```
2023-03-07T17:08:23.043Z	xxxxxxxxx-xxxx-xxxx-xxxxxxxxxxxxx	ERROR	Error: Failed to get the current sub/segment from the context.
    at Object.contextMissingLogError [as contextMissing] (/var/task/index.js:15505:23)
    at getSegment (/var/task/index.js:15569:49)
    at ProviderService.getSegment (/var/task/index.js:20115:51)
    at Tracer2.getSegment (/var/task/index.js:20539:39)
    at open (/var/task/index.js:20841:32)
    at captureLambdaHandlerBefore (/var/task/index.js:20852:11)
    at runMiddlewares (/var/task/index.js:24918:23)
    at async runRequest (/var/task/index.js:24876:5)

2023-03-07T17:08:23.043Z	xxxxxxxxx-xxxx-xxxx-xxxxxxxxxxxxx	ERROR	Error: Failed to get the current sub/segment from the context.
    at Object.contextMissingLogError [as contextMissing] (/var/task/index.js:15505:23)
    at getSegment (/var/task/index.js:15569:49)
    at ProviderService.getSegment (/var/task/index.js:20115:51)
    at Tracer2.getSegment (/var/task/index.js:20539:39)
    at Tracer2.addErrorAsMetadata (/var/task/index.js:20209:33)
    at captureLambdaHandlerError (/var/task/index.js:20867:18)
    at runMiddlewares (/var/task/index.js:24918:23)
    at runRequest (/var/task/index.js:24902:13)

2023-03-07T17:08:23.044Z	xxxxxxxxx-xxxx-xxxx-xxxxxxxxxxxxx	ERROR	Invoke Error 	{
    "errorType": "Error",
    "errorMessage": "Failed to get the current sub/segment from the context.",
    "originalError": {
        "errorType": "Error",
        "errorMessage": "Failed to get the current sub/segment from the context.",
        "stack": [
            "Error: Failed to get the current sub/segment from the context.",
            "    at Tracer2.getSegment (/var/task/index.js:20541:17)",
            "    at open (/var/task/index.js:20841:32)",
            "    at captureLambdaHandlerBefore (/var/task/index.js:20852:11)",
            "    at runMiddlewares (/var/task/index.js:24918:23)",
            "    at async runRequest (/var/task/index.js:24876:5)"
        ]
    },
    "stack": [
        "Error: Failed to get the current sub/segment from the context.",
        "    at Tracer2.getSegment (/var/task/index.js:20541:17)",
        "    at Tracer2.addErrorAsMetadata (/var/task/index.js:20209:33)",
        "    at captureLambdaHandlerError (/var/task/index.js:20867:18)",
        "    at runMiddlewares (/var/task/index.js:24918:23)",
        "    at runRequest (/var/task/index.js:24902:13)"
    ]
}
```

dreamorosi:

As reported in #1356, the current behavior of the Tracer and corresponding middleware is fairly unforgiving when it comes to handling missing segments. Prior to this PR, when calling the `getSegment` method, if a segment is not found, an error would be thrown.

This can happen either when users try to retrieve the current segment outside of the context of a function invocation or when the trace data is malformed and the X-Ray SDK is not able to derive the current segment.

This behavior, if not handled properly, can cause a function invocation to fail, which might not be desirable given that for many customers tracing is considered best-effort.

This PR:
- aligns the Tracer utility with the X-Ray SDK so that if a segment is not found, a warning is logged instead of throwing
- improves the handling of missing segments in the `captureLambdaHandler` middleware in all stages to be more resilient
- bumps the `aws-xray-sdk-node` dependency to the latest version
- refactor and extracts some of the internal logic around current segment retrieval from the `Tracer` class to the `ProviderService` one

This PR closes #1356.

### How to verify this change

Check the existing tests as well as the newly added ones in the checks under this repo. Check also the results of [this successful integration test](https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/4407673701).

### Related issues, RFCs

**Issue number:** #1356

## Checklist

### Breaking change checklist

---

### DIFF
--- a/packages/tracer/src/Tracer.ts
+++ b/packages/tracer/src/Tracer.ts
@@ -512,7 +512,7 @@ class Tracer extends Utility implements TracerInterface {
   }
   
   /**
-   * Get the active segment or subsegment in the current scope.
+   * Get the active segment or subsegment (if any) in the current scope.
    * 
    * Usually you won't need to call this method unless you are creating custom subsegments or using manual mode.
    *
@@ -531,7 +531,7 @@ class Tracer extends Utility implements TracerInterface {
    * }
    * ```
    * 
-   * @returns segment - The active segment or subsegment in the current scope.
+   * @returns segment - The active segment or subsegment in the current scope. Will log a warning and return `undefined` if no segment is found.
    */
   public getSegment(): Segment | Subsegment | undefined {
     if (!this.isTracingEnabled()) {
--- a/packages/tracer/src/Tracer.ts
+++ b/packages/tracer/src/Tracer.ts
@@ -153,6 +153,7 @@ class Tracer extends Utility implements TracerInterface {
     * @see https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-errors
     *
     * @param error - Error to serialize as metadata
+    * @param [remote] - Whether the error was thrown by a remote service. Defaults to `false`
     */
   public addErrorAsMetadata(error: Error, remote?: boolean): void {
     if (!this.isTracingEnabled()) {
--- a/packages/tracer/src/Tracer.ts
+++ b/packages/tracer/src/Tracer.ts
@@ -531,7 +531,7 @@ class Tracer extends Utility implements TracerInterface {
    * }
    * ```
    * 
-   * @returns segment - The active segment or subsegment in the current scope. Will log a warning and return `undefined` if no segment is found.
+   * @returns The active segment or subsegment in the current scope. Will log a warning and return `undefined` if no segment is found.
    */
   public getSegment(): Segment | Subsegment | undefined {
     if (!this.isTracingEnabled()) {
--- a/packages/tracer/src/Tracer.ts
+++ b/packages/tracer/src/Tracer.ts
@@ -38,7 +38,7 @@ import { Segment, Subsegment } from 'aws-xray-sdk-core';
  * 
  * const tracer = new Tracer({ serviceName: 'serverlessAirline' });
  * 
-* const lambdaHandler = async (_event: any, _context: any) => {
+ * const lambdaHandler = async (_event: any, _context: any) => {
  *   ...
  * };
  * 
@@ -154,19 +154,24 @@ class Tracer extends Utility implements TracerInterface {
     *
     * @param error - Error to serialize as metadata
     */
-  public addErrorAsMetadata(error: Error): void {
+  public addErrorAsMetadata(error: Error, remote?: boolean): void {
     if (!this.isTracingEnabled()) {
       return;
     }
 
     const subsegment = this.getSegment();
+    if (subsegment === undefined) {
+
+      return;
+    }
+
     if (!this.captureError) {
       subsegment.addErrorFlag();
 
       return;
     }
 
-    subsegment.addError(error, false);
+    subsegment.addError(error, remote || false);
   }
 
   /**
@@ -527,13 +532,15 @@ class Tracer extends Utility implements TracerInterface {
    * 
    * @returns segment - The active segment or subsegment in the current scope.
    */
-  public getSegment(): Segment | Subsegment {
+  public getSegment(): Segment | Subsegment | undefined {
     if (!this.isTracingEnabled()) {
       return new Subsegment('## Dummy segment');
     }
     const segment = this.provider.getSegment();    
     if (segment === undefined) {
-      throw new Error('Failed to get the current sub/segment from the context.');
+      console.warn(
+        'Failed to get the current sub/segment from the context, this is likely because you are not using the Tracer in a Lambda function.'
+      );
     }
  
     return segment;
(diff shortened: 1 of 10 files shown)


packages/tracer/src/Tracer.ts
```diff
@@ -154,19 +154,24 @@ class Tracer extends Utility implements TracerInterface {
     *
     * @param error - Error to serialize as metadata
     */
-  public addErrorAsMetadata(error: Error): void {
+  public addErrorAsMetadata(error: Error, remote?: boolean): void {
```
am29d, line 157: missing JSDocs, not clear why we need it.

packages/tracer/src/Tracer.ts
```diff
@@ -527,13 +532,15 @@ class Tracer extends Utility implements TracerInterface {
    * 
    * @returns segment - The active segment or subsegment in the current scope.
```
am29d, line 533: might need signal here that it can be undefined, so the caller can deal with the result.

am29d (commented):
I like the cleanup it tests 👌, much easier to follow now. 

The implications of additional `undefined` are bigger than I thought. It tears through many methods, similar to `isTracingEnabled` and reminds me of java or golang null checks. But I have no immediate answer to make it cleaner.

dreamorosi:
> I like the cleanup it tests 👌, much easier to follow now.

Thanks!

> The implications of additional `undefined` are bigger than I thought. It tears through many methods, similar to `isTracingEnabled` and reminds me of java or golang null checks. But I have no immediate answer to make it cleaner.

Indeed, my attempt at somewhat centralising this behavior was to extract it in the provider class. This is also something I've been meaning to do for a while as it'd help decouple the two and open the door for other providers in the future.

am29d (approved):

PR #1416, opened 2023-04-24
feat(idempotency): implement IdempotencyHandler

### LINKED ISSUE #1303: Feature request: implement `IdempotencyHandler`
### Use case

All the usages of the Idempotency utility described in #447 share most of their internal logic. This is because in all cases, there are some actions to be taken before and after the function being made idempotent.

For this reason the logic that handles idempotency can be extracted in its own class and reused across all access patterns. This class should be called `IdempotencyHandler`.

Note that there's already an initial basic implementation of the class under `packages/idempotency/src`. As a part of this issue, the resolver should revisit this function and align the implementation with the [version seen in Powertools for Python](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/utilities/idempotency/base.py#L43).

### Solution/User Experience

Users should not use this class directly. The class will be used under the hood by the decorator, Middy middleware, and wrapper function.

### Alternative solutions

_No response_

### Acknowledgment

- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [Java](https://github.com/awslabs/aws-lambda-powertools-java/)

am29d:
This is a small PR for the #1303 implementation. Most of the `IdempotencyHandler` was already implemented and aligned with decorator and function wrapper. In addition, I have added a try mechanism to mitigate rare race conditions error, similar to python implementation. 

### Related issues, RFCs

**Issue number:** closes #1303 

## Checklist

### Breaking change checklist

---

### DIFF
--- a/packages/idempotency/src/IdempotencyHandler.ts
+++ b/packages/idempotency/src/IdempotencyHandler.ts
@@ -1,41 +1,76 @@
+import type { AnyFunctionWithRecord, IdempotencyOptions } from './types';
 import { IdempotencyRecordStatus } from './types';
-import type {
-  AnyFunctionWithRecord,
-  IdempotencyOptions,
-} from './types';
 import {
+  IdempotencyAlreadyInProgressError,
   IdempotencyInconsistentStateError,
   IdempotencyItemAlreadyExistsError,
-  IdempotencyAlreadyInProgressError,
-  IdempotencyPersistenceLayerError
+  IdempotencyPersistenceLayerError,
 } from './Exceptions';
-import { IdempotencyRecord } from './persistence/IdempotencyRecord';
+import { IdempotencyRecord } from './persistence';
 
 export class IdempotencyHandler<U> {
   public constructor(
     private functionToMakeIdempotent: AnyFunctionWithRecord<U>,
-    private functionPayloadToBeHashed: Record<string, unknown>, 
+    private functionPayloadToBeHashed: Record<string, unknown>,
     private idempotencyOptions: IdempotencyOptions,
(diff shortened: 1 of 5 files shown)


packages/idempotency/src/IdempotencyHandler.ts
```diff
@@ -1,41 +1,84 @@
...
+  }
+
+  /**
+   * Main entry point for the handler
+   * IdempotencyInconsistentStateError can happen under rare but expected cases
+   * when persistent state changes in the small time between put & get requests.
+   * In most cases we can retry successfully on this exception.
+   */
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
```
dreamorosi, line 56: Before we merge, can we fix this return type and remove the comment?

Based on what I see, we either need to change the return type to `Promise<U | undefined>` or make sure that the function returns something in the `catch` block when the `if` statement is skipped.

dreamorosi (commented):
Looking good, I just would like to address one return type and remove a `ts-ignore` before merging

packages/idempotency/src/IdempotencyHandler.ts
am29d: Hmm this is  weird construct we have there. Adding `undefined` to the Promise would leak it further to `makeIdempotentFunction` and also higher to the user api, which I'd like to avoid. TS complains about missing return statement outside of the for-loop. But adding a statement makes it impossible to test, because this return statement is unreachable. I have added a throw with coverage ignore.

dreamorosi (approved):

Issue #1378, opened 2023-03-21
Feature request: Add isTraceSampled method to Tracer


icholy:
### Use case

I'd like a attribute in my logs to indicate whether or not a request has been sampled. Something like this:

``` json
{
  "level": "ERROR",
  "msg": "something went wrong",
  "trace_id": "1-6419d889-20172e50462fac2c7c481454",
  "trace_sampled": true
}
```

I can populate the `trace_id` using `Tracer#getRootXrayTraceId` , but there's no way to find out if the trace is sampled.

### Solution/User Experience

A new `isTraceSampled` method on the `Tracer` class.

``` typescript
log({
  level: "ERROR",
  msg: "something went wrong",
  trace_id: tracer.getRootXrayTraceId(),
  trace_sampled: tracer.isTraceSampled(),
})
```

### Alternative solutions

Implement my own `isTraceSampled` function which parses the `_X_AMZN_TRACE_ID` header value.

### Acknowledgment

- [X] This feature request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/#tenets)
- [x] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [Java](https://github.com/awslabs/aws-lambda-powertools-java/), and [.NET](https://github.com/awslabs/aws-lambda-powertools-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

icholy:
* Return type of the method, considering that the Trace data might be undefined:
  * `boolean | undefined`
* Where/how this new method should be implemented (I see some common logic with this existing method)
  * Add a `EnvironmentVariableService#getXrayTraceSampled` method.
  * Use that method inside the `Tracer#isTraceSampled`
* Agree on a name for the method that conveys its meaning:
  * `isSampled`
  * `isTraceSampled`
  * `getTraceSampled`
  * `getRootXrayTraceSampled`
* What type of changes should be made [to the documentation](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/tracer/#access-aws-x-ray-root-trace-id) to articulate the feature/method
  *  Not sure

dreamorosi:
> * Return type of the method, considering that the Trace data might be undefined:
>   
>   * `boolean | undefined`
>     
>     * I think we could get away with just `boolean`. If there's no trace, then it's not sampled.
[...]

PR #1435, opened 2023-05-04
feat(tracer): add isTraceSampled method

icholy:

This change adds a new `isTraceSampled` method to `Tracer`. It lets users find out if the current trace is sampled or not. This information may be included in structured logs along with the Trace ID.

### How to verify this change

Set the `Sampled=1` flag in the `_X_AMZN_TRACE_ID` environment variable and assert that `isTraceSampled` returns `true`.
Example:    `Root=1-5759e988-bd862e3fe1be46a994272793;Parent=557abcec3ee5a047;Sampled=1`

### Related issues, RFCs

**Issue number:** #1378

## Checklist

### Breaking change checklist

---

### DIFF
--- a/packages/commons/src/config/EnvironmentVariablesService.ts
+++ b/packages/commons/src/config/EnvironmentVariablesService.ts
@@ -61,6 +61,20 @@ class EnvironmentVariablesService extends ConfigService {
   }
 
   /**
+   * It returns true if the Sampled flag is set in the _X_AMZN_TRACE_ID environment variable.
+   * 
+   * The AWS X-Ray Trace data available in the environment variable has this format:
+   * `Root=1-5759e988-bd862e3fe1be46a994272793;Parent=557abcec3ee5a047;Sampled=1`,
+   *
+   * @returns {string}
+   */
+  public getXrayTraceSampled(): boolean {
+    const xRayTraceData = this.getXrayTraceData();
+
+    return xRayTraceData[2] === 'Sampled=1';
+  }
+
+  /**
    * It returns true if the string value represents a boolean true value.
    *
    * @param {string} value
--- a/packages/commons/src/config/EnvironmentVariablesService.ts
+++ b/packages/commons/src/config/EnvironmentVariablesService.ts
@@ -55,9 +55,7 @@ class EnvironmentVariablesService extends ConfigService {
   public getXrayTraceId(): string | undefined {
     const xRayTraceData = this.getXrayTraceData();
 
-    if (xRayTraceData.length === 0) return undefined;
-
-    return xRayTraceData[0].replace('Root=', '');
+    return xRayTraceData?.Root;
   }
 
   /**
@@ -71,7 +69,7 @@ class EnvironmentVariablesService extends ConfigService {
   public getXrayTraceSampled(): boolean {
     const xRayTraceData = this.getXrayTraceData();
 
-    return xRayTraceData[2] === 'Sampled=1';
+    return xRayTraceData?.Sampled === '1';
   }
 
   /**
@@ -86,12 +84,27 @@ class EnvironmentVariablesService extends ConfigService {
     return truthyValues.includes(value.toLowerCase());
   }
 
-  private getXrayTraceData(): string[] {
+  /**
+   * It parses the key/value data present in the _X_AMZN_TRACE_ID environment variable
+   * and returns it as an object when available.
+   */
+  private getXrayTraceData(): Record<string, string> | undefined {
     const xRayTraceEnv = this.get(this.xRayTraceIdVariable);
 
-    if (xRayTraceEnv === '') return [];
+    if (xRayTraceEnv === '') return undefined;
 
-    return xRayTraceEnv.split(';');
+    if (!xRayTraceEnv.includes('=')) return { Root: xRayTraceEnv };
+
+    const xRayTraceData: Record<string, string> = {};
+
+    xRayTraceEnv.split(';').forEach((field) => {
+
+      const [ key, value ] = field.split('=');
+
+      xRayTraceData[key] = value;
+    });
+
+    return xRayTraceData;
   }
 }
 
--- a/packages/commons/src/config/EnvironmentVariablesService.ts
+++ b/packages/commons/src/config/EnvironmentVariablesService.ts
@@ -66,7 +66,7 @@ class EnvironmentVariablesService extends ConfigService {
    * The AWS X-Ray Trace data available in the environment variable has this format:
    * `Root=1-5759e988-bd862e3fe1be46a994272793;Parent=557abcec3ee5a047;Sampled=1`,
    *
-   * @returns {string}
+   * @returns {boolean}
    */
   public getXrayTraceSampled(): boolean {
     const xRayTraceData = this.getXrayTraceData();
--- a/packages/commons/src/config/EnvironmentVariablesService.ts
+++ b/packages/commons/src/config/EnvironmentVariablesService.ts
@@ -53,11 +53,11 @@ class EnvironmentVariablesService extends ConfigService {
    * @returns {string}
    */
   public getXrayTraceId(): string | undefined {
-    const xRayTraceId = this.get(this.xRayTraceIdVariable);
+    const xRayTraceData = this.getXrayTraceData();
 
-    if (xRayTraceId === '') return undefined;
+    if (xRayTraceData.length === 0) return undefined;
 
-    return xRayTraceId.split(';')[0].replace('Root=', '');
+    return xRayTraceData[0].replace('Root=', '');
   }
 
   /**
@@ -72,6 +72,13 @@ class EnvironmentVariablesService extends ConfigService {
     return truthyValues.includes(value.toLowerCase());
   }
 
+  private getXrayTraceData(): string[] {
+    const xRayTraceEnv = this.get(this.xRayTraceIdVariable);
+
+    if (xRayTraceEnv === '') return [];
+
+    return xRayTraceEnv.split(';');
+  }
 }
 
 export {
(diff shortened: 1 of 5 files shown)


dreamorosi:
Hi @icholy thank you for taking the time to open this PR!

Please allow us some time to go through the changes and review the PR.

packages/tracer/src/Tracer.ts
```diff
@@ -547,6 +547,19 @@ class Tracer extends Utility implements TracerInterface {
...
+   * Get the current value of the AWS X-Ray Sampled flag.
+   * 
+   * Utility method that returns the current AWS X-Ray Sampled flag.
+   *
+   * @see https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-traces
+   * 
+   * @returns boolean - `true` if the trace is sampled, `false` otherwise.
+   */
+  public isTraceSampled(): boolean {
+    return this.envVarsService.getXrayTraceSampled();
```
dreamorosi, line 560:
```suggestion
    if (!this.isTracingEnabled()) return false;

    return this.envVarsService.getXrayTraceSampled();
```

packages/tracer/src/Tracer.ts
```diff
@@ -547,6 +547,19 @@ class Tracer extends Utility implements TracerInterface {
...
   }
 
+  /**
+   * Get the current value of the AWS X-Ray Sampled flag.
+   * 
+   * Utility method that returns the current AWS X-Ray Sampled flag.
+   *
+   * @see https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-traces
+   * 
+   * @returns boolean - `true` if the trace is sampled, `false` otherwise.
```
dreamorosi, line 557:
```suggestion
   * @returns boolean - `true` if the trace is sampled, `false` if tracing is disabled or the trace is not sampled.
```

packages/commons/src/config/EnvironmentVariablesService.ts
```diff
@@ -72,6 +86,13 @@ class EnvironmentVariablesService extends ConfigService {
     return truthyValues.includes(value.toLowerCase());
   }
 
+  private getXrayTraceData(): string[] {
+    const xRayTraceEnv = this.get(this.xRayTraceIdVariable);
+
+    if (xRayTraceEnv === '') return [];
```
dreamorosi, line 92:
```suggestion
    if (xRayTraceEnv === '' || !xRayTraceEnv.includes(';')) return;;
```

packages/commons/src/config/EnvironmentVariablesService.ts
```diff
@@ -72,6 +86,13 @@ class EnvironmentVariablesService extends ConfigService {
     return truthyValues.includes(value.toLowerCase());
   }
 
+  private getXrayTraceData(): string[] {
```
dreamorosi, line 89:
```suggestion
private getXrayTraceData(): { root: string, sampled: string } | undefined {
```

icholy: The `string[]` return type is the signature you prescribed in the original issue.

packages/commons/src/config/EnvironmentVariablesService.ts
```diff
@@ -72,6 +86,13 @@ class EnvironmentVariablesService extends ConfigService {
...
+    const xRayTraceEnv = this.get(this.xRayTraceIdVariable);
+
+    if (xRayTraceEnv === '') return [];
+
+    return xRayTraceEnv.split(';');
```
dreamorosi, line 94:
```suggestion
    const traceFields = xRayTraceEnv.split(';');
    const rootField = traceFields.find((field) => field.startsWith('Root=')) || '';
    const sampledField = traceFields.find((field) => field.startsWith('Sampled=')) || '';

    return {
      root: rootField.includes('=') ? rootField.split('=')[1] : undefined,
      sampled: sampledField.includes('=') ? rootField.split('=')[1] : undefined,
    };
```

packages/commons/src/config/EnvironmentVariablesService.ts
```diff
@@ -53,11 +53,25 @@ class EnvironmentVariablesService extends ConfigService {
...
+   * 
+   * The AWS X-Ray Trace data available in the environment variable has this format:
+   * `Root=1-5759e988-bd862e3fe1be46a994272793;Parent=557abcec3ee5a047;Sampled=1`,
+   *
+   * @returns {boolean}
+   */
+  public getXrayTraceSampled(): boolean {
+    const xRayTraceData = this.getXrayTraceData();
+
+    return xRayTraceData[2] === 'Sampled=1';
```
dreamorosi, line 74:
```suggestion
    return xRayTraceData.sampled === '1';
```

icholy:
@dreamorosi seems like you have a very specific implementation in mind. I think it would be easier for both if you just added your commits on top of mine.

packages/commons/src/config/EnvironmentVariablesService.ts
```diff
@@ -53,11 +53,25 @@ class EnvironmentVariablesService extends ConfigService {
...
    */
   public getXrayTraceId(): string | undefined {
-    const xRayTraceId = this.get(this.xRayTraceIdVariable);
+    const xRayTraceData = this.getXrayTraceData();
 
-    if (xRayTraceId === '') return undefined;
+    if (xRayTraceData.length === 0) return undefined;
 
-    return xRayTraceId.split(';')[0].replace('Root=', '');
+    return xRayTraceData[0].replace('Root=', '');
```
dreamorosi, line 60:
```suggestion
    return xRayTraceData.root;
```

packages/commons/src/config/EnvironmentVariablesService.ts
```diff
@@ -53,11 +53,25 @@ class EnvironmentVariablesService extends ConfigService {
    * @returns {string}
    */
   public getXrayTraceId(): string | undefined {
-    const xRayTraceId = this.get(this.xRayTraceIdVariable);
+    const xRayTraceData = this.getXrayTraceData();
 
-    if (xRayTraceId === '') return undefined;
+    if (xRayTraceData.length === 0) return undefined;
```
dreamorosi, line 58:
```suggestion
```

packages/commons/src/config/EnvironmentVariablesService.ts
dreamorosi: True, my bad.

After looking at the full implementation I think parsing and looking for the values would make for a more solid implementation rather than relying on the order of the fields. This way if fields get added or the order changes for some reason it won't immediately break.

What do you think?

icholy:
Your logic makes sense, but personally I think you're giving `getXrayTraceData` too much responsibility. It should only understand parsing the key/value syntax. IE: `getXrayTraceData(): Record<string, string> | undefined` .

So parsing `Root=1-5759e988-bd862e3fe1be46a994272793;Parent=557abcec3ee5a047;Sampled=1` would result in an object like this:

```
{
  Root: "1-5759e988-bd862e3fe1be46a994272793",
  Parent: "557abcec3ee5a047",
  Sampled: "1"
}
```

For backwards compatibility, it would place the whole string in the `Root` key if it doesn't contain key/value pairs. 

Then the `getXrayTraceId` and `getXrayTraceSampled` contain logic for interpreting the parsed data. So `getXrayTraceData` deals with syntax, and `getXrayTraceId` & `getXrayTraceSampled` deal with semantic.

dreamorosi:
You're right, and thanks for hearing me out.

Let's go with your proposal, after this I'm looking forward to merge it right away.

icholy:
@dreamorosi ready for another review round.

packages/commons/src/config/EnvironmentVariablesService.ts
```diff
@@ -72,6 +84,28 @@ class EnvironmentVariablesService extends ConfigService {
...
+  /**
+   * It parses the key/value data present in the _X_AMZN_TRACE_ID environment variable
+   * and returns it as an object when available.
+   */
+  private getXrayTraceData(): Record<string, string> | undefined {
+    const xRayTraceEnv = this.get(this.xRayTraceIdVariable);
+
+    if (xRayTraceEnv === '') return undefined;
+
+    if (!xRayTraceEnv.includes('=')) return { Root: xRayTraceEnv };
```
icholy, line 96: Note: I'm checking for `=` instead of `;` so that an expression like `Root=123` still parses correctly.

icholy:
I checked the CI logs, but I can't figure out why they're failing.

packages/tracer/src/Tracer.ts
```diff
@@ -547,6 +547,21 @@ class Tracer extends Utility implements TracerInterface {
...
+   * Get the current value of the AWS X-Ray Sampled flag.
+   * 
+   * Utility method that returns the current AWS X-Ray Sampled flag.
+   *
+   * @see https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-traces
+   * 
+   * @returns boolean - `true` if the trace is sampled, `false` if tracing is disabled or the trace is not sampled.
+   */
+  public isTraceSampled(): boolean {
+    if (!this.isTracingEnabled()) return false;
```
icholy, line 560: Should `getRootXrayTraceId` also have a check this like?

dreamorosi: Possibly yes, it could save some cycles and avoid having to parse a potentially nonexistent variable

dreamorosi:
I think there's one line/code branch not being covered by the tests

packages/tracer/tests/unit/Tracer.test.ts
```diff
@@ -279,6 +279,23 @@ describe('Class: Tracer', () => {
 
   });
 
+  describe('Method: isTraceSampled', () => {
+
+    test('when called, it returns true if the Sampled flag is set', () => {
+
```
icholy, line 285: I can't figure out where `_X_AMZN_TRACE_ID` is being set for this test.

dreamorosi (approved):
Thank you for contributing the feature and for addressing the comments during the review!

Issue #447, opened 2022-01-10
RFC: idempotency utility


walmsles:
## Description of the feature request

**Problem statement**
Idempotency is a core Cloud issue that needs a solution to enable stable, fault-tolerant systems that can be affected by repeated transactions.  A true idempotency solution as a utility for Powertools would be really useful for this project based on the function/features of the existing AWS Lambda Powertools for python

**Summary of the feature**
Link to Python Lambda Powertools documentation as a good example to follow: https://awslabs.github.io/aws-lambda-powertools-python/latest/utilities/idempotency

**Benefits for you and the wider AWS community**
Supply a good idempotent solution "Out of the Box" with Powertools for Typescript developers.

**Describe alternatives you've considered**
Looked at this: [https://www.npmjs.com/package/middy-idempotent](https://www.npmjs.com/package/middy-idempotent) but relies on redis for storage and the Idempotent Key generation is not ideal and looks like it will be problematic.

**Additional context**
https://awslabs.github.io/aws-lambda-powertools-python/latest/utilities/idempotency/

### Related issues, RFCs

None.

dreamorosi:
Hi @walmsles thank you for the request. Feature parity with the Python version is definitely something that we have on our radars. Middy-compliant middlewares are only one of the usages that we want to cover at this stage (with the others being Class method decorators & manual instrumentation) so the feature deserves a closer look.
Please give us some time to discuss this internally.

ijemmy:
Hi everyone, here is a design proposal. Would appreciate your comments on this (especially around the **Utility interface** as it's different from Python and Java which have no constraint on decorator usage.) 

## Design proposal (Request for comments)

### 1. Summary
The goal of this document is to propose the scope and design of Idempotency utility for Powertools for TypeScript. The utility has been implemented in the Python and Java version. We will use the current [Python implementation](https://awslabs.github.io/aws-lambda-powertools-python/latest/utilities/idempotency/) (https://awslabs.github.io/aws-lambda-powertools-python/latest/utilities/idempotency/) as a baseline, and describe only the differences we will make in TypeScript. Anything not discussed here will be the same as in Python version.

This RFC assumes that you are familiar with Python’s implementation. If you aren’t, please check [the documentation of Python Idempotency utility](https://awslabs.github.io/aws-lambda-powertools-python/latest/utilities/idempotency/) (https://awslabs.github.io/aws-lambda-powertools-python/latest/utilities/idempotency/) first.

[...]

PR #1442, opened 2023-05-11
test(idempotency): add e2e tests for idempotency

am29d:

This PR introduces e2e tests for idempotency using decorator and function wrapper. Several changes were also introduced to the interface and options type, here is the breakdown. 

### `IdempotencyOptions`

First, we have unwrapped the options object before passing it to the `IdempotencyHandler`, this helps to keep it generic and decide which parameters we should pass from decorator or wrapper. We also expanded the type to contain `dataKeywordArgument` and `config` as optional parameters. This enables us to apply decorator to lambda handler or any other method. In addition, we can use the same interface for decorator and wrapper. 

### Test cases

The test cases should help to test basic functionality. We have much more possible tests with permutations of input, config, persistence layer options and sdk options. Not all of them are covered yet and I suggest to use the beta phase to add more tests down the road. The current test cases are:

* decorator
  * call method twice sequentially, the response value is in ddb
  * call method twice parallel, the response value is in ddb and we got one error during second call
  * call method once with custom persistence layer options
  * call method that fails, there is no ddb entry
  * call method with keyword argument with two different key/value pairs, the response values are in ddb
* wrapper
  * call method twice sequentially, the response value is in ddb 
  * call method with custom persistence layer options

### Related issues, RFCs

**Issue number:** #447 

## Checklist

### Breaking change checklist

---


packages/idempotency/src/types/IdempotencyOptions.ts
```diff
@@ -1,9 +1,11 @@
 import type { Context } from 'aws-lambda';
 import { BasePersistenceLayer } from '../persistence/BasePersistenceLayer';
+import { IdempotencyConfig } from 'IdempotencyConfig';
 
-type IdempotencyOptions = {
-  dataKeywordArgument: string
+type IdempotentOptions = {
   persistenceStore: BasePersistenceLayer
+  dataKeywordArgument?: string
```
dreamorosi, line 7: Can we please add docstrings for each property like below? This way when someone hovers it it's clear what they are.

packages/idempotency/src/IdempotencyHandler.ts
```diff
@@ -1,25 +1,25 @@
...
 import { IdempotencyRecordStatus } from './types';
 import {
   IdempotencyAlreadyInProgressError,
   IdempotencyInconsistentStateError,
   IdempotencyItemAlreadyExistsError,
   IdempotencyPersistenceLayerError,
 } from './Exceptions';
-import { IdempotencyRecord } from './persistence';
+import { BasePersistenceLayer, IdempotencyRecord } from './persistence';
+import { IdempotencyConfig } from 'IdempotencyConfig';
```
dreamorosi, line 10:
```suggestion
import { IdempotencyConfig } from './IdempotencyConfig';
```

packages/idempotency/src/IdempotencyHandler.ts
```diff
@@ -1,25 +1,25 @@
...
 
 export class IdempotencyHandler<U> {
   public constructor(
     private functionToMakeIdempotent: AnyFunctionWithRecord<U>,
     private functionPayloadToBeHashed: Record<string, unknown>,
-    private idempotencyOptions: IdempotencyOptions,
+    private config: IdempotencyConfig,
```
dreamorosi, line 16: A few comments:
- `config` is now unused, is this right? should it be removed?
- Should we make all these arguments an object i.e. `config` like everywhere else? In JS positional arguments are hard to deal with
- Regardless of the above, I think we should change this class declaration to follow the same style as the rest of the project and have the property declared above the constructor, with the constructor actually assigning them as needed

packages/idempotency/tests/e2e/makeFunctionIdempotent.test.ts
```diff
@@ -0,0 +1,101 @@
...
+import { deployStack, destroyStack } from '../../../commons/tests/utils/cdk-cli';
+import { GetCommand, ScanCommand } from '@aws-sdk/lib-dynamodb';
+import { createIdempotencyResources } from '../helpers/idempotencyUtils';
+
+const runtime: string = process.env.RUNTIME || 'nodejs18x';
+
+if (!isValidRuntimeKey(runtime)) {
+  throw new Error(`Invalid runtime key value: ${runtime}`);
+}
+const stackName = generateUniqueName(RESOURCE_NAME_PREFIX, v4(), runtime, 'makeFnIdempotent');
```
dreamorosi, line 21: The `uuid` should be the same across the same test, so that we can identify resources related to the same test run by the prefix.

The way it is now each resource name will have its own prefix because we generate a new uuid for each name.

packages/idempotency/tests/e2e/makeFunctionIdempotent.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    });
+
+  }, TEST_CASE_TIMEOUT);
+
+  test('when called with customized function wrapper, it creates ddb entry with custom attributes', async () => {
+    const payload = { records: [ { id: 1, foo: 'bar' }, { id: 2, foo: 'baq' }, { id: 3, foo: 'bar' } ] };
+    const payloadHash = createHash('md5').update('"bar"').digest('base64');
+
+    const invocationLogsCustmozed = await invokeFunction(functionNameCustom, 2, 'SEQUENTIAL', payload, false);
+    await ddb.send(new GetCommand({
```
dreamorosi, line 84: You can't/shouldn't use `await` with `.then` afterwards.

Change this (& the other occurrences) to `const data = await ddb.send(...`

packages/idempotency/tests/e2e/makeFunctionIdempotent.test.ts
```diff
@@ -0,0 +1,101 @@
...
+const makeFunctionIdepmpotentFile = 'makeFunctionIdempotent.test.FunctionCode.ts';
+
+const app = new App();
+let stack: Stack;
+
+describe('Idempotency e2e test, default settings', () => {
+  const ddb = new DynamoDBClient({ region: 'eu-west-1' });
+  stack = new Stack(app, stackName);
+
+  const functionNameDefault = generateUniqueName(RESOURCE_NAME_PREFIX, v4(), runtime, 'default');
```
dreamorosi, line 31: I'd move all these outside of the `describe` to match the other tests

packages/idempotency/tests/e2e/makeFunctionIdempotent.test.FunctionCode.ts
```diff
@@ -0,0 +1,67 @@
...
+const logger = new Logger();
+
+const processIdempotently = makeFunctionIdempotent(
+  processRecord,
+  {
+    persistenceStore: dynamoDBPersistenceLayer,
+    dataKeywordArgument: 'foo'
+  });
+
+function processRecord(record: Record<string, unknown>): string {
```
dreamorosi, line 35:
```suggestion
const processRecord = (record: Record<string, unknown>): string => {
```

packages/idempotency/tests/e2e/makeFunctionIdempotent.test.FunctionCode.ts
```diff
@@ -0,0 +1,67 @@
+import { DynamoDBPersistenceLayer } from '../../src/persistence';
+import { makeFunctionIdempotent } from '../../src/makeFunctionIdempotent';
+import { Logger } from '@aws-lambda-powertools/logger';
+import { Context } from 'aws-lambda';
+
+const IDEMPOTENCY_TABLE_NAME = process.env.IDEMPOTENCY_TABLE_NAME;
```
dreamorosi, line 6:
```suggestion
const IDEMPOTENCY_TABLE_NAME = process.env.IDEMPOTENCY_TABLE_NAME || 'table_name';
```

dreamorosi (changes requested):
Thank you for this PR and for the hard work, I can see (and I know) you've put a lot of effort into this - I appreciate it.

Overall it's looking good - I have left some comments that I'd like to discuss before moving forward.

packages/idempotency/src/types/IdempotencyOptions.ts
```diff
@@ -1,9 +1,11 @@
 import type { Context } from 'aws-lambda';
 import { BasePersistenceLayer } from '../persistence/BasePersistenceLayer';
+import { IdempotencyConfig } from 'IdempotencyConfig';
 
-type IdempotencyOptions = {
-  dataKeywordArgument: string
+type IdempotentOptions = {
```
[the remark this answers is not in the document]
am29d: Forgot to push changes yesterday. Changed to `IdempotencyConfig`.

packages/idempotency/tests/e2e/makeFunctionIdempotent.test.ts
am29d: oooh that makes sense, will adjust

packages/idempotency/tests/e2e/idempotencyDecorator.test.FunctionCode.ts
```diff
@@ -0,0 +1,95 @@
...
+    logger.info(`Got test event: ${JSON.stringify(_event)}`);
+    for (const record of _event.records) {
+      logger.info(`Processing event: ${JSON.stringify(record)}`);
+      await this.process(record);
+    }
+
+    return 'Hello World Keyword Argument';
+  }
+
+  @idempotent({ persistenceStore: dynamoDBPersistenceLayer, dataKeywordArgument: 'foo' })
```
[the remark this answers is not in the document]
am29d: I agree to introduce dedicated decorator for the handler to make the input more explicit for each case. Let's have `idempotentLambdaHandler` and `idempotentMethod` similar to the Tracer. 

> As a matter of fact, the dataKeywordArgument (at least the way I understand it in Python) refers to the name of the argument of the function that contains the data to be used for the idempotency, in this case it'd be record. Not the key (i.e. foo) in the payload.

This was [discussed in the RFC previously](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/447#issuecomment-1348681945). The agreement was enforce the signature of the function wrapper to a record and use `datakeywordArgument` to decide which key/value to use for idempotency.

packages/idempotency/tests/e2e/makeFunctionIdempotent.test.FunctionCode.ts
am29d: Why should we add default?

dreamorosi: TypeScript is complaining because environment variables can be `string | undefined`. We know it's there because we created the function, but it doesn't know it.

To make it shut up we have three options:
- Add a default (we do this in all other e2e tests) so it's always a string
- Change the declaration to `process.env.IDEMPOTENCY_TABLE_NAME!` - note the exclamation mark which tells TS this value is for sure there
- Introduce some `if` statement that check whether the variable it's defined or not, and if not throw an error

Number three is verbose, number 2 would generate a warning from eslint which we'd have to suppress, so we are left with number one.

packages/idempotency/tests/e2e/idempotencyDecorator.test.FunctionCode.ts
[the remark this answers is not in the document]
dreamorosi: Okay, I'm still a bit on the fence about this and not 100% sure about the agreement that was reached in the RFC now that I understand better the Idempotency utility.

I propose this, let's put a pin on the `dataKeywordArgument` topic for now and follow the RFC, so that we don't block the development. When we are done with all the features let's try to use it in real codebases and see how the DX is, as well as hearing customer feedback. If needed we'll change it then.

dreamorosi:
You can run `npm run lint-fix -w packages/commons` from the root dir to fix all the linting stuff that popped out. I didn't think it would affect your PR as I forgot the `commons` package was touched, my bad.

packages/idempotency/src/IdempotencyHandler.ts
```diff
@@ -40,10 +40,31 @@ export class IdempotencyHandler<U> {
           `There is already an execution in progress with idempotency key: ${idempotencyRecord.idempotencyKey}`
         );
       }
-    } else {
-      // Currently recalling the method as this fulfills FR1. FR3 will address using the previously stored value https://github.com/awslabs/aws-lambda-powertools-typescript/issues/447
-      return this.functionToMakeIdempotent(this.fullFunctionPayload);
     }
+
+    return idempotencyRecord.getResponse() as U;
```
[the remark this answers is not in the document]
am29d: Ah this thing gave me a bit of headache. Imho we can't guess the return type of the function, so at this point the result will be `unknown`. What do you think?

packages/idempotency/src/types/IdempotencyOptions.ts
```diff
@@ -1,9 +1,11 @@
...
   persistenceStore: BasePersistenceLayer
+  dataKeywordArgument?: string
+  config?: IdempotencyConfig
```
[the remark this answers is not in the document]
am29d: I need to revisit this part. The config is relevant for the persistence layer, so we don't need it here at all.

packages/idempotency/tests/e2e/makeFunctionIdempotent.test.FunctionCode.ts
am29d: Interesting, I don't have the warning :) I will add a default value.

dreamorosi: image

I hope that after applying the linting/formatting repo-wide we'll have the same configs & see same warnings.

packages/idempotency/src/IdempotencyHandler.ts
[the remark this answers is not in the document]
am29d: Let's revisit this topic in a separate issue.

packages/idempotency/src/types/IdempotencyOptions.ts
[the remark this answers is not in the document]
am29d: ok nvm, we pass config to the `IdempotencyHandler` and need to configure the persistence layer within the constructor.

packages/idempotency/src/IdempotencyHandler.ts
am29d: * Need to add `persistanceStore.configure(config)` to the constructor
* Do you mean wrapping the signature into an interface `IdempotencyHandlerOptions`?
* agree

dreamorosi (approved):
Thank you for working on this, I've run the integration tests on this branch and it's passing.

I'm gonna fix the linting in a subsequent PR and also open an issue to remember ourselves to go back to the types once we are done with the main implementation.

Thank you again Alex, enjoy the weekend!

Issue #1293, opened 2023-02-17
Feature request: idempotency Middy middleware


saragerion:
### Use Case

Users who already use Middy middleware in their functions should be able to make their function `handler` idempotent via a `makeHandlerIdempotent` Middy-compatible middleware.

### Solution/User Experience

```ts
import { 
  makeHandlerIdempotent,
  DynamoDBPersistenceLayer,
  IdempotencyConfig
} from '@aws-lambda-powertools/idempotency';
import middy from '@middy/core';

const config = new IdempotencyConfig({...});
const persistenceLayer = new DynamoDBPersistenceLayer({...});

const lambdaHandler = async (_event: any, _context: any): Promise<void> => {
  /* ...Function logic here... */
}

export const handler = middy(lambdaHandler)
  .use(makeHandlerIdempotent({ config, persistenceLayer });
```

The `makeHandlerIdempotent` middleware should work with both `sync` and `async` function `handler`s. The middleware should accept an object with mandatory `persistenceStore` and an optional `config` one. The former should be an instance of any class that extends `BasePersistenceLayer`, while the latter should be an instance of the class `IdempotencyConfig`.

Following the Powertools for Python decorator implementation, the middleware should:
- return early if the `POWERTOOLS_IDEMPOTENCY_DISABLED` env variable has a truthy value (using `EnvironmentVariableService`) - `onBefore`
- use the provided config object or instantiate a new one if none is passed - `onBefore`
- register the Lambda context into the config object (used to manage timeouts) - `onBefore`
- instantiate an `IdempotencyHandler` - `onBefore`
- call & return the `IdempotencyHandler.handle()` method - `onBefore`

This last step will ensure that the `IdempotencyHandler` will perform all the actions needed to make the function idempotent.

Given how Middy handles requests and errors, there's a chance that the current public methods of `IdempotencyHandler` might not be enough and we'd need to add additional public methods to call in the `onAfter` or `onError` phase.

### Acknowledgment

- [X] This feature request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/#tenets)
- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [Java](https://github.com/awslabs/aws-lambda-powertools-java/)

PR #1474, opened 2023-05-24
feat(idempotency): `makeHandlerIdempotent` middy middleware

dreamorosi:

This PR introduces a new Middy middleware that allows to wrap a Lambda handler and make it idempotent.

The middleware can be used like this:

```ts
import {
  makeHandlerIdempotent,
  DynamoDBPersistenceLayer,

const dynamoDBPersistenceLayer = new DynamoDBPersistenceLayer({
  tableName: 'idempotencyTable',
});

const lambdaHandler = async (_event: unknown, _context: unknown) => {
  //...
};

  .use(makeHandlerIdempotent({ persistenceStore: dynamoDBPersistenceLayer }));
```

#### Implementation details

The first detail to notice is that contrary to the function wrapper and the class method decorator, this middleware doesn't leverage fully the `IdempotencyHandler` class. The reason for this is that Middy has a specific lifecycle when it comes to logic running before, after, and in case of error thrown by the handler. The current implementation of the `IdempotencyHandler` groups certain actions together and expects certain signatures and a class instantiation and as such it's not possible for the middleware to use the same method all the time.

As a result, the middleware interacts with the persistence store directly and is in charge of handling/throwing the errors. This allows the middleware to follow its own patterns, however moving forward we will need to make sure we keep the two implementations in sync.

### Related issues, RFCs

**Issue number:** #1293

## Checklist

### Breaking change checklist

---

### DIFF
--- a/packages/idempotency/src/middleware/makeHandlerIdempotent.ts
+++ b/packages/idempotency/src/middleware/makeHandlerIdempotent.ts
@@ -165,7 +165,9 @@ const makeHandlerIdempotent = (
     };
   } else {
     return {
-      before: () => ({}),
+      before: () => {
+        return undefined;
+      },
     };
   }
 };
--- a/packages/idempotency/src/middleware/makeHandlerIdempotent.ts
+++ b/packages/idempotency/src/middleware/makeHandlerIdempotent.ts
@@ -66,6 +66,7 @@ const makeHandlerIdempotent = (
    * cases we can safely retry the handling a few times.
    *
    * @param request - The Middy request object
+   * @param retryNo - The number of times the handler has been retried
    */
   const before = async (
     request: MiddyLikeRequest,
@@ -156,11 +157,17 @@ const makeHandlerIdempotent = (
     }
   };
 
-  return {
-    before,
-    after,
-    onError,
-  };
+  if (idempotencyConfig.isEnabled()) {
+    return {
+      before,
+      after,
+      onError,
+    };
+  } else {
+    return {
+      before: () => ({}),
+    };
+  }
 };
 
 export { makeHandlerIdempotent };
--- a/packages/idempotency/src/middleware/makeHandlerIdempotent.ts
+++ b/packages/idempotency/src/middleware/makeHandlerIdempotent.ts
@@ -4,8 +4,10 @@ import { cleanupMiddlewares } from '@aws-lambda-powertools/commons/lib/middlewar
 import {
   IdempotencyItemAlreadyExistsError,
   IdempotencyPersistenceLayerError,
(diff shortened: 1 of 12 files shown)


packages/idempotency/src/middleware/makeHandlerIdempotent.ts
```diff
@@ -0,0 +1,144 @@
...
+ * export const handler = middy(lambdaHandler)
+ *  .use(makeHandlerIdempotent({ persistenceStore: dynamoDBPersistenceLayer }));
+ * ```
+ *
+ * @param options - Options for the idempotency middleware
+ */
+const makeHandlerIdempotent = (
+  options: IdempotencyLambdaHandlerOptions
+): MiddlewareLikeObj => {
+  const idempotencyConfig = options.config
```
dreamorosi, line 43: Is it okay to instantiate this once and reuse them throughout different requests?

Asking because in the decorator and function wrapper we create a new one for each request.

packages/idempotency/src/middleware/makeHandlerIdempotent.ts
```diff
@@ -0,0 +1,144 @@
...
+   * execution in progress for the given idempotency key. If there is, we
+   * need to determine its status and return the appropriate response or
+   * throw an error.
+   *
+   * If there is no execution in progress, we need to save a record to the
+   * idempotency store to indicate that an execution is in progress.
+   *
+   * @param request - The Middy request object
+   */
+  const before = async (request: MiddyLikeRequest): Promise<unknown | void> => {
```
dreamorosi, line 64: Currently the middleware doesn't implement the retry mechanism that is [present in `IdempotencyHandler.handle`](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/244cfd2b42d24318ab1d149ab694c68109aef342/packages/idempotency/src/IdempotencyHandler.ts).

I am not 100% sure whether this can be done in the context of a middleware and before being sure I need to better understand when is that we can/need retry. Could we have a quick chat so maybe you can help me?

packages/idempotency/src/middleware/makeHandlerIdempotent.ts
dreamorosi: Confirmed that this is fine.

packages/idempotency/src/middleware/makeHandlerIdempotent.ts
dreamorosi: The retry mechanism is already handled by `IdempotencyHandler`, so there's no additional implementation to do here.

am29d (approved):

PR #1225, opened 2023-01-11
refactor(all): add middy-like types to commons

dreamorosi:
## Description of your changes

Following the discussion in the linked issue (#1068), as well as other issues (#1080, #1081), customers who are using Powertools and bundling their TypeScript project with `tsc`, but that were not using any of the Middy-compatible middleware were having compile issues due to the dependency being missing.

We have investigated several workarounds:
- Adding the `@middy/core` dependency (#1218)
- Changing the exports to isolate the middleware & thus the dependency ([#1068#issuecomment-1377575328](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/1068#issuecomment-1377575328))
- Reworking the types (this PR)

The first option was a no-go as it would have introduced an unnecessary dependency for those users having the issue. The second one would have constituted a breaking change, requiring a major release.

The last option (**credits to @saragerion for proposing it**) allows us to maintain backwards compatibility while solving the issue described above.

The solution that this PR proposes (again, credits to @saragerion), introduces some of the types previously imported from `@middy/core` into `@aws-lambda-powertools/commons`. By bringing these types into the project, we can remove the dependency to `@middy/core` from our codebase and allow customers to bundle using `tsc` as well as `esbuild`.

> **Note**
> The `@middy/core` dependency was used only for its types. Powertools still supports Middy middleware, and customers who want to use them still need to opt-in by adding the dependency to their Lambda functions.

### How to verify this change

See existing unit tests below this PR and successful integration test run: https://github.com/awslabs/aws-lambda-powertools-typescript/actions/runs/3894429621

### Related issues, RFCs

**Issue number:** #1068

### PR status

***Is this ready for review?:*** YES  
***Is it a breaking change?:*** NO

## Checklist

- [x] [My changes meet the tenets criteria](https://awslabs.github.io/aws-lambda-powertools-typescript/#tenets)
- [x] I have performed a *self-review* of my own code
- [x] I have *commented* my code where necessary, particularly in areas that should be flagged with a TODO, or hard-to-understand areas
- [ ] I have made corresponding changes to the *documentation*
- [ ] I have made corresponding changes to the *examples*
- [x] My changes generate *no new warnings*
- [x] The *code coverage* hasn't decreased
- [ ] I have *added tests* that prove my change is effective and works
- [x] New and existing *unit tests pass* locally and in Github Actions
- [ ] Any *dependent changes have been merged and published*
[...]

Issue #1068, opened 2022-08-22
Bug: unable to build with `tsc` when not using `@middy/core`


rreddypally:
## Bug description

EDIT: see [this comment](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/1068#issuecomment-1227047896) for updated info after triage.

---

The import statement in Tracer.ts for `TracerInterface` is importing all the modules from the root. https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/tracer/src/Tracer.ts#L3.

I am trying to use tracer module without the need to install middy.

### Expected Behavior

Should be able to use Tracer without the need to import middy when its not used.

### Current Behavior

This behavior is forcing to import middy middleware

### Possible Solution

import { TracerInterface } from './TracerInterface';

### Steps to Reproduce

1. Install tracer npm module `npm i @aws-lambda-powertools/tracer`
2. Import tracer `import { Tracer } from "@aws-lambda-powertools/tracer/lib/Tracer";`
3. Add tracer annotation and build fails with `error TS2307: Cannot find module '@middy/core' or its corresponding type declarations`.

### Environment

* **Powertools version used**: 1.1.1
* **Packaging format (Layers, npm)**: npm
* **AWS Lambda function runtime**: Node 16

dreamorosi:
Hi @rreddypally thank you for opening this issue.

I have tried reproducing the issue you describe but unfortunately I am not able to.

In [the docs](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/tracer/#usage) we recommend to import the Tracer directly (i.e. `import { Tracer } from '@aws-lambda-powertools/tracer';`), however even if I import it like you suggest I am still able to bundle the function, run it successfully on AWS Lambda, and see the annotated segment in the traces.

I have made a screen record in which I create a new project from zero and I set up the latest version of Tracer in a function, then bundle it, and then deploy & run it. As mentioned, I have tested both `import` statements (the recommended one, and the scoped one), and both build & run successfully.

As you'll see in the video, I never install Middy and I only install `esbuild` to bundle, and `@types/node` for types (probably redundant anyway).

https://user-images.githubusercontent.com/7353869/185918256-4d69291f-d083-45e4-a1c8-129e7de1fd13.mp4

[...]

Issue #617, opened 2022-03-03
Maintenance: simplify and revisit tsconfig files & settings 


saragerion:
## Description of the feature request

Right now the tsconfig files located in the folder of each utility (Logger, Metrics, Tracer) have repeated settings with some differences. 
Example: `resolveJsonModule` in the Logger utility, but there might be others.
Would be good to revisit the current tsconfig setup and the way we transpile our code. 

**Problem statement**

Inconsistent, potentially unnecessary tsconfig settings across the same utility and different utilities.

**Summary of the feature**

Simplify tsconfig files & settings.

**Code examples**

I like the way the AWS SDK v3 does it, extending one another:
https://github.com/aws/aws-sdk-js-v3/tree/main/clients/client-s3

**Benefits for you and the wider AWS community**

Simplification and better governance.

**Describe alternatives you've considered**

N/A

**Additional context**

This issue affects all existing utilities.

goverdhan07:
Hi @saragerion! as far as I have understood, this issue is regarding making one common file for tsconfig for all three folders(Logger, Metrics, Tracer) which has similar settings, and extend individual/different settings in each of the folder through new tsconfig files. ryt? please let me know if I'am missing something here :) Thanks!

dreamorosi:
@goverdhan07 thanks for commenting the issue, I understand that what you're mentioning is correct but there might also be some settings in the `tsconfig.json` files that are unnecessary and that might be possible to remove. Ideally we'd need to go through all of them, extend when possible and remove when redundant.

awsdb:
Also note that there are some additional `tsconfig` files like `tsconfig-dev.json` and `tsconfig.es.json`. If they are not needed, the could be deleted. And another note: `tsconfig.json` only includes `"src/**/*"` but not `tests`, so all the compiler settings (like e.g. `"strict": true`) won't be applied to the test code which is a bit unfortunate (e.g. no consistent live linting in VSCode for tests).

dreamorosi:
Not sure if related, but in #1373, and specifically in point 4 of the issue it's mentioned that even though `strict: true` is applied in the config, interfaces are not flagging errors when a class that implements them does so incorrectly, i.e.:
```ts
const MetricUnit = {
    Bites: 'Bites'
} as const;
type MetricUnit = typeof MetricUnit;

interface MetricsInterface {
    addMetric(name: string, unit: MetricUnit, value: number): void
}

class Metric implements MetricsInterface {
[...]

PR #1485, opened 2023-06-07
docs(idempotency): API docs for idempotency

am29d:

* bump typedoc deps
* add typedoc config per package with a base configuration in the root folder, this helps us to control the `entryPoints` for each package
* add api docs for idempotency
* changed idempotency README to specifics of that package and additional information to the root README. 

### Related issues, RFCs

**Issue number:** #1104 

## Checklist

### Breaking change checklist

---


tsconfig.json
dreamorosi, line 1: Why was this needed here?

I don't disagree that we should also centralize the `tsconfig` and apply the same pattern of extending it (#617), just trying to understand if there's a dependency between the changes made in this PR and this file being at the root.

typedoc.base.json
dreamorosi, line 1: What's the difference between `typedoc.json` and `typedoc.base.json` and do we need both?

packages/idempotency/src/persistence/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './BasePersistenceLayer';
 export * from './BasePersistenceLayerInterface';
 export * from './IdempotencyRecord';
+export * from './DynamoDBPersistenceLayer';
```
dreamorosi, line 4: I left this out and exposed it via a separate import on purpose, and I feel strongly (but are willing to discuss) about keeping it that way.

At the moment we only have one persistence layer, so having it exported here is fine.

If (or actually when) we add more persistence layers (i.e. ElastiCache or others), this file will become:

```ts
// packages/idempotency/src/persistence/index.ts
export * from './DynamoDBPersistenceLayer';
export * from './OtherPersistenceLayer';
```

However **customers using one won't need the other**.

By having it like this, whenever customers do this: `import { DynamoDBPersistenceLayer } from '@aws-lambda-powertools/idempotency/persistence';` or this: `import { OtherPersistenceLayer } from '@aws-lambda-powertools/idempotency/persistence';` both are imported.

This a few undesirable effects:

- If any of the persistence layers have unique dependencies, compilers like `tsc` will require them to be installed in the project, even if you are not explicitly using the code (because you are importing it) - this has happened in #1068, resulted in #1225, and is something that we [want to change in v2](https://github.com/awslabs/aws-lambda-powertools-typescript/discussions/1269#discussioncomment-4891876) for Logger, Tracer, and Metrics.
- Some bundlers like `esbuild` cannot always tree shake the unused module, because it's always being imported
- If the initialization of the module is expensive (which is **not** the case with DynamoDB), then customers would be paying this price for both persistence layers.

By having them separate instead, customers have to explicitly opt-in and import whichever persistence layer they are actually using, i.e. `import { DynamoDBPersistenceLayer } from '@aws-lambda-powertools/idempotency/dynamodb';` or `import { OtherPersistenceLayer } from '@aws-lambda-powertools/idempotency/other';`. This way the imports are isolated from each other and: 1/ dependencies for each one can be installed selectively, 2/ tree-shaking and bundling can be done effectively.

This can be achieved with customizing the exports ([see here](https://github.com/awslabs/aws-lambda-powertools-typescript/blob/main/packages/idempotency/package.json#L44-L47)), and is also why in Parameters we don't have an `index.ts` file that exports all the different providers.

packages/idempotency/src/persistence/IdempotencyRecord.ts
```diff
@@ -1,15 +1,31 @@
...
   public inProgressExpiryTimestamp?: number;
+  /**
+   * The hash of the payload of the request, used for comparing requests.
+   */
   public payloadHash?: string;
+  /**
+   * The response data of the request, this will be returned if the payload hash matches.
+   */
   public responseData?: Record<string, unknown>;
   private status: IdempotencyRecordStatus;
```
dreamorosi, line 30: Can we add a comment here like the others & maybe link to `IdempotencyRecordStatus`?

packages/idempotency/src/index.ts
```diff
@@ -2,3 +2,5 @@ export * from './Exceptions';
 export * from './IdempotencyConfig';
 export * from './idempotentDecorator';
 export * from './makeFunctionIdempotent';
+export * from './middleware';
```
dreamorosi, line 5: See comment I left under `packages/idempotency/src/persistence/index.ts` about imports/exports - TL;DR; the omission was intentional and I believe should stay like that.

packages/idempotency/README.md
```diff
@@ -1,71 +1,131 @@
...
 
-The [Serverless TypeScript Demo](https://github.com/aws-samples/serverless-typescript-demo) shows how to use Powertools for AWS Lambda (TypeScript).  
-You can find instructions on how to deploy and load test this application in the [repository](https://github.com/aws-samples/serverless-typescript-demo).
+const processIdempotently = makeFunctionIdempotent(proccessingFunction, {
+  persistenceStore: dynamoDBPersistenceLayer,
+  dataKeywordArgument: "transactionId"
+});
+
+export const handler = async (
+  _event: Event,
```
dreamorosi, line 121: Ideally these examples should be copy-paste-able. If we I do that I don't have imports for `Event` & `Context`.

We can either add the imports from `aws-lambda`, or use `unknown` like done in the rest of the docs/examples.

dreamorosi (changes requested):
Hey Alex, thanks for the great work - I appreciate that you took the time to improve the API docs for the whole project and standardized the configs!

I have left a few comments, mostly minor ones, except one related to the imports and exports. See more details below - happy to discuss it further later if needed.

typedoc.base.json
am29d: `typedoc.base.json` is the common configuration for individual packages. We need to the separation because otherwise typedoc will have a circular dependency if you run it on the root level: `typedoc.json` -> `package/logger/typedoc.json-> `typedoc.json`. The error sounds something like this:

> [error] Project at ./packages/logger has entryPointStrategy set to packages, but nested packages are not supported.

packages/idempotency/src/persistence/index.ts
am29d: Hmm I see, I was wondering about the inconsistency of the exports. I agree that we should be mindful of the exports. 

I added it to collect it for the typedoc output, since the entrypoint is `index.ts`. I will revert the exports to the previous state.

packages/idempotency/README.md
am29d: you are right, I always complained about missing import in any other docs, now I did the same 😅

typedoc.json
```diff
@@ -0,0 +1,23 @@
+{
+  "entryPoints": ["packages/*"],
+  "entryPointStrategy": "packages",
+  "name": "AWS Lambda Powertools Typescript API Reference",
```
dreamorosi, line 4:
```suggestion
  "name": "Powertools for AWS Lambda (Typescript) API Reference",
```

am29d (commented):
good catch

dreamorosi (approved):
Good work with this PR, the new docs are looking nice.

I think we should include the `BaseProvider` class for Parameters (so that customers can extend it and build their own provider) and check that `DynamoDBPersistenceLayer` is included in the exports - but we'll do it in separate PRs.

Issue #1509, opened 2023-06-19
Maintenance: phase out docs publishing to GitHub Pages


dreamorosi:
### Summary

To complete the migration to the new GitHub Org and docs infrastructure, we need to remove from the Docs workflow all the steps that push the docs to GitHub pages, as well as disabling the feature in the repo settings.

### Why is this needed?

So that the docs are published only to the new destination.

### Which area does this relate to?

Automation

### Acknowledgment

- [X] This request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda-typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

dreamorosi:
The `versions.json` [file in the repo](https://github.com/aws-powertools/powertools-lambda-typescript/blob/backup-gh-pages/versions.json) doesn't match with the one in S3 (`1.9.0` and its `latest` alias are missing).

This results in the docs not being available for the latest version.

PR #1513, opened 2023-06-19
chore(ci): remove GH pages

sthulb:

Remove GH pages from GH actions. We no longer use GH pages, as we have migrated to our new docs hosting already.

### Related issues, RFCs

**Issue number:** #1509

## Checklist

### Breaking change checklist

---

### DIFF
--- a/.github/workflows/reusable-publish-docs.yml
+++ b/.github/workflows/reusable-publish-docs.yml
@@ -85,21 +85,6 @@ jobs:
         run: |
           rm -rf api
           npm run docs-generateApiDoc
-      - name: Release API docs
-        uses: peaceiris/actions-gh-pages@bd8c6b06eba6b3d25d72b7a1767993c0aeee42e7 # v3.9.2
-        with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          publish_dir: ./api
-          keep_files: true
-          destination_dir: ${{ env.VERSION }}/api
-      - name: Release API docs to latest
-        if: ${{ inputs.alias == 'latest' }}
-        uses: peaceiris/actions-gh-pages@bd8c6b06eba6b3d25d72b7a1767993c0aeee42e7 # v3.9.2
-        with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          publish_dir: ./api
-          keep_files: true
-          destination_dir: latest/api
       - name: Configure AWS credentials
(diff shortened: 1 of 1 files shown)


dreamorosi:
@sthulb are we still keeping the `gh-pages` branch around or just building the docs and publishing them?

Also can you please take a look at the comment in #1509 and sync the versions? I could manually push the `versions.json` file to the bucket but that might not be the best way to fix it.

sthulb:
> @sthulb are we still keeping the `gh-pages` branch around or just building the docs and publishing them?

just building and pushing to S3. There’s no need to keep the branch. The backup-GH-pages branch exists as a part of the transition. 

> 
> Also can you please take a look at the comment in #1509 and sync the versions? I could manually push the `versions.json` file to the bucket but that might not be the best way to fix it.

versions.json will be addressed in a PR tomorrow. I have to do the same for python too

dreamorosi:
Ok, I don't know `mkdocs` and `mike` enough so I'll leave this to you.

dreamorosi (commented):
Removing these steps only removes the publishing of the API folder (aka API docs), the actual docs written with `mkdocs` are still being pushed to the `gh-pages` (currently `backup-gh-pages`) branch by `mike` at [L81-83](https://github.com/aws-powertools/powertools-lambda-typescript/blob/5d035a1d4597d9036cdc4ea58a10b046b8a67f52/.github/workflows/reusable-publish-docs.yml#L81-L84).

Did you leave them on purpose?

Also, if we remove the branch entirely (which I think we should), I'm not sure how the workflow is going to be able to manage the `versions.json`. My understanding is that currently `mike` relies on the presence of this file in the other branch, and when creating a new version/alias it appends it to the file.

If the branch is removed, then we need to make sure `mike` is able to handle the file that is now present in S3.

dreamorosi:
@sthulb given #1521 should we remove `mike` entirely from both the workflow & the `docs/requirements.txt` file?

dreamorosi:
Hey @sthulb, can we remove `mike deploy --push --update-aliases --no-redirect ${{ env.VERSION }} ${{ env.ALIAS }} --branch backup-gh-pages` and `mike set-default --push latest --branch backup-gh-pages` or at least its `--push` flag?

We have discussed to keep around the `backup-gh-pages` branch, which I agree with, but I'm not sure whether we still want to continue pushing to it.

If yes, then we need to restore the changes you're proposing in this PR (& close it), so that both docs & API docs are published.

If no, then we can remove the `--push` flag from the commands above or (better) remove the commands entirely.

Let me know if I'm missing something, looking forward to merge/close the PR.

Issue #1320, opened 2023-02-21
Docs: add example for create your own provider


dreamorosi:
### What were you searching in the docs?

The current version of the Parameters docs doesn't include an example of creating a custom provider for the utility. This section would be useful because it showcases the versatility & extensibility of the utility by allowing customers to create their own provider by extending the `BaseProvider`.

In doing so, customers can retrieve parameters/secrets/configs from 3rd party or unsupported services and still leverage transform and caching from Parameters.

### Is this related to an existing documentation section?

This [section exists in the docs](https://awslabs.github.io/aws-lambda-powertools-python/2.9.0/utilities/parameters/#create-your-own-provider) of the utility in Powertools for Python.

### How can we improve?

We should create a similar section that shows at least one custom provider.

### Got a suggestion in mind?

N/A

### Acknowledgment

- [X] I understand the final update might be different from my proposed suggestion, or refused.

PR #1528, opened 2023-06-20
feat(parameters): review types and exports

### LINKED ISSUE #1424: Feature request: improve return types for AppConfig Parameters
### Use case

> **Note**
> This is a follow up issue to #1386 and aims at continuing the work started in #1387

Currently when retrieving parameter values from AppConfig, the Parameter utility has somewhat generic return types.

Given that the utility knows both the types returned by the API and the transformation to be applied, there's an opportunity to improve the return types by applying some heuristics.

### Solution/User Experience

TBD

### Alternative solutions

_No response_

### Acknowledgment

- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [Java](https://github.com/awslabs/aws-lambda-powertools-java/), and [.NET](https://github.com/awslabs/aws-lambda-powertools-dotnet/)

### Future readers

### LINKED ISSUE #1425: Feature request: improve return types for DynamoDB Parameters
### Use case

> **Note**
> This is a follow up issue to #1386 and aims at continuing the work started in #1387

Currently when retrieving parameter values from DynamoDB, the Parameter utility has somewhat generic return types.

Given that the utility knows both the types returned by the API and the transformation to be applied, there's an opportunity to improve the return types by applying some heuristics.

### Solution/User Experience

TBD

### Alternative solutions

_No response_

### Acknowledgment

- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [Java](https://github.com/awslabs/aws-lambda-powertools-java/), and [.NET](https://github.com/awslabs/aws-lambda-powertools-dotnet/)

### Future readers

dreamorosi:

This PR touches most of the files in the Parameters utility with the goal of streamlining the types of the inputs & outputs of each provider, as well as finalizing the exports that the utility will make available to customers. Additionally, it also adds a few types/[type guards](https://blog.logrocket.com/how-to-use-type-guards-typescript/) in the commons utility which are currently used by the parameters utility but that will be used by others as well.

### Exports

As mentioned in previous issues/PRs/discussions, we want customers to be able to tree-shake the library as much as possible and be able to build the project without the need of installing dependencies that might needed by parts of the code that they are not using.

For example, in Parameters, customers using the `SecretsProvider` might not be interested in bringing into their project `@aws-sdk/client-appconfigdata`, a dependency needed by the `AppConfigProvider`. To allow for this, we need to create granular exports in the `package.json` of the utility.

The exports are governed by the `exports` and `typeVersions` sections of the `package.json` of the utility. The first one tells Node.js where to resolve a certain path in the `node_modules` entry of the utility, whereas the `typeVersions` section tells TypeScript where to find the types for that same path.

The exports that we are offering are:
```ts
import { Transform } from '@aws-lambda-powertools/parameters';
import { GetParameterError } from '@aws-lambda-powertools/parameters/errors';
import { ... } from '@aws-lambda-powertools/parameters/<PROVIDER>';
import type { ... } from '@aws-lambda-powertools/parameters/<PROVIDER>/types';
```

The first export, and only top-level one, is a readonly object that contains the different transform available so that customers can use it like `Transform.JSON` or `Transform.AUTO` and not have to worry about spelling or case (although the regular string is still allowed).

The second export, under `/errors` exposes the types of errors that can be thrown by the utility. This can be used by customers to catch & handle errors or during tests.

The other exports are all specific to the providers, for example, the SSM provider, exports the following things:
```ts
import {
  getParameter,
  getParameters,
  getParametersByName,
  SSMProvider,
 } from '@aws-lambda-powertools/parameters/ssm';
import type { ... } from '@aws-lambda-powertools/parameters/ssm/types'; // all the types specific to SSMProvider & functions
```

The other providers follow the same logic. We are also exposing the `BaseProvider` under the `/base` path, this way customers will be able to extend this class to create their own custom provider.

In order to support these exports I had to partially reorganize the folder structure of the utility; this is the main driver behind the large diff.

#### Base Provider

Before this PR, the `BaseProvider` was quite strict in the return types for the `get` and `getMultiple` methods. Specifically, it allowed only to return `string`, `Uint8Array`, `Record<string, unknown>`, and `undefined`. These types were overly restrictive and would have hindered the extensibility of the provider.

When researching HashiCorp Vault for #1320 I discovered that this particular store returned types that were incompatible with the types described above, this sparked also the discussion in aws-powertools/powertools-lambda-python#2250.

Starting from this PR, the `BaseProvider` is free to return a promise that resolves to `unknown`. The `BaseProvider` handles caching and transformations only, so it doesn't need to know the types it's handling when returning. The actual value retrieval is done by the `_get` & `_getMultiple` methods, which are `abstract` and leave the actual implementation to more specific providers.

With this in mind, `BaseProvider` can return `unknown` and not care about the return types as long as the downstream providers (i.e. `SecretsProvider`, `DynamoDBProvider`, or `VaultProvider`) handle and validate the types they retrieve and return as needed.

Once made this change I had to do some light rework of the types & add some runtime type-checks to the transform function so that we are able to work with unknown types most of the time and still apply the transform only on types we want.

#### Adaptive types for `DynamoDBProvider` and `AppConfigProvider`

Continuing the work started in previous releases for the other providers, I have applied generics to these two providers so that we can make some assumptions on the return type based on the transformations applied, or allow customers to specify a type if they know more than the compiler.

This closes #1425, and closes #1424

#### Updates to JSON transform types

Before this PR I mistakenly assumed that when applying a JSON transform the result type would always be an object (`{}` / `Record<string, unknown>`. This was a naïve assumption.

In reality, the JSON specification allows to parse and considers many/most primitive types as valid JSON documents. For instance, `"hello"`, `5`, `true`, and `[]` among others, are all valid JSON documents that can be processed with `JSON.parse()`.

For this reason I introduced some new types in the `commons` utility that represent these primitives. The new `JSONValue`, which includes the combinations of primitives as well as arrays and objects containing these primitives, is now used as default return type whenever a JSON transform is applied.

As a side note, the types just discussed will be useful also for the Idempotency utility and for other utilities handling unknown payloads that can be JSON parsed/stringified.

#### Miscellaneous 

While working on the PR I also had to do changes here & there, these don't fit in any specific category:
- Shortened/standardized the names of most customer-facing types/interfaces (i.e. `SSMGetOptionsUnion` becomes simply `SSMGetOptions`)
- Fixed some formatting issues in the docstrings as I encountered & updated the `typedoc.json` config to reflect the new exports
- Updates to unit & integration tests to reflect the changes above (mostly imports & types-related)

### Related issues, RFCs

**Issue number:** #1424

## Checklist

### Breaking change checklist

---

### DIFF
--- a/packages/parameters/src/appconfig/AppConfigProvider.ts
+++ b/packages/parameters/src/appconfig/AppConfigProvider.ts
@@ -7,7 +7,8 @@ import {
 import type { StartConfigurationSessionCommandInput } from '@aws-sdk/client-appconfigdata';
 import type {
   AppConfigProviderOptions,
-  AppConfigGetOptionsInterface,
+  AppConfigGetOptions,
+  AppConfigGetOutput,
 } from '../types/AppConfigProvider';
 
 /**
@@ -31,7 +32,10 @@ import type {
  * ```typescript
  * import { AppConfigProvider } from '@aws-lambda-powertools/parameters/appconfig';
  *
- * const configProvider = new AppConfigProvider();
+ * const configProvider = new AppConfigProvider({
+ *   application: 'my-app',
+ *   environment: 'prod',
+ * });
  *
  * export const handler = async (): Promise<void> => {
  *   // Retrieve a configuration profile
@@ -52,7 +56,10 @@ import type {
  * ```typescript
  * import { AppConfigProvider } from '@aws-lambda-powertools/parameters/appconfig';
  *
- * const configProvider = new AppConfigProvider();
+ * const configProvider = new AppConfigProvider({
+ *   application: 'my-app',
+ *   environment: 'prod',
+ * });
  *
  * export const handler = async (): Promise<void> => {
  *   // Retrieve a configuration profile and cache it for 10 seconds
@@ -67,7 +74,10 @@ import type {
  * ```typescript
  * import { AppConfigProvider } from '@aws-lambda-powertools/parameters/appconfig';
  *
- * const configProvider = new AppConfigProvider();
+ * const configProvider = new AppConfigProvider({
+ *   application: 'my-app',
+ *   environment: 'prod',
+ * });
  *
  * export const handler = async (): Promise<void> => {
  *   // Retrieve a config and always fetch the latest value
@@ -84,7 +94,10 @@ import type {
  * ```typescript
  * import { AppConfigProvider } from '@aws-lambda-powertools/parameters/appconfig';
  *
- * const configProvider = new AppConfigProvider();
+ * const configProvider = new AppConfigProvider({
+ *   application: 'my-app',
+ *   environment: 'prod',
+ * });
  *
  * export const handler = async (): Promise<void> => {
  *   // Retrieve a JSON config or Feature Flag and parse it as JSON
@@ -98,7 +111,10 @@ import type {
  * ```typescript
  * import { AppConfigProvider } from '@aws-lambda-powertools/parameters/appconfig';
  *
- * const configProvider = new AppConfigProvider();
+ * const configProvider = new AppConfigProvider({
+ *   application: 'my-app',
+ *   environment: 'prod',
+ * });
  *
  * export const handler = async (): Promise<void> => {
  *   // Retrieve a base64-encoded string and decode it
@@ -114,7 +130,10 @@ import type {
  * ```typescript
  * import { AppConfigProvider } from '@aws-lambda-powertools/parameters/appconfig';
  *
- * const configProvider = new AppConfigProvider();
+ * const configProvider = new AppConfigProvider({
+ *   application: 'my-app',
+ *   environment: 'prod',
+ * });
  *
  * export const handler = async (): Promise<void> => {
  *   // Retrieve a config and pass extra options to the AWS SDK v3 for JavaScript client
@@ -140,6 +159,8 @@ import type {
  * import { AppConfigProvider } from '@aws-lambda-powertools/parameters/appconfig';
  *
  * const configProvider = new AppConfigProvider({
+ *   application: 'my-app',
+ *   environment: 'prod',
  *   clientConfig: { region: 'eu-west-1' },
  * });
  * ```
@@ -155,6 +176,8 @@ import type {
  *
  * const client = new AppConfigDataClient({ region: 'eu-west-1' });
  * const configProvider = new AppConfigProvider({
+ *   application: 'my-app',
+ *   environment: 'prod',
  *   awsSdkV3Client: client,
  * });
  * ```
@@ -204,7 +227,10 @@ class AppConfigProvider extends BaseProvider {
    * ```typescript
    * import { AppConfigProvider } from '@aws-lambda-powertools/parameters/appconfig';
    *
-   * const configProvider = new AppConfigProvider();
+   * const configProvider = new AppConfigProvider({
+   *   application: 'my-app',
+   *   environment: 'prod',
+   * });
    *
    * export const handler = async (): Promise<void> => {
    *   // Retrieve a configuration profile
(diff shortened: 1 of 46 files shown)


dreamorosi:
Apologies in advance for the big diff, the PR body explains why it had to be this big (changes in exports & types). This should be the last major PR for this utility though.

packages/parameters/src/appconfig/AppConfigProvider.ts
```diff
@@ -222,35 +248,43 @@ class AppConfigProvider extends BaseProvider {
...
 
   /**
    * Retrieving multiple configurations is not supported by AWS AppConfig.
    */
-  public async getMultiple(
-    path: string,
-    _options?: unknown
-  ): Promise<undefined | Record<string, unknown>> {
-    return super.getMultiple(path);
+  public async getMultiple(path: string, _options?: unknown): Promise<void> {
```
am29d, line 275: Should the return type here be `Promise<unknown>` ? 

Also, can't we provide the same option for the caller to pass explicit types to be expected, similar to `get`? I see, that it's not trivial to handle mixed return types, but getting list of the same type is possible. 

I see you did it for DynamoDBProvider.

packages/parameters/src/appconfig/AppConfigProvider.ts
```diff
@@ -321,7 +355,7 @@ class AppConfigProvider extends BaseProvider {
   protected async _getMultiple(
     _path: string,
     _sdkOptions?: unknown
-  ): Promise<Record<string, string | undefined>> {
+  ): Promise<void> {
```
am29d, line 358: Also here, unclear why it is `void`

packages/parameters/src/base/transformValue.ts
```diff
@@ -0,0 +1,64 @@
...
+import { isString } from '@aws-lambda-powertools/commons';
+import { fromBase64 } from '@aws-sdk/util-base64-node';
+import { TRANSFORM_METHOD_BINARY, TRANSFORM_METHOD_JSON } from '../constants';
+import { TransformParameterError } from '../errors';
+import type { TransformOptions } from '../types/BaseProvider';
+
+/**
+ * Utility function to transform a value.
+ *
+ * It supports JSON and binary transformations, as well as an 'auto' mode that will try to transform the value based on the key.
```
am29d, line 11: I'd add signals on the logic of this transformation when parameter ends with `.json`, `.binary`, so the user can understand which inputs will be transformed. Not too details of course to expose and repeat entire implementation.

packages/parameters/src/types/DynamoDBProvider.ts
```diff
@@ -71,38 +72,120 @@ type DynamoDBProviderOptions =
...
+type DynamoDBGetOptions =
+  | DynamoDBGetOptionsTransformNone
+  | DynamoDBGetOptionsTransformJson
+  | DynamoDBGetOptionsTransformBinary
+  | undefined;
+
+/**
+ * Generic output type for DynamoDBProvider get method.
+ */
+type DynamoDBGetOutput<
```
am29d, line 110: Quite an artistic way for a type definition, forces one to reread multiple times. 

So I will give a try to read the priority:

1. ExplicitUserProvidedType, if not undefined
2. InferredFromOptionsType, if not undefined
   2.1 JSONValue, if undefined or TransformJson or TransformNone
   2.2 String, if TransformBinary
   2.3 never, else
   
Is that right?

am29d (changes requested):
Just a minor question on the void return type for `getMultiple`

packages/parameters/src/appconfig/AppConfigProvider.ts
dreamorosi: `void` is used when there's no `return` keyword. `unknown` would be a wider type in this case.

For `AppConfigProvider` specifically, the `getMultiple` operation is not supported at all and this method, when called, is guaranteed to throw (see `_getMultiple`). 

So since we are not returning anything, we use `void`.

packages/parameters/src/appconfig/AppConfigProvider.ts
dreamorosi: Similar to [here](https://github.com/aws-powertools/powertools-lambda-typescript/pull/1528#discussion_r1238421232), `void` is used when there's no return. This method always throws an error.

Looking at these two functions:
```ts
const functionA = (): void => {
  console.log('hi');
}

const functionB = (): undefined => {
  console.log('hi');

  return;
}
```

In the example above, `functionA` never returns anything, as far as TypeScript is concerned. The other one, `functionB` instead returns an `undefined` (empty = `undefined`).

In JS terms, the return is implicit, so in reality both return `undefined`, however this tells the TS compiler that the function is not expected to return anything at all.

Using `undefined`, tells the caller that there is a value, and this value is `undefined`, so I can do `const result = functionB();` and know that `result` here is undefined.

Conversely, if I did `const restulA = functionA();`, I would get an error because `functionA()` is not expected to return any value.

packages/parameters/src/types/DynamoDBProvider.ts
dreamorosi: That's correct!

2.3 should never happen at runtime, hence the `never`.

As a side note, I agree that these types are on the ugly side. Unfortunately at the moment I don't have a better way of expressing them.

packages/parameters/src/appconfig/AppConfigProvider.ts
am29d: ok I wasn't aware of that, so it means user can call `getMultiple` and nothing happens and there will be no error or indication, except method doc. Seems like there is no way to prevent the method call 😕

packages/parameters/src/base/transformValue.ts
dreamorosi: Updated in the last commit, explaining the flow. Good suggestion

am29d (approved):

PR #1540, opened 2023-06-22
fix(idempotency): pass lambda context remaining time to save inprogress

### LINKED ISSUE #1482: Bug: `idempotencyHandler` not using registered context when saving in progress
### Expected Behaviour

When using the `idempotentLambdaHandler`, `idempotentFunction` decorators, or `makeFunctionIdempotent` function wrapper the underlying idempotency handler should use the Lambda context to determine the remaining milliseconds in the current execution when saving a record as in-progress. This way, if an execution times out while processing it can be handled correctly later on.

### Current Behaviour

The context is never used in the `IdempotencyHandler` module and the `idempotentLambdaHandler` never calls `idempotencyConfig.registerLambdaContext(context)` - those are two separate but related issues.

As a result, even though the operation is made idempotent, the handler logs this warning:
> 2023-06-02T16:42:39.239Z	7bac60e0-3084-403c-9c69-be9f20d0e761	WARN	Could not determine remaining time left. Did you call registerLambdaContext on IdempotencyConfig?

### Code snippet

```typescript
import type { Context } from "aws-lambda";
import type { LambdaInterface } from "@aws-lambda-powertools/commons";
import { idempotentLambdaHandler } from "@aws-lambda-powertools/idempotency";
import { IdempotencyConfig } from "@aws-lambda-powertools/idempotency";

type EventRecord = {
  foo: string;
};

const logger = new Logger();

const dynamoDBPersistenceLayer = new DynamoDBPersistenceLayer({
  tableName:
    process.env.IDEMPOTENCY_TABLE_NAME || "lambda-powertools-playground-table",
});
const idempotencyConfig = new IdempotencyConfig({});

  @idempotentLambdaHandler({
    persistenceStore: dynamoDBPersistenceLayer,
  })
  public async handler(event: EventRecord, _context: Context): Promise<string> {
    logger.info("Got test event:", { event });
    return "Processing done: " + event.foo;
  }
}

const defaultLambda = new Lambda();
export const handler = defaultLambda.handler.bind(defaultLambda);
```

### Steps to Reproduce

1. Deploy the code
2. Run it
3. Observe the warning in the logs & the absence of the `in_progress_expiry_attr` value in the DynamoDB Table

### Possible Solution

_No response_

latest

18.x

### Packaging format used

npm

### Execution logs

_No response_

am29d:
In this PR we fix the issue to pass lambda context remaining execution time to save in progress. This only applies to function wrapper for now. We need to rework the entire decorator signature to extract the context, this will be handled in a separate PR. 

### Related issues, RFCs

**Issue number:** closes #1482 

## Checklist

### Breaking change checklist

---

### DIFF
--- a/packages/idempotency/src/idempotentDecorator.ts
+++ b/packages/idempotency/src/idempotentDecorator.ts
@@ -5,6 +5,16 @@ import {
 } from './types';
 import { IdempotencyHandler } from './IdempotencyHandler';
 import { IdempotencyConfig } from './IdempotencyConfig';
+import { Context } from 'aws-lambda';
+
+const isContext = (arg: unknown): arg is Context => {
+  return (
+    arg !== undefined &&
+    arg !== null &&
+    typeof arg === 'object' &&
+    'getRemainingTimeInMillis' in arg
+  );
+};
 
 /**
  * use this function to narrow the type of options between IdempotencyHandlerOptions and IdempotencyFunctionOptions
@@ -28,14 +38,20 @@ const idempotent = function (
     descriptor: PropertyDescriptor
   ) {
     const childFunction = descriptor.value;
-    descriptor.value = function (record: GenericTempRecord) {
+    descriptor.value = function (
+      record: GenericTempRecord,
+      ...args: unknown[]
+    ) {
       const functionPayloadtoBeHashed = isFunctionOption(options)
         ? record[(options as IdempotencyFunctionOptions).dataKeywordArgument]
         : record;
-
       const idempotencyConfig = options.config
         ? options.config
         : new IdempotencyConfig({});
+      const context = args[0];
+      if (isContext(context)) {
+        idempotencyConfig.registerLambdaContext(context);
+      }
       const idempotencyHandler = new IdempotencyHandler<GenericTempRecord>({
         functionToMakeIdempotent: childFunction,
         functionPayloadToBeHashed: functionPayloadtoBeHashed,
--- a/packages/idempotency/src/idempotentDecorator.ts
+++ b/packages/idempotency/src/idempotentDecorator.ts
@@ -32,6 +32,7 @@ const idempotent = function (
       const functionPayloadtoBeHashed = isFunctionOption(options)
         ? record[(options as IdempotencyFunctionOptions).dataKeywordArgument]
         : record;
+
       const idempotencyConfig = options.config
         ? options.config
         : new IdempotencyConfig({});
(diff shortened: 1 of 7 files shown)


packages/idempotency/tests/unit/makeFunctionIdempotent.test.ts
```diff
@@ -25,6 +27,12 @@ const mockGetRecord = jest
   .spyOn(BasePersistenceLayer.prototype, 'getRecord')
   .mockImplementation();
 
+const mockLambaContext: Context = {
```
dreamorosi, line 30: Nit: consider importing the dummy context [from here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/commons/src/samples/resources/contexts/hello-world.ts) like we do [in other tests](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/metrics/tests/unit/Metrics.test.ts#L6-L10).

packages/idempotency/src/idempotentDecorator.ts
```diff
@@ -32,6 +32,7 @@ const idempotent = function (
       const functionPayloadtoBeHashed = isFunctionOption(options)
         ? record[(options as IdempotencyFunctionOptions).dataKeywordArgument]
         : record;
+
```
dreamorosi, line 35: Even with the current signature (& the comment about future refactoring), can we not do any assertion when `idempotentLambdaHandler` is used on the second argument like we do for the fn wrapper?

Like, getting the second argument which might or might not be context ([example](https://github.com/am29d/aws-lambda-powertools-typescript/blob/a35c8ed7154aa834fd50f61dc171b012f0af4874/packages/idempotency/src/makeFunctionIdempotent.ts#L61)), and then doing an assertion on its content ([example](https://github.com/am29d/aws-lambda-powertools-typescript/blob/a35c8ed7154aa834fd50f61dc171b012f0af4874/packages/idempotency/src/makeFunctionIdempotent.ts#L68-L71)).

Is it worth it? Should we not do it?

packages/idempotency/tests/unit/makeFunctionIdempotent.test.ts
am29d: ah, I was looking for something similar. We need to move this thing to a better place, somewhere else.

packages/idempotency/src/idempotentDecorator.ts
am29d: Good point, adding args to check the context works. I have added a copy of `isContext`, it needs a better place. Exporting would mix wrapper and decorator imports. What do you think?

dreamorosi: It's fine for now, we should probably move it to `commons` together with the other type guards I'm adding in #1528.

Let's leave this for now.

dreamorosi (approved):
Thanks for fixing this!

PR #1567, opened 2023-06-29
chore(maintenance): add powertools to user-agent in SDK clients

### LINKED ISSUE #1533: Maintenance: append user agent to AWS SDK clients
### Summary

The AWS SDK v3 allows to add user agent headers to the requests made using its clients via its middleware stack.

### Why is this needed?

So that AWS can attribute requests made to its API when using Powertools.

Governance

### Solution

Expose a shared function that takes a client and adds a middleware to it that will attach Powertools-related user agents to the request. This shared function can then be used by other Powertools utilities that make requests to AWS APIs.

### Acknowledgment

### Future readers

am29d:

In this PR we introduce our custom middleware that can be added to any middleware stack of AWS SDK v3 clients. We decided to use middleware instead of `customUserAgent` options, because we might get SDK client from the user. Because `config` field of the client is readonly, we can't add change it, after the client has been initialised. 

Any feature using an SDK client can call like this:

```ts
const ddbClient = new DynamoDBClient(options);

addUserAgentMiddleware(ddbClient, 'my-feature);
```

this will modify the `user-agent` field by appending  `PT/my-feature/1.10.0 PTEnv/NA`  at the end of the value. 

I was about to add the new functionality to idempotency, but we will do it in a separate PR. Meanwhile, I have removed the singleton for the client in the `DynamoDBPersistenceLayer.ts`, similar how we use it parameters utility. 

### Related issues, RFCs

**Issue number:** closes #1533

## Checklist

### Breaking change checklist

---


packages/idempotency/src/persistence/DynamoDBPersistenceLayer.ts
```diff
@@ -230,14 +223,6 @@ class DynamoDBPersistenceLayer extends BasePersistenceLayer {
     );
   }
 
-  private getClient(): DynamoDBClient {
-    if (!this.client) {
-      this.client = new DynamoDBClient(this.clientConfig);
```
dreamorosi, line 235: The idea here was to initialize the client lazily.

Couldn't we have added the middleware at this point?

```ts
if (!this.client) {
  this.client = new DynamoDBClient(this.clientConfig);
  this.client.middleware.add(...);
}
```

and do the same in the `constructor` with the one that they pass us?

packages/commons/src/version.ts
```diff
@@ -0,0 +1,2 @@
+// this file is auto generated, do not modify
+module.exports = { version: '1.10.0' };
```
dreamorosi, line 2:
```suggestion
export const PT_VERSION = '1.10.0';
```

dreamorosi: This way we are not mixing exports type, the one with `module.exports` is CJS only.

Also if you export it with the final name, you don't have to alias it with `as PT_VERSION` when you import it, same for the object vs `string`

packages/commons/src/middleware/userAgentMiddleware.ts
```diff
@@ -0,0 +1,38 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-nocheck
```
dreamorosi, line 2: I don't feel very comfortable with this.

The easy way would be to just import `@aws-sdk/types` which has all the types we need. However I'm not sure we should be adding it to the bundle.

After the release I'll give it a try and see if I can do some simplified types to avoid the `@ts-nocheck` to the entire file

packages/commons/src/index.ts
```diff
@@ -5,3 +5,4 @@ export * as ContextExamples from './samples/resources/contexts';
 export * as Events from './samples/resources/events';
 export * from './types/middy';
 export * from './types/utils';
+export * from './middleware';
```
dreamorosi, line 8: This was excluded purposefully. 

The `middleware` folder is specific to Middy-related stuff, which might/might not bring some dependencies.

We don't want to put this in the default export and we want consumers to be explicitly importing the path `@aws-lambda-powertools/commons/lib/middleware` (for now, `@aws-lambda-powertools/commons/middleware` in v2).

dreamorosi: They are both middleware, but of two different kinds, I would consider creating another subfolder called `user-agent` or something that points to the AWS SDK instead.

packages/commons/tests/unit/userAgentMiddleware.test.ts
```diff
@@ -0,0 +1,56 @@
...
+const logHeadersMiddleware = (next, _context) => async (args) => {
+  console.log(args.request.headers);
+
+  return await next(args);
+};
+
+describe('Function: addUserAgentMiddleware', () => {
+  it('adds powertools user agent to request header at the end', async () => {
+    const lambdaClient = new LambdaClient({
+      logger: console,
```
dreamorosi, line 20: I think this line is not needed anymore

packages/idempotency/src/persistence/DynamoDBPersistenceLayer.ts
am29d: I wanted to have one place in constructor to do that. Is the performance noticeable?

packages/commons/src/version.ts
am29d: agree

packages/commons/src/middleware/userAgentMiddleware.ts
am29d: I did not manage to add something like `T extends Client`, thus the shortcut. Curious to see how it can be fixed and fully agree that it's not the best way to handle it.

packages/idempotency/src/persistence/DynamoDBPersistenceLayer.ts
dreamorosi: I think it's not, let's go with your version.

Let's streamline the constructor altogether. Basically let's do the same as the Parameters util & remove the `private client!` assertion, since we know it'll always be set before the constructor returns.

packages/commons/src/middleware/userAgentMiddleware.ts
dreamorosi: I think worst case scenario we can add `// @ts-ignore` to specific lines if needed, but keep the rest of the file typed. 

Let's see, it might not be worth creating custom types to avoid issues in case the SDK changes the interfaces. What do you think?

packages/idempotency/src/persistence/DynamoDBPersistenceLayer.ts
am29d: Alright, I'll revert to the previous version with lazy init.

packages/commons/src/index.ts
am29d: Good point, I will put it on the top level

packages/commons/src/middleware/userAgentMiddleware.ts
am29d: I have added explicit `// @ts-ignore` for two lines instead of entire file.

Do you mean to create a union type with all the clients we currently support? Something like:

`type SupportedSdkClients = LambdaClient | DynamoDBClient | S3Client;` 

It would imply that we use the `@aws-sdk/types` package, right? Or is there other way to narrow the types of the clients passed to the middleware?

am29d:
* Fixed tests
* added try/catch so we don't block user code if adding user agent fails
* added parametrised tests for each client we use in our utilities

packages/commons/tests/unit/userAgentMiddleware.test.ts
```diff
@@ -0,0 +1,109 @@
...
+  region: 'us-east-1',
+  endpoint: 'http://localhost:9001',
+  credentials: {
+    accessKeyId: 'test',
+    secretAccessKey: 'test',
+    sessionToken: 'test',
+  },
+};
+
+describe('Given a client of instance: ', () => {
```
dreamorosi, line 23: Very clean, I like it

dreamorosi (approved):
Thanks for working on this.

I have run the integration tests on this branch and [they are passing](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/5447643933).

PR #1577, opened 2023-07-04
chore(parameters): apply UA to AWS SDK clients used by Parameters

### LINKED ISSUE #1537: Maintenance: apply UA to AWS SDK clients used by Parameters
### Summary

In #1533 we have discussed the possibility to add user agents to requests made by AWS SDK clients that are passed to or instantiated by Powertools. As part of this issue we should use the shared function introduced in the PR linked to that issue and use it in the Parameters utility.

### Why is this needed?

So that AWS can attribute requests made to its API when using Powertools. This helps support the project and drive internal support.

Parameters

### Solution

_No response_

### Acknowledgment

### Future readers

am29d:
Add user agent to clients user by parameters utility.

### Related issues, RFCs

**Issue number:** closes #1537 

## Checklist

### Breaking change checklist

---

### DIFF
--- a/packages/parameters/tests/unit/AppConfigProvider.test.ts
+++ b/packages/parameters/tests/unit/AppConfigProvider.test.ts
@@ -8,11 +8,12 @@ import { ExpirableValue } from '../../src/base/ExpirableValue';
 import { AppConfigProviderOptions } from '../../src/types/AppConfigProvider';
 import {
   AppConfigDataClient,
-  StartConfigurationSessionCommand,
   GetLatestConfigurationCommand,
+  StartConfigurationSessionCommand,
 } from '@aws-sdk/client-appconfigdata';
 import { Uint8ArrayBlobAdapter } from '@aws-sdk/util-stream';
 import { mockClient } from 'aws-sdk-client-mock';
+import * as UserAgentMiddleware from '@aws-lambda-powertools/commons/lib/userAgentMiddleware';
 import 'aws-sdk-client-mock-jest';
 
 describe('Class: AppConfigProvider', () => {
@@ -34,6 +35,11 @@ describe('Class: AppConfigProvider', () => {
         environment: 'MyAppProdEnv',
       };
 
+      const userAgentSpy = jest.spyOn(
+        UserAgentMiddleware,
+        'addUserAgentMiddleware'
+      );
+
       // Act
       const provider = new AppConfigProvider(options);
 
@@ -43,6 +49,8 @@ describe('Class: AppConfigProvider', () => {
           serviceId: 'AppConfigData',
         })
       );
+
+      expect(userAgentSpy).toHaveBeenCalled();
     });
 
     test('when the user provides a client config in the options, the class instantiates a new client with client config options', async () => {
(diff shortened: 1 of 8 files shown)


am29d:
e2e tests ✅https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/5454144683

am29d:
Good catch with duplicate attachment. 

> Does this attach the middleware twice? If so, does this result in the UA being appended twice? Same applies to Idempotency and all other providers.

It fails to add two middlewares with identical name, we catch the error and continue with user code execution. 

We could extend the header for each utility, similar how .NET does. For example `PT/parameters/ PT/idempotency`, but it needs more consideration.

dreamorosi:
> It fails to add two middlewares with identical name, we catch the error and continue with user code execution.

It's good that we don't halt execution, but we still log a warning that could be misleading. I'm going to open an issue to remember working on this asap.

packages/parameters/tests/unit/AppConfigProvider.test.ts
```diff
@@ -43,6 +49,8 @@ describe('Class: AppConfigProvider', () => {
           serviceId: 'AppConfigData',
         })
       );
+
+      expect(userAgentSpy).toHaveBeenCalled();
```
dreamorosi, line 53: The `client` inside the provider is public and can be accessed via `provider.client`. This means you should be able to test like you did in the `commons` module by doing:

```ts
expect(provider.client.middlewareStack.identify()).toContain(
  'addPowertoolsToUserAgent: POWERTOOLS,USER_AGENT'
);
```

I think that would be preferable to changing the import (`* as ..`) and selecting the module in the spy.

dreamorosi (changes requested):
I have left a comment on the testing to improve the assertion on the newly added middleware.

Also I have opened an issue #1578 to continue the discussion above about multiple `addUserAgentMiddleware` calls.

packages/parameters/tests/unit/AppConfigProvider.test.ts
am29d: > I think that would be preferable to changing the import (* as ..) and selecting the module in the spy. 

Can you elaborate? Curious to hear about the benefits. 

I want to avoid to test the internals of the `commons` module and the dependencies behaviour. Imho, we should only check the call has happened downstream, otherwise we replicate the assertion logic in different places which makes it harder to change in the future. For instance, if we rename the middleware from `addPowertoolsUserAgent` to something else, we'd need to revisit every utility that called this method.

dreamorosi: Based on the way that the Jest spy is written, if we rename `addPowertoolsUserAgent` to something else the test would also break:

```ts
const userAgentSpy = jest.spyOn(
  UserAgentMiddleware,
  'addUserAgentMiddleware' // if function name changes, this spy also breaks
);
```

However I see your point and agree. I tend to avoid setting spies on imports but I can't think of a better way of doing it without reworking the constructor (which si not worth it), so let's leave it this way!

dreamorosi (approved):

am29d:
In case of #1578 we agreed to attach only one user agent middleware for a client. If the client is used in multiple utilities, we check if a middleware is present and skip the second call, first come first serve. 

We will revisit the strategy at later points to align with other languages.

PR #1583, opened 2023-07-05
chore(maintenance): avoid attaching two middlewares to ua

### LINKED ISSUE #1578: Maintenance: avoid attaching two middlewares when appending UA to AWS SDK
### Summary

In #1577, #1575, and #1567 we have added logic that appends a user agent to AWS SDK requests by leveraging the middleware stack of the AWS SDK v3. The current setup however does not take in account the fact that customers might pass us the same client more than once (i.e. create a client > instrument it with Tracer > pass it to Parameters / Idempotency).

### Why is this needed?

The current implementation will log a warning that informs the user that Powertools was not able to apply the middleware, this could be misleading.

Other

### Solution

We should make sure that this case is handled correctly by either: 1/ making the operation idempotent (like the [X-Ray SDK itself does](https://github.com/aws/aws-xray-sdk-node/blob/449e3b2e5552b640df341a8b17c124efda5086d3/packages/core/lib/patchers/aws3_p.ts#L201-L204) when instrumenting), 2/ appending the second user agent (like [suggested here](https://github.com/aws-powertools/powertools-lambda-typescript/pull/1577#issuecomment-1620499258)), 3/ converting operation to a no-op if the middleware is already there.

### Acknowledgment

### Future readers

am29d:

This PR adds a check before adding user agent middleware, to avoid attaching two or more middlewares and fire a warning in the catch block. I have also refactored the tests, to make it a bit more readable. 

### Related issues, RFCs

**Issue number:** closes #1578 

## Checklist

### Breaking change checklist

---

### DIFF
--- a/packages/commons/src/userAgentMiddleware.ts
+++ b/packages/commons/src/userAgentMiddleware.ts
@@ -33,6 +33,13 @@ const customUserAgentMiddleware = (feature: string) => {
 // @ts-ignore
 const addUserAgentMiddleware = (client, feature: string): void => {
   try {
+    if (
+      client.middlewareStack
+        .identify()
+        .includes('addPowertoolsToUserAgent: POWERTOOLS,USER_AGENT')
+    ) {
+      return;
+    }
     client.middlewareStack.addRelativeTo(
       customUserAgentMiddleware(feature),
       middlewareOptions
--- a/packages/commons/tests/unit/userAgentMiddleware.test.ts
+++ b/packages/commons/tests/unit/userAgentMiddleware.test.ts
@@ -3,12 +3,26 @@ import { InvokeCommand, LambdaClient } from '@aws-sdk/client-lambda';
 import { DynamoDBClient } from '@aws-sdk/client-dynamodb';
 import { ScanCommand } from '@aws-sdk/lib-dynamodb';
 import { GetParameterCommand, SSMClient } from '@aws-sdk/client-ssm';
-import { version as PT_VERSION } from '../../package.json';
+import { PT_VERSION } from '../../src/version';
 import { AppConfigDataClient } from '@aws-sdk/client-appconfigdata';
 import {
   GetSecretValueCommand,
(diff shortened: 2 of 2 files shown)


dreamorosi (approved):
Thank you for fixing this!

packages/commons/src/userAgentMiddleware.ts
```diff
@@ -33,6 +33,13 @@ const customUserAgentMiddleware = (feature: string) => {
 // @ts-ignore
```
sthulb, line 33: not sure I like ignoring things in the type system

dreamorosi: Acknowledged and agree.

In this instance we decided to use an `any` as a type for the `client` because we didn't want to bring in the `@aws-sdk/types` package into the utility. This code resides in the `commons` utility, which is pulled in by all the other Powertools utilities.

Given that there's all the core utilities don't use any AWS SDK, using the `@aws-sdk/types` in this module would require customers to install it as a developer dependency when they develop or build their code using Powertools.
This is a problem because when building with TypeScript, the type-checking engine will block the compilation is the dependency is missing.

To avoid this, we decided to go for an `any` type (which is discouraged in our codebase - hence the `@ts-ignore`) and also put [unit tests](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/commons/tests/unit/userAgentMiddleware.test.ts#L89-L114).

We have considered writing our own minimal `Client` type, and the option is still in the cards, but have decided to not include it for now given the limited scope of the feature (i.e. getting a client from a 3rd party library and calling a method on it) and the fact that we implemented the function in a way that even if you pass a random object (aka not a client) it won't throw an error but just log a warning.

I am still strongly inclined to review this type as soon as we have bandwidth and remove the ignore.

sthulb: I think my preference would be to write our own types def.

Issue #1586, opened 2023-07-07
Feature request: implement `BasePartialBatchProcessor ` class


dreamorosi:
### Use Case

The Python version of Powertools for AWS features a classes called `BasePartialBatchProcessor` and `BatchProcessor`, and its respective interfaces. These class is used to provide a Batch Processor utility in Powertools.

### Solution/User Experience

Refer to the implementation [found here](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/utilities/batch/base.py#L222), and specifically to the `BasePartialBatchProcessor` class and `BasePartialProcessor` interface.

Refer to the implementation [found here](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/utilities/batch/base.py#L366), and specifically to the `BatchProcessor` class.

### Acceptance criteria

- The class covers all the features found in the reference implementation
- The class and its type are exported
- Unit tests are implemented and have 100% coverage
- Classes and methods are documented via docstrings, following project's conventions as much as possible

### Notes

The reference implementation also implements an `AsyncBatchProcessor` and both classes have `_process_record` and `_async_process_record` methods.

I'd like to dedicate some time investigating whether we can merge these two classes & methods into a single one.

The ideal DX would have the customer simply importing `BatchProcessor` and passing it a function reference when calling `process_partial_response`. The utility should be smart enough to understand whether the function passed is synchronous or asynchronous and handle it accordingly (i.e. if it's async, then `await` the record handler).

If this is not possible, or too complex / unreliable, or the resulting types are confusing or overly broad, then let's stick with two separate entities.

erikayao93:
Just opened PR for this issue!

PR #1588, opened 2023-07-07
feat(batch): Implementation of base batch processing classes

erikayao93:

- Added `BasePartialProcessor` abstract class with basic processing framework
- Added `BasePartialBatchProcessor` abstract class with processing framework specific to SQS, Kinesis, and DynamoDB
- Added `BatchProcessor` class with processing implementation
- Added unit tests with 100% coverage on new classes

### Related issues, RFCs

**Issue number: #1586** 

## Checklist

### Breaking change checklist

---


dreamorosi:
@mergify rebase

packages/batch/src/BatchProcessor.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * Process native partial responses from SQS, Kinesis Data Streams, and DynamoDB
+ */
+import {
+  BasePartialBatchProcessor,
+  BaseRecord,
+  FailureResponse,
+  SuccessResponse,
+} from '.';
```
dreamorosi, line 9: While at this stage using this type of import is harmless, importing from the default `index.ts` could lead to unintended side effects in the future as the codebase grows.

When it comes to internal files referencing themselves, we prefer to scope imports as much as possible, for example:
```ts
import { BasePartialBatchProcessor } from './BasePartialBatchProcessor';
import type { BaseRecord, FailureResponse, SuccessResponse } from './types';
```

By doing this the file imports only and exactly the entities that are needed for it to work. While this might result in us having to change the imports if we change the folder structure, it'll allow us to be intentional about what we are importing.

Importing from the main file (or in general from files that themselves contain multiple exports) could cause unintended side effects that are hard to track & reproduce, for example imagine we have two classes: `ClassA` and `ClassB`. Both of them are re-exported in `index.ts`:
```ts
export * from './ClassA';
export * from './ClassB';
```

Imagine we now have a third file (`otherModule.ts`) where we need to import `ClassA` only. By using the imports from `index.ts` JS will be requiring and importing both since they are both present in `index.ts`. Generally this is fine, however if `ClassB` is expensive to instantiate we would be putting unnecessary overhead. Additionally, depending on the bundling strategy used by our customers, importing unnecessary modules could make [tree-shaking](https://webpack.js.org/guides/tree-shaking/) more challenging or impossible, further increasing the unnecessary overhead.

packages/batch/src/BatchProcessor.ts
```diff
@@ -0,0 +1,31 @@
...
+   */
+  public async processRecord(
+    record: BaseRecord
+  ): Promise<SuccessResponse | FailureResponse> {
+    try {
+      const data = this.toBatchType(record, this.eventType);
+      const result = await this.handler(data);
+
+      return this.successHandler(record, result);
+    } catch (e) {
```
dreamorosi, line 25:
```suggestion
    } catch (error) {
```

dreamorosi: Please consider applying thsi for all occurrences of `catch (e)`.

We discourage using single-char variables to encourage using meaningful names for all variables.

Even though in this case it's pretty clear that `e` stands for error, depending on the proficiency/level/experience of the reader this meaning could get lost. Writing the full name disambiguates what `e` is.

packages/batch/src/types.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * Types for batch processing utility
+ */
+import { DynamoDBRecord, KinesisStreamRecord, SQSRecord } from 'aws-lambda';
+
+// types from base.py
+type EventSourceDataClassTypes =
```
dreamorosi, line 7: Let's rename this to `EventSourceTypes`. 

The "data class" nomenclature is specific to Python and doesn't add much to JS/TS developers.

packages/batch/src/types.ts
```diff
@@ -0,0 +1,31 @@
...
+  | KinesisStreamRecord
+  | DynamoDBRecord;
+
+type RecordValue = unknown;
+type BaseRecord = { [key: string]: RecordValue } | EventSourceDataClassTypes;
+
+type ResultType = unknown;
+type SuccessResponse = [string, ResultType, EventSourceDataClassTypes];
+
+type FailureResponse = [string, string, EventSourceDataClassTypes];
```
dreamorosi, line 18: Similar to `SuccessResponse`, can we narrow down the first `string` to `fail`?

packages/batch/src/BasePartialBatchProcessor.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    return failures;
+  }
+
+  /**
+   * Determines whether all records in a batch failed to process
+   * @returns true if all records resulted in exception results
+   */
+  public entireBatchFailed(): boolean {
+    return this.exceptions.length == this.records.length;
```
dreamorosi, line 111:
```suggestion
return this.exceptions.length === this.records.length;
```

packages/batch/tests/helpers/factories.ts
```diff
@@ -0,0 +1,69 @@
...
+    messageAttributes: {},
+    md5OfBody: 'e4e68fb7bd0e697a0ae8f1bb342846b3',
+    eventSource: 'aws:sqs',
+    eventSourceARN: 'arn:aws:sqs:us-east-2:123456789012:my-queue',
+    awsRegion: 'us-east-1',
+  };
+};
+
+export const kinesisEventFactory = (body: string): KinesisStreamRecord => {
+  let seq = '';
```
dreamorosi, line 24: What is `seq` here?

packages/batch/tests/helpers/factories.ts
```diff
@@ -0,0 +1,69 @@
...
+  for (let i = 0; i < 52; i++) {
+    seq = seq + Math.floor(Math.random() * 10);
```
dreamorosi, line 26: If possible, let's replace `Math.random()` with `randomInt()` from [`node:crypto`](https://nodejs.org/api/crypto.html#cryptorandomintmin-max-callback).

packages/batch/tests/helpers/factories.ts
```diff
@@ -0,0 +1,69 @@
...
+      StreamViewType: 'NEW_AND_OLD_IMAGES',
+      SequenceNumber: seq,
+      SizeBytes: 26,
+    },
+    awsRegion: 'us-west-2',
+    eventName: 'INSERT',
+    eventSourceARN: 'eventsource_arn',
+    eventSource: 'aws:dynamodb',
+  };
+};
```
dreamorosi, line 69: Let's move the exports at the bottom of the file to follow convention.

packages/batch/tests/unit/BatchProcessor.test.ts
```diff
@@ -0,0 +1,477 @@
...
+    test('Batch processing SQS records with no failures', async () => {
+      // Prepare
+      const firstRecord = sqsEventFactory('success');
+      const secondRecord = sqsEventFactory('success');
+      const records = [firstRecord, secondRecord];
+      const processor = new BatchProcessor(EventType.SQS);
+
+      // Act
+      processor.register(records, sqsRecordHandler);
+      const processedMessages = await processor.process();
```
dreamorosi, line 45: If `sqsRecordHandler` is synchronous, do we need `await`?

packages/batch/tests/unit/BatchProcessor.test.ts
```diff
@@ -0,0 +1,477 @@
...
+
+      // Assess
+      expect(processedMessages).toStrictEqual([
+        ['success', (firstRecord as SQSRecord).body, firstRecord],
```
dreamorosi, line 49: Since the return type of the factory is already `SQSRecord` perhaps we don't need a typecast here.

dreamorosi (commented):
Thank you for the work on this PR, I have left a number of comments but 9/10 are minor ones and at times related to convention/styling.

The implementation is going in the right direction and I look forward to get it merged!

packages/batch/src/errors.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * Batch processing exceptions
+ */
+
+class BaseBatchProcessingError extends Error {
```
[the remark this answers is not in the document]
erikayao93: I followed the implementation pattern from the Python repo, which has the error implemented like this. I assume it is to allow the implementation to be more extensible, as you suggested, so I figured I'd keep the same pattern in the case that the Python side decides on an extension.

packages/batch/tests/unit/BatchProcessor.test.ts
erikayao93: The `process()` function is asynchronous so that it can handle the asynchronous functions, I'm not sure if there's an alternative for that.

dreamorosi: Let's take these two example handlers:
```ts
const mySyncHandler = (record) => {
  // do something
};

const myAsyncHandler = async (record) => {
  // do something async
};
```

Depending on which one of the two I pass to `processor.register()` we should have to `await` the `processor.process()` function or not.

For instance:
```ts
processor.register(records, mySyncHandler);
const processedMessages = processor.process(); // no await, since mySyncHandler is synchronous

processor.register(records, myAsyncHandler);
const processedMessagesAsync = await processor.process() // using await since myAsyncHandler is async (aka it returns a promise)
```

We should aim at supporting this type of behavior, which might require to be creative not only with the implementation but also with the return types (i.e. return type of `process()` method depend on type of the second arg of `register()` method).

I'm pointing this out at this stage to set direction, however if we don't have a solution right away I am okay with putting a pin on it and opening an issue. This way we can address this later on.

erikayao93: I'll have a look today to see if I can figure anything out and aim for a revision by EOD for me for you to be able to review tomorrow. I'll leave a comment here if I manage to implement something that works.

erikayao93: From what I've looked at, it seems like our options are either to create a separate set of functions (like what Python has) or to always await (what we have now).

Theoretically, if we wanted to keep it as a single method, we might be able to do something like [this](https://stackoverflow.com/questions/72202437/refactor-both-sync-and-async-method-implementations), but we'd have to create another level on top of `process()` that returns a list or a promised list. Then, we'd be able to have an additional `AsyncBatchProcessor` with an async `process()` implementation that generates that list with a `Promise` wrapper properly.

dreamorosi: As discussed during our meeting for now we will leave the implementation as-is and we will revisit the topic once the utility is in a more advanced state. I'll open an issue and link it here shortly.

dreamorosi: Issue created #1603, resolving the conversation.

Issue #1607, opened 2023-07-13
Feature request: support accessing Lambda context in record handler


dreamorosi:
### Use Case

In the Python version of Powertools for AWS, when using Batch Processing and specifying a record handler customers can optionally configure the utility to pass the Lambda context to the record handler. This is useful in those cases in which the record handler needs to be aware of info from the context or needs to be able to [retrieve the amount of time](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-context.html) left before the function times out.

### Solution/User Experience

Refer to the implementation [described here](https://docs.powertools.aws.dev/lambda/python/latest/utilities/batch/#accessing-lambda-context).

In Python the feature is possible thanks to the existence of keyword arguments, customers can simply specify an argument named `lambda_context` in the record handler function and the Batch Processing utility will propagate the context from the function to it. In JS there's no notion of keyword arguments so we'll have to find an alternative solution. Proposals/ideas are welcome.

### Acceptance criteria

- Unit tests are implemented with 100% coverage
- Added code is documented via docstrings, following project's conventions as much as possible

erikayao93:
Picking up implementation on this!

PR #1609, opened 2023-07-14
feat(batch): Support for Lambda context access in batch processing

erikayao93:
- Added BatchProcessingOptions type to hold context and future configuration options
- Added optional parameters to carry `options` parameter
- Added unit tests to check use of `options` parameter

### Related issues, RFCs

**Issue number: #1607** 

## Checklist

### Breaking change checklist

---

### DIFF
--- a/packages/batch/src/BatchProcessor.ts
+++ b/packages/batch/src/BatchProcessor.ts
@@ -19,7 +19,13 @@ class BatchProcessor extends BasePartialBatchProcessor {
   ): Promise<SuccessResponse | FailureResponse> {
     try {
       const data = this.toBatchType(record, this.eventType);
-      const result = await this.handler(data);
+
+      let result: unknown;
+      if (this.options) {
+        result = await this.handler(data, this.options);
+      } else {
+        result = await this.handler(data);
+      }
 
       return this.successHandler(record, result);
     } catch (e) {
--- a/packages/batch/src/BatchProcessor.ts
+++ b/packages/batch/src/BatchProcessor.ts
@@ -20,12 +20,7 @@ class BatchProcessor extends BasePartialBatchProcessor {
     try {
       const data = this.toBatchType(record, this.eventType);
 
-      let result: unknown;
-      if (this.options) {
-        result = await this.handler(data, this.options);
-      } else {
-        result = await this.handler(data);
-      }
+      const result = await this.handler(data, this.options);
 
       return this.successHandler(record, result);
     } catch (e) {
--- a/packages/batch/src/processPartialResponse.ts
+++ b/packages/batch/src/processPartialResponse.ts
@@ -28,13 +28,7 @@ const processPartialResponse = async (
     );
   }
 
-  const records = event['Records'];
-
-  if (options) {
-    processor.register(records, recordHandler, options);
-  } else {
-    processor.register(records, recordHandler);
-  }
+  processor.register(event.Records, recordHandler, options);
 
   await processor.process();
 
--- a/packages/batch/src/processPartialResponse.ts
+++ b/packages/batch/src/processPartialResponse.ts
@@ -1,6 +1,7 @@
 import {
   BasePartialBatchProcessor,
   BaseRecord,
+  BatchProcessingOptions,
   EventType,
   PartialItemFailureResponse,
 } from '.';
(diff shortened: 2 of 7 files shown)


packages/batch/src/processPartialResponse.ts
```diff
@@ -28,7 +30,12 @@ const processPartialResponse = async (
 
   const records = event['Records'];
 
-  processor.register(records, recordHandler);
+  if (options) {
+    processor.register(records, recordHandler, options);
```
am29d, line 34: We can safely pass the options to the processor register method because the check on the options is already there, so the entire `if-block` can be simplified to:

`processor.register(event.Records, recordHandler, options);`

packages/batch/src/BatchProcessor.ts
```diff
@@ -19,7 +19,13 @@ class BatchProcessor extends BasePartialBatchProcessor {
...
     try {
       const data = this.toBatchType(record, this.eventType);
-      const result = await this.handler(data);
+
+      let result: unknown;
+      if (this.options) {
+        result = await this.handler(data, this.options);
+      } else {
+        result = await this.handler(data);
+      }
```
am29d, line 28: I am curious, why can't we pass undefined options to the handler? The handler provided by the user have to check the options before using, right? Maybe I am missing something. 

In case there needs to be an if-branch based on the options, I'd suggest to simplify to a one line:

```const result = this.options? this.handler(data, this.options) : this.handler(data);```

File: packages/batch/tests/unit/BatchProcessor.test.ts [hunk 3]
```diff
@@ -418,4 +426,42 @@ describe('Class: BatchProcessor', () => {
...
+
+    test('Batch processing when context is provided and handler does not accept', async () => {
+      // Prepare
+      const firstRecord = sqsRecordFactory('success');
+      const secondRecord = sqsRecordFactory('success');
+      const records = [firstRecord, secondRecord];
+      const processor = new BatchProcessor(EventType.SQS);
+
+      // Act
+      processor.register(records, sqsRecordHandler, options);
```
am29d, line 457: Should we call `handlerwithContext` here? Is the test designed to pass options even though we don't expect it to be used?

packages/batch/tests/unit/BatchProcessor.test.ts
```diff
@@ -418,4 +426,42 @@ describe('Class: BatchProcessor', () => {
       );
     });
   });
+
+  describe('Batch processing with Lambda context', () => {
```
am29d, line 430: We need addiotnal tests for malformed context, because we cant be certain what the user will pass. I suggest to have tests with empty and malformed options, you can cast it to `as Context`. I'd suggest to have an error message that outputs the context, so the user can inspect it. and copare what was passed to the processor. What do you think?

packages/batch/tests/unit/BatchProcessor.test.ts
erikayao93: This test is to check that if users register with context, even if the handler doesn't accept context it will still run. I think TS/JS is built so that the handler (without options parameter defined) will still work even if the processor tries to pass options into it, but I wasn't aware of this so I checked the behavior. This resolves the last conversation as well, so I'll make the change to simplify the logic, but I think the test here is fine?

packages/batch/tests/unit/BatchProcessor.test.ts
erikayao93: I suppose I may not have full understanding of `Context`, but the batch processing classes are merely passing the `options` passed by a user around, until it can get to the handler on each invocation. But since the handler is user defined, we may not necessarily be able to check when processing is failing directly because of malformed `Context`. I can try adding in a check on required fields and do more research on that side though.

am29d: Ah, sorry I have poorly phrased my question.

You are absolutely right, we can't asses the core issue with malformed `Context`. 

I want to capture the situation in a test, when user passes the `Context` to the batch processor, but the `Context` is malformed. It can be the user made a mistake, or something happened to the `Context` in between, we don't know. But if this happens, we want to provide a good error message that at shows what was inside the `Context`. This will help the user to troubleshoot the problem further.

packages/batch/tests/unit/BatchProcessor.test.ts
am29d: Yep, makes sense, we can keep it.

packages/batch/tests/unit/BatchProcessor.test.ts
am29d: Exactly, I think it'd be an urge to output and debug the problem. By printing error with the context we can help them to recover.

am29d:
Great work @erikayao93 ! Thanks a lot for this contribution and bringing batch utility closer to release 🙏 .

am29d (approved):

Issue #1603, opened 2023-07-11
Maintenance: revisit async/sync processing for batch utility


dreamorosi:
### Summary

The current implementation of the Batch Processing utility allows you to register an handler function that is executed for each record in a batch. This function can be both synchronous and asynchronous. In order to allow for async functions the current `process()` method has been implemented as `async` and always returns a `Promise`.

This requires the caller to have to `await` the result and handle the promise. When the caller passes a synchronous handler this might result counterintuitive, even though the addition of a `Promise` technically is a no-op.

This discussion arose during the review of the first PR for the utility ([here](https://github.com/aws-powertools/powertools-lambda-typescript/pull/1588#discussion_r1258374579)), however at that time it was too early to make a decision and so we have decided to postpone the topic until later in the implementation.

### Why is this needed?

So that we can review whether the current implementation is appropriate and decide if we need to change the types, add a new methods to distinguish between sync/async processing, or split the classes entirely. 

### Which area does this relate to?

Other

### Acknowledgment

- [X] This request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda-typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

PR #1616, opened 2023-07-16
feat(batch): add async processor

dreamorosi:

This PR introduces a new class and a new high-level function to the Batch Processing utility that allows to process a batch using asynchronous functions. The new class is called `AsyncBatchProcessor` and uses most of the same logic as the regular `BatchProcessor`, except that it process the records using an asynchronous handler and it uses `Promise.all` to run all the promises in parallel. The new companion high-level function, called `asyncProcessPartialResponse()` follows the same logic as the `processPartialResponse()` one in that it wraps the class and handles the orchestration before returning any failures.

### Related issues, RFCs

**Issue number:**  #1603

## Checklist

### Breaking change checklist

---

### DIFF
--- a/packages/batch/src/BasePartialProcessor.ts
+++ b/packages/batch/src/BasePartialProcessor.ts
@@ -1,15 +1,15 @@
-/**
- * Abstract class for batch processors
- */
-import {
+import type {
   BaseRecord,
   BatchProcessingOptions,
   EventSourceDataClassTypes,
   FailureResponse,
   ResultType,
   SuccessResponse,
-} from '.';
+} from './types';
 
+/**
+ * Abstract class for batch processors.
+ */
 abstract class BasePartialProcessor {
   public exceptions: Error[];
 
@@ -35,6 +35,15 @@ abstract class BasePartialProcessor {
   }
 
   /**
+   * Process a record with an asyncronous handler
+   *
+   * @param record Record to be processed
+   */
+  public abstract asyncProcessRecord(
+    record: BaseRecord
+  ): Promise<SuccessResponse | FailureResponse>;
+
+  /**
    * Clean class instance after processing
    */
   public abstract clean(): void;
@@ -50,7 +59,6 @@ abstract class BasePartialProcessor {
     exception: Error
   ): FailureResponse {
     const entry: FailureResponse = ['fail', exception.message, record];
-    // console.debug('Record processing exception: ' + exception.message);
     this.exceptions.push(exception);
     this.failureMessages.push(record);
 
@@ -66,13 +74,45 @@ abstract class BasePartialProcessor {
    * Call instance's handler for each record
    * @returns List of processed records
    */
-  public async process(): Promise<(SuccessResponse | FailureResponse)[]> {
+  public process(): (SuccessResponse | FailureResponse)[] {
+    /**
+     * If this is an async processor, user should have called processAsync instead,
+     * so we call the method early to throw the error early thus failing fast.
+     */
+    if (this.constructor.name === 'AsyncBatchProcessor') {
+      this.processRecord(this.records[0]);
+    }
     this.prepare();
 
     const processedRecords: (SuccessResponse | FailureResponse)[] = [];
     for (const record of this.records) {
-      processedRecords.push(await this.processRecord(record));
+      processedRecords.push(this.processRecord(record));
     }
+
+    this.clean();
+
+    return processedRecords;
+  }
+
+  /**
+   * Call instance's handler for each record
+   * @returns List of processed records
+   */
+  public async processAsync(): Promise<(SuccessResponse | FailureResponse)[]> {
+    /**
+     * If this is an sync processor, user should have called process instead,
+     * so we call the method early to throw the error early thus failing fast.
+     */
+    if (this.constructor.name === 'BatchProcessor') {
+      await this.asyncProcessRecord(this.records[0]);
+    }
+    this.prepare();
+
+    const processingPromises: Promise<SuccessResponse | FailureResponse>[] =
+      this.records.map((record) => this.asyncProcessRecord(record));
+
+    const processedRecords: (SuccessResponse | FailureResponse)[] =
+      await Promise.all(processingPromises);
 
     this.clean();
 
@@ -85,7 +125,7 @@ abstract class BasePartialProcessor {
    */
   public abstract processRecord(
     record: BaseRecord
-  ): Promise<SuccessResponse | FailureResponse>;
+  ): SuccessResponse | FailureResponse;
 
   /**
    * Set class instance attributes before execution
--- a/packages/batch/src/constants.ts
+++ b/packages/batch/src/constants.ts
@@ -1,8 +1,8 @@
-/**
- * Constants for batch processor classes
- */
 import { DynamoDBRecord, KinesisStreamRecord, SQSRecord } from 'aws-lambda';
-import type { PartialItemFailureResponse, EventSourceDataClassTypes } from '.';
+import type {
+  PartialItemFailureResponse,
+  EventSourceDataClassTypes,
+} from './types';
 
 const EventType = {
   SQS: 'SQS',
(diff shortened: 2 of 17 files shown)


packages/batch/src/BasePartialProcessor.ts
```diff
@@ -66,13 +74,45 @@ abstract class BasePartialProcessor {
...
+    this.clean();
+
+    return processedRecords;
+  }
+
+  /**
+   * Call instance's handler for each record
+   * @returns List of processed records
+   */
+  public async processAsync(): Promise<(SuccessResponse | FailureResponse)[]> {
```
erikayao93, line 101: Should this be named to `asyncProcess()` to be consistent with the method name `asyncProcessRecord()`?

packages/batch/src/BasePartialProcessor.ts
```diff
@@ -66,13 +74,45 @@ abstract class BasePartialProcessor {
...
+    /**
+     * If this is an sync processor, user should have called process instead,
+     * so we call the method early to throw the error early thus failing fast.
+     */
+    if (this.constructor.name === 'BatchProcessor') {
+      await this.asyncProcessRecord(this.records[0]);
     }
+    this.prepare();
+
+    const processingPromises: Promise<SuccessResponse | FailureResponse>[] =
```
erikayao93, line 111: I might be lacking understanding on how to handle asynchronous processes here, so I'm glad to learn more. 
Is there a reason not to do `push(await this.asyncProcessRecord)` to push a `SuccessResponse | FailureResponse` (matching the syntax of the synchronous processing) instead of doing a `Promise.all`?

packages/batch/src/constants.ts
```diff
@@ -1,8 +1,8 @@
-/**
```
erikayao93, line 1: Curious why we remove the documentation here?

packages/batch/tests/unit/AsyncBatchProcessor.test.ts
```diff
@@ -0,0 +1,237 @@
+/**
+ * Test AsyncAsyncBatchProcessor class
```
erikayao93, line 2: Double `Async` typo

packages/batch/tests/unit/AsyncBatchProcessor.test.ts
```diff
@@ -0,0 +1,237 @@
...
+ *
+ * @group unit/batch/class/asyncAsyncBatchProcessor
```
erikayao93, line 4: Should this have the tests for Lambda context as well?

packages/batch/tests/unit/SqsFifoPartialProcessor.test.ts
```diff
@@ -69,26 +57,22 @@ describe('Class: SqsFifoBatchProcessor', () => {
     });
   });
 
-  describe('Asynchronous SQS FIFO batch processing', () => {
-    test('SQS FIFO Batch processor with no failures', async () => {
+  describe('hronous SQS FIFO batch processing', () => {
```
erikayao93, line 60: typo, but also:
We can probably remove this whole set of tests since we aren't doing the asynchronous processing anymore and this now just does the same thing as the first set of tests.

packages/batch/tests/unit/asyncProcessPartialResponse.test.ts
```diff
@@ -0,0 +1,79 @@
+/**
+ * Test processPartialResponse function
+ *
+ * @group unit/batch/function/processpartialresponse
```
erikayao93, line 4: Same as comment for `asyncBatchProcessor` tests -- do we need tests for Lambda context here?

packages/batch/tests/unit/asyncProcessPartialResponse.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    jest.clearAllMocks();
+    jest.resetModules();
+    process.env = { ...ENVIRONMENT_VARIABLES };
+  });
+
+  afterAll(() => {
+    process.env = ENVIRONMENT_VARIABLES;
+  });
+
+  describe('Process partial response function call tests', () => {
```
erikayao93, line 29: Note test for asynchronous in description?

packages/batch/tests/unit/asyncProcessPartialResponse.test.ts
```diff
@@ -0,0 +1,79 @@
...
+        batch,
+        asyncSqsRecordHandler,
+        processor
+      );
+
+      // Assess
+      expect(ret).toStrictEqual({ batchItemFailures: [] });
+    });
+
+    test('Process partial response through handler for SQS records with incorrect event type', async () => {
```
erikayao93, line 50: The original tests were split into
- function call tests
- call through handler tests

Should we have the same split here? And have tests for each type of event to match the synchronous tests?

packages/batch/src/BasePartialProcessor.ts
erikayao93: Ah, the comment didn't leave particularly where I intended it to be -- I'm referring to lines 111-115 here.

dreamorosi: No problem, let me explain: if we did `push(await this.asyncProcessRecord)` then the result of each `this.asyncProcessRecord` would be `await`ed before being added to the array. This would allow the customer to pass an async handler but the records would still be processed sequentially.

Using this method we are instead iterating the records and adding the unresolved promise to the array (notice the lack of `await`). After all the promises are in the array we call `Promise.all` which starts starts all the promises and resolves itself when all the underlying promises have either resolved or rejected.

This allows us to essentially process the whole batch concurrently (aka in parallel) rather than sequentially. I implemented the method this way to match the use case of the one [found in Python](https://docs.powertools.aws.dev/lambda/python/latest/utilities/batch/#processing-messages-asynchronously).

This detail (concurrent vs parallel) is also the reason why there's no `AsybcSQSFifoProcessor`, with FIFO queues we must respect the order of the records, so we can only process them sequentially.

packages/batch/src/constants.ts
dreamorosi: I removed the comments at the top of the files as it's not a convention we use in this codebase. We tend to prefer adding comments to specific entities rather than on the file itself. This is something that [I pointed out](https://github.com/aws-powertools/powertools-lambda-typescript/pull/1588#discussion_r1258312831) during the review of your first PR for your utility.

The only exception to this are test files, in those files the header comment is used by [jest-runner-groups](https://www.npmjs.com/package/jest-runner-groups) to identify and run tests of the same group.

packages/batch/tests/unit/AsyncBatchProcessor.test.ts
dreamorosi: Good spot, removed all occurrences.

packages/batch/tests/unit/SqsFifoPartialProcessor.test.ts
dreamorosi: Good point, removed the set.

See [here](https://github.com/aws-powertools/powertools-lambda-typescript/pull/1616#discussion_r1267773334) for why we don't have async behavior with SQSFifoHandler.

packages/batch/tests/unit/AsyncBatchProcessor.test.ts
dreamorosi: Added the tests!

packages/batch/src/BasePartialProcessor.ts
dreamorosi: I wasn't sure at first but I think you are right, I have changed it.

packages/batch/tests/unit/asyncProcessPartialResponse.test.ts
dreamorosi: Added!

packages/batch/src/types.ts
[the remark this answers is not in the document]
dreamorosi: Fixed

PR #1624, opened 2023-07-24
improv(tests): adopt new aws cdk cli library

dreamorosi:
## Description of your changes

This PR introduces a new package to the workspace dedicated to testing utilities. Historically the `commons` package has been a catch-all package for all shared utilities and code that didn't belong to any specific package. This has caused the package to sprawl (#1354) and while this hasn't always had an impact on what gets published to npm in the `commons` utility, the repository could use some reorganization. 

In the case of integration/e2e tests specifically, there's a lot of room for improvement (#1314) since one of the most critical pieces of our testing infrastructure currently relies on an internal API from CDK that could (and has) change(d) without notice.

As a part of the new internal & unpublished `testing` utility, the PR introduces a new class called `TestStack` that relies on the `@aws-cdk/cli-lib-alpha` module, which allows to interact with the AWS CDK CLI programmatically. This class incorporates many of the lessons learned and aims at streamlining the boilerplate code needed to setup the infrastructure for a test, while standardizing the resources deployed as part of it.

#### Technical deep dive

To articulate the benefits of the new `TestStack` class, let's try to partially refactor the [`makeHandlerIdempotent` tests](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/idempotency/tests/e2e/makeHandlerIdempotent.test.ts). This test suite is fairly recent and presents many of the issues and sprawl that the new class aims to solve.

At a high level, our integration tests can be divided (like many tests of this type) in three discrete areas:
- Setup
- Testing (aka the test cases)
- Teardown

While this PR focuses on the first and last phases, and specifically on the "Setup" one, the new class could address some of the optimizations in the "Testing" phase in future PRs (see "Future improvements" section below).

In order to setup the architecture needed for the tests, this specific one as well as all others, all have a version of the code below:
> **Note**
> Comments were added for clarity, but are not present in the source

```ts
// Validate that the specified runtime is valid (present in all tests)
const runtime: string = process.env.RUNTIME || 'nodejs18x';

if (!isValidRuntimeKey(runtime)) {
  throw new Error(`Invalid runtime key value: ${runtime}`);
}

// Generate an unique name for the test stack (this allows to run multiple e2e tests at the same time)
const uuid = v4();
const stackName = generateUniqueName(
  RESOURCE_NAME_PREFIX,
  uuid,
  runtime,
  'makeFnIdempotent'
);

// Specify the filename for the Lambda that will be deploy to run the test (this is unique to each test suite)
[...]

Issue #1623, opened 2023-07-24
Maintenance: adopt AWS CDK cli library


dreamorosi:
### Summary

To run our integration (e2e) tests we deploy resources to an AWS account via CDK. These resources are spun up right before the test and then destroyed immediately afterwards. When we first wrote this infrastructure, CDK didn't allow programmatic use and so we had to rely on some workarounds. In the past months however CDK has released a new alpha module called `@aws-cdk/cli-lib-alpha` ([link](https://docs.aws.amazon.com/cdk/api/v2/docs/cli-lib-alpha-readme.html)) that could potentially replace our workarounds.

We should look into this module and create an utility that can be used in all the tests that require AWS resources.

### Why is this needed?

The current implementation relies on [an internal CDK API](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/commons/tests/utils/cdk-cli.ts). As such, its implementation can change at any time and break our test infrastructure. This has happened in one of the latest releases of CDK.

### Which area does this relate to?

Tests

### Acknowledgment

- [X] This request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda-typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

PR #1633, opened 2023-07-27
improv(tests): adopt aws cdk cli lib

dreamorosi:

> **Note**
> This PR supersedes #1624 and proposes a much reduced scope compared to that PR.

This PR introduces a new `testing` utility that at the moment contains only a new class called `TestStack`. This class is a simple wrapper around the `@aws-cdk/cli-lib-alpha` module and provides only the minimum amount of methods needed to synth, deploy, and destroy a CDK Stack.

This driver behind this change, as detailed in the linked issue, stems from the reliance of our current integration/e2e tests on an internal CDK command line which has changed in a recent release and thus has locked us out of newer and patched versions of CDK.

Since our first implementation of the integration tests over a year ago, CDK has released a new package (`@aws-cdk/cli-lib-alpha`) that allows to interact with the CLI programmatically, and that can replace our current implementation. Even though the package is in alpha, we consider this still to be an improvement over reliance on internal API, and we hope that through our usage we can help shape the package.

As mentioned above, this is the second PR that tries to apply this change. The previous PR was much larger in scope and in addition to introducing the new CLI wrapper, it also added a ton of new utilities. After discussing internally we have however decided to leave these changes out and work on a RFC before attempting to implement them.

The changes in this PR constitute a stop-gap solution that allows us to move away from the internal CDK API and access current versions of CDK. Once the RFC mentioned will be out and discussed, I am open to decommission the changes introduced this PR in favor of a more solid implementation.

While working on the PR I prioritized keeping changes in the codebase as low as possible, even though there are several areas (i.e. standard resource factories) that could have resulted in a more streamlined implementation.

#### Detailed list of changes:
- Removed of all `aws-cdk`-related (i.e. `http-proxy`, `vm2`, etc.) dependencies from the project
- Removed `cdk-cli` util from `commons` [1]
- Created a new internal testing utility
- Added `aws-cdk-lib` as dependency scoped to the testing module
- Added `aws-cdk` & `aws-cdk-lib` as scoped dependencies to the layers package [2]
- Minimal changes to `tracer` e2e tests to use the new package + added the new package as `devDependency`
- Minimal changes to `logger` e2e tests to use the new package + added the new package as `devDependency`
- Minimal changes to `metrics` e2e tests to use the new package + added the new package as `devDependency`
- Minimal changes to `idempotency` e2e tests to use the new package + added the new package as `devDependency`
- Minimal changes to `parameters` e2e tests to use the new package + added the new package as `devDependency`
- Changes to `layers` e2e tests to use the new package + added the new package as `devDependency` [3]

_**Notes:**_
[1] This was an internal module that was never published to npm, so I don't think it constitutes a breaking change
[2] Since the CLI is used as an actual CLI in the CI for this package, we don't need the additional dependencies and we are not relying on any internal
[3] The diff here is slightly higher compared to the other packages because this e2e test deploys two CDK apps/stacks

### Related issues, RFCs

**Issue number:** #1623

## Checklist

### Breaking change checklist

---


dreamorosi:
Integration test, using the new module, ran successfully: https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/5682665966

dreamorosi:
Also, SonarCloud is going to most probably bring up a ton of findings because the PR touches many existing files that have a lot of code.

I will happily address any finding on the newly added code (`packages/testing/*`), but in order to keep the diff contained & in light of what discussed in the PR body as well as the fact that these are internal test files, I would suggest to not dedicate time to rework other findings.

packages/testing/src/TestStack.ts
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Test stack that can be deployed to the selected environment.
+ */
+class TestStack implements ICloudAssemblyDirectoryProducer {
+  /**
+   * Reference to the AWS CDK App object.
+   * @default new App()
+   */
+  public appRef: App;
```
am29d, line 16: Is there any specific reason for `Ref` suffix? Can we keep it as `app` and `stack`?

packages/testing/src/TestStack.ts
```diff
@@ -0,0 +1,83 @@
...
+    ];
+  }
+
+  /**
+   * Destroy the test stack.
+   */
+  public async destroy(): Promise<void> {
+    await this.#cli.destroy({
+      stacks: [this.stackRef.stackName],
+      requireApproval: false,
```
am29d, line 62: This is odd, the CLI command `destroy` doesn't have `require-approval` but rather `force` option. You can even see that the type of the same field is different between `deploy` and `destroy`. I will submit an issue in the cdk repo.

am29d (commented):
Thanks for the PR! The new cli interface looks clean and we finally have a dedicated place for the testing lib. 

I have left only one small comment.

am29d:
> I will happily address any finding on the newly added code (packages/testing/*), but in order to keep the diff contained & in light of what discussed in the PR body as well as the fact that these are internal test files, I would suggest to not dedicate time to rework other findings.

I agree, let's address the changes in a dedicated PR.

packages/testing/src/TestStack.ts
am29d: It still works, though the CDK team remap the parameter to `--force`

https://github.com/aws/aws-cdk/blob/b1ca3c09e68a2c1f5bf5ce4c9c40f12db7f1767f/packages/%40aws-cdk/cli-lib-alpha/lib/cli.ts#L233C10-L233C33

dreamorosi: Agree that we should submit an issue (& maybe PR). I was also confused by the naming, probably they it's a mistake.

packages/testing/src/TestStack.ts
dreamorosi: This is probably a mental model I brought from React where references to DOM nodes [are called `ref`](https://legacy.reactjs.org/docs/refs-and-the-dom.html).

In this specific case I wanted to clearly signal that these are references to the CDK `Stack` and `App` objects, rather than something internal that we create.

It's also partially an answer to your feedback in the previous PR about name overloading.

Should I change them and remove the `Ref` suffix?

am29d: Ah makes sense. I don't have a strong opinion on that. I see that we actually create the object in the constructor (TIL about `??` coalesce 👌). Imho we don't lose any value if we remove it, especially because there is no concept or usage of `Ref` suffix within CDK domain language (I might be wrong, just never had a chance to see it).

dreamorosi: I have removed the suffix and re-ran the integration tests, which [are passing](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/5692859819).

am29d (approved):
🚀

Issue #1664, opened 2023-08-29
Maintenance: drop support for Node.js 14


dreamorosi:
### Summary

> [!important]
> On November 23, 2023 Powertools for AWS Lambda (TypeScript) will end support for Node.js 14.

This action follows Node.js 14 reaching End-Of-Life (EOL) on April 30, 2023 and is scheduled just one business day before the date in which AWS Lambda will no longer apply security patches and other updates to the runtime.

We recommend that you upgrade your existing Node.js 14 functions to Node.js 18 before November 23, 2023.

End of support does not impact existing functions and functions using versions of Powertools for AWS Lambda (TypeScript) released before November 23, 2023, however starting from this date we will not guarantee compatibility with Node.js 14.

You can use the following AWS CLI command to list all functions in a specific region using Node.js 14, including published function versions:

> [!Note]
> Make sure to replace the value of the `--region` flag and run this command in each region you have Lambda functions workloads

```sh
aws lambda list-functions --function-version ALL --region <Region> --output text --query "Functions[?Runtime=='nodejs14.x'].FunctionArn"
```

### Why is this needed?

Starting from November 27, 2023 we will no longer be able to deploy new Lambda functions using the Node.js 14 runtime, this means we won't be able to run integration tests for new code changes.

With this in mind, we will stop supporting the runtime altogether.

### Which area does this relate to?

Governance

### Acknowledgment

- [X] This request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda-typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

Issue #1375, opened 2023-03-17
Maintenance: test repo & packages with TypeScript 5.x


dreamorosi:
### Summary

On March 16th [TypeScript 5.0 was released](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/). Among other things, the release has some changes related to decorators support.

We should invest some time to test the project with the new version, understand if there's any implication, and if any change is required.

### Why is this needed?

We can expect customers will want to start using the new TS version and so we need to test compatibility.

### Which area does this relate to?

Other

### Acknowledgment

- [X] This request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/#tenets)
- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [Java](https://github.com/awslabs/aws-lambda-powertools-java/), and [.NET](https://github.com/awslabs/aws-lambda-powertools-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

dreamorosi:
After some initial research, it appears that the signature/type of how decorators are defined in TS has changed with the new version. This is unfortunate but also not surprising since decorators as a feature in TypeScript/JavaScript are still experimental.

Prior to TS5.0, the signature was this:
```ts
type MethodDecorator = <T>(
  target: Object,
  propertyKey: string | symbol,
  descriptor: TypedPropertyDescriptor<T>
) => TypedPropertyDescriptor<T> | void;
```

Now, with TS5.0, the signature looks like this:
```ts
type ClassMethodDecorator = (
  value: Function,
  context: {
    kind: "method";
    name: string | symbol;
    access: { get(): unknown };
    static: boolean;
    private: boolean;
    addInitializer(initializer: () => void): void;
  }
) => Function | void;
```

The main differences that I can see will affect us are:
- The decorated method is now `value` (first argument) instead of `descriptor.value` (3rd argument, which we use now)
- Number of arguments goes from 3 to 2
- The name of the decorated method/function is now `context.name` and not `propertyKey` (which has now been removed)

The next step should be to create a small PoC to validate the findings above.

dreamorosi:
In #1667 we have revisited and simplified the `tsconfig.json` files in order to have a main one in the root of the project and then have each package in the workspace extend that one and apply their own specific configuration.

[...]

PR #1672, opened 2023-09-12
chore(maintenance): adopt TypeScript 5.x

dreamorosi:

This PR bumps the TypeScript version for the project to the latest available, which at the time of writing is 5.2.2.

As detailed in the linked issue, one of the main changes brought by TypeScript 5.x is a new implementation and signature for class method decorators [1][2]. This new implementation is more in line with the latest version of [the tc39 proposal](https://github.com/tc39/proposal-decorators) which is currently at Stage 3 (of 4). The same implementation however is **not** compatible with the project's implementation and signature due to Powertools having based its decorator on another implementation (found in [Babel](https://babeljs.io/docs/babel-plugin-proposal-decorators)) that instead is based on an even more experimental stage 2 implementation.

Until TypeScript 4.9.5 (latest version of the 4.x major) the `experimentalDecorators` [compiler option](https://www.typescriptlang.org/docs/handbook/decorators.html) allowed developers to enable support for decorators. Without this option enabled, code that used decorators would not be transpiled by TypeScript and cause a build error. For all intense and purposes up until this version decorators were an opt-in feature in TypeScript.

With TypeScript 5.x decorators have finally become a first-class citizen and the feature is enabled by default. The `experimentalDecorators` [compiler option](https://www.typescriptlang.org/tsconfig#experimentalDecorators) is still present and its meaning can be considered as unchanged, even though enabling/disabling it yields a different outcome than in 4.x.

Starting from 5.0 setting the `experimentalDecorators` option to `false` causes TypeScript to transpile decorators according to the stage 3 implementation (aka the new one).  Setting the option to `true` instead causes TypeScript to transpile using the older stage 2 implementation.

With the above in mind, in upgrading the project to support TypeScript 5.x we had to choose whether to adopt the new implementation or stick with the old. As many things in this ecosystem, and especially when it comes to experimental features, each option comes with its own tradeoffs.

As mentioned previously, the decorators 5.x implementation is more formally correct and for this reason our initial plan was to move all Powertools decorators to it. This is detailed in the linked issue and given the different signature it would have resulted in a breaking change & consequently landed in the next major release.

While we were able to successfully create a PoC of the new implementation based on decorators 5.x, we soon realized that the new stage 3 implementation is **not supported** by tools like `esbuild` (https://github.com/evanw/esbuild/issues/104#issuecomment-1585516166) and its derivatives (i.e. `vitest`). This means that at the time of writing it's not possible to use stage 3 decorators supported by TypeScript with `esbuild`, AWS CDK, and AWS SAM CLI.

For this reason, and also because we rely on both `esbuild` and CDK to run our integration tests, we have decided to stick with the stage 2 implementation which allows us to land TypeScript 5.x support without causing breaking changes.

We acknowledge that this is not an optimal solution due to the stage 2 implementation being now outdated, however we consider this tradeoff worthy for the following reasons:
- `esbuild` is widely used by Serverless developers due to it being used under the hood by both AWS SAM CLI and AWS CDK, as well as being mentioned as a recommended option to bundle and transpile TypeScript to JavaScript in the [AWS Lambda Developer Guide](https://docs.aws.amazon.com/lambda/latest/dg/lambda-typescript.html#)
- this option requires developers to enable `experimentalDecorators` in their `tsconfig.json` file which is already required for customers using Powertools' decorators, thus maximing backwards compatibility
- the new implementation supported by TypeScript 5.x is still considered non-final and subject to change before finally reaching stage 4 and becoming part of the language. 

In order to bump the TS version I also had to upgrade other development dependencies, namely: `jest`, `ts-jest`, `@types/jest`, `typedoc` (and its plugin dependencies), `eslint` (and its plugin dependencies). Note that I was not able to upgrade `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` to their latest versions due to them having dropped support for Node.js 14. As soon as we start working on a release that is no longer targeting that runtime we should upgrade them (#1664).

> [!IMPORTANT]
> After merging this PR we will continue to listen for customers feedback and remain open to revisit the choice and discuss alternatives as long as there is significant demand, so if anyone reading this in the future uses/wants to use Powertools and also authors decorators using the stage 3 implementation and is blocked by this choice, please consider opening an issue or getting in touch via our usual channels.

[1] https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#decorators 
[2] https://2ality.com/2022/10/javascript-decorators.html

### Related issues, RFCs

**Issue number:** #1375

## Checklist

### Breaking change checklist

---


am29d (approved):
Thanks for this upgrade Andrea! 

>While we were able to successfully create a PoC of the new implementation based on decorators 5.x, we soon realized that the new stage 3 implementation is not supported by tools like esbuild (https://github.com/evanw/esbuild/issues/104#issuecomment-1585516166) and its derivatives (i.e. vitest). This means that at the time of writing it's not possible to use stage 3 decorators supported by TypeScript with esbuild, AWS CDK, and AWS SAM CLI.
> 
> For this reason, and also because we rely on both esbuild and CDK to run our integration tests, we have decided to stick with the stage 2 implementation which allows us to land TypeScript 5.x support without causing breaking changes.

Let's setup a regular reminder to check the [esbuild feature request for stage 3 decorators](https://github.com/evanw/esbuild/issues/104) and subscribe to the issue.

dreamorosi:
I am subscribed but looking at the maintainer's comments in the issue it's unlikely they will support the feature until it reaches stage 4 but maybe they'll change their mind based on demand.

For the record: in order to reach stage 4, a feature needs to be implemented & shipped in at least two major browsers, after which the implementation will be considered final. Currently as far as I can tell only Firefox has started working on this.

Hopefully now that TS has adopted stage 3 other browsers will follow suit and include decorators as part of their 2024 plans but in the best of cases I think we are looking at 12-18 months before anything changes.

This is less than ideal, but we have to live with the decision of supporting an experimental feature and this is the best we can do with the current state of things.

TheHolyWaffle:
@dreamorosi esbuild now officially supports TC39 decorators since version 0.21.0:

https://github.com/evanw/esbuild/issues/104#issuecomment-2097359571

Can we start migrating the decorators to the official approach? It'd be handy to drop middy fully. But until decorators are re-implemented, we cannot.

dreamorosi:
Hi @TheHolyWaffle, yes we're aware of the esbuild now supporting the new syntax however making the change will likely require a breaking change, and a major version release, on our side.

We release v2 at the end of Q1 2024 and according to our current [versioning policy](https://docs.powertools.aws.dev/lambda/typescript/latest/versioning/) we officially marked v1 as end of like only one month ago.

With this in mind, unless we can adopt the new decorators in a backwards compatible way, I don't see us releasing a v3 before early Q2 2025. Between now and then, the team will be fully focused on maintaining and improving the existing features and start working on the next upcoming feature which is Event Handler as stated in to [our roadmap](https://docs.powertools.aws.dev/lambda/typescript/latest/roadmap/#event-handler-rest).

We are tracking the discussion and demand for this feature in #3085 - feel free to add your 👍 and share any insight if you know how to remove the backwards compatibility blocker.

Issue #1678, opened 2023-09-13
Maintenance: expose record status & expiry config + make DynamoDB deps optional


dreamorosi:
### Summary

While working on #1484 I realized that in order to make it easy to create your own Idempotency provider there are a few items that need to be tweaked:
- Some persistence layers require a default ttl setting when instantiating a client, this info is already available in the `BasePersistenceLayer` class that is extended when creating your own provider, but it's not exposed to children
- The `IdempotencyRecordStatus` is an object that maps constants to values (aka similar to an Enum). Currently this is part of the `types` folder & exports, which is confusing and also wrong, since it's not a type.
- When not using the `DynamoDBPersistenceLayer` there's no reason why customers should install the `@aws-sdk/lib-dynamodb` module

This issue serves to track the work to change these items.

### Why is this needed?

So that customers creating their own persistence layer for Idempotency or extending the default one can have a better DX.

### Which area does this relate to?

Idempotency

### Acknowledgment

- [X] This request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda-typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

PR #1679, opened 2023-09-14
improv(idempotency): expose record status & expiry config + make DynamoDB Client optional

dreamorosi:

As discussed in the linked issue this PR targets three areas of the Idempotency package. All the change aim at improving the DX for customers who decide to bring their own persistence layer, the documentation on how to actually write your own persistence layer will be addressed in a future PR.

#### DynamoDB Client

The `DynamoDBPersistenceLayer` relies on two AWS SDK modules: `@aws-sdk/client-dynamodb` and `@aws-sdk/lib-dynamodb`.

Before this PR the latter was listed as dependency while the other was only a development dependency. This mean that one was installed automatically upon installing the Idempotency utility while the other was to be installed manually by customers.

This PR marks both modules as `devDependencies` so that both must be installed explicitly by customers. This not only is in line with all other Powertools utilities, but also allows customers developing their own persistence layers to avoid installing an unnecessary dependency.

A third AWS SDK module (`@aws-sdk/util-base64-node`) is still listed as regular dependency, this is because the module is used to generate idempotency keys regardless of the persistence store used, so the module is considered as required.

This PR also introduces the concept of `peerDependencies` ([learn more](https://nodejs.org/en/blog/npm/peer-dependencies)). This allows us to tell npm which versions of these optional dependencies the utility is compatible with. If customers try to install a version outside the range npm will emit a warning and exit the process.

#### `IdempotencyRecordStatus` as constant

The `IdempotencyRecordStatus` object is a map/dictionary of value that facilitate working with the states that an idempotency record can assume throughout its lifecycle:

```ts
const IdempotencyRecordStatus = {
  INPROGRESS: 'INPROGRESS',
  COMPLETED: 'COMPLETED',
  EXPIRED: 'EXPIRED',
} as const;
```

Before this PR the object was included as part of the `types` folder even though it's not a type. When building custom providers the dictionary must be imported, however doing so from an import like `@aws-lambda-powertools/idempotency/types` can be confusing.

This PR moves the object to the constants file and exposes it via the default export: `@aws-lambda-powertools/idempotency`. In doing so, the PR also modifies all the occurrences of imports throughout the package - this has caused the diff to include more files than expected.

#### `getExpiresAfterSeconds()` method

`expiresAfterSeconds` is a private property of the `BasePersistenceLayer` class. The value is configurable when initializing a new persistence layer and indicates the amount of time a record is kept around for idempotency - this is also known as `ttl`.

The class uses this value when creating a new `IdempotencyRecord`, this is fine for persistence stores like DynamoDB that use one of the items attributes as `ttl` key. Some other stores instead set the property either or both at the time of instantiating a new client or when writing an item.

In order to facilitate the former case, this PR exposes the property via a new `getExpiresAfterSeconds()` method that can be used by classes that extend the `BasePersistenceLayer` to get the value before seeing any record.

### Related issues, RFCs

**Issue number:** #1678

## Checklist

### Breaking change checklist

---

### DIFF
--- a/packages/idempotency/package.json
+++ b/packages/idempotency/package.json
@@ -80,9 +80,13 @@
   },
   "dependencies": {
     "@aws-lambda-powertools/commons": "^1.12.1",
-    "@aws-sdk/lib-dynamodb": "^3.231.0",
     "@aws-sdk/util-base64-node": "^3.209.0",
     "jmespath": "^0.16.0"
+  },
+  "peerDependencies": {
+    "@aws-sdk/client-dynamodb": ">=3.360.0",
+    "@aws-sdk/lib-dynamodb": ">=3.410.0",
+    "@middy/core": ">=3.x <4.x"
   },
   "keywords": [
     "aws",
@@ -95,6 +99,7 @@
   "devDependencies": {
     "@aws-lambda-powertools/testing-utils": "file:../testing",
     "@aws-sdk/client-dynamodb": "^3.360.0",
+    "@aws-sdk/lib-dynamodb": "^3.410.0",
     "@types/jmespath": "^0.15.0",
     "aws-sdk-client-mock": "^2.2.0",
     "aws-sdk-client-mock-jest": "^2.2.0"
--- a/packages/idempotency/package.json
+++ b/packages/idempotency/package.json
@@ -84,8 +84,8 @@
     "jmespath": "^0.16.0"
   },
   "peerDependencies": {
-    "@aws-sdk/client-dynamodb": ">=3.360.0",
-    "@aws-sdk/lib-dynamodb": ">=3.410.0",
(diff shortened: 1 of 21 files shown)


packages/idempotency/package.json
```diff
@@ -80,10 +80,25 @@
   },
   "dependencies": {
     "@aws-lambda-powertools/commons": "^1.12.1",
-    "@aws-sdk/lib-dynamodb": "^3.231.0",
     "@aws-sdk/util-base64-node": "^3.209.0",
     "jmespath": "^0.16.0"
   },
+  "peerDependencies": {
```
am29d, line 86: This is great, we should replicate it to other utilities.

dreamorosi: Yep, I'll do it in the next PR

am29d (approved):

Issue #1707, opened 2023-09-24
Maintenance: refactor AWS SDK client check in Parameters and Idempotency


dreamorosi:
### Summary

Customers using the Idempotency and Parameters utilities can optionally pass an AWS SDK client to be used to communicate with AWS APIs. In order to ensure that the object passed at runtime is correct we must perform some conditional logic which at the moment is based on `instanceof`.

When working with Node.js based AWS Lambda functions, the AWS SDK can come from at least 4 known locations (not in order):
1. The Lambda runtime
2. The Lambda Layer (`/opt/nodejs/node_modules/@aws-sdk/`)
3. The customer provided zip archive (`/var/task/node_modules/@aws-sdk/`)
4. The customer provided bundled code (i.e. a single file bundled with tools like `esbuild` and placed in `/var/task/index.js`)

Powertools doesn't ship with the AWS SDK, this allows the user to have control over where the AWS SDK is sourced from among the locations above.

When both Powertools and the SDK are bundled in the same file (4) then Powertools will import/require the SDK from the bundle itself since they are colocated. If both Powertools and the SDK are archived as part of the same zip archive deployed to Lambda (3), Powertools will most likely import the one and the same `node_module` folder. If instead no SDK is provided, Powertools will use the one present in the Lambda runtime (1).

In all the cases above, **since Powertools is always colocated with the customer code** the SDK imported/required in the customer code will be the same that Powertools also requires internally. This allows the runtime `instanceof` check to be successful.

If instead Powertools is shipped as part of a Lambda Layer and the customer brings their own AWS SDK (cases 3 or 4) the `instanceof` check will fail. This happens because Powertools will always default to using a copy of the SDK from either the Layer itself or from the runtime.

Albeit confusing, since the AWS SDK module is imported from two different locations Node.js will consider it as two separate modules for all intense and purposes and so checks like `client instanceof SdkClient` will fail because the two objects are not the same.

We should consider adopting alternative means to check whether an object passed to us is indeed the AWS SDK client that we expect so that the edge cases above will also work.

### Why is this needed?

We need to work on this because we are having customer demand (#1703) for adding utilities that include this logic to our Lambda Layer. At present this strict check would prevent using our Lambda Layers in some edge cases.

Besides Lambda Layers, once we will start shipping hybrid builds with both CJS and ESM the issue might become ever more likely to happen since we won't be able to guarantee the import paths of code that we don't control.

### Which area does this relate to?

Other

### Solution

[...]

PR #1708, opened 2023-09-24
refactor(commons): aws sdk util + identity check

dreamorosi:

This PR introduces changes that tweak the logic used to identify an object passed to a Powertools utility as an AWS SDK client. This work is needed so that we can include Parameters, Idempotency, and Batch to the Lambda Layers as described in the linked issue.

#### AWS SDK util

Since all of a sudden several Powertools utilities have the necessity of knowing whether an unknown object they received is an instance of an AWS SDK client, I have added a new set of utility dedicated to interacting with AWS SDK as part of the commons package/utility. This set of utility will include also the already existing UA middleware as well as the new ones being introduced here.

The main components of this util are:
- Types to express a bare-bone AWS SDK client & some minimal user agent arguments
- Type guard that receives an object and asserts its shape **at runtime** based on the type described above
- User Agent middleware (already existing, simply refactored to use the new things introduced above, and moved into the folder)

For the purpose of our utilities, we will consider a viable AWS SDK client any object that has a function called `send`, a `config` object, and a `middlewareStack` object that contains two functions named `identify` and `addRelativeTo` respectively.

```ts
interface SdkClient {
  send: (args: unknown) => Promise<unknown>;
  config: {};
  middlewareStack: {
    identify: () => string[];
    addRelativeTo: (middleware: unknown, options: unknown) => void;
  };
}
```

The interface above *is not* enough for the client to function properly (aka make requests to AWS APIs) nor for the middleware stack to actually work end to end, however if these keys are not present then there's no point for us in continuing and we can discard the object as invalid.

##### User Agent Middleware

With the introduction of the SDK util and the type guard just mentioned, we can also finally improve (& simplify) the implementation of the UA middleware. When we first implemented the middleware we decided to not include any type from the AWS SDK, in hindsight this was the correct choice since it would have suffered from the same issue we are trying to solve.

In doing so however, we weren't able to narrow down the type of the parameters/objects in the middleware. This lead to using several `// @ts-ignore` comments to suppress issues. At the time we experimented with creating a custom type and a type guard similar to the one introduced in this PR but we decided to not use it to keep things simple.

With the type guard available however, we can reliably assert the type of the object passeed to the middleware factory and remove all the comments that suppressed TS issues. The liberal usage of these comments was also something that @sthulb raised during a review, so the change should be a net positive across the board.

With the types and arguments being less ambigous I have also refactored the unit tests for the middleware so that we are testing only the inputs/outputs and number of calls, rather than the AWS SDK middleware engine behavior (aka we test our own logic, no theirs)..

#### Determining the client type

After introducing the `isSdkClient` helper function described above I initially tried to find a reliable way of determining the client type. Essentially I wanted to ensure that not only an object was a valid AWS SDK v3 client, but also that the client passed belonged to the correct service (i.e. `DynamoDBPersistenceLayer` should receive an instance of `DynamoDBClient` and not `SSMClient`).

As discussed in the linked issue, class instance (i.e. `instanceof DynamoDBClient`) and prototype checks (i.e. `client.prototype.name === 'DynamoDBClient'`) are out of the question. So the next thing I tried to do was to try finding some properties/attributes within the instantiated class that would allow us to reliably do this.

##### Option 1: `serviceId`

All clients appear to have a `serviceId` string within the `config` whose value is set to the service name by default. This was promising, however after looking at the docs I realized that we cannot rely on it since customers can modify its value by passing a custom one when instantiating a new client. This is useful in those cases where customizing the labels in the CloudWatch Service Map is desirable:

```ts
const client = new DynamoDBClient({});
console.log(client.config.serviceId);
// 'DynamoDB' <- label used in Service Map to represent this node

const clientWithCustomService = new DynamoDBClient({
  serviceId: 'PersistenceStore',
});
console.log(client.config.serviceId);
// 'PersistenceStore' <- label used in Service Map to represent this node
```

##### Option 2: `defaultSigningName`

Another candidate that I found was another property called `defaultSigningName` also found within the `config` object. This string represents the value used when signing a request with Sigv4 (see "service name" string in [diagram here](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-authentication-HTTPPOST.html#sigv4-signature-calc-post)). This property is not configurable so it appeared to be the perfect candidate for this purpose.

After running a [few tests](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/6291493245) however I found out that this property was introduced somewhat recently or anyway in a version that **postdates** the one present in the Lambda runtime (3.188.0).

With the above in mind we cannot rely on this field either since it wouldn't work with clients instantiated using the SDK version bundled in the runtime.

##### Option 3: `endpoint`

The next, and last, option would be using the `endpoint` method, which is an **async function** that returns the API endpoint that the client will send request to. Given that all AWS endpoints have unique hostnames we could have used this value to determine somewhat reliably the type of client.

This option however had two problems: 1/ it's an async method, 2/ endpoint is also user-configurable.

In the current implementation the logic for checking the client is found in the constructor of the Powertools utilities. This means we cannot make asynchronous calls since JS allows only synchronous code to be run in a class constructor (for a good reason).

We could have circumvented the issue by refactoring the client check & instantiation to happen asynchronously, kind of like lazy-loading the client only when the first call is made, however this would have complicated the overall implementation for little benefit since the endpoint itself is not a silver bullet.

When creating an SDK client customers can specify a custom endpoint, this is useful mainly for testing but can be used also for advanced use cases that require proxies or custom endpoints. In all these cases we wouldn't be able to make assumptions on the service type based on the hostname.

Given that the use cases described above are also some of the primary reasons why someone would want to use a custom SDK client with the Powertools utilities, I decided to discard this option and refactor the client check/instantiation.

##### Option 4: do nothing

Before starting this refactor, relaxing the checks was a last resort in my view. After investigating all the previous ones however I am now convinced that it's the most sensible option to avoid getting in the way of customers.

The reasons for this change of mind are:
- passing a custom SDK client is an advanced and opt-in feature, customers using this feature have likely read the documentation or IDE docstrings and know what to pass
- the parameter is strongly typed, while this doesn't prevent runtime errors in itself both IDE and compile steps will fail unless all safety measures are explicitly disabled (i.e. `// @ts-ignore`)
- follow-up of the above, we don't perform runtime checks in other similar places like when the customer passes us client configurations
- the AWS SDK itself doesn't do this type of checks. For instance, doing this is allowed: `DynamoDBDocumentClient.from(new SSMClient({}))` since they also [don't perform](https://github.com/aws/aws-sdk-js-v3/blob/0ce9353d5723f8ba5e29e5c67495c7bb0d82530c/lib/lib-dynamodb/src/DynamoDBDocumentClient.ts#L149) runtime checks on this type of thing

For the reasons above, at least for the time being, I suggest we simply try to check whether the object looks like an AWS SDK client and if so, trust that it is. If the provided object has the shape of a client but ends up not being one, this will result in a runtime error.

So to sum up, the current implementation looks like this:
```ts
if (config?.awsSdkv3Client && isSdkClient(config.awsSdkv3Client)) {
  // use this client
} else {
  // create a new client
}
```

I also expect that if the logic above is too coarse we'll listen to customer feedback and refine it based on their usage.

#### Unit Tests

Additionally, the unit tests for all providers have been modified to use scoped module mocking when testing the user agent middleware. Prior to this PR the whole module was imported and then a spy was placed on the specific function. This is not bad in itself, however with the export having now changed, importing everything with `*` would have been wasteful. This method is also more in line with other tests (i.e. Tracer Provider test) and doesn't require any extra spy to be created.

#### Other minor changes

The PR also includes the following minor changes:
- Remove the now redundant `devDependencies` from the commons package as result of the change
- Add two AWS SDK related `devDependencies` to the testing package since it's the package that directly uses them
- Remove the example of install command from the docstring of the Parameters utility since it was breaking intellicode tooltips in VSCode
- Apply native scoped mock of `addUserAgentMiddleware` function to Tracer unit tests

### Related issues, RFCs

**Issue number:** #1707

## Checklist

### Breaking change checklist

---


dreamorosi:
Successful integration test run: https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/6290906353

dreamorosi:
I have also published a test Lambda Layer to our beta account, this is the partial ARN `arn:aws:lambda:eu-west-1:************:layer:1707-pr:2` (replace the Account ID).

The Layer was created using the content of this branch using the script below (see details).

<details>

```sh
#!/usr/bin/env bash

set -e

# Clean up tmp folder and create nodejs folder
rm -rf tmp/*
mkdir -p tmp/nodejs

# Build and package all utilities
## Commons
npm run build -w packages/commons
npm pack -w packages/commons
mv aws-lambda-powertools-commons-*.tgz tmp/nodejs
## Logger
npm run build -w packages/logger
npm pack -w packages/logger
mv aws-lambda-powertools-logger-*.tgz tmp/nodejs
## Metrics
npm run build -w packages/metrics
npm pack -w packages/metrics
mv aws-lambda-powertools-metrics-*.tgz tmp/nodejs
## Tracer
npm run build -w packages/tracer
npm pack -w packages/tracer
mv aws-lambda-powertools-tracer-*.tgz tmp/nodejs
## Parameters
npm run build -w packages/parameters
npm pack -w packages/parameters
mv aws-lambda-powertools-parameters-*.tgz tmp/nodejs
## Batch
npm run build -w packages/batch
npm pack -w packages/batch
mv aws-lambda-powertools-batch-*.tgz tmp/nodejs
## Idempotency
npm run build -w packages/idempotency
npm pack -w packages/idempotency
mv aws-lambda-powertools-idempotency-*.tgz tmp/nodejs

cd tmp/nodejs
npm init -y

# Install utilities & dependencies
npm i \
  aws-lambda-powertools-commons-*.tgz \
  aws-lambda-powertools-logger-*.tgz \
  aws-lambda-powertools-metrics-*.tgz \
  aws-lambda-powertools-tracer-*.tgz \
  aws-lambda-powertools-idempotency-*.tgz \
  aws-lambda-powertools-parameters-*.tgz \
  aws-lambda-powertools-batch-*.tgz

# Remove unnecessary files
rm -rf node_modules/@types \
  aws-lambda-powertools-*.tgz \
  node_modules/@aws-lambda-powertools/*/lib/*.d.ts \
  node_modules/@aws-lambda-powertools/*/lib/*.d.ts.map \
  node_modules/**/README.md \
  node_modules/@smithy \
  node_modules/@aws-sdk \
  node_modules/.bin/semver \
  node_modules/async-hook-jl/test \
  node_modules/shimmer/test \
  node_modules/jmespath/artifacts \
  package.json \
  package-lock.json

# Restore to root
cd ../..

# Restore changes to package.json
git restore packages/*/package.json
```
</details>

dreamorosi:
Apparently the version of the AWS SDK (`3.188.0`) present in the Lambda runtime (`nodejs18x`) doesn't have the `defaultSigningName` key in the config object as it was introduced sometime afterwards.

I need to find another way to identify the client so I'm moving this back to a draft.

dreamorosi:
Successful integration test run: https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/6296842563

am29d (commented):
Absolutely incredible PR which reads like a good thriller. I was pondering the ideas over the weekend and ended up with similar conclusion: either check the shape or remove the check. 

The PR looks great, thanks a lot for putting so much work into it and documenting the decision process. I have just one small suggestion. 

In case customers pass a wrong client with the assumption that their client is actually correct and the utility should accept it. Right now, we silently default to a valid SDK client, without emitting any information about this step. Can we add `console.warn` in the constructor to surface this information and reduce potential time to troubleshoot?

dreamorosi:
Thanks for the good feedback, it's nice to see you appreciated it :D 

I have added the warning log as you suggested. Initially I opted not to do it to avoid complicating the implementation but I agree it's worth it.

To avoid a double nested if/else statement I've changed the logic to optimistically instantiate a new client and replace it with the provided one only when a valid one is passed. As far as I can tell the client instantiation is cheap so the impact should be negligible.

am29d (approved):

Issue #1703, opened 2023-09-21
Feature request: Please release @aws-lambda-powertools/idempotency within the lambda layer


HaaLeo:
### Use case

The current lambda layer version 19 (`arn:aws:lambda:us-east-1:094274105915:layer:AWSLambdaPowertoolsTypeScript:19`)  does not contain the latest released idempotency package `@aws-lambda-powertools/idempotency`.

### Solution/User Experience

Please release a new layer version including the `@aws-lambda-powertools/idempotency` package.

### Acknowledgment

- [X] This feature request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda-typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

am29d:
Hey @HaaLeo 

thanks for raising the issue, and apologies for missing idempotency package in the layer. I will add the package quickly and release a new version.

am29d:
Waiting for #1707 to resolve SDK import conflicts. After that we can include all utilities into the layer.

PR #1710, opened 2023-09-25
chore(layers): bundle assets from source when testing

### LINKED ISSUE #1709: Maintenance: build Lambda Layers from source during tests
### Summary

Currently the `LayerPublisher` construct that we use to deploy Lambda Layers during integration tests and new releases always defaults to installing the Powertools utilities from npmjs.com. This means that we don't have any way to testing the code that we haven't released in Lambda Layers.

This issue proposes to introduce a way to run the integration tests for the `layers` workspace using Powertools utilities that are build from source rather than downloaded from npmjs.com (see "Solution" section below for visual representation of the proposal).

This will allow us to easily test changes that we make to: 
- the utilities & their exports/build process
- the way we package & deploy layers
- test behaviors that are present only in Layers

Without the change above we have no way of covering these cases unless we make a release first.

### Why is this needed?

Up until now our Lambda Layers usage was fairly straightforward as it involved only our core utilities (Tracer, Metrics, Logger) which always shipped with their own dependencies.

In the past couple of weeks alone however the complexity has increased significantly due to separate work streams that involve deploying unpublished versions:
- #1687
- #1708

For this reason we should have a way to run integration tests for the `layer` workspace using Powertools utilities built from source rather then the ones from the last published release.

Automation

### Solution

Visual representation of current process:

```mermaid
graph TB
  A[Run layers e2e tests] --> B[Create tmp/nodejs folder]
  B --> C[Download utilities from registry]
  C --> D[Create zip archive]
  D --> E[Deploy Layers]
  E --> F[Run tests in a function that uses Layer]
```

Visual representation of proposed process:

```mermaid
graph TD
  A[Run layers e2e tests] --> B[Create tmp/nodejs folder]
  B --> C[Build utilities from source]
  C --> D[Install built utilities from tar.gz]
  D --> E[Create zip archive]
  E --> F[Deploy Layers]
  F --> G[Run tests in a function that uses Layer]
```

### Acknowledgment

### Future readers

dreamorosi:

This PR introduces some changes to the way the Lambda Layers are built when running integration tests. The changes should be backward compatible with the current release process, which should work as before for the most part.

Below a matrix of the changes and the areas they are supposed to impact:
|  | Integration | Release |
|--------|--------|--------|
| Bundling Script | ✅  | ✅  |
| Lambda Function | ✅ | ✅ |
| Build Steps | ✅ | ❌  |

#### Bundling Script

Before this PR we had a bash script present in the `.github/scripts/` folder that had to be run before deploying the layers or running tests.

This was fine in the CI environment since it was only an additional step. When working locally however we didn't have any automated way of running it and contributors/maintainers had to remember to run it before being able to run the integration tests for the `layers` workspace.

This PR removes this script in favor of using CDK local bundling. The `Code.fromAsset()` method that we use to reference the layer content accepts two arguments: the location of the asset and a bundling configuration object. 

Even though this is not really well documented, it's possible to force the bundling to happen locally and without Docker. This is exactly what the `NodejsFunction` L3 construct does under the hood and also what [described here](https://aws.amazon.com/blogs/devops/building-apps-with-aws-cdk/) in passing.

The command looks something like this, but with more `execSync` commands:
```ts
code: Code.fromAsset(resolve(__dirname), {
  bundling: {
    // This is here only because is required by CDK, however it is not used since the bundling is done locally
    image: Runtime.NODEJS_18_X.bundlingImage,
    // We need to run a command to generate a random UUID to force the bundling to run every time
    command: [`echo "${randomUUID()}"`],
    local: {
      tryBundle(outputDir: string) {
        execSync('...'); // bash command to execute as part of the bundling
      }
   }
}  
```

With this method we don't have to maintain two scripts and the build process is always run automatically as part of the synth process 🚀 .

#### Lambda Function

Before this release we were performing the following checks:
- The version in the `package.json` file of one utility (Logger) matches the expected version
- Import & initialize the utilities
- Test Logger by emitting a debug log
- Test Metrics by adding a metric (`ColdStart`)
- Test Tracer by interacting with a segment

This PR changes the first and last actions to:
- Check that the version in the `package.json` files of each utility present in the layer matches the expected one
- Test that the tracer actually works by logging the segment's content, so that it can be asserted in the unit tests

The first one is mainly for an excess of caution, however the second was needed because most of the methods in the Tracer are treated as non-critical and as such they are not meant to throw if they don't succeed. By logging the content of a segment we can test that the actions we perform actually happened.

#### Build Steps

As mentioned in the linked issue, before this PR we didn't have any means to test the Powertools utilities in a Layer without making a release. This PR introduces a new CDK context variable called `BuildFromSource` that allows to tell the CDK construct to build the packages from sources before deploying the Lambda Layer (defaults to `false`).

This behavior is set to `true` in the integration tests but defaults to `false` in the release pipeline. When `BuildFromSource` is `true`, the build process runs `npm run build -w packages/<utility>` and `npm pack -w packages/<utility>`, then uses the resulting tarballs to install the utility in the layer folder.

This not only allows us to test the code we are working on in the integration tests, but also serves as an early canary for the build/pack process itself.

### Related issues, RFCs

**Issue number:** closes #1709

## Checklist

### Breaking change checklist

---


am29d:
Ah, I have discovered exactly the same issue when working on #1703 because we can't test layer without having a release.

am29d (approved):
Outstanding refactoring! I was working on improving the shell script, but you took it a step further, well done!

I was tempted to add the packages and adjust the function code here, but I will address it in a separate PR.

dreamorosi:
Thanks, glad you agree!

I thought about that too but the issue was assigned to you and saw you were already pretty far ahead with the tests so you should take credit for that work.

Issue #1714, opened 2023-09-26
RFC: Powertools for AWS (TypeScript) v2


dreamorosi:
### Which Powertools for AWS Lambda (TypeScript) utility does this relate to?

Other

### Summary

As already indicated in the [public roadmap](https://docs.powertools.aws.dev/lambda/typescript/latest/roadmap/#version-2-release), in the coming months we will be releasing a new major version.

This will allow us to both introduce new features as well as quality of life improvements that might require breaking changes or otherwise special attention.

### Use case

The next major release will have two goals:
- Introduce new features that require breaking changes
- Improve the three existing core utilities with practices that we have learned but haven't applied to avoid changes

The two areas that will definitely require breaking changes are:
- ES Modules support (#521): adding this will require changes in the way we bundle & export modules, for this reason we will classify as breaking change even though the imports and interface might remain equivalent
- Improving Logger (#1261, #1136, #1500, TBC): there are a series of changes that we want to make to improve Logger and make it more powerful and extensible. To do so we will have to refactor the internals of the utility which might impact some advanced users in a breaking way.

In addition to the above we are also going to take the opportunity to improve other areas of the codebase. Some of these changes might also require breaking changes. For each one of them we are committed to consider the tradeoffs and keep this issue updated as we investigate.

### Proposal

All the work for this release will happen in a separate branch `feat/v2`, this will allow us to publish alpha releases to the npm registry independently from the main version (1.x). This will allow customers to get early access to the new features and optionally start applying the required changes or simply give feedback on the new items.

#### Quick summary

| Item | Issue | Status | Code change required | Notes |
|--------|--------|--------|--------|--------|
| ES Modules support | #521 | ✅ | | |
| Drop Node.js 14 support | #1686 | ✅ | | |
| Tracer capture sync methods | #1204 | ❌ | | no demand |
| Logger format all attributes | #1261 | ✅ | Yes | |
| Logger observability providers | #1500 | ✅ | | |
| Logger widen arguments types | #1777 | ❌ | | more feedback needed |
| Tracer `captureLambdaHandler` higher order function | #1833 | ❌ | | more feedback needed | 
| Logger sampling - dynamically change to `debug` | #1136 | ✅  | Yes | |
| Logger exotic types - support `Set` and `Map` | #1649 | ❌ | | closed in favor of #1776 |
| JMESPath functions | #1298 | 🚧 | | |
[...]

PR #1734, opened 2023-10-11
feat(logger): Add dual exports with esmodule support on Logger

antstanley:

This change adds dual package exports to the Logger package, so it now exports CommonJs and ES Modules. The intention is that this PR is more for **discussion purposes** at this point, as I've made quite a few decisions to implement this efficiently. 

The changes are primarily isolated to this package, and does not break any other packages or require changes on other packages.

Hopefully it informs the best approach to do this on all the packages.

To implement this I've had to do the following

**Change the package type to `module` as default**

- Add the `type` field to [`package.json`](packages/logger/package.json) and set it to `module`
- Update all the imports in the `src` and `test` folders to include the `.js` extension
- Reference: [https://www.totaltypescript.com/relative-import-paths-need-explicit-file-extensions-in-ecmascript-imports](https://www.totaltypescript.com/relative-import-paths-need-explicit-file-extensions-in-ecmascript-imports)
- tldr: TypeSciprt won't export a pure ESM file without `"type": "module"`, and when that is set it won't compile with the `.js` extension on imports.
 
**Add dual exports to [`package.json`](packages/logger/package.json)**

- Add the `exports` property, and use conditional exports to specify which set of files to use when importing or requiring
- Set `require` to use `lib/cjs/index.js`
- Set `import` to use `lib/esm/index.js`
- Reference: [https://www.typescriptlang.org/docs/handbook/esm-node.html](https://www.typescriptlang.org/docs/handbook/esm-node.html)

**Add `tsconfig.esm.json` to monorepo root and to package root**

- ESM specific TypeScript configs to enable building of ES Modules
- Root [`tsconfig.esm.json`](./tsconfig.esm.json) extends the base [`tsconfig.json`](./tsconfig.json) file and overrides the module specific settings
- Package [`tsconfig.esm.json`](packages/logger/tsconfig.esm.json) extends the root ESM config by specifying the `outDir` as `lib/esm`
- Package  [`tsconfig.json`](packages/logger/tsconfig.json) is updated to set the `outDir` to `lib/cjs`

```json
{
  "extends": "./tsconfig.json",
  "compilerOptions": {
    "module": "NodeNext",
    "moduleResolution": "NodeNext", 
    "noEmitHelpers": true 
  }
}
```

**Update `build` scripts**

- Now three build scripts in [`package.json`](packages/logger/package.json)
- `build:cjs` - Build the CommonJs build using the default `tsconfig.json`
- `build:esm` - Build the ESM build using  [`tsconfig.esm.json`](packages/logger/tsconfig.esm.json)
- `build` - Calls both build scripts

**Add `postbuild` script to define module types in `lib/cjs` and `lib/esm`**

- ~~Created a simple JS script to create `package.json` files in the output directories~~
- ~~[`post build script`](packages/logger/scripts/buildExportPackageJson.js) - Creates a `package.json` file in `lib/esm` and `lib/cjs` to set the module type to the correct type respectively for each folder~~
- The package root `package.json` file has property `type` set to `module` to direct TypeScript and Node to parse file as ESM, the problem with this is all `.js` files will be seen as ESM files, so the files in `lib/cjs` which are CommonJS files, will be seen as the wrong file type. By putting a secondary `package.json` in `lib/cjs` with the property `"type": "commonjs"` it will direct Node to interpret those files as CommonJS files.  
- ~~The post build script creates `package.json` files in `lib/cjs` and `lib/esm`, but it only really needs to create one in `lib/cjs` for those CommonJS files. I decided to add a `package.json` file to `lib/esm` with `"type": "module"` to be explicit, and to mirror file structure of `lib/cjs`, though technically its not needed, as the package's root `package.json` already has the `type` property set correctly for ESM files.~~

- **NOTE** the postbuild script has been replaced by appending `&& echo '{ \"type\": \"commonjs\" }' > lib/cjs/package.json` to the end of `build:*` scripts

#### Enhancements / Fixes for tests

**Removed all barrel files**
- `jest` / `ts-jest` was giving errors when trying to interpret any `export * from 'someModule'` lines, which happens mostly in barrel files.
- Barrel files have a performance impact on import and load times.
- Instead of spending the time debugging `jest`/`ts-jest` configs I decided to remove all barrel files and switch to direct imports
- Reference: [Speeding up the JavaScript ecosystem - The barrel file debacle](https://marvinh.dev/blog/speeding-up-javascript-ecosystem-part-7/)
- Reference: [Performance comparison on direct imports vs barrel files](https://twitter.com/JoshuaKGoldberg/status/1711389528181666284)

### Related issues, RFCs

RFC: #1714 

**Issue number:** #521 

## Checklist

### Breaking change checklist

NO

---


packages/logger/README.md
dreamorosi, line 1: Would it be possible to revert this? Not sure the changes were intended

antstanley, line 1: I ran prettier on `packages/logger` and it updated the README for some reason.

packages/logger/src/Logger.ts
```diff
@@ -210,7 +213,7 @@ class Logger extends Utility implements ClassThatLogs {
    * @returns {void}
    */
   public addPersistentLogAttributes(attributes?: LogAttributes): void {
-    merge(this.persistentLogAttributes, attributes);
+    merge(this.persistentLogAttributes!, attributes);
```
dreamorosi, line 216: Why are we adding this assertion here?

packages/logger/README.md
antstanley: Prettier did me in :-(, and I didn't notice... will revert

packages/logger/src/Logger.ts
antstanley: got a TS error during testing, but not during compile for some reason. Quickly added the assertion to remove the error, instead of debugging the `jest` / `ts-jest` config

antstanley: Just re-run tests, and the error is gone without the assertion. Clearly fixed my config after this. Will revert.

antstanley:
> Thanks for this amazing PR, I really appreciate it!
> 
> One question, perhaps out of ignorance. Is it normal that in the ESM build, the files that contain only types/interfaces are empty? I.e.
> 
> ```
> cat packages/logger/lib/esm/log/LogItemInterface.js 
> export {};
> ```

So this won't have an impact on the running code, and the types are in the declaration file. It's normal behaviour for the TypeScript compiler `tsc`. There are a few long running issues over on the TypeScript repo over this behaviour because it can break some front end project builds. At this moment there is not flag to disable it.

packages/logger/README.md
antstanley: Done. It was my own VS Code formatting defaults for markdown in the end.

packages/logger/package.json
```diff
@@ -16,19 +16,35 @@
...
     "watch": "jest --watch --group=unit",
-    "build": "tsc --build --force",
+    "build:cjs": "tsc --build --force",
+    "build:esm": "tsc --project tsconfig.esm.json",
+    "build": "npm run build:esm & npm run build:cjs",
     "lint": "eslint --ext .ts,.js --no-error-on-unmatched-pattern .",
     "lint-fix": "eslint --fix --ext .ts,.js --no-error-on-unmatched-pattern .",
     "prebuild": "rimraf ./lib",
+    "postbuild": "node ./scripts/buildExportPackageJson.js",
     "prepack": "node ../../.github/scripts/release_patch_package_json.js ."
```
dreamorosi, line 26: can we add this ` && rimraf ./lib/tsconfig.*.tsbuildinfo` so we remove the incremental build files from the packed output?

packages/logger/package.json
```diff
@@ -16,19 +16,35 @@
...
   },
   "homepage": "https://github.com/aws-powertools/powertools-lambda-typescript/tree/main/packages/logger#readme",
   "license": "MIT-0",
-  "main": "./lib/index.js",
-  "types": "./lib/index.d.ts",
+  "type": "module",
+  "exports": {
+    "node": {
+      "require": {
+        "types": "./lib/index.d.ts",
```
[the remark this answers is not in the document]
antstanley: ahh yes! you're right. Copy paster error! Will fix.

packages/logger/package.json
antstanley: Added `&& rimraf ./lib/*.tsbuildinfo` to remove the incremental build files

packages/logger/src/middleware/middy.ts
dreamorosi, line 1: In v2 we would like to remove this from the main export and move it to a dedicated one so that customers can do this:
```ts
import { injectLambdaContext } from '@aws-lambda-powertools/logger/middleware';
```

instead of this

```ts
import { injectLambdaContext } from '@aws-lambda-powertools/logger';
```

The reasoning here is that not all customers use Middy and so this code, alongside its dependencies, doesn't have to be loaded for them. We had some problems with this in the past that caused us to write our own types, but for v2 we'd like to separate the imports.

This is something we are already doing in all the newer packages ([example](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/parameters/package.json#L48)). 

I have tried to do it in this branch by changing the `exports` field to this, but it doesn't work once it's built and packed:
```json
"exports": {
    ".": {
      "require": {
        "types": "./lib/index.d.ts",
        "default": "./lib/cjs/index.js"
      },
      "import": {
        "types": "./lib/index.d.ts",
        "default": "./lib/esm/index.js"
      }
    },
    "./middleware": {
      "require": {
        "types": "./lib/cjs/middleware/middy.d.ts",
        "default": "./lib/cjs/middleware/middy.js"
      },
      "import": {
        "types": "./lib/esm/middleware/middy.d.ts",
        "default": "./lib/esm/middleware/middy.js"
      }
    },
    "./types": {
      "require": {
        "types": "./lib/cjs/types/index.d.ts",
        "default": "./lib/cjs/types/index.js"
      },
      "import": {
        "types": "./lib/esm/types/index.d.ts",
        "default": "./lib/esm/types/index.js"
      }
    }
  },
```

Do you know how can we do this?

tsconfig.cjs.json
antstanley, line 1: This actually isn't needed. I put it there for future proofing, with the idea that CJS builds start to refer to this file, and tsconfig.json is the base config without module settings.

packages/logger/scripts/buildExportPackageJson.js
shdq, line 1: Hey @antstanley thank you for moving this issue forward. I don't know if you discussed it with @dreamorosi or not, but what do you think of just coping `package.json` files instead of creating them with the script, as I suggested [here](https://github.com/aws-powertools/powertools-lambda-typescript/issues/521#issuecomment-1625260251)? If you stick to the script, maybe it makes sense to move it to commons because it will be the same across all the packages?

dreamorosi: I don't have a preference on whether they should be generated or copied over, however for the location of the script itself we will for sure move it to `.github/scripts/` as this is where we keep scripts that are used during build/pack/etc.

packages/logger/src/middleware/middy.ts
dreamorosi: Ok, I think I might have found a way to make it work:

```json
"exports": {
    ".": {
      "require": {
        "types": "./lib/cjs/index.d.ts",
        "default": "./lib/cjs/index.js"
      },
      "import": {
        "types": "./lib/esm/index.d.ts",
        "default": "./lib/esm/index.js"
      }
    },
    "./middleware": {
      "import": "./lib/esm/middleware/middy.js",
      "require": "./lib/cjs/middleware/middy.js"
    },
    "./types": {
      "import": "./lib/esm/types/index.js",
      "require": "./lib/cjs/types/index.js"
    }
  },
  "typesVersions": {
    "*": {
      "middleware": [
        "lib/cjs/middleware/middy.d.ts",
        "lib/esm/middleware/middy.d.ts"
      ],
      "types": [
        "lib/cjs/types/index.d.ts",
        "lib/esm/types/index.d.ts"
      ]
    }
  }
```

@antstanley & @shdq - do you see any issues with doing this?

packages/logger/scripts/buildExportPackageJson.js
antstanley: The idea I had was that there was only one `package.json` to be maintained by the maintainers, as the `lib/esm/package.json` and `lib/cjs/package.json` would be minimal package.json files with the sole purpose of telling Node which module system to use. The package will still be shipping the package.json from the package root

Instead of the `postbuild` JS script I could easily have done the following. 

```json
{
  "scripts": {
    "build:cjs": "tsc --build --force && echo '{ "type": "commonjs" }' > lib/cjs/package.json,
    "build:esm": "tsc --project tsconfig.esm.json && echo '{ "type": "module" }' > lib/esm/package.json",
    "build": "npm run build:esm & npm run build:cjs",
  }
}

```

In hindsight, this is potentially better as it leaves less code to be maintained, and I can ditch the build script.

packages/logger/src/middleware/middy.ts
antstanley: Based on this [https://github.com/nodejs/node/issues/33305#issuecomment-625938633](https://github.com/nodejs/node/issues/33305#issuecomment-625938633) the following should work

```json
// package.json

"exports": {
    ".": {
      "require": {
        "types": "./lib/index.d.ts",
        "default": "./lib/cjs/index.js"
      },
      "import": {
        "types": "./lib/index.d.ts",
        "default": "./lib/esm/index.js"
      }
    },
    "./middleware": {
      "require": {
        "types": "./lib/cjs/middleware/middy.d.ts",
        "default": "./lib/cjs/middleware/"
      },
      "import": {
        "types": "./lib/esm/middleware/middy.d.ts",
        "default": "./lib/esm/middleware/"
      }
    },
    "./types": {
      "require": {
        "types": "./lib/cjs/types/index.d.ts",
        "default": "./lib/cjs/types/index.js"
      },
      "import": {
        "types": "./lib/esm/types/index.d.ts",
        "default": "./lib/esm/types/index.js"
      }
    }
  },

```

```ts
import { injectLambdaContext } from '@aws-lambda-powertools/logger/middleware/middy.js';

```

Will experiment and get back to you

packages/logger/scripts/buildExportPackageJson.js
dreamorosi: I like the reasoning about less files to maintain, thanks for the thought!

Do you think we could potentially do the same with the tsconfig files and move to use flags directly?

packages/logger/src/middleware/middy.ts
dreamorosi: I was able to make the one above ([here](https://github.com/aws-powertools/powertools-lambda-typescript/pull/1734#discussion_r1355340723)) work by:
- adding a `typesVersions` section
- creating a barrel file for the `types` folder (this is needed only for the `/types` subpath)

With these two I am able to import:
```ts
import { injectLambdaContext } from '@aws-lambda-powertools/logger/middleware';
import type {
  LogAttributeValue,
  LogAttributes,
  UnformattedAttributes,
} from "@aws-lambda-powertools/logger/types";
```

In this specific instance (`types/index.ts`) I am inclined to bring back the barrel file. I completely agree with removing them elsewhere and overall moving away from them across the repo wherever possible. 

In this case given the limited amount of exports/files present in the barrel file, as well as the fact that we are dealing with type definition that don't import modules, I'd like to prioritise DX and allow customers to import any of the types under the same subpath export.

Reading the resources you mentioned in the OP the consensus seems to be that if the number of files imported is limited the impact will also be limited.

antstanley: Ok. Got this working on my fork, and pushed here. I've also published an npm package under `@antstanley/logger` you can install and test this out.

`exports` looks like this

```json

"exports": {
    ".": {
      "require": {
        "types": "./lib/cjs/index.d.ts",
        "default": "./lib/cjs/index.js"
      },
      "import": {
        "types": "./lib/esm/index.d.ts",
        "default": "./lib/esm/index.js"
      }
    },
    "./middleware": {
      "require": {
        "types": "./lib/cjs/middleware/middy.d.ts",
        "default": "./lib/cjs/middleware/middy.js"
      },
      "import": {
        "types": "./lib/esm/middleware/middy.d.ts",
        "default": "./lib/esm/middleware/middy.js"
      }
    }
  },

```

In my fork I remove barrel files, so on exports I set the `./middleware` export property to `lib/esm/middy.js`, `lib/cjs/middy.js` etc

This enables you to do imports and requires on both root and `/middleware`

```js
// ES Module
import { Logger } from '@antstanley/logger'
import { injectLambdaContext } from '@antstanley/logger/middleware'
```

```js
// CommonJs
const { injectLambdaContext } = require('@antstanley/logger/middleware')
const { Logger } = require('@antstanley/logger')
```

packages/logger/scripts/buildExportPackageJson.js
antstanley: Ok. I've dropped the postbuild JS script, and replaced it with the extended `build:*` commands in this commit [https://github.com/aws-powertools/powertools-lambda-typescript/pull/1734/commits/cffad2ecd8306f2dd1d4325802a4e25670871ed9](https://github.com/aws-powertools/powertools-lambda-typescript/pull/1734/commits/cffad2ecd8306f2dd1d4325802a4e25670871ed9)

Less code to maintain!

packages/logger/src/middleware/middy.ts
dreamorosi: Thanks for looking into it and also for publishing the test package!

I can confirm that this work as long as `module` is set to `NodeNext` (or equivalent) but doesn't if I use `commonjs`:
image

If instead I put this in the `package.json` (by manually editing the file under `node_modules/@antstanley/logger/package.json`), it works with both module resolutions:

```json
"typesVersions": {
    "*": {
      "middleware": [
        "lib/cjs/middleware/middy.d.ts",
        "lib/esm/middleware/middy.d.ts"
      ]
    }
  }
```

With this in mind, I'd say let's add the `typesVersions` part & repeat the same for the `/types` subpath export.

After that I think we're good to merge this!

packages/logger/scripts/buildExportPackageJson.js
antstanley: > I like the reasoning about less files to maintain, thanks for the thought!
> 
> Do you think we could potentially do the same with the tsconfig files and move to use flags directly?

I initially went that route, using flags on `build:esm` but it became a really long string, even though its only four properties to change. Also having an explicit `tsconfig.esm.json` that other tools can use is useful.

packages/logger/src/middleware/middy.ts
antstanley: Changes done in this commit [8b695c7](https://github.com/aws-powertools/powertools-lambda-typescript/pull/1734/commits/8b695c7c897fcb296aca6aabb960f785101f6a4f)

Also published a new version of this in my package at `@antstanley/logger@1.16.2` with these changes

dreamorosi: Confirmed that it works, thanks a lot!

dreamorosi (approved):
@antstanley, thank you so much for this PR!

I really appreciate you stepping up for the community and contributing this PR, as well as addressing and iterating very rapidly.

Let's merge it!

antstanley:
So happy to see this in! Great work @dreamorosi and @shdq!

Issue #1136, opened 2022-10-27
Feature request: Always log message even when sampling is enabled


bestickley:
### Use case

> [!important]
> See this comment for an updated scope & context on the issue.

I'd like to use Lambda Powertools Logger to log audit information. This info needs to be logged 100% of the time whereas my other log information doesn't. When I enable a sample rate, I'd like to be able to be sure the audit information is logged. 

### Solution/User Experience

```ts
const logger = new Logger({
  serviceName: "MyService",
});
// don't care about "always", could be whatever you think makes sense
logger.always.info("Here is my important log information");
```

### Acknowledgment

- [X] This feature request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/#tenets)
- [X] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [Java](https://github.com/awslabs/aws-lambda-powertools-java/)

dreamorosi:
Hi @bestickley thank you for opening this issue.

Please allow us some time to review this and come up with an answer.

dreamorosi:
At the moment we don't support this type of behavior, however I think it's worth discussing and I would like also to hear the opinion of other maintainers and customers.

In the meantime, I would like to suggest an alternative to achieve a similar behavior. The alternative revolves around using two logger instances:
```ts
import { Logger, injectLambdaContext } from "@aws-lambda-powertools/logger";
import middy from "@middy/core";

const logger = new Logger({
  serviceName: "MyService",
  sampleRateValue: 0.01,
});
const auditLogger = logger.createChild({
  sampleRateValue: 1,
});

export const handler = middy(async () => {
  logger.info("I log sometimes");
  auditLogger.info("I always log");
})
  .use(injectLambdaContext(logger))
  .use(injectLambdaContext(auditLogger));
``` 

These should produce the following logs:
```
{
    "cold_start": false,
    "function_arn": "arn:aws:lambda:eu-west-1:536254204126:function:lambda-powertools-playground-fn",
    "function_memory_size": 128,
    "function_name": "lambda-powertools-playground-fn",
    "function_request_id": "6bb4c509-31d0-4dc7-9e16-f08f22a70b59",
    "level": "INFO",
    "message": "I log sometimes",
    "sampling_rate": 0.01,
    "service": "MyService",
    "timestamp": "2022-11-13T16:47:24.460Z",
[...]

PR #1744, opened 2023-10-13
feat(logger): align sampling debug logs feature implementation with the other runtimes

shdq:

This PR introduces a new implementation for the sampling debug logs feature. It aligns with other runtimes, where log sampling decision is done at the per-function level.

**Main changes:**
- `logsSampled` field was removed in favor of the `sample_rate`
- `setInitialSampleRate` method initializes `sampleRateValue` in order like in initial log level method:

1. constructor
2. custom config
3. env var
4. default value (zero), it's never `undefined` now.
- validation added to `sampleRateValue` to be in range from `0.0` to `1.0`. Before it could be any number. It doesn't throw an error and silently ignores invalid values. This could be improved on the types level, to only allow values in this range, but it's not trivial.

### Related issues, RFCs

**Issue number:** #1136 

## Checklist

### Breaking change checklist

---


packages/logger/src/Logger.ts
```diff
@@ -240,7 +238,7 @@ class Logger extends Utility implements ClassThatLogs {
       logFormatter: this.getLogFormatter(),
     };
     const parentsPowertoolsLogData = this.getPowertoolLogData();
-    const childLogger = this.createLogger(
```
dreamorosi, line 243: If possible, we should try to keep this method as it allows people to more easily extend the logger. 

If we instantiate the logger as `const childLogger = new Logger()`, even if I create my own logger super class (i.e. `class MyOwnLogger extends Logger {}`), its children will always be of type `Logger` (the original).

While we don't strictly need it, by providing this method, people creating `MyOwnLogger` can override it and make it return an instance of `MyOwnLogger` like so:

```ts
class MyOwnLogger extends Logger {
  protected createLogger(options?: ConstructorOptions): MyOwnLogger {
    return new MyOwnLogger(options);
  }
}
```

More context here, where the issue was [discussed](https://github.com/aws-powertools/powertools-lambda-typescript/issues/1307#issuecomment-1625421828) & added.

dreamorosi:
Hey Sergei, thanks a lot for the PR!

I need a bit more time to review it and try the code, but at first look it seems like you are in going in the right direction 😄 

To answer your questions:
> I removed `refreshSampleRateCalculation()` for now, should we make it an alias to `configureSampling()`?

[In Python](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/logging/logger.py#L325) the method seems to be exclusively private (it starts with `_`), so maybe we don't need to.

> Should we pollute log output when the log level is changed to DEBUG because of the sampling rate?
`this.debug('Setting log level to DEBUG due to sampling rate')` This behavior is ported from Python.

I'm not sure yet to be honest. On one hand, I can see how this could be useful to let people know that the log level has changed because of this feature.

On the other hand, I do kind of agree with your choice of wording (`polluting`), so I'm not sure. 

Let me think about it a bit longer and I'll give you an answer when I review more in depth and run the code a few times, either way it should be an easy addition.

packages/logger/src/Logger.ts
shdq: Thanks for the explanation. I'll reverse this change, maybe it make sense to add a link to this issue in the method comment section as a memo.

dreamorosi: I think you're right that we should document/explain this in the comment.

I'd rather though write a comment that explains how to use the method and what it's for rather than linking the issue.

We can reuse and rephrase the explanation/example I mentioned in my previous comment.

shdq: Based on the discussion and your comment, I reversed this change with a detailed comment.

shdq:
> > I removed `refreshSampleRateCalculation()` for now, should we make it an alias to `configureSampling()`?
> 
> [In Python](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/logging/logger.py#L325) the method seems to be exclusively private (it starts with `_`), so maybe we don't need to.

My concern is that `refreshSampleRateCalculation()` is exposed in [docs](https://docs.powertools.aws.dev/lambda/typescript/latest/core/logger/#sampling-logs), so it would be another breaking change if we remove it.

dreamorosi:
> My concern is that `refreshSampleRateCalculation()` is exposed in [docs](https://docs.powertools.aws.dev/lambda/typescript/latest/core/logger/#sampling-logs), so it would be another breaking change if we remove it.

You are right, I would say let's keep it for now and adopt your suggestion.

packages/logger/src/Logger.ts
```diff
@@ -908,6 +882,41 @@ class Logger extends Utility implements ClassThatLogs {
...
+   * 1. Constructor value
+   * 2. Custom config service value
+   * 3. Environment variable value
+   * 4. Default value (zero)
+   *
+   * @private
+   * @param {number} [sampleRateValue]
+   * @returns {void}
+   */
+  private setInitialSampleRate(sampleRateValue?: number): void {
```
dreamorosi, line 896: As we discussed on Discord, I think we could simplify this method & merge it with `configureSampling` one:

```ts
private setInitialSampleRate(sampleRateValue?: number): void {
    this.powertoolLogData.sampleRateValue = 0;
    const constructorValue = sampleRateValue;
    const customConfigValue =
      this.getCustomConfigService()?.getSampleRateValue();
    const envVarsValue = this.getEnvVarsService().getSampleRateValue();
    for (const value of [constructorValue, customConfigValue, envVarsValue]) {
      if (this.isValidSampleRate(value)) {
        this.powertoolLogData.sampleRateValue = value;

        if (value && randomInt(0, 100) / 100 <= value) {
          this.setLogLevel('DEBUG');
          this.debug('Setting log level to DEBUG due to sampling rate');
        }

        return;
      }
    }
  }
```

Note that I have also swapped the order of the `this.setLogLevel('DEBUG');` and the actual `this.debug()` call. This is because in the previous order the log was never emitted if the log level was not already `DEBUG`.

In order for this to work, we'll also have to change the order of the function calls in the `setOptions` method to this:

```ts
this.setConsole();
this.setCustomConfigService(customConfigService);
this.setInitialLogLevel(logLevel);
this.setLogFormatter(logFormatter);
this.setPowertoolLogData(serviceName, environment);
this.setInitialSampleRate(sampleRateValue);
// .. others
```

This is because in order to set the level and log, we need both the log level, the log formatter, and the Powertools log data to be set.

packages/logger/src/Logger.ts
```diff
@@ -558,34 +522,31 @@ class Logger extends Utility implements ClassThatLogs {
...
-   * @param {ConstructorOptions} [options]
-   * @returns {Logger}
-   */
-  protected createLogger(options?: ConstructorOptions): Logger {
-    return new Logger(options);
-  }
-
-  /**
-   * Decides whether the current log item should be printed or not.
+   * **Please do not remove this method during refactoring!**
```
dreamorosi, line 529: I wouldn't add this to the comments - let's just explain what this method is for & how it's used (which you did). This way future readers will see that it has an actual usage & it won't be removed.

packages/logger/src/Logger.ts
```diff
@@ -603,6 +564,17 @@ class Logger extends Utility implements ClassThatLogs {
     });
   }
 
+  /**
+   * It dynamically sets log level based on sampling rate value
+   */
+  private configureSampling(): void {
```
dreamorosi, line 570: With the comment about `setInitialSampleRate`, we can remove this.

packages/logger/src/Logger.ts
```diff
@@ -311,15 +309,6 @@ class Logger extends Utility implements ClassThatLogs {
...
 
-  /**
-   * It returns a boolean value, if true all the logs will be printed.
-   *
-   * @returns {boolean}
-   */
-  public getLogsSampled(): boolean {
-    return this.logsSampled;
-  }
-
```
dreamorosi, line 322: Putting this comment here because the line is not part of the diff, but I would modify also the `createChild` method to make it more explicit that the child is getting the `sampleRateValue` from the parent, to this:

```ts
const parentsOptions: ConstructorOptions = {
  logLevel: this.getLevelName(),
  customConfigService: this.getCustomConfigService(),
  logFormatter: this.getLogFormatter(),
  sampleRateValue: this.powertoolLogData.sampleRateValue,
};
```

dreamorosi (commented):
This is looking really good!

I have pulled the repo and made some suggestions. Also, as discussed on Discord, I'm sharing the modifications to adapt the integration tests.

In the `packages/logger/tests/e2e/sampleRate.decorator.test.ts` file, in the test case that starts with `describe('Enabling sample rate', () => {` (around L70), change the  `if/else` to this:
```ts
if (logMessages.length === 1 && logMessages[0].includes('ERROR')) {
  countNotSampled++;
} else if (
  logMessages.length === 5 &&
  logMessages[0].includes(
    'Setting log level to DEBUG due to sampling rate'
  )
) {
  countSampled++;
} else {
  console.error(`Log group ${logGroupName} contains missing log`);
  throw new Error(
    'Sampled log should have either 1 error log or 4 logs of all levels'
  );
}
```

As you can see the test is the same, but now we are looking for 5 log entries when log sampling is enabled. The new log entry is the one that Logger emits to notify the user that debug log sampling is enabled, so we assert that it contains the correct string.

In order for this to work, we need to do a slight modification to the `TestInvocationLogs` class, which is used in this test (`invocationLogs[i].getFunctionLogs()`). The modification is needed because up until now we were looking only at log entries between the `START` and `END` logs of the invocation (which are the logs emitted within the scope of the function handler).

Now however we need also look at the ones emitted during the initialisation phase, since the debug log (`Setting log level to DEBUG due to sampling rate`) is emitted outside of the handler.

To do this we need to modify the file `packages/testing/src/TestInvocationLogs.ts` to the content below (collapsed for brevity).

<details>

```ts
import { LogLevel } from './constants.js';
import type { FunctionLog } from './types.js';

const CloudWatchLogKeywords = {
  END: 'END RequestId',
  INIT_START: 'INIT_START',
  REPORT: 'REPORT RequestId',
  START: 'START RequestId',
  XRAY: 'XRAY TraceId',
} as const;

class TestInvocationLogs {
  public static LEVEL = LogLevel;

  /**
   * Array of logs from invocation.
   *
   * The first element is START, and the last two elements are END, and REPORT.
   *    [
   *      'START RequestId: c6af9ac6-7b61-11e6-9a41-93e812345678 Version: $LATEST',
   *      '{"cold_start":true,"function_arn":"arn:aws:lambda:eu-west-1:561912387782:function:loggerMiddyStandardFeatures-c555a2ec-1121-4586-9c04-185ab36ea34c","function_memory_size":128,"function_name":"loggerMiddyStandardFeatures-c555a2ec-1121-4586-9c04-185ab36ea34c","function_request_id":"7f586697-238a-4c3b-9250-a5f057c1119c","level":"INFO","message":"This is an INFO log with some context and persistent key","service":"logger-e2e-testing","timestamp":"2022-01-27T16:04:39.323Z","persistentKey":"works"}',
   *      '{"cold_start":true,"function_arn":"arn:aws:lambda:eu-west-1:561912387782:function:loggerMiddyStandardFeatures-c555a2ec-1121-4586-9c04-185ab36ea34c","function_memory_size":128,"function_name":"loggerMiddyStandardFeatures-c555a2ec-1121-4586-9c04-185ab36ea34c","function_request_id":"7f586697-238a-4c3b-9250-a5f057c1119c","level":"INFO","message":"This is an INFO log with some context","service":"logger-e2e-testing","timestamp":"2022-01-27T16:04:39.323Z","persistentKey":"works","additionalKey":"additionalValue"}',
   *      '{"cold_start":true,"function_arn":"arn:aws:lambda:eu-west-1:561912387782:function:loggerMiddyStandardFeatures-c555a2ec-1121-4586-9c04-185ab36ea34c","function_memory_size":128,"function_name":"loggerMiddyStandardFeatures-c555a2ec-1121-4586-9c04-185ab36ea34c","function_request_id":"7f586697-238a-4c3b-9250-a5f057c1119c","level":"ERROR","message":"There was an error","service":"logger-e2e-testing","timestamp":"2022-01-27T16:04:39.323Z","persistentKey":"works","error":{"name":"Error","location":"/var/task/index.js:2778","message":"you cannot prevent this","stack":"Error: you cannot prevent this\\n    at testFunction (/var/task/index.js:2778:11)\\n    at runRequest (/var/task/index.js:2314:36)"}}',
   *      'END RequestId: c6af9ac6-7b61-11e6-9a41-93e812345678',
   *      'REPORT RequestId: c6af9ac6-7b61-11e6-9a41-93e812345678\tDuration: 2.16 ms\tBilled Duration: 3 ms\tMemory Size: 128 MB\tMax Memory Used: 57 MB\t',
   *    ]
   * See https://docs.aws.amazon.com/lambda/latest/dg/nodejs-logging.html for details
   */
  private logs: string[];

  public constructor(logResult: string) {
    const rawLog = Buffer.from(logResult, 'base64').toString('utf-8').trim();
    this.logs = rawLog.split('\n');
  }

  /**
   * Find all functional logs whether it contains a given text
   * @param text
   * @param levelToFilter level to filter
   * @returns
   */
  public doesAnyFunctionLogsContains(
    text: string,
    levelToFilter?: keyof typeof LogLevel
  ): boolean {
    const filteredLogs = this.getFunctionLogs(levelToFilter).filter((log) =>
      log.includes(text)
    );

    return filteredLogs.length > 0;
  }

  /**
   * Return all the log of the function
   * @returns Array of function logs
   */
  public getAllFunctionLogs(): string[] {
    return this.logs;
  }

  /**
   * Return the index of the log that contains `END RequestId`
   * @param logs
   * @returns {number} index of the log that contains END RequestId
   */
  public static getEndLogIndex(logs: string[]): number {
    return logs.findIndex((log) => log.startsWith(CloudWatchLogKeywords.END));
  }

  /**
   * Return only logs from function, exclude INIT_START, START, END, REPORT,
   * and X-Ray log generated by the Lambda service.
   *
   * @param {typeof LogLevel} [levelToFilter] - Level to filter the logs
   * @returns Array of function logs, filtered by level if provided
   */
  public getFunctionLogs(levelToFilter?: keyof typeof LogLevel): string[] {
    const exclusionKeywords = Object.values(CloudWatchLogKeywords);
    let filteredLogs = this.logs.filter(
      (log) => !exclusionKeywords.some((keyword) => log.startsWith(keyword))
    );

    if (levelToFilter) {
      filteredLogs = filteredLogs.filter((log) => {
        try {
          const parsedLog = TestInvocationLogs.parseFunctionLog(log);

          return parsedLog.level == levelToFilter;
        } catch (error) {
          // If log is not from structured logging : such as metrics one.
          return (
            (log.split('\t')[2] as keyof typeof LogLevel) === levelToFilter
          );
        }
      });
    }

    return filteredLogs;
  }

  /**
   * Return the index of the log that contains `INIT_START`
   * @param logs
   * @returns {number} index of the log that contains `INIT_START`
   */
  public static getInitLogIndex(logs: string[]): number {
    return logs.findIndex((log) =>
      log.startsWith(CloudWatchLogKeywords.INIT_START)
    );
  }

  /**
   * Return the index of the log that contains `REPORT RequestId`
   * @param logs
   * @returns {number} index of the log that contains `REPORT RequestId`
   */
  public static getReportLogIndex(logs: string[]): number {
    return logs.findIndex((log) =>
      log.startsWith(CloudWatchLogKeywords.REPORT)
    );
  }

  /**
   * Return the index of the log that contains `START RequestId`
   * @param logs
   * @returns {number} index of the log that contains `START RequestId`
   */
  public static getStartLogIndex(logs: string[]): number {
    return logs.findIndex((log) => log.startsWith(CloudWatchLogKeywords.START));
  }

  /**
   * Return the index of the log that contains `XRAY TraceId`
   * @param logs
   * @returns {number} index of the log that contains `XRAY TraceId`
   */
  public static getXRayLogIndex(logs: string[]): number {
    return logs.findIndex((log) => log.startsWith(CloudWatchLogKeywords.XRAY));
  }

  /**
   * Each of log message contains a JSON with the structured Log object (e.g. {\"cold_start\":true, ..})
   * @param log
   */
  public static parseFunctionLog(log: string): FunctionLog {
    return JSON.parse(log);
  }
}

export { TestInvocationLogs };
```

</details>

After making the change in the testing package, if you want to run the tests, you'll have to run `npm run build -w pacakges/testing` first.

Hope this helps!

shdq:
Oh, one last thing to update is the docs section.

dreamorosi (approved):
I have just run the integration tests and they're passing 🎉 
image

As discussed on Discord, I'm going to take care of the docs in a separate PR since I have to also create the v2 update guide.

Thanks A TON for the work on this, it's highly appreciated!

dreamorosi:
The CI checks are failing due to an extra comma that I introduced a while ago in a previous commit - this is due to the version of Prettier on `main` being older than the one in this branch and it's formatting in that specific case having changed.

I have already fixed the issue in another commit on the branch being merged into ([here](https://github.com/aws-powertools/powertools-lambda-typescript/commit/0b32e113c3a77778e6199cfd1f5775b2f254b460)), and the file wasn't touched in this PR, so merging anyway.

PR #1780, opened 2023-11-02
fix(metrics): deduplicate dimensions when serialising

### LINKED ISSUE #1773: Bug: Override of default dimension with new value results into wrong CloudWatch dimension output
### Expected Behaviour

If I set a default dimension, e.g. `type` with the value `meeting`, and then set through singleMetric with the same dimension name set to another value, the output to CloudWatch contains twice the dimension `type`. This is incorrect in my understanding.

### Current Behaviour

-

### Code snippet

```javascript

import {Metrics, MetricUnits} from "@aws-lambda-powertools/metrics";

const metrics = new Metrics({
    namespace: 'testing',
    serviceName: 'serviceA',
});

export const handler = (event, handler) => {
  metrics.setDefaultDimensions({ type: "meeting" });
  const singleMetric = metrics.singleMetric();
  singleMetric.addDimension("type", "a");
  singleMetric.addMetric("hello", MetricUnits.Count, 1);
};
```

Output: 

```
{"_aws":{"Timestamp":1698769257078,"CloudWatchMetrics":[{"Namespace":"testing","Dimensions":[["service","type","type"]],"Metrics":[{"Name":"hello","Unit":"Count"}]}]},"service":"serviceA","type":"a","hello":1}
```

The `Dimensions` field contains `type` twice, which should be once in my understanding.

### Steps to Reproduce

Run code snippet below and watch output.

### Possible Solution

When writing the dimension you could use `new Set` to ensure the array does not contain duplicates.

latest

18.x

### Packaging format used

npm

### Execution logs

_No response_

am29d:
This PR fixes the issue with duplicate dimensions when we serialise metrics. The quick fix is to filter unique keys for dimensions when serialising the metrics. 

I have noticed that the implementation is a bit different compared to python version. In python we use `dimension_set` which has unique values from default dimension and dimensions. So every change of `defaultDimension` needs to be reflected in the `dimension` field. There is no clear easier approach, in both cases you need to clean up and merge values at some point, either in `setDefaultDimension` `clearDefaultDimension` or in `serialise`. 

I have briefly considered to apply the same strategy, but it would require a more complex refactoring. I consider the `Set` operation performance wise negligible, since we are only doing it once during serialisation. 

### Related issues, RFCs

**Issue number:** closes #1773 

## Checklist

### Breaking change checklist

---


dreamorosi (approved):
Integration [tests are green](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/6734706555), let's ship it!

Issue #1766, opened 2023-10-25
Feature request: TypeScript: Log Metrics with JSON (not console.log)


cortexcompiler:
### Use case

We would like our Metrics to be logged in pure JSON to match the [EMF document structure](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format_Specification.html#CloudWatch_Embedded_Metric_Format_Specification_structure) for `addMetric()` and the `logMetrics()` middleware.  This will allow the logs to be machine readable and ingested by aggregators.  Instead we are seeing the EMF JSON prefixed with the timestamp, request Id, and log level, because the code uses [console.log()](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/metrics/src/Metrics.ts#L391)

We use the [Powertools for AWS Lambda (TypeScript) Metrics](https://docs.powertools.aws.dev/lambda/typescript/latest/core/metrics/) and [Powertools for AWS Lambda (TypeScript) Logger](https://docs.powertools.aws.dev/lambda/typescript/latest/core/logger/).  The Logger works fine because it uses its own [console instance](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/logger/src/Logger.ts#L123).

This is an example of one of the log events from CloudWatch:
```
2023-10-25T17:38:05.273Z	274cced9-0439-4a76-befe-eeef0ec34460	INFO	
{
    "_aws": {
        "Timestamp": 1698255485273,
        "CloudWatchMetrics": [
            {
                "Namespace": "xxxxxx",
                "Dimensions": [
                    [
                        "service",
                        "environment",
                        "function_name"
                    ]
                ],
                "Metrics": [
                    {
                        "Name": "ColdStart",
                        "Unit": "Count"
                    }
                ]
            }
        ]
    },
    "service": "xxxxxx",
    "environment": "test",
    "function_name": "xxxxxx",
    "ColdStart": 1
}
```

### Solution/User Experience

I would like to see the metrics logged with just JSON similar to how the logs are output from the Powertools Logger package.

- **Option 1:** Stop using standard `console.log()` and do something similar to the Powertools Logger, or even re-use it.
   - Con: This would not technically be backwards compatible and could in theory break folks if they are parsing their logs in a specific way
   - Pro: Everyone now gets easily machine readable metrics logs
[...]

PR #1786, opened 2023-11-08
feat(metrics): log directly to stdout 

dreamorosi:

This PR changes the console object used in the Metrics utility from the global one (aka `console.log`) to one instantiated by the utility itself. For most customers this change will have no impact given that the EMF object is going to continue to be parsed as usual, however those customers who send their logs to other destinations other than CloudWatch will now be able to parse the log as a well-formed JSON rather than a log that contains a timestamp, a request ID, and the EMF object.

The implementation follows the one in the Logger utility, which means that by default it will use a new instance of the `Console` object from the `node:console` module. When the `POWERTOOLS_DEV` environment variable is set to a truthy value, the both the Metrics & Logger utility will instead revert back to the global `console` object. This allows customers to easily set spies in their tests.

I have extracted the logic that checks this variable from the Logger to the Commons package, so that the Metrics utility can reuse it.

### Related issues, RFCs

**Issue number:** #1766

## Checklist

### Breaking change checklist

---


dreamorosi:
Moving this back to draft as I have to fix integration tests in the Layers that are now failing.

Metrics integration tests are passing because we are asserting against the metrics retrieved from CloudWatch, this shows that the change is successful. Layers on the other hand are failing because we are doing an assertion on the shape of the logs, which has now changed.

Tomorrow I'll fix and mark it ready.

dreamorosi:
I have integration tests passing: 
https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/6816958295

am29d (approved):

Issue #1783, opened 2023-11-04
Maintenance: Support for Node.js 20


dreamorosi:
### Summary

Node.js 20 official docker image is available in preview: https://github.com/aws/aws-lambda-base-images/issues/91#issuecomment-1737694321. We can start preparing to support the new runtime in Powertools for AWS Lambda (TypeScript).

### Why is this needed?

Because we aim at supporting the new runtime the same day it's launched.

### Acknowledgment

- [X] This request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

dreamorosi:
This is now released under [v1.15.0](https://github.com/aws-powertools/powertools-lambda-typescript/releases/tag/v1.15.0) version!

PR #1790, opened 2023-11-13
feat(maintenance): add support for nodejs20.x runtime

dreamorosi:

This PR makes the necessary changes to support the Node.js 20 runtime in AWS Lambda.

The PR also moves the development and CI environments to Node.js 20.

> [!important]
> After this PR is merged, please remember to switch your local dev environment to the version specified in the `.nvmrc` file. You can do so by running `nvm use` or equivalent.

### Related issues, RFCs

**Issue number:** #1783

## Checklist

### Breaking change checklist

---


.devcontainer/devcontainer.json
```diff
@@ -8,16 +8,15 @@
...
+      "VARIANT": "20-bullseye"
     }
   },
   "customizations": {
     // Add the ids of extensions you want installed when the container is created.
     "extensions": [
       "dbaeumer.vscode-eslint",
       "esbenp.prettier-vscode",
-      "firsttris.vscode-jest-runner",
-      "amazonwebservices.aws-toolkit-vscode"
```
am29d, line 20: Is this intentional and we no longer need the plugin?

.github/workflows/run-e2e-tests.yml
```diff
@@ -53,11 +53,11 @@ jobs:
       - name: Setup NodeJS
         uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3.6.0
         with:
-          node-version: ${{ matrix.version }}
+          node-version: '20'
```
am29d, line 56: Should this be `${{ matrix.version }}`?

.github/workflows/run-e2e-tests.yml
```diff
@@ -53,11 +53,11 @@ jobs:
...
       - name: Setup dependencies
         uses: ./.github/actions/cached-node-modules
         with:
-          nodeVersion: ${{ matrix.version }}
+          nodeVersion: '20'
```
am29d, line 60: Should this be `${{ matrix.version }}`?

docs/snippets/idempotency/templates/tableSam.yaml
```diff
@@ -17,7 +17,7 @@ Resources:
   HelloWorldFunction:
     Type: AWS::Serverless::Function
     Properties:
-      Runtime: python3.11
```
am29d, line 20: 🫢

am29d (commented):
Thanks for the PR!

I assume some of the test refactoring were necessary for the new version? For example`packages/logger/tests/unit/formatter/PowertoolLogFormatter.test.ts` .

.github/workflows/run-e2e-tests.yml
dreamorosi: In an earlier version of this PR I left this as it was and integration tests were failing on Node.js 14 (see https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/6856373318).

This was because of CDK, which I had to upgrade in order to add support for the new runtime.

In upgrading the dependency however, we bumped to a version that dropped support for Node.js (the node version in the CI, not the Lambda runtime).

In order for Powertools to support the new Lambda runtime we needed the integration tests to be passing. At the same time, we can't drop the Node.js 14 runtime for 2 more weeks, so we also need the integration tests for that one to be running.

After some consideration I found an alternative that would have allowed us to keep the older CDK version and reference the new Node.js 20 runtime. Instead of using `Runtime.NODEJS_20_X` which is introduced in the new version, we can instead treat this as a custom runtime and use `new Runtime('nodejs20.x', RuntimeFamily.NODEJS)`.

This would have allowed us to keep the older version of CDK, run on Node.js 14 in the CI for a couple more weeks, and still deploy a function using the new runtime since CloudFormation would have recognized the `nodejs20.x` runtime at deployment time.

This worked fine (see https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/6856924987) but after thinking about it for some more time I also realized that the version of Node.js used in the CI doesn't really matter for integration tests.

The primary goal of these tests is to assert that our code is compatible with a certain Lambda runtime. To this end, we use CDK to deploy some ephemeral infrastructure, use `esbuild` to bundle the function code, and finally use the AWS SDK to interact with the deployed resources.

With this in mind, the only Node.js version that matters in this test is the one used in the Lambda runtime, which is completely independent from the one used in the CI. Using one Node.js version or the other in the CI doesn't affect the CDK deployment, nor the bundle (settings are codified in the `tsconfig.json`), or the SDK calls.

Given the above, I consider it safe to use the latest Node.js version in the CI while still using the matrix value in the Lambda runtime. This change does not affect the unit test matrix, which does run the test within the CI and as such still needs to be using the target Node.js versions.

What do you think? Am I missing something?

.github/workflows/run-e2e-tests.yml
dreamorosi: See [here](https://github.com/aws-powertools/powertools-lambda-typescript/pull/1790#discussion_r1392204289) for more details.

.devcontainer/devcontainer.json
dreamorosi: Yes, it's intentional. 

I've remove it to align this config with the Gitpod one. There I also removed the extension a while ago because it now requires a login due to CodeWhisperer.

Ultimately I think we should just include the bare minimum of extensions needed and this is not one of them.

In the future I'd like to go even further and move away completely from pre installing any extension and instead use the [recommended extension system](https://code.visualstudio.com/docs/editor/extension-marketplace#_recommended-extensions) which would simply prompt the user and let them choose which one they want to install. This is for a future PR though.

dreamorosi:
> I assume some of the test refactoring were necessary for the new version? For example`packages/logger/tests/unit/formatter/PowertoolLogFormatter.test.ts` . 

That's correct, I tried to keep the changes at minimum as much as possible.

In the example you mentioned, it seems that the new Node.js version has changed the way `AssertionError` formats errors. The new version includes more information in the error message, which required me to change the test to allow for this new behavior.

In doing so, I also realized that there was no need to actually throw the errors in order to test the formatter, and that we could just instantiate the error object and pass it directly into the method.

This change, albeit not strictly necessary for this PR, seemed like a low hanging fruit that would simply the overall test file, so I went with it.

Another refactor I had to make was in the Metrics integration tests. In this case I noticed that when testing the Node.js 20 runtime and CI combination, the assertions on the dimension array of a metric were failing. Upon debugging I saw that they were failing due to the assertions using `strictEqual` and the arrays having same members but different ordering.

Since for our purposes the order of the dimensions doesn't matter, I wrote a small helper function that sorts the dimension array based on the name of the dimension before doing the assertion.

.github/workflows/run-e2e-tests.yml
am29d: > With this in mind, the only Node.js version that matters in this test is the one used in the Lambda runtime, which is completely independent from the one used in the CI. Using one Node.js version or the other in the CI doesn't affect the CDK deployment, nor the bundle (settings are codified in the tsconfig.json), or the SDK calls.

Makes sense, good change!

am29d (approved):
Lets goooooo 🚀

dreamorosi:
Integration [tests passing](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/6861764190), merging the PR.

PR #1831, opened 2023-12-21
feat(parser): implement parser decorator

### LINKED ISSUE #1816: Feature request: implement parser decorator
### Use case

Similar to middy middleware we want to use build-in schemas and envelopes to parse and extract events using decorators. 

### Solution/User Experience

Based on the RFC the decorator should take model and envelope as an param object:

```ts
const order = z.object({
  id: z.number(),
  description: z.string(),
});

type Order = z.infer<typeof order>;

  // Decorate your handler class method
  @logger.injectLambdaContext()
  @parser({eventBridgeEnvelope, order})
  public async handler(
    _event: OrderModel,
    _context: Context
  ): Promise<void> {
    logger.info(`Received order model ${_event}`);
  }
}
```

we should also be able to pass only the model, which can be a custom event or an extension of the built-in schema i.e. 

```ts
const order = z.object({
  id: z.number(),
  description: z.string(),
});

const customEBschema = EventBridgeSchema.extend({
  detail: order,
});
```

### Alternative solutions

_No response_

### Acknowledgment

### Future readers

am29d:

This PR adds parser decorator based on the RFC and #1816 . 

### Related issues, RFCs

**Issue number:** closes #1816 

## Checklist

### Breaking change checklist

---


packages/parser/src/parser.ts
```diff
@@ -0,0 +1,59 @@
...
+ */
+const parser = <S extends ZodSchema>(
+  options: ParserOptions<S>
+): HandlerMethodDecorator => {
+  return (_target, _propertyKey, descriptor) => {
+    const original = descriptor.value!;
+
+    const { schema, envelope } = options;
+
+    descriptor.value = function (
```
dreamorosi, line 42: Does this still keep the decorated method async? I haven't stepped through with the debugger, but I think with this implementation we are implicitly replacing an async method with a sync one and when calling it with the `apply()` we are returning an unresolved promise.

Not sure if this is an issue but I think this is why we made all other decorators async.

What do you think?

dreamorosi (commented):
Left a comment about the decorator implementation, other than that it's looking good

packages/parser/src/parser.ts
am29d: Great observation! I checked in the debugger and the call `original.apply()` remains async, while `descriptor.value` is of course sync function and thus all the method of the class will be replaced with a sync function, changing return type value instead of Promise.

We could check `original.constructor.name === AsyncFunction` to decide what we should return. The more interesting case will bubble up once users want to bring their custom `transforms` or `refines`  which include async code. This would force us to expose `parseAsync` and branch the parsing for envelopes and decorator/middy. 

I'd suggest to change the modified `descriptor.value` to async, to keep it consistent with other decorator and we will circle back in a separate issue on this matter. We planned to address the sync/async decorators anyway.

dreamorosi: Yea I agree that we should stick to async and keep things simple and consistent.

We discussed doing runtime checks for the constructor type to determine if it's sync or async when we implemented the Batch Processor and we eventually decided to drop the idea.

The main argument against this is that even though within the function body you might be able to do this check, the function signature will always have to be `async`, which then leads to the return type always being a `Promise<T>`.

dreamorosi (approved):
Thank you for working on this and addressing the comment!

Let's merge it!

PR #1835, opened 2024-01-04
docs(parser): add docs for parser utility

### LINKED ISSUE #1796: Docs: write main docs for parser

The documentation should include a section dedicated to the upcoming parser utility.

_No response_

### How can we improve?

```Markdown
N/A
```

### Got a suggestion in mind?

_No response_

### Acknowledgment

am29d:

This PR adds documentation section for the parser utility. The main structure is aligned with the current python documentation, except "Advanced use cases", since we don't generate models from JSON/object strucutre. 

I have also used this PR to adjust the exports. 

### Related issues, RFCs

**Issue number:** closes #1796 

## Checklist

### Breaking change checklist

---


docs/utilities/parser.md
```diff
@@ -0,0 +1,216 @@
...
+    1. Use `z.infer` to extract the type of the schema
+    2. `event` is typed as `Order` object
+    3. we can infer deeply nested types, here `Order` and `OrderItem` are inferred
+
+### Compatibility with @types/aws-lambda
+
+The package `@types/aws-lambda` contains type definitions for AWS service event invocations. 
+Powertools parser utility also bring AWS Lambda event types based on the built-in schema definitions. 
+Both libraries try to provide types for the same event structure, but they might not be compatible with each other yet. 
+We are working on a sustainable solution to make them compatible and avoid any breaking changes in the future.
```
dreamorosi, line 211: Reading this I am a bit confused (as a customer), mainly because I don't understand whether this is an issue with Powertools or `@types/aws-lambda`.

As a maintainer, while I know we have an issue open for this, I think we should agree & articulate our stance on this topic a bit better to avoid future confusion amongst us and towards customers.

The main point here, besides calling out potential discrepacies between us & the other library, is in my opinion to establish our strategy towards event types as a whole. This will then influence which actions we'll take now and in the future, as well as when discrepacies are detected.

Currently the `@types/aws-lambda` types are community maintained and manually authored. AWS events can change and the types can contain issues.

While I agree with the sentiment of maintaining these types and our schemas (and in the future JSON schemas as well) up to date and aligned, I don't think this ultimately guarantees that the types/schemas are actually correct.

As a think big idea, we should at least consider the option of instead investing efforts in creating automations to periodically test against real AWS events to get the actual shapes of the events sent to Lambda. Based of those, we could then derive the schemas as well as the TypeScript types.

Once we detect drift (aka changes in the events), we could then propagate them to the appropriate destinations be it within the scope of Powertools or `@types/aws-lambda`.

I am not expecting the whole industry to move over our own types (i.e. `@aws-lambda-powertools/events` or `@aws-lambda-powertools/types`) but I would still aim at becoming the source of truth for these events/schemas/types.

The DefinitelyTyped repo (aka the one that hosts `@types/*` packages) already has precedents of depending on external type definition, one of them was added in the past 6 months to the `@types/node` types to support the `fetch` module. This module was originally implemented under `undici` and later incorporated into Node.js. The source still resides in the `undici` repo, so to add support for these types they started publishing type definitions for under `undicy-types` that are now a direct depenencey pf the `@types/node` package (see [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/5684af052501705829b574de2394a6c570aaed56/types/node/package.json#L19)).

While it'll be a lot of work, I believe this would bring a lot of value to customers and eventually would help with establishing a single and authoritative source of truth for AWS events that today does not exist. Doing this would benefit not only Powertools TS or the Node.js community but also Powertools Python and the Python community as a whole via Pydantic. Additionally, it would open the door to future use cases like testing factories (#445), integrations with SAM or CDK, and similar.

---

Now, back to this PR, the TL;DR; here is that at this stage and in this section I would simply call out that there could be discrepancies between `@types/aws-lambda` and our schemas. At the moment we don't have any automatic way of detecting these differences and because of this we: 1/ recommend customers to use our types whenever possible, and 2/ consider opening an issue to us whenever they detect issues.

But overall I wouldn't say that we aim at being compatible with these types because they themselves might be incorrect, and while we will definitely contribute back when finding issues, ultimately we don't have control over them.

dreamorosi (changes requested):
Thank you for the amazing work here so far.

I have left a series of comments throughout the PR, while some of them are clear cut (I think), others are more opinionated and I am open to discuss them.

In the code snippets, please review the type imports and add an `import type` whenever we are importing a type. For example, when importing the `Context` from `aws-lambda`. Also I would suggest we use Logger whenever we are using `console.log`, even if it adds a few extra lines to the snippet.

Also I didn't leave a comment in each one of them, but in the unit tests when importing schemas or envelopes from their respective barrel files, we should use the full relative path. So for example `import { SnsSchema } from '../../../src/schemas';` would become `import { SnsSchema } from '../../../src/schemas/index.js';` and so on.

docs/utilities/parser.md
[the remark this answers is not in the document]
am29d: I have simplified the schema to have a single schema now, instead of two. This way we have only one schema and one type.

docs/utilities/parser.md
am29d: Ok, refreshed this part. We recommend to use out types, but if there any errors, here is the link to raise an issue. I have removed that compatibility part to not overpromise. 

I agree that we can maintain types in PT in the long run.

docs/utilities/parser.md
```diff
@@ -0,0 +1,216 @@
...
+* Built-in envelopes to unwrap and validate popular event sources payloads
+* Extend and customize envelopes to fit your needs
+* Available for middy middleware and TypeScript method decorators
+
+## Getting started
+
+### Install
+
+```bash
+npm install @aws-lambda-powertools/parser zod
```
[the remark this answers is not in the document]
am29d: addde `zod~3` to install promt

docs/utilities/parser.md
```diff
@@ -0,0 +1,216 @@
...
+```
+
+This utility supports zod v3.0.0 and above.
```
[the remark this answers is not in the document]
am29d: Not a big fan, especially because it's the only information in this paragraph. It'd become otherwise empty.

docs/utilities/parser.md
```diff
@@ -0,0 +1,216 @@
...
+
+## Define schema
+
+You can define your schema using zod:
```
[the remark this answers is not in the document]
am29d: changed

docs/utilities/parser.md
```diff
@@ -0,0 +1,216 @@
...
+This is a schema for order and order items using zod. 
+You can create more complex schemas using zod, such as nested objects, arrays, unions, etc. see [zod documentation](https://zod.dev) for more details.
+
+## Parse events
+
+You can parse inbound events using `parser` decorator or middy middleware. 
+Both are also able to parse either an object or JSON string as an input.
+
+???+ note
+    The decorator and middleware will replace the event object with the parsed schema if successful. This means you might be careful when nesting other decorators that expect event to have a specific structure.
```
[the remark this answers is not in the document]
am29d: Yeah, I read it few times, feels terrible, added new version

docs/utilities/parser.md
```diff
@@ -0,0 +1,216 @@
...
+
+The package `@types/aws-lambda` contains type definitions for AWS service event invocations. 
+Powertools parser utility also bring AWS Lambda event types based on the built-in schema definitions. 
+Both libraries try to provide types for the same event structure, but they might not be compatible with each other yet. 
+We are working on a sustainable solution to make them compatible and avoid any breaking changes in the future. 
+
+We recommend to use the types provided by the parser utility. 
+
+## Error handling
+We don't have any error handling in the utility and propagate all errors from zod, which are thrown as `ZodError`.
```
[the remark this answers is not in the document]
am29d: I will remove error handling for now. I need to add our custom `ParseError` in a separate PR and will also add docs. Atm we throw generic `Error` in the envelope.

docs/index.md
dreamorosi, line 1: Rebase conflict probably - should've been the v2

docs/snippets/package.json
```diff
@@ -2,11 +2,11 @@
   "name": "docs",
   "version": "2.0.3",
   "description": "A collection code snippets for the Powertools for AWS Lambda (TypeScript) docs",
+  "type": "module",
   "author": {
     "name": "Amazon Web Services",
     "url": "https://aws.amazon.com"
   },
-  "private": true,
```
dreamorosi, line 9: Why removing this?

am29d: woops, unintentional, the merge was a wild ride.

docs/index.md
am29d: What do you mean?

docs/utilities/parser.md
```diff
@@ -0,0 +1,258 @@
...
+* Extend and customize envelopes to fit your needs
+* Safe parsing option to avoid throwing errors and custom error handling
+* Available for Middy.js middleware and TypeScript method decorators
+
+## Getting started
+
+### Install
+
+```bash
+npm install @aws-lambda-powertools/parser zod~3
```
dreamorosi, line 27:
```suggestion
npm install @aws-lambda-powertools/parser zod@~3
```

docs/snippets/package.json
dreamorosi: If we remove it, this package will get published to npm and make the release fail.

docs/index.md
dreamorosi, line 1: Please revert this too, and I think we're done

dreamorosi (changes requested):
image

I think there's something wrong with the formatting of the main index docs page (that's what I was trying to point out in the past few comments).

Not sure if intentional or automatic, but removing these indentations will break all the callouts.

am29d:
> I think there's something wrong with the formatting of the main index docs page (that's what I was trying to point out in the past few comments).
> 
> Not sure if intentional or automatic, but removing these indentations will break all the callouts.

Ah, there was some weird indentation during merge. I have checked the collapsing callouts in the `index.md` it should be good now.

docs/index.md
```diff
@@ -289,11 +289,12 @@ Core utilities such as Tracing, Logging, and Metrics will be available across al
 | [Parameters](./utilities/parameters.md)   | High-level functions to retrieve one or more parameters from AWS SSM Parameter Store, AWS Secrets Manager, AWS AppConfig, and Amazon DynamoDB                     |
 | [Idempotency](./utilities/idempotency.md) | Class method decorator, Middy middleware, and function wrapper to make your Lambda functions idempotent and prevent duplicate execution based on payload content. |
 | [Batch Processing](./utilities/batch.md)  | Utility to handle partial failures when processing batches from Amazon SQS, Amazon Kinesis Data Streams, and Amazon DynamoDB Streams.                             |
+| [Parser](./utilities/parser.md)           | Utility to parse and validate AWS Lambda event payloads using Zod, a TypeScript-first schema declaration and validation library.                                  |
 
 ## Environment variables
 
 ???+ info
-	Explicit parameters take precedence over environment variables
+Explicit parameters take precedence over environment variables
```
dreamorosi, line 297: image

should be

image

dreamorosi (changes requested):
Found another tooltip that got broken

docs/index.md
am29d: tooltip hunter

dreamorosi (approved):
Thanks for the work on this PR, happy to merge it!

Issue #1799, opened 2023-11-23
Maintenance: Improve security posture by addressing OpenSSF results


sthulb:
### Summary

We recently added OpenSSF Scorecard to the repository, on the initial scan, it found a few actionable items.

### Why is this needed?

OpenSSF Scorecard scanner found minor issues with the project that we should improve on.

### Which area does this relate to?

Governance

### Acknowledgment

- [X] This request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [X] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

dreamorosi:
The "on new PR" workflow is failing due to it attempting to obtain more elevated permissions than the caller workflow - see [an example here](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/7263820844/workflow).

Would appreciate if you could take a look. 

cc @am29d.

dreamorosi:
The "On PR merge" workflow is also failing due to permission issues: https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/7292228413/job/19872848955

dreamorosi:
The issues to the PR automations have been fixed in #1978.

@am29d I'm wondering if you'd be okay with us removing the `Create Artifact` step ([here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/b487b878e78267ba2fab93a40df069c7474ac62d/.github/workflows/reusable_publish_docs.yml#L99)) from the docs workflows.

The step requires `actions: write` which apparently lowers our scorecard score, and since it's used in three different workflows (the one that uses it plus two caller ones) it penalizes us 3 times.

With this in mind I would like us to consider removing it from the workflow and troubleshoot in other ways when needed.

What do you think?

dreamorosi:
With the release workflow merged we have completed most of the work needed for the OpenSSF scorecard.

After discussing internally, these are the pending items for the next iteration, after which we can close the topic:
[...]

PR #1978, opened 2024-01-26
chore(ci): refactor workflows to scope permissions

dreamorosi:

This PR attempts to address some of the OpenSSF Scorecard recommendations by modifying the GitHub Actions workflows.

Most of the changes introduced in this PR are ported from the Powertools Python repository, which has already gone through this same process.

This PR only addresses some of the recommendations, other more complex/critical workflows like release, integration testing, and layers will be addressed in dedicated PRs.

In addition to changes directly aimed at addressing the scorecard, this PR also makes the following changes:
- rename some of the workflow to conform with naming in the Python repo - this should help removing cognitive load
- add documentation for each modified workflow as comment in the workflow file
- remove `mike` step from docs workflow, this step is likely redundant since we don't use its versioning as well as not publish to GitHub Pages anymore. If the change is successful we'll remove the dependency in a future PR

### Related issues, RFCs

**Issue number:** #1799

## Checklist

### Breaking change checklist

---


.github/workflows/dispatch_analytics.yml
```diff
@@ -28,6 +26,20 @@ jobs:
       group: analytics
     runs-on: ubuntu-latest
     environment: analytics
+    permissions:
```
am29d, line 29: Is the level of permission block affecting the score card?

.github/workflows/measure-packages-size.yml
```diff
@@ -1,45 +0,0 @@
-name: Measure packages size
```
am29d, line 1: I liked this buddy from time to time, and we are not using this workflow often. Not relevant for this PR, but what do you think to introduce package changes to releases?

.github/workflows/on-doc-v2-merge.yml
```diff
@@ -1,25 +0,0 @@
-name: Docs v2 Publish (merge)
```
am29d, line 1: No longer needed?

am29d (commented):
Nice work, great to have more context for the permissions, wasn't always clear why we need it.

.github/workflows/on-doc-v2-merge.yml
dreamorosi: On second thought I think we don't.

I created it but it doesn't work, and since we are near merging the v2 branch I don't want to invest time in figuring out what's happening with the jq & `versions.json` file.

We are going to publish the migration guide (which I'll finish soon) on `main` and with an item in the navbar so customers have a chance to see it soon.

For changes that are specific to utility pages, we can just use v2 and use local preview for the next week or two before we finally merge v2.

.github/workflows/measure-packages-size.yml
dreamorosi: I like it too, but it uses an action from a fork owned by someone who left the team/company, so we can't use it as-is for now.

Additionally, it uses SAM to bundle which in turns uses Docker. Now that we have a better understanding of `esbuild` (see Layers work) and headless CDK (see `testing` package) we can do the same with local bundling and measure that, which is much simpler and faster.

We'll reintroduce this in the near-ish future, how soon depends on how the quarter/year planning goes in the coming days & weeks.

.github/workflows/dispatch_analytics.yml
dreamorosi: Yep, check [this section](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) from their docs, the relevant sentence is this:

> The highest score is awarded when the permissions definitions in each workflow's yaml file are set as read-only at the [top level](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions) and the required write permissions are declared at the [run-level](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idpermissions). One point is reduced from the score if all jobs have their permissions defined but the top level permissions are not defined. This configuration is secure, but there is a chance that when a new job is added to the workflow, its job permissions could be left undefined because of human error.

So basically we need to set a permission at the file (aka workflow) level that is limited (`read`) and then define appropriate permissions at the job level. This way other jobs can't inadvertently access data they're not supposed to.

dreamorosi: I should also note that I'm still getting up to speed with the scorecard system, for now I'm using the Py repo as baseline since I confirmed with them that they don't have findings in these areas.

am29d: that makes sense, I just wasn't sure if it applies for only one job in a workflow, but I see the value in scoping it down to individual job.

.github/workflows/measure-packages-size.yml
am29d: agree

am29d (approved):

PR #2006, opened 2024-02-02
chore(ci): add versioning workflow for pre-release

am29d:
This PR adds a workflow to run versioning before a release, which solves two issues. 1/ `lerna version` does not push the commit back to `main`, 2/ we can review version and changelog changes before the release, avoiding accidentally wrong version bumps based on conventional commits. 

While we previously had a tag during versioning, I suggest keeping the tag after a successful package publish step in the release workflow. 

### Related issues, RFCs

**Issue number:** #1799

## Checklist

### Breaking change checklist

---


.github/actions/create-pr/create-pr/action.yml
dreamorosi, line 1: is the nested `create-pr/created-pr` structure intentional?

.github/workflows/make-version.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Checkout code
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11  # v4.1.1
+        with:
+          ref: ${{ github.ref }}
+      - name: Setup NodeJS
+        uses: actions/setup-node@b39b52d1213e96004bfcb1c61a8a6fa8ab84f3e8 # v3.6.0
+        with:
+          node-version: "20"
+          cache: "npm"
+      - name: Setup auth tokens
```
dreamorosi, line 29: Before this step we need this:

```yml
- name: Setup dependencies
```

otherwise it will get `lerna` from npm, which can be a problem.

(In the v2 branch we are also [adding a flag](https://github.com/aws-powertools/powertools-lambda-typescript/blob/fc59672e2bf87cd565be6880c694fc83b102e055/.github/actions/cached-node-modules/action.yml#L7-L9) to that custom action to skip building the packages, which is not needed in this context, so it'll be faster)

dreamorosi: image

am29d: Do you mean after this step similar to what we have in `make-release.yml`?

dreamorosi: Yep, you can replace the setup node with this one, like in the other workflows

.github/actions/create-pr/action.yml
```diff
@@ -0,0 +1,88 @@
...
+runs:
+  using: "composite"
+  steps:
+    - id: adjust-path
+      run: echo "${{ github.action_path }}" >> $GITHUB_PATH
+      shell: bash
+    - id: setup-git
+      name: Git client setup and refresh tip
+      run: |
+        git config user.name "Powertools for AWS Lambda (TypeScript) bot"
```
dreamorosi, line 66: Can/should we change this user to 

```sh
git config --global user.name 'github-actions[bot]'
git config --global user.email 'github-actions[bot]@users.noreply.github.com'
```

like below?

Reason why I suggest this is that since that user doesn't actually exist it shows as an empty box everywhere in the GH UI

image

image

.github/workflows/make-version.yml
am29d: done

dreamorosi: forgot to commit?

am29d: not it's right below, no?

dreamorosi (approved):
Thank you for working on this. Looks great and I have tested it in a fork.

Would like to hear from @sthulb if we could move this action (aka only what's under `.github/actions/create-pr`) under the dedicated `actions` repo we have in the org.

This action is an almost exact copy [of the one here](https://github.com/aws-powertools/powertools-lambda-python/tree/develop/.github/actions/create-pr) from the Python repo. The changes we made here should be able to be refactored so that they're parameters/inputs of the action.

 This way we can have a single place & maintain only once rather than have the same / slightly similar action in two projects.

.github/actions/create-pr/action.yml
```diff
@@ -0,0 +1,92 @@
...
+#     echo "Branch: ${BRANCH}"
+#   env:
+#     PR_ID: ${{ steps.create-pr.outputs.pull_request_id}}
+#     BRANCH: ${{ steps.create-pr.outputs.temp_branch}}
+
+inputs:
+  files:
+    description: "Files to add separated by space"
+    required: true
+  temp_branch_prefix:
```
sthulb, line 39: Couldn't this be generated?

.github/actions/create-pr/action.yml
```diff
@@ -0,0 +1,92 @@
...
+    description: "Prefix for temporary git branch to be created, e.g, ci-docs"
+    required: true
+  pull_request_title:
+    description: "Pull Request title to use"
```
sthulb, line 43: This could be generated: "chore(ci): Bump version from X to Y"

.github/actions/create-pr/action.yml
```diff
@@ -0,0 +1,92 @@
...
+runs:
+  using: "composite"
+  steps:
+    - id: adjust-path
+      run: echo "${{ github.action_path }}" >> $GITHUB_PATH
+      shell: bash
+    - id: setup-git
+      name: Git client setup and refresh tip
+      run: |
+        git config --global user.name 'github-actions[bot]'
```
sthulb, line 70: we also have the PT bot user: aws-powertools-bot

.github/actions/create-pr/action.yml
dreamorosi: This is designed to be a reusable action, that's why many of these are designed to be inputs defined by the caller.

For now (aka this PR) we are using the action only as part of the versioning process, so hardcoding the PR message makes sense.

In a future PR we will use this same action to open a PR that updates the documentation with the new Lambda Layer version (for example). 

With this in mind, allowing the caller to define these fields makes sense.

.github/actions/create-pr/action.yml
dreamorosi: Strictly speaking, yes, however having a human readable prefix helps maintainers. Refer to [this comment](https://github.com/aws-powertools/powertools-lambda-typescript/pull/2006/files/3c849206fb967799e5141323e8f1521892e4a678#r1478038479) as to why these are inputs defined by the caller workflow.

.github/actions/create-pr/action.yml
```diff
@@ -0,0 +1,92 @@
...
+#   run: |
+#     echo "PR number: ${PR_ID}"
+#     echo "Branch: ${BRANCH}"
+#   env:
+#     PR_ID: ${{ steps.create-pr.outputs.pull_request_id}}
+#     BRANCH: ${{ steps.create-pr.outputs.temp_branch}}
+
+inputs:
+  files:
+    description: "Files to add separated by space"
```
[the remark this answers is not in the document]
dreamorosi: With the current usage, nothing. 

In the future we'll reuse this same action in another workflow - see [here](https://github.com/aws-powertools/powertools-lambda-typescript/pull/2006/files/3c849206fb967799e5141323e8f1521892e4a678#r1478038479).

Strictly speaking we could also remove the parameter and always run `git add .` to add everything, but having this as an input makes the action more flexible for future usage.

Unless you think there're security concerns stemming from the parameter I'd be inclined keep it.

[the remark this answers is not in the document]
am29d: You can either pass specific files or leave it blank than it takes all the files that were staged.

.github/actions/create-pr/action.yml
am29d: we could hard code the prefix for now, since only pt and ts teams use it and they can be the same.

.github/actions/create-pr/action.yml
am29d: ah this is much better, will change

dreamorosi: Using that (non existent) user makes the UI broken
image

Unless we create an actual bot user (aka a GH App), the user won't exist so better to use the action user which exists.

.github/actions/create-pr/action.yml
dreamorosi: I think it's still useful to have a prefix defined by the caller:
i.e. `ci-version-???` or `ci-bump-???` for branches created as part of version bump vs `ci-docs-???` for branches created as part of the docs layer version (this latter will come in a future PR).

If for some reason the workflow(s) fail and leave dangling branches it'll be easier for us to tell them apart instead of having to scan the list, open each one, and see what's in it (which we would have to do if the name is opaque).

dreamorosi:
@sthulb what about the question [here](https://github.com/aws-powertools/powertools-lambda-typescript/pull/2006#pullrequestreview-1861127602) on moving the action under the `actions` repo?

.github/actions/create-pr/action.yml
am29d: This is a GitHub user, not a GitHub App, thus we can't use it. I will keep github-actions bot for the config.

.github/actions/create-pr/action.yml
am29d: Keeping this as an input option for now.

dreamorosi (approved):
thanks for raising the bar on this!

PR #2028, opened 2024-02-06
chore(ci): refactor release workflow

am29d:

In this PR we refactor the release workflow to reduce the permission scope and add additional review step for the layer arn docs changes. The changes include

* [x] add `create-tag` because we don't do it with lerna any more
* [x] add `create-pr` step after layer `prod` deployment to review doc changes
* [x] use `github.sha` for commit hash checkout
* [x] publish `from-package` in detached head with a sha, because we tag in the next
* [x] change layer doc updates to PR review  
* [x] scope down permissions to `contents: read` 

### Related issues, RFCs

**Issue number:** #1799 

## Checklist

### Breaking change checklist

---

### DIFF
--- a/.github/workflows/make-release.yml
+++ b/.github/workflows/make-release.yml
@@ -53,7 +53,6 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11  # v4.1.1
         with:
-          token: ${{ secrets.GH_PUBLISH_TOKEN }}
           ref: ${{ github.sha }}
       - name: Git client setup
         run: | 
--- a/.github/workflows/make-release.yml
+++ b/.github/workflows/make-release.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11  # v4.1.1
         with:
-          ref: ${{ github.ref }}
+          ref: ${{ github.sha }}
       - name: Setup NodeJS
         uses: actions/setup-node@b39b52d1213e96004bfcb1c61a8a6fa8ab84f3e8 # v4.0.1
         with:
@@ -54,7 +54,7 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11  # v4.1.1
         with:
           token: ${{ secrets.GH_PUBLISH_TOKEN }}
-          ref: ${{ github.ref }}
+          ref: ${{ github.sha }}
       - name: Git client setup
         run: | 
           git config --global user.name 'aws-powertools-bot'
--- a/.github/workflows/make-release.yml
+++ b/.github/workflows/make-release.yml
@@ -35,15 +35,9 @@ jobs:
           cache: "npm"
       - name: Setup auth tokens
         run: |
-          git config --global user.name 'github-actions[bot]'
-          git config --global user.email 'github-actions[bot]@users.noreply.github.com'
-          git remote set-url origin https://x-access-token:${{ secrets.GH_PUBLISH_TOKEN }}@github.com/$GITHUB_REPOSITORY
           npm set "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}"
       - name: Setup dependencies
         uses: ./.github/actions/cached-node-modules
-      - name: Version
-        run: |
-          npx lerna version minor --force-publish --no-commit-hooks --yes
       - name: Publish to npm
         run: |
           NPM_CONFIG_PROVENANCE=true npx lerna publish from-git --yes
@@ -52,7 +46,28 @@ jobs:
         run: |
           VERSION=$(cat lerna.json | jq .version -r)
           echo RELEASE_VERSION="$VERSION" >> "$GITHUB_OUTPUT"
-
+  
+  create_tag: 
+    needs: [publish-npm]
+    permissions:
+      contents: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11  # v4.1.1
+        with:
+          token: ${{ secrets.GH_PUBLISH_TOKEN }}
+          fetch-depth: 0
+      - name: Git client setup
+        run: | 
+          git config --global user.name 'aws-powertools-bot'
+          git config --global user.email '151832416+aws-powertools-bot@users.noreply.github.com'
+          git config remote.origin.url >&-
+      - name: Create git tag
+        run : |
+          git tag -a v${{ needs.publish-npm.outputs.RELEASE_VERSION }} -m "Release v${{ needs.publish-npm.outputs.RELEASE_VERSION }}"
+          git push origin v${{ needs.publish-npm.outputs.RELEASE_VERSION }}
+   
   # NOTE: Watch out for the depth limit of 4 nested workflow_calls.
   # publish_layer -> reusable_deploy_layer_stack -> reusable_update_layer_arn_docs
   publish_layer:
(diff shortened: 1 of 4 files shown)


heitorlessa:
just wanted to say... great work @am29d !! 

Abiding to our implicit tenet of working in public, we briefly discussed about the seal+checkout process, so here's the reasoning:

* **Prevent in-between commits to slip in**. One of the issues we've had having versioning being bumped outside the release process was that we didn't have guarantees that another commit could've been pushed between bump and release. The solution was to always checkout code from the release commit (immutable), bump the version in the first step, seal the changed source code along w/ integrity hash, and restore it in subsequent steps.
* **GitHub requires a `checkout` in the build process**. I've bumped into mysterious issues with the git client, credentials, and actions in general when I tried to restore the sealed artifact without a `checkout`. The solution was to always checkout followed by a restore of the sealed artifact which would overwrite what's being checked out, thus bringing the newly bumped version we are meant to release.

Whichever solution you find that prevents accidental code being added to the release after a version bump I'll be happy with it. Ping @sthulb for a security review once this is settled.

We're here to help!

dreamorosi:
> great work @am29d !!
+1 on this

> Prevent in-between commits to slip in

We have opted to go in this direction because we want to delegate the version increase to Lerna rather than having to manually compute it and input it. Already in its current form, Lerna scans the messages of all the commits made since the previous release and based on conventional commits it decides the next version. 

At the very beginning of the project (mid 2021) this was done manually but given the modular nature of the project we found that this put a significant cognitive load on the maintainers. To obviate this we adopted the `lerna version` command which creates the changelog for each package in its respective directory, updates all the `package.json` files, and commits the changes.

Given that we want to remove the `contents: write` permission and we no longer want our CI to mutate the repository without the maintainers supervision, we need to surface the commit into a PR. This also gives us the chance to review the new version and its contents before the actual release process starts.

With this in mind, we decided to split the workflow in two:
- one that versions the project (#2006)
- and another that does the release on npm as well as Layers (this PR)

Our original plan was to have a maintainer take the commit hash of the versioning commit and use it as an input of the release workflow. After your comment however I wonder if this might not be enough to establish strong provenance.

I don't want to complicate the workflows further, but at least for now I'm inclined to keep the versioning & release steps separate unless there's no other way that satisfies governance & security. To this end, I see two options:
1. add an initial step to the release workflow that checks the input commit hash & determines if it comes from the correct PR (aka the latest versioning one).
2. use the "Versioning PR merged" event as a trigger of the release workflow

The second option is probably the simpler (in terms of implementation) and guarantees a stronger link between versioning & release. The only problem I see with it is that if for any reason the release workflow fails, we'd have to revert the versioning commit & run it again, but I think it's an acceptable tradeoffs in the name of security (as long as we have a runbook ).

What do you all think?

heitorlessa:
looking 👀

.github/workflows/make-release.yml
```diff
@@ -24,35 +24,47 @@ jobs:
...
           npm set "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}"
       - name: Setup dependencies
         uses: ./.github/actions/cached-node-modules
-      - name: Version
-        run: |
-          npx lerna version minor --force-publish --no-commit-hooks --yes
       - name: Publish to npm
         run: |
-          NPM_CONFIG_PROVENANCE=true npx lerna publish from-git --yes
+          NPM_CONFIG_PROVENANCE=true npx lerna publish from-package --git-head ${{ github.sha }} --yes
```
heitorlessa, line 40: NIIIIIIICE! Such a tiny change with huge value <3 thank you

.github/workflows/make-release.yml
```diff
@@ -24,35 +24,47 @@ jobs:
...
+  create_tag: 
+    needs: [publish-npm]
+    permissions:
+      contents: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11  # v4.1.1
+        with:
+          token: ${{ secrets.GH_PUBLISH_TOKEN }}
```
heitorlessa, line 56: why do you need an user-defined token since this is a read-only operation? what am I missing?

heitorlessa (changes requested):
one question about the usage of a custom token on `action/checkout`.

Ask: could you please document the overall process of `make-release.yml`?

Example: https://github.com/aws-powertools/powertools-lambda-python/blob/9b00d0a690077cc6df227556ad45055b8febc4b3/.github/workflows/release.yml#L3

.github/workflows/make-release.yml
am29d: Missed this one, it's a legacy config, I have removed it.

.github/workflows/make-release.yml
dreamorosi: Can you please test this locally against [`verdaccio`](https://verdaccio.org) to see how it behaves when re-run? 

Since we publish multiple packages there could be cases in which some packages are published successfully and others aren't and we need to repeat the run.

Basically I'd like to make sure the release workflow is idempotent. At least based on the docs for these flags this seems the most appropriate command but I'd like to err on the caution side and test it, if possible.

am29d: I used both, `verdaccio` locally and private playground repo to publish, it worked and it is also idempotent. The command will compare the package version with the registry and publish only the package versions that are higher. Here is an example with two packages, where one is already publish on NPM and the other is not:

```bash
Found 1 package to publish:
lerna info auto-confirmed 
lerna info publish Publishing packages to npm...
 - @am29d/commons => 1.0.13
```

This raises a more general question on how we should see the release operation in case of partial failure. If it's atomic, we'd need to investigate the issue and make another release of all packages. Or, we'd need to treat each package as an individual release with the convention to have the same version across all packages (which is an indirect dependency). 

TLDR: rerunning the job will publish only unpublished packages with the higher local version compared to NPM registry.

dreamorosi: > This raises a more general question on how we should see the release operation in case of partial failure.

If the failure is in the CI pipeline (for whatever reason), then a retry, even if it involves commits to fix our workflows, should not warrant a version bump imo. 

I see a lot of projects that bump the version whenever there's a change in dev dependencies (on the project's side) or in their pipelines, and imo this is wrong. Version bumps should happen only when the code published to npm changes.

With this in mind, retrying should be ok, as long as the goal is to have all the packages of a given version published to npm.

> TLDR: rerunning the job will publish only unpublished packages with the higher local version compared to NPM registry.

I think this is technically not accurate, and if it was, it would prevent us from ever publishing a v1 security release once we release v2. After reading the wording on the `lerna version` docs my understanding is that lerna simply reads the version in th `package.json`, then checks if that version is available on the registry, and if not it publishes it.

Which one is it?

dreamorosi: Also, sorry for not mentioning this in my previous message, but thanks a lot for taking the time to test this. Amazing work!

am29d: I might be wrong, but as far as I understood, `lerna version` operates on the repo, not on the NPM. Did you mean `lerna publish`? Then this is correct, we can not publish v1 fixes using `from-package` once v2 is published in NPM, **using this workflow**. 

We can't have both, using `from-package` helps us to rerun the workflow, using `from-git` would publish any version we have in the `package.json`. 

If the intention is to have one workflow for both releases, then I need to revisit the workflow.

heitorlessa (approved):
great work @am29d !!!

.github/workflows/make-release.yml
dreamorosi: Sorry, it was a typo. I meant `publish`, and I was referring to the sentence: `rerunning the job will publish only unpublished packages with the higher local version compared to NPM registry.`.

If this was true then once we have say `2.0.0` on npm, then attempting to publish `1.20.0` would not succeed because the local version is lower than the one on the registry.

But this is not the case, the `lerna` docs say this:
> the list of packages to publish is determined by inspecting each package.json and determining if any package version is not present in the registry. Any versions not present in the registry will be published.

So we are good

dreamorosi (approved):
Amazing work, looking forward to test this!

Thanks for raising the bar on this!

Issue #2045, opened 2024-02-09
Maintenance: add badges to the main README


am29d:
### Summary

Badges are useful indicators for quick information and python project utilises some of them, like openssf scorecard, supported runtimes, etc. We should review which badges we can use similar to python project or any other specific badges for javascript ecosystem. We should ensure the balance between quantity and quality of the information. 

### Why is this needed?

To provide immediate visual cues on project status and metrics and enhance credibility. 

### Which area does this relate to?

Other

### Acknowledgment

- [X] This request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

PR #2072, opened 2024-02-14
chore(docs): include badges in readme

hjgraca:

- Add badges to readme

### Related issues, RFCs

**Issue number:** #2045 

## Checklist

### Breaking change checklist

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,9 +6,6 @@
 [![Quality Gate Status](https://sonarcloud.io/api/project_badges/measure?project=aws-powertools_powertools-lambda-typescript&metric=alert_status)](https://sonarcloud.io/summary/new_code?id=aws-powertools_powertools-lambda-typescript)
 [![Security Rating](https://sonarcloud.io/api/project_badges/measure?project=aws-powertools_powertools-lambda-typescript&metric=security_rating)](https://sonarcloud.io/summary/new_code?id=aws-powertools_powertools-lambda-typescript)
 [![OpenSSF Scorecard](https://api.securityscorecards.dev/projects/github.com/aws-powertools/powertools-lambda-typescript/badge)](https://api.securityscorecards.dev/projects/github.com/aws-powertools/powertools-lambda-typescript)
-![NPM Downloads](https://img.shields.io/npm/dw/%40aws-lambda-powertools%2Ftracer?style=flat-square&label=Tracer)
-![NPM Downloads](https://img.shields.io/npm/dw/%40aws-lambda-powertools%2Flogger?style=flat-square&label=Logger)
-![NPM Downloads](https://img.shields.io/npm/dw/%40aws-lambda-powertools%2Fmetrics?style=flat-square&label=Metrics)
 [![Join our Discord](https://dcbadge.vercel.app/api/server/B8zZKbbyET?style=flat-square)](https://discord.gg/B8zZKbbyET)
 
 Powertools for AWS Lambda (TypeScript) is a developer toolkit to implement Serverless [best practices and increase developer velocity](https://docs.powertools.aws.dev/lambda/typescript/latest/#features).
--- a/README.md
+++ b/README.md
@@ -1,6 +1,5 @@
 # Powertools for AWS Lambda (TypeScript) <!-- omit in toc -->
 
-[![Build](https://github.com/aws-powertools/powertools-lambda-typescript/actions/workflows/pr-run-linting-check-and-unit-tests.yml/badge.svg)]([https://github.com/aws-powertools/powertools-lambda-typescript/actions/workflows/pr-run-linting-check-and-unit-tests.yml])
 ![NodeSupport](https://img.shields.io/static/v1?label=node&message=%2016|%2018|%2020&color=green?style=flat-square&logo=node)
 ![GitHub Release](https://img.shields.io/github/v/release/aws-powertools/powertools-lambda-typescript?style=flat-square)
 [![Quality Gate Status](https://sonarcloud.io/api/project_badges/measure?project=aws-powertools_powertools-lambda-typescript&metric=alert_status)](https://sonarcloud.io/summary/new_code?id=aws-powertools_powertools-lambda-typescript)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,6 +1,15 @@
...
+![NodeSupport](https://img.shields.io/static/v1?label=node&message=%2016|%2018|%2020&color=green?style=flat-square&logo=node)
+![GitHub Release](https://img.shields.io/github/v/release/aws-powertools/powertools-lambda-typescript?style=flat-square)
+[![Quality Gate Status](https://sonarcloud.io/api/project_badges/measure?project=aws-powertools_powertools-lambda-typescript&metric=alert_status)](https://sonarcloud.io/summary/new_code?id=aws-powertools_powertools-lambda-typescript)
+[![Security Rating](https://sonarcloud.io/api/project_badges/measure?project=aws-powertools_powertools-lambda-typescript&metric=security_rating)](https://sonarcloud.io/summary/new_code?id=aws-powertools_powertools-lambda-typescript)
+[![OpenSSF Scorecard](https://api.securityscorecards.dev/projects/github.com/aws-powertools/powertools-lambda-typescript/badge)](https://api.securityscorecards.dev/projects/github.com/aws-powertools/powertools-lambda-typescript)
+![NPM Downloads](https://img.shields.io/npm/dw/%40aws-lambda-powertools%2Ftracer?style=flat-square&label=Tracer)
```
dreamorosi, line 9: I would remove these, we have 3 more packages and soon we'll have more. I think listing only the core ones reinforces the notion that we only do observability - which we are trying to get away from.

In the future we could use a custom badge that aggregates all of them, but overall I think it's ok not to show this for now.

README.md
```diff
@@ -1,6 +1,15 @@
 # Powertools for AWS Lambda (TypeScript) <!-- omit in toc -->
 
-[![Join our Discord](https://dcbadge.vercel.app/api/server/B8zZKbbyET)](https://discord.gg/B8zZKbbyET)
+[![Build](https://github.com/aws-powertools/powertools-lambda-typescript/actions/workflows/pr-run-linting-check-and-unit-tests.yml/badge.svg)]([https://github.com/aws-powertools/powertools-lambda-typescript/actions/workflows/pr-run-linting-check-and-unit-tests.yml])
```
dreamorosi, line 3: I would also remove this one, and rename back to the original name the workflow.

The workflow linked here is the one that runs on new PRs or when new code is pushed to an existing PR. The one that is run when a PR is merged would be [this other](https://github.com/aws-powertools/powertools-lambda-typescript/blob/01d3dc4f0e25edbf1cc86e34b09e1a9fbc4b281d/.github/workflows/on_merged_pr.yml) one.


Overall I think this is a bit redundant anyway since the same info on the last merged changes on `main` is right next to the files:
Screenshot 2024-02-14 at 19 24 17

dreamorosi (changes requested):
Hi @hjgraca thank you for the PR!

I left comments related to the "build" and "downloads" badges, which I think we should remove and potentially reconsider in a future PR if applicable.

README.md
am29d: We initially discussed the option of adding all packages, but it would be too much information overload. I agree with you on an unintentional signal for "observability only", and change my mind to remove the badges for downloads for now.

README.md
am29d: Imho it does not hurt to have redundant information, there are no costs for having this badge. Overall, we have many places where the information is redundant, and it's often desirable. 

Regarding the name, I think we should go a step further and rename the workflow to "Quality check", to abstract the steps, instead of `lint-unit-e2e-fuzzy-special-checks`. Then we can keep it for both PR merge semantic and the badge.

dreamorosi: OK, but it's not just about the name.

The workflow linked in this badge runs on PR so showing the badge doesn't make sense since it'll show the status of the last check ran on any of the open PRs.

If we want to add the badge, then we need link it to a workflow that runs on the latest code currently on `main`.

am29d: > If we want to add the badge, then we need link it to a workflow that runs on the latest code currently on main.

Ah, I missed this point, sorry. You are right, we should only include the latest merge to main.

dreamorosi: Np, I was also talking about two things in the same comment and could've been clearer.

So I'm on board with adding the badge and using a more meaningful name like you suggested, however I think we should address this in a separate PR.

After looking at the workflows again, we don't really have a true "code quality" workflow that runs on merged PRs - which is a problem on its own.

To verify this, I just merged this Dependabot PR, and these below were the workflows that run as a result:
image

- Scorecard
- Release drafter
- Record PR details + label related issue with `status/pending-release`

The Python repo on the other hand, has a workflow called `quality-check.yml` ([here](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/.github/workflows/quality_check.yml)) that runs both on PRs and `push` events (aka merge commits on `main` in our case).

The closest thing we have, is the "On PR update" workflow ([here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/.github/workflows/pr-run-linting-check-and-unit-tests.yml#L3-L5)), that in turn calls the test & lint reusable workflow.

I don't have a strong opinion on whether we should modify the "On PR update" workflow to behave like the "Quality Check" in Python, or leave that alone & add our own version of "Quality Check" that runs on `push` events on `main`. I also don't know if we should run integration tests whenever we merge on `main` like Python does (possibly yes?).

I do think however this is a larger discussion and in the interest of keeping the scope of this PR focused, I suggest we remove this one badge from this PR and open a dedicated issue that we can pick up right away after merging this PR.

Once we have the appropriate workflows in place, they're documented, and know they're working consistently then we can add the badge back to the README.

What do you both think?

hjgraca: I agree, this needs a bigger discussion that is outside the scope of this PR. Will remove the badge and undo the workflow rename

dreamorosi (approved):
Thank you for working on this PR and addressing all the comments!

Issue #1103, opened 2022-09-29
Docs: refresh examples


ijemmy:
## Description of the improvement

**Summary of the proposal**
The repository contains - at present - examples in CDK and SAM that showcase how to use the Powertools utilities available. The new idempotency utility should be part of these example.

**How, where did you look for information**
In the CDK & SAM examples.

**Missing or unclear documentation**
N/A

**Improvement**
Include examples of the new utility.

### Related existing documentation 
https://github.com/awslabs/aws-lambda-powertools-typescript/tree/main/examples

### Related issues, RFCs

#447

dreamorosi:
As part of this issue I was thinking of addressing a couple of points:
a. add the Idempotency utility to the examples
b. refactor the part of the example that covers Parameters
c. add Batch Processing
d. improve the README

### a. Add Idempotency

As part of this issue we should add something about the Idempotency utility. Based on the current sample I think the most logical place to add the utility would be in the [put-item function](https://github.com/aws-powertools/powertools-lambda-typescript/blob/02cd84bbe8505345c7f6bf8d7cdf123d3608daa0/examples/cdk/functions/put-item.ts) as specifically modify the handler in such a way that the Idempotency utility is used to ensure that the same item is inserted only once.

Currently the code that inserts the item is part of the handler and looks like this:

```ts
const body = JSON.parse(event.body);
const { id, name } = body;

await docClient.send(
  new PutCommand({
    TableName: tableName,
    Item: {
      id,
      name,
    },
  })
);
```

I propose to extract the logic above to its own function and use the `makeIdempotent` function wrapper on it using all or some of the arguments.

### b. Refactor Parameters

A couple of months ago we added the Parameters utility to the examples in conjunction with the GA of the utility. At the time I recall us struggling to find a way to use it and eventually we settled on creating a second REST API endpoint and using the utility to retrieve an arbitrary value which is then returned in the response.

After looking at the usage again, and also listening [customer feedback](https://github.com/aws-powertools/powertools-lambda-typescript/discussions/1659), I think there's an opportunity to improve this area and use the utility for something more useful.

My first choice would have been to store the name of the Idempotency table (from the previous point) in a SSM parameter and then use the Parameters utility to retrieve it. This however is not possible at the moment given that we must instantiate the `DynamoDBPersistenceLayer` outside of the function handler and using `await` at the top-level requires ESM.

For context, and to maybe revisit this once we finally support ESM in the coming months, this is what I mean:

[...]

PR #2117, opened 2024-02-20
feat!: version 2, the ESM update

dreamorosi:

This PR includes all the changes for the next major version of Powertools for AWS Lambda (TypeScript).

The changes in the PR support the features described [here](https://docs.powertools.aws.dev/lambda/typescript/latest/upgrade/), however the size of the PR is due to two factors:
- this is a long running branch (6+ months)
- to allow for ESM+CJS build we had to touch almost all files under `packages/*` and modify all imports and exports

The PR also includes changes to the docs & related code snippets.

Many of the changes in this branch have been reviewed in isolation when they were merged in the `feat/v2` branch.

### Related issues, RFCs

**Issue number:** #1714

## Checklist

### Breaking change checklist

***Is it a breaking change?:*** YES

- [x] I have documented the migration process

---


heitorlessa:
check PR title :D

dreamorosi:
> check PR title :D

Open to change the description, but the `feat!` part is intentional and it's what will trigger a major bump.


(sorry I mistakenly clicked the edit button rather than the quote reply one from mobile. Didn't mean to edit your comment!)

heitorlessa:
Aha didn't know the `!`. By the power of grayskull I'll probably trust these 404 files are fine.

dreamorosi:
I have just run the integration tests on this branch & they are green: https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/7994389117

> By the power of grayskull I'll probably trust these 404 files are fine.

giphy

examples/cdk/functions/common/powertools.ts
dreamorosi, line 1: For now both CDK & SAM examples have been updated minimally and only so that they can be deployed.

We have another issue (#1103) that I want to pick up in the next iteration to actually leverage some of the new features like top-level await.

packages/batch/jest.config.cjs
```diff
@@ -5,6 +5,9 @@ module.exports = {
   },
   runner: 'groups',
   preset: 'ts-jest',
+  moduleNameMapper: {
+    '^(\\.{1,2}/.*)\\.js$': '$1',
+  },
```
dreamorosi, line 10: This is addition to all the jest config files is needed because we are now importing packages via relative path making the extension explicit.

This is a Typescript requirement and not a style choice, read [here for more info](https://www.totaltypescript.com/relative-import-paths-need-explicit-file-extensions-in-ecmascript-imports) on why.

The new `moduleNameMapper` key tells Jest how to interpret these new paths, and it tells it to remove the `.js` extension. This is because Jest doesn't work with ESM at all so we are basically treating the tests as CJS. In the near/medium future we're going to probably move away from Jest entirely in favor of Vite, but for now this is how we make it work.

packages/batch/package.json
```diff
@@ -13,25 +13,50 @@
...
-    "test:e2e:nodejs14x": "echo 'Not Implemented'",
     "test:e2e:nodejs16x": "echo 'Not Implemented'",
     "test:e2e:nodejs18x": "echo 'Not Implemented'",
     "test:e2e:nodejs20x": "echo 'Not Implemented'",
     "test:e2e": "echo 'Not Implemented'",
     "watch": "jest --watch",
-    "build": "tsc --build --force",
+    "build:cjs": "tsc --build tsconfig.json && echo '{ \"type\": \"commonjs\" }' > lib/cjs/package.json",
+    "build:esm": "tsc --build tsconfig.esm.json && echo '{ \"type\": \"module\" }' > lib/esm/package.json",
+    "build": "npm run build:esm & npm run build:cjs",
```
dreamorosi, line 23: While before we were building a single output under `lib`, now we are generating two of them: one for CJS and one for ESM, respectively under `lib/cjs/*` and `lib/esm/*`.

To do this, we are now maintaining two `tsconfig.json` files, one for CJS and another for ESM, and using them to create their respective build.

In addition to this, we are also dynamically creating a `package.json` file under each one of the versions. This file contains a single section that sets the `type` to either `commonjs` or `module` (ESM). This is what allows Node.js, TS, and bundlers to recognize the folder as a module root.

packages/batch/src/index.ts
dreamorosi, line 1: This is an example of how we have changed our export files to be more specific & intentional, which will facilitate tree shaking and help us keep our API surface tight.

Instead of exporting (aka making available) everything like we did before, we are specifying only the things that are meant to be used externally (aka by packages that use this as dependency).

These exports can be used, but are not primarily intended, for importing things within the package itself. For example if `packages/batch/src/foo.ts` needs to import a utility from `packages/batch/src/bar.ts` they should do so directly rather than relying on the default exports.

We were doing this already for the most part across the project, but I have spent a good amount of time reviewing every single import and making sure this is the case. Moving forward, if you see instances of this when the import could be relative within a package, please fix it.

Additionally, we are also removing some of the exports entirely as they have been moved in their own subpath (i.e. `./types`).

packages/batch/tsconfig.esm.json
dreamorosi, line 1: This is an example of one of the `tsconfig.json` files we are maintaining in each package. This one specifically is for the ESM build and it extends the one found at the root of the project.

The main tasks for this config are to:
- bring in the larger config via the `extends` keyword
- set paths for this specific package relative to the package root (i.e. `packages/batch`).

Note that we are also setting a `tsBuildInfoFile`. This is needed because we need to maintain two of these - one for ESM and another for CJS. This is a newer feature that we are introducing with this version which basically allows TS to maintain a build cache and perform incremental builds.

This should help during development as it allows us to build only files that have changed. In the future I would like to explore storing these aritifacts in the GitHub Actions cache so that CI builds can benefit too, but I haven't done any work on that direction and this is mainly a forward looking change to give us options in case our codebase grows in the future.

packages/batch/tsconfig.json
dreamorosi, line 1: Everything that I said in [this comment](https://github.com/aws-powertools/powertools-lambda-typescript/pull/2117/files#r1499280448) applies here, except that this file is the corresponding one but for CJS.

packages/commons/src/awsSdkUtils.ts
dreamorosi, line 1: The commons package was somewhat messy as we tended to use it as a catch-all place to put utilities that didn't fit anywhere else and were shared across two or more utilities.

This is still the case and will be for the foreseeable future, however with this PR I have taken the liberty to reorganize the file structure according to their use, so that we can neatly export things using subpaths.

packages/commons/src/samples/resources/events/aws/apigateway-authorizer.json
dreamorosi, line 1: Before this version we were shipping all the AWS events JSON files. This was not documented anywhere and most customers would not have them in their final deployment package thanks to bundling.

Some customers however complained (rightfully so) about them being there so we are removing them for good.

This had to be done in a major version even though we don't have any evidence of them being used anywhere by anyone.

packages/logger/src/types/Log.ts
dreamorosi, line 1: Some of the types have been slightly renamed or otherwise consolidated to reduce the amount of files.

tsconfig.json
dreamorosi, line 1: This is the main `tsconfig.json` file that is used & extended throughout the other utilities in the repo. Since it's the default one it's pretty much the same as before and most of the new features are in the `tsconfig.esm.json` one.

tsconfig.esm.json
dreamorosi, line 1: This is the main `tsconfig.json` that serves as base for the ESM ones that extend it in the project.

As you can see it inherits the configs in the main one (the CJS one) but overwrites some compiler options. These are the ones that are ESM specific.

docs/index.md
```diff
@@ -26,7 +26,7 @@ You can use Powertools for AWS Lambda (TypeScript) in both TypeScript and JavaSc
 
 You can install Powertools for AWS Lambda (TypeScript) using one of the following options:
 
-* **Lambda Layer**: [**arn:aws:lambda:{region}:094274105915:layer:AWSLambdaPowertoolsTypeScript:29**](#){: .copyMe}:clipboard:
+* **Lambda Layer**: [**arn:aws:lambda:{region}:094274105915:layer:AWSLambdaPowertoolsTypeScriptV2:1**](#){: .copyMe}:clipboard:
```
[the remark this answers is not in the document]
am29d: We don't need to update the script.

am29d (commented):
Only few small questions.

examples/cdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cdk-sample",
-  "version": "1.18.1",
+  "version": "1.18.0",
```
dreamorosi, line 3:
```suggestion
  "version": "1.18.1",
```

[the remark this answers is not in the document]
dreamorosi: Good one, and you were right, it was probably because of a merge conflict gone wrong.

I have reverted the change!

package.json
```diff
@@ -66,7 +66,6 @@
     "lerna": "^8.1.2",
     "lint-staged": "^15.2.2",
     "prettier": "^3.2.5",
-    "rimraf": "^5.0.5",
```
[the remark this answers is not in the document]
dreamorosi: Before this PR we had a command in each package that explicitly deleted the contents of the `packages/*/lib/*` folder before doing a build ([example](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/metrics/package.json#L25)).

Starting from this PR we have enabled incremental builds (in the main `tsconfig.json` files), so cleaning the build would be counterproductive. 

Given the above, we no longer use the package and so I removed it to have one last dependency.

dreamorosi:
Thanks a lot for the review @am29d, really appreciate you pouring time & effort to go through it.

I have addressed both your comments. Please feel free to start resolving any open conversion that you're aligned with & have no more questions on.

am29d (approved):
Incredible work Andrea, thank you for your time, energy and endurance to make V2 possible. Looking forward to finally release it!

Issue #1645, opened 2023-08-02
Maintenance: jmespath utility


rjmackay:
### Summary

The recently added idempotency feature adds a dependency on jmespath. While that package has many downloads, it hasn't been updated in 2 years. 

This seems like a concern for the powertools tenets:

> Keep it lean. Additional dependencies are carefully considered for security and ease of maintenance, and prevent negatively impacting startup time.

The key selection that jmespath is used for could easily be implemented without this dependency (see https://github.com/aws-powertools/powertools-lambda-typescript/issues/1644)

### Why is this needed?

jmespath seems unmaintained and not ideal for a production dependency.

### Which area does this relate to?

JMESPath

### Solution

- Remove jmespath and replace with a much simpler method of selecting the idempotency key
- Replace jmespath with a more well maintained implementation? (https://www.npmjs.com/package/@metrichor/jmespath is slightly more recent and in typescript)
- Fork and own jmespath?

### Acknowledgment

- [X] This request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda-typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

PR #2135, opened 2024-02-24
feat(jmespath): add JMESPath utility

dreamorosi:

This PR introduces a new Powertools utility for JMESPath which will be published under `@aws-lambda-powertools/jmespath`.

The utility is primarily aimed at being a dependency of the Idempotency utility but can be used as standalone as well as in future utilities that revolve around JSON like the Validation utility (#508).

As discussed in one of the linked issues (#1645) the [official JMESPath module](https://www.npmjs.com/package/jmespath) does not support custom functions ([example in Powertools Python](https://docs.powertools.aws.dev/lambda/python/latest/utilities/jmespath_functions/#built-in-jmespath-functions)) and all other existing versions and forks are largely unmaintained or owned by individuals.

To tackle the concerns and limitations above, as well as providing a modern implementations fully written in TypeScript we decided to implement our own version based of the Python [jmespath module](https://pypi.org/project/jmespath/) which is MIT licensed.

The JMESPath specification offers a [compliance test suite](https://jmespath.org/compliance.html) that is used by all libraries who want to offer spec-compliance, the utility we are introducing in this PR adopts that same suite and thus can be considered fully spec compliant. 

In addition to the utility and corresponding documentation, this PR also introduces a new helper function that allows to extract data from envelopes when using JSON. Together with this, it introduces also a series of envelopes from common data sources similar to [what found in Powertools Python](https://docs.powertools.aws.dev/lambda/python/latest/utilities/jmespath_functions/#extracting-data).

```ts
import {
  extractDataFromEnvelope,
  SQS,
} from '@aws-lambda-powertools/jmespath/envelopes';
import type { SQSEvent } from 'aws-lambda';

const logger = new Logger();

type MessageBody = {
  customerId: string;
};

export const handler = async (event: SQSEvent): Promise<void> => {
  const records = extractDataFromEnvelope<Array<MessageBody>>(event, SQS);
  for (const record of records) {
    // records is now a list containing the deserialized body of each message
    const { customerId } = record;
    logger.appendKeys({ customerId });
  }
};
```

Note that while there are similarities with the Parser utility being implemented in parallel with this utility, the main difference is the way the problem is approached. While the Parser utility uses Zod to validate and parse a payload using Zod schemas, the `extractDataFromEnvelope()` function and its envelopes is instead used to _simply_ extract a subset of data from a JSON payload while also not applying any kind of validation.

Together with these utilities we also provide some custom JMESPath functions that complement the existing JMESPath [built-in functions](https://jmespath.org/specification.html#built-in-functions). Using these functions customers can reach into encoded data structures within their JSON payload.

For example, when dealing with an event coming to API Gateway or SQS, the body of the message is a stringified version of a JSON. To access it customers have to apply their own parsing, but with these custom functions they do so within the same JMESPath expression, and potentially even extract data from within the parsed object.

```ts
import type { APIGatewayEvent } from 'aws-lambda';
import { randomUUID } from 'node:crypto';
import {
  makeIdempotent,
  IdempotencyConfig,

  tableName: 'IdempotencyTable',
});

export const handler = makeIdempotent(
  async (event: APIGatewayEvent) => {
    const body = JSON.parse(event.body || '{}');
    const { user, productId } = body;

    const result = await createSubscriptionPayment(user, productId);

    return {
      statusCode: 200,
      body: JSON.stringify({
        paymentId: result.id,
        message: 'success',
      }),
    };
  },
  {
    persistenceStore,
    config: new IdempotencyConfig({
      eventKeyJmesPath: 'powertools_json(body)',
    }),
  }
);

declare function createSubscriptionPayment(
  user: string,
  productId: string
): Promise<{ id: string; message: string }>;
```

Finally, to support this utility I have made some small addition to the commons utility. Some of these, like the utilities to parse to and from base64 encoding can and will be adopted by the Parameters utility as well, allowing us to drop a dependency on a deprecated AWS SDK package.

This utility will be adopted by the Idempotency utility in a future PR. Some of the features mentioned above will allow customers to select parts of payloads to be used for idempotency that were not accessible before, like the SQS example mentioned above.

### Related issues, RFCs

**Issue number:** #1645

## Checklist

### Breaking change checklist

---


docs/requirements.txt
dreamorosi, line 1: I am not sure why this file is being shown here in the diff, there's no difference I can see.

I also tried to copy the raw version of the file present on `main` and paste it back in this branch but it's still here.

I think it's due to the rebases, please ignore it.

dreamorosi:
I closed this PR in favor of: #2187, #2189, #2192, #2212, #2213, #2214, #2259, #2264, #2266, #2265, #2271

PR #1838, opened 2024-01-08
feat(parser): add types for built-in schemas

am29d:
## Description of your changes

This PR adds type exports for the built in schemas. I have slightly changed the names by adding `Event` suffix. This is similar to the data classes used in python and supports the typing of the `event` object of the handler.

We don't need all the schema object to be reexported as types, because `z.infer` will catch all the nested structure implicitly. 

### Related issues, RFCs

**Issue number:** closes #1837 

## Checklist

- [x] [My changes meet the tenets criteria](https://docs.powertools.aws.dev/lambda/typescript/#tenets)
- [x] I have performed a *self-review* of my own code
- [x] I have *commented* my code where necessary, particularly in areas that should be flagged with a TODO, or hard-to-understand areas
- [ ] I have made corresponding changes to the *documentation*
- [x] My changes generate *no new warnings*
- [ ] I have *added tests* that prove my change is effective and works
- [x] The PR title follows the [conventional commit semantics](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/.github/semantic.yml#L2)

### Breaking change checklist

***Is it a breaking change?:*** NO

- [ ] I have documented the migration process
- [ ] I have added, implemented necessary warnings (if it can live side by side)

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Disclaimer**: We value your time and bandwidth. As such, any pull requests created on non-triaged issues might not be successful.

PR #2179, opened 2024-03-08
chore(parser): add parser subpath exports to package.json

### LINKED ISSUE #2177: Maintenance: add subpath exports to package.json
### Summary

We currently don't export subpaths for `envelopes`, `schemas` and `types` in our `package.json`, and we should add them, so customers can import them.

### Why is this needed?

So customers can granularly import schemas and envelopes following the same standard that we have introduced with in v2.

_No response_

### Solution

_No response_

### Acknowledgment

### Future readers

am29d:
This PR adds exports to `package.json` following the convention we use since v2. 

### Related issues, RFCs

**Issue number:** closes #2177 

## Checklist

### Breaking change checklist

---


packages/parser/package.json
```diff
@@ -37,6 +37,38 @@
...
+    "./middleware": {
+      "require": "./lib/cjs/middleware/parser.js",
+      "import": "./lib/esm/middleware/parser.js"
+    },
+    "./schemas": {
+      "require": "./lib/cjs/schemas/index.js",
+      "import": "./lib/esm/schemas/index.js"
+    },
+    "./envelopes": {
+      "require": "./lib/cjs/index.js",
```
dreamorosi, line 50: I think we're missing a level (`envelopes`) in the path here.

packages/parser/package.json
```diff
@@ -37,6 +37,38 @@
...
+      "import": "./lib/esm/index.js"
+    }
```
dreamorosi, line 52: We're also missing a `exports.types` to match the `typeVersions.*.types` one.

packages/parser/src/schemas/index.ts
dreamorosi, line 1: I've spot checked some of the files and they all seem to have only export, which makes the `*` safe at this point.

I'd still however go with an explicit named export like `export { fooBar } from './alb.js';` to avoid future issues.

With the `*` export, if we add any exported module to any one of the files it'll end up exposed & potentially in customers bundles.

I think overall being explicit > than terse, even at the risk of being pedantic here.

packages/parser/src/envelopes/index.ts
dreamorosi, line 1: Same comment I left for `packages/parser/src/schemas/index.ts` applies here.

packages/parser/src/types/index.ts
```diff
@@ -0,0 +1,3 @@
+export type { ParserOptions } from './ParserOptions.js';
+export type { Envelope } from './envelope.js';
+export * from './schema.js';
```
dreamorosi, line 3: Besides the `*` (see comment under `packages/parser/src/schemas/index.ts`), why are we exporting the schemas here?

These are Zod objects so it's better if we don't mix them with pure TS types. One benefit of this is that in the future when we put this in the Lambda layer we can cut off anything under `packages/parser/src/types` and make the layer smaller and more performant.

As a rule of thumb, if we're not doing `export type ...`, it shouldn't be under `pacakges/*/src/types` but either be under the default export or its own subpath.

dreamorosi (changes requested):
Left a couple of comments, looking forward to merge this once they're addressed.

Thank you for breaking down the items in a separate PR!

packages/parser/tests/unit/parser.decorator.test.ts
```diff
@@ -6,11 +6,11 @@
 
 import { LambdaInterface } from '@aws-lambda-powertools/commons/lib/esm/types';
```
dreamorosi, line 7: nit: this an the next line could be `import type ...`.

packages/parser/src/schemas/index.ts
am29d: agree on explicit exports, I was too lazy, changed in `schemas` and `envelopes`

packages/parser/src/types/index.ts
am29d: Probably the wording is misleading, these are not schema types, but types that we infer from the built-in schema, that can use during development (#1838). If customers bring `@types/aws-lambda` there might be a mismatch between development and runtime. 

```ts
import type { ALBEvent } from "@aws-lambda-powertools/parser/types";

@parser({ schema: AlbSchema })
public async handler(event: ALBEvent, _context: Context): Promise<unknown> {
  return event;
}
```

dreamorosi: Ok my bad here.

I saw `./schema.js` and assumed it was `packages/parser/src/schema` rather than `packages/parser/src/types/schema.ts`.

Makes sense, I'd just remove the `*` then but the import is fine.

am29d:
I have fixed the imports in tests to reach the target coverage and also removed unused exports, good thing to be explicit!

packages/parser/package.json
am29d: Added types block

dreamorosi (approved):
Thanks for this PR and for the interactions.

Issue #2188, opened 2024-03-08
Feature request: add helper function to decode base64 encoded strings


dreamorosi:
### Use case

Across the project we decode base64 encoded strings in two places: the Parameters utility when fetching configurations from AppConfig, and the upcoming JMESPath utility when using the `powertools_base64()` custom function.

To avoid code duplication we should extract this functionality in the commons package so that it can be reused by both consumers.

### Solution/User Experience

The utility should have a small footprint and an isolated export so that it can be tree shaken and removed from bundle when not used.

### Acknowledgment

- [X] This feature request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

PR #2189, opened 2024-03-08
improv(commons): expand type utils functions

dreamorosi:
### Related issues, RFCs

**Issue number:** #2188

## Checklist

### Breaking change checklist

---


packages/commons/src/typeUtils.ts
```diff
@@ -0,0 +1,177 @@
+/**
+ * Returns true if the passed value is a record (object).
+ *
+ * @param value The value to check
+ */
+const isRecord = (value: unknown): value is Record<string, unknown> => {
```
am29d, line 6: the keys in the `Record` doesn't have to strings only, afaik we can have `string | number | symbol`

packages/commons/src/typeUtils.ts
```diff
@@ -0,0 +1,177 @@
...
+const isIntegerNumber = (value: unknown): value is number => {
+  return isNumber(value) && Number.isInteger(value);
+};
+
+/**
+ * Check if a value is truthy.
+ *
+ * @param value The value to check
+ */
+const isTruthy = (value: unknown): boolean => {
```
am29d, line 45: `NaN` and functions would not pass here, but I am not sure we need them, just asking if left out this edge cases on purpose.

packages/commons/tests/unit/typeUtils.test.ts
```diff
@@ -0,0 +1,329 @@
...
+        // Act
+        const result = isTruthy(value);
+
+        // Assess
+        expect(result).toBe(true);
+      }
+    );
+
+    it.each(['', 0, false, [], {}, Symbol])(
+      'returns true when the passed value is falsy',
```
am29d, line 64:
```suggestion
      'returns false when the passed value is falsy',
```

packages/commons/src/typeUtils.ts
```diff
@@ -0,0 +1,177 @@
...
+  }
+};
+
+/**
+ * Check if a value is null.
+ *
+ * @param value The value to check
+ */
+const isNull = (value: unknown): value is null => {
+  return Object.is(value, null);
```
am29d, line 67: what is the difference between `value === null` and `Object.is`? It looks like they yield same result, is it rather a preference?

am29d (changes requested):
Nice packaged PR, just a minor question on the `NaN` edge case.

packages/commons/src/typeUtils.ts
dreamorosi: True.

I don't think we need to include them as they're not JSON serializable anyway.

If in the future we need to check truthiness of these values we can still expand this function.

packages/commons/src/typeUtils.ts
dreamorosi: In the context of `null` the two notations are exactly equivalent, there are [other cases](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is#description) where `Object.is` is preferable but this is not one of them.

In this case it was just preference. In my mind it _reads_ a bit more clear since it's closer to plain English.

packages/commons/src/typeUtils.ts
am29d: > they're not JSON serializable anyway. 

makes sense

am29d (dismissed):

am29d (dismissed):

Issue #2206, opened 2024-03-12
Feature request: JMESPath built-in functions


dreamorosi:
### Use case

As part of the JMESPath utility we need to implement the logic needed to support all the built-in functions (i.e. `avg`, `max`, `min`, `ceil`, etc.) defined in the [JMESPath specification](https://jmespath.org/specification.html#built-in-functions).

### Solution/User Experience

Customers should be able to use these functions in their expressions like:

```ts
const payload = {
  grades: [1, 5, 2, 10],
};

const maxGrade = search('max(grades)', payload); // 10
const minGrade = search('min(grades)', payload); // 1
const avgGrade = search('avg(grades)', payload); // 4.5
const roundedUpAvgGrade = search('ceil(avg(grades))', payload); // 5
const roundedDownAvgGrade = search('floor(avg(grades))', payload); // 4
// ...
```

All the built-in functions defined in the spec should be supported.

All types, classes, and functions should be documented and exposed to customers via API docs.

### Acknowledgment

- [X] This feature request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

PR #2259, opened 2024-03-21
feat(jmespath): add built in functions

dreamorosi:

This PR adds a new `Functions` class to the JMESPath utility. This class holds the implementation for all the built-in functions listed in the [JMESPath specification](https://jmespath.org/specification.html#built-in-functions) as well as a class method decorator to validate function arguments.

Customers are not expected to use this class directly, instead the logic in it will allow them to use function expressions in their JMESPath expressions, for example:

```ts
const payload = {
  grades: [1, 5, 2, 10],
};

const maxGrade = search('max(grades)', payload); // 10
const minGrade = search('min(grades)', payload); // 1
const avgGrade = search('avg(grades)', payload); // 4.5
const roundedUpAvgGrade = search('ceil(avg(grades))', payload); // 5
const roundedDownAvgGrade = search('floor(avg(grades))', payload); // 4
```

This part of the code has been implemented as a class purposefully so that we, as well as customers, can extend the class and add new functions to the utility.

To facilitate this, the `Functions` class provides also a class method decorator defined as static method. Using this decorator a method implementing a JMESPath function can have its arguments validated at runtime for types, number, and whether or not the [function is variadic](https://en.wikipedia.org/wiki/Variadic_function).

For example:

```ts
class CustomFunctions extends PowertoolsFunctions {
  @PowertoolsFunctions.signature({ // (1)!
    argumentsSpecs: [['string']],
    variadic: false,
  })
  public funcDecodeBrotliCompression(value: string): string { // (2)!
    const encoded = fromBase64(value, 'base64');
    const uncompressed = brotliDecompressSync(encoded);

    return uncompressed.toString();
  }
}
```

To keep the PR focused I am adding only the base functions, however the next PR will add some Powertools functions that are equivalent to the [ones found in Python](https://docs.powertools.aws.dev/lambda/python/latest/utilities/jmespath_functions/#built-in-jmespath-functions).

### Related issues, RFCs

**Issue number:** #2206

## Checklist

### Breaking change checklist

---

### DIFF
--- a/packages/jmespath/src/Functions.ts
+++ b/packages/jmespath/src/Functions.ts
@@ -1,4 +1,605 @@
-// This is a placeholder for the real class. The actual implementation will be added in a subsequent PR.
-export class Functions {
-  public iAmAPlaceholder = true;
+import type {
+  JSONArray,
+  JSONObject,
+  JSONValue,
+} from '@aws-lambda-powertools/commons/types';
+import {
+  getType,
+  isNumber,
+  isRecord,
+} from '@aws-lambda-powertools/commons/typeutils';
+import type { Expression } from './Expression.js';
+import { JMESPathTypeError } from './errors.js';
+import type {
+  FunctionSignatureDecorator,
+  FunctionSignatureOptions,
+  JSONObject as JSONObjectType,
+} from './types.js';
+import { arityCheck, typeCheck } from './utils.js';
+
+/**
+ * A class that contains the built-in JMESPath functions.
+ *
+ * The built-in functions are implemented as methods on the Functions class.
+ * Each method is decorated with the `@Function.signature()` decorator to enforce the
+ * arity and types of the arguments passed to the function at runtime.
+ *
+ * You can extend the Functions class to add custom functions by creating a new class
+ * that extends Functions and adding new methods to it.
+ *
+ * @example
+ * ```typescript
+ * import { search } from '@aws-lambda-powertools/jmespath';
+ * import { Functions } from '@aws-lambda-powertools/jmespath/functions';
+ *
+ * class MyFunctions extends Functions {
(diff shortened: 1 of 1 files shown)


packages/jmespath/src/Functions.ts
```diff
@@ -1,4 +1,605 @@
...
+ * that extends Functions and adding new methods to it.
+ *
+ * @example
+ * ```typescript
+ * import { search } from '@aws-lambda-powertools/jmespath';
+ * import { Functions } from '@aws-lambda-powertools/jmespath/functions';
+ *
+ * class MyFunctions extends Functions {
+ *   ⁣@Functions.signature({
+ *     argumentsSpecs: [['number'], ['number']],
```
am29d, line 37: shouldn't the argument spec be `array-number`?

am29d (commented):
What happens when the argumentSpecs do not match with the signature?

packages/jmespath/src/Functions.ts
dreamorosi: In this case, we are specifically asking for at least 1 and at most 2 arguments (`variadic: true`) both of which must be `number`. In the method we use `Array<number>` because we are getting the arguments (`args` plural) of the method.

If we put `array-number` then it'd be 1 single argument with type `Array<number>`.

For example, we mean this: `my_method(1, 2)` and not this `my_method([1, 2])`.

dreamorosi:
> What happens when the argumentSpecs do not match with the signature?

It will throw a runtime `JMESPathTypeError`, all the logic for that is implemented from [this line onwards](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/jmespath/src/utils.ts#L127).

am29d (approved):

Issue #1619, opened 2023-07-20
Feature Request: support for tracing NodeJs 18 native `fetch` calls


RaphaelManke:
### Expected Behaviour

In our Team we use the native nodeJS 18 fetch module to make api calls. 

We were expecting that fetch calls would be traced the same as they where when we used `node-fetch`. 

### Current Behaviour

The fetch calls are not traced whereas the node-fetch calls are traced. 

The x-ray traces show only the api calls made with node-fetch. 

![image](https://github.com/aws-powertools/powertools-lambda-typescript/assets/15867688/b3199dea-1b45-448a-9730-6030346e874d)

### Code snippet

```ts
import { LambdaInterface } from "@aws-lambda-powertools/commons";
import { Tracer } from "@aws-lambda-powertools/tracer";
import nodeFetch from "node-fetch";

const tracer = new Tracer();

class Lambda implements LambdaInterface {
  // Decorate your handler class method

  @tracer.captureMethod()
  private async nativeFetch() {
    const response = await fetch("https://api.github.com/users/aws-samples");
    const json = await response.json();
    return json;
  }

  @tracer.captureMethod()
  private async nodeFetch() {
    const response = await nodeFetch(
      "https://api.github.com/users/aws-samples"
    );
    const json = await response.json();
    return json;
  }

  public async handler(_event: unknown, _context: unknown): Promise<void> {
    await this.nativeFetch();
    await this.nodeFetch();
  }
}

const handlerClass = new Lambda();
export const handler = handlerClass.handler.bind(handlerClass); //

```

### Steps to Reproduce

1. clone the bug repo https://github.com/RaphaelManke/aws-powertools-node-18-tracing-fetch-bug
2. npm install
3. npm run cdk deploy
4. invoke the lambda

### Powertools for AWS Lambda (TypeScript) version

1.11.1

### AWS Lambda function runtime

18.x

### Packaging format used

npm

am29d:
Hey @RaphaelManke, thanks for raising the issue. After a closer look it seems like this is x-ray SDK specific, since we are using it. The related issue and discussion is here https://github.com/aws/aws-xray-sdk-node/issues/531 .

dreamorosi:
We have been tracking this for a while and at the moment we are blocked by the X-Ray SDK that we use for our Tracer utility, which does not support this.

[...]

PR #2293, opened 2024-03-28
feat(tracer): instrument fetch requests

dreamorosi:

This PR introduces to Tracer the ability to instrument and trace requests made with the `fetch` global module.

Given that the `fetch` implementation present in Node.js is based on `undid`, the implementation added in PR uses `diagnostics_channel`, which [according to `undici`'s docs is the preferred way](https://github.com/nodejs/node/blob/27493a1dd73f41c07f04ca5d89d001e340a9feec/deps/undici/src/docs/api/DiagnosticsChannel.md?plain=1#L6) to instrument requests made using the modules. 

This implementation - based on a pub/sub model - allows us to completely decouple the tracing logic from the module, which removes the need of monkey patching or wrapping the module with our implementation. The result consists in a series of event handlers that are subscribed to certain channels as soon as the `Tracer` class is instantiated.

To support the implementation the PR also extends some types of the `Subsegment` class as well as adding a handful of types and helpers. This opens the door to also replace the instrumentation of the `http` and `https` modules with a similar pattern in the future since these modules also publish events to [their own `diagnostics_channel`](https://nodejs.org/docs/latest/api/diagnostics_channel.html#http).

A special thanks goes to @RaphaelManke, who suggested this idea and created the first proof of concept for it.

### Related issues, RFCs

**Issue number:** #1619

## Checklist

### Breaking change checklist

---

### DIFF
--- a/packages/tracer/src/Tracer.ts
+++ b/packages/tracer/src/Tracer.ts
@@ -16,7 +16,7 @@ import type {
   CaptureMethodOptions,
 } from './types/Tracer.js';
 import { ProviderService } from './provider/ProviderService.js';
-import type { ProviderServiceInterface } from './types/ProviderServiceInterface.js';
+import type { ProviderServiceInterface } from './types/ProviderService.js';
 import type { Segment, Subsegment } from 'aws-xray-sdk-core';
 import xraySdk from 'aws-xray-sdk-core';
 const { Subsegment: XraySubsegment } = xraySdk;
@@ -153,6 +153,7 @@ class Tracer extends Utility implements TracerInterface {
     this.provider = new ProviderService();
     if (this.isTracingEnabled() && this.captureHTTPsRequests) {
       this.provider.captureHTTPsGlobal();
+      this.provider.instrumentFetch();
     }
     if (!this.isTracingEnabled()) {
       // Tell x-ray-sdk to not throw an error if context is missing but tracing is disabled
--- a/packages/tracer/src/provider/ProviderService.ts
+++ b/packages/tracer/src/provider/ProviderService.ts
@@ -21,6 +21,60 @@ const {
   setLogger,
 } = xraySdk;
 import { addUserAgentMiddleware } from '@aws-lambda-powertools/commons';
+import { subscribe } from 'node:diagnostics_channel';
+
+const decoder = new TextDecoder();
+
+/**
+ * The `fetch` implementation based on `undici` includes the headers as an array of encoded key-value pairs.
+ * This function finds the header with the given key and decodes the value.
+ *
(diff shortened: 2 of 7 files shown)


dreamorosi:
Integration tests passing on all managed runtimes: https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/8522716494

RaphaelManke:
What about the error channel? Do you think the "headers" channel who'll be called when there is an error like request timeout or hang up ? I think in case of an error the segment needs to be closed at least.

dreamorosi:
> What about the error channel? Do you think the "headers" channel who'll be called when there is an error like request timeout or hang up ? I think in case of an error the segment needs to be closed at least.

Good catch, I only tested for cases where the server returns a 4xx or 5xx response which is handled by the `onResponse` handler. I am adding an additional handler for `onError` which will cover cases where the client is unable to establish a connection.

dreamorosi:
@RaphaelManke I have added additional logic to handle the case and close the subsegment + restore the parent.

Thank you again for spotting this!

packages/tracer/tests/unit/random.test.ts
am29d, line 1: What's that? 🤔

am29d (changes requested):
Great work Andrea! 

Small ask, otherwise it's good to merge. 

We already have fetch in our examples, let's remove the rest of axios in the [batch readme](https://github.com/aws-powertools/powertools-lambda-typescript/blob/b908aa1111d332fcf3638a77f24a545b85b4c103/packages/batch/README.md?plain=1#L180).

packages/tracer/tests/unit/random.test.ts
dreamorosi: It's our new Tracer testing strategy, instead of mocking every single method (which makes things verbose) we'll actually use the context storage.

I was doing some tests and left the file by mistake, removed it.

am29d:
> > We already have fetch in our examples, let's remove the rest of axios in the [batch readme](https://github.com/aws-powertools/powertools-lambda-typescript/blob/b908aa1111d332fcf3638a77f24a545b85b4c103/packages/batch/README.md?plain=1#L180).
> 
> I found another file that also uses `axios` in the batch package that is not used at all, plus I found a few line highlight that are wrong in the Batch page - if you're okay I'll open a separate issue and tackle all these there, so we keep this PR focused.

Sounds good to me, removing `random.test.ts` and we are good to merge

dreamorosi:
@am29d - I removed the file by pushing an empty comment, not sure what happened.

Should now be fine.

am29d (approved):

Issue #2383, opened 2024-04-17
Bug: jmespath custom functions introspection breaks when minifying


dreamorosi:
### Expected Behaviour

When using the JMESPath utility I should be able to minify my code and all the utility's features should work.

### Current Behaviour

When the function including the utility is minified the utility is not able to introspect its own methods and correctly register the functions available, which causes a runtime error:

```sh
2024-04-17T08:03:11.119Z        a71a8744-df90-4187-bff6-d8f8e0d0478d    ERROR   Invoke Error    {"errorType":"TypeError","errorMessage":"t.introspectMethods is not a function or its return value is not iterable","stack":["TypeError: t.introspectMethods is not a function or its return value is not iterable","    at Object.introspectMethods (/var/task/index.js:2:12877)","    at l.introspectMethods (/var/task/index.js:2:12877)","    at U.introspectMethods (/var/task/index.js:2:12877)","    at new ue (/var/task/index.js:2:16337)","    at he.search (/var/task/index.js:2:20094)","    at Le (/var/task/index.js:2:30000)","    at Runtime.Sr [as handler] (/var/task/index.js:2:31173)","    at Runtime.handleOnceNonStreaming (file:///var/runtime/index.mjs:1173:29)"]}
```

This affects both ESM and CJS builds and can be mitigated only by not aliasing functions/methods names in the minified bundle - in `esbuild` this corresponds to the [`keepNames` field](https://esbuild.github.io/api/#keep-names).

### Code snippet

```ts
import { search } from "@aws-lambda-powertools/jmespath";
import { PowertoolsFunctions } from "@aws-lambda-powertools/jmespath/functions";

export const handler = async (event: any): Promise<any> => {
  return search("powertools_json(foo).bar", event, {
    customFunctions: new PowertoolsFunctions(),
  });
};
```

### Steps to Reproduce

Bundle the function above and minify it:

```sh
npx esbuild --bundle index.ts --platform=node --outdir=out --format=esm --minify
```

Then deploy and observe the error.

On the other hand, if you also enable the `--keepNames` flag, the error won't occur.

### Possible Solution

Refactor [this method](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/jmespath/src/Functions.ts#L531-L558) used to register the available methods to not rely on the prototype, or statically define the methods beforehand if that's not possible.

### Powertools for AWS Lambda (TypeScript) version

latest

### AWS Lambda function runtime

20.x

[...]

PR #2384, opened 2024-04-17
fix(jmespath): refactor custom function introspection to work with minification

dreamorosi:

This PR fixes a bug that caused JMESPath custom functions to fail to be detected and throw a runtime error when running on minified functions.

The previous logic relied on the name of the class (via `prototype.constructor.name`) which caused it to fail when the code is minified since entities are aliased to use shorter names.

The new logic introduced in this PR uses an identity check (`value instanceof Class`) which minification-proof since even when aliased both classes will still maintain their inheritance.

### Related issues, RFCs

**Issue number:** #2383

## Checklist

### Breaking change checklist

---


am29d (approved):
great catch, we might need to inspect other places where we use prototype checks, we might discover more over time.

PR #2429, opened 2024-04-24
fix(parser): set etag optional for delete object notifications

### LINKED ISSUE #2410: Bug: Parser's S3 event notifications schemas
### Summary

Based on the finding in aws-powertools/powertools-lambda-python#4156 it appears that when S3 sends a notification for an object deleted and triggers a Lambda function, the `etag` field is optional.

In looking at our implementation vs the types in `@types/aws-lambda` it seems that we have [a single `S3EventNotificationEventBridgeDetailSchema` schema](https://github.com/aws-powertools/powertools-lambda-typescript/blob/d767d61e74ad49b8b9bcf9c809a1ebef68fad8f9/packages/parser/src/schemas/s3.ts#L68) while DefinitelyTyped has multiple kinds, one for each notification type. 

I'm not sure if we should add schemas for each event type like DefinitelyTyped does or have a single one, however having multiple might help with addressing the subtle differences in field presence (i.e. `etag` not present in delete notifications) & values (i.e. `deletion-type` equals `Permanently Deleted`).

For now I'm opening this as a feature request, but if we decide to just make `etag` optional and keep the schema as-is, we can convert this as maintenance type.

### Why is this needed?

So that customers can use the schema to parse their events.

Parser

### Solution

_No response_

### Acknowledgment

### Future readers

am29d:

In this PR we set etag to optional. We initially discussed in #2410 that it'd be best to split s3 event notification schema into separate objects for event types (delete, create, update, lifecycle, etc) similar to how `@types/aws-lambda` does it.

However `@types/aws-lambda` is not correct for the s3 notification, i.e. etag is a required field in `DeleteObject` events, thus we can't rely on them. In addition, splitting the event types would introduce either redundancy between the events of the common fields, or additional complexity if we decide to split them based on base and specific fields. It'd be hard to read them.

### Related issues, RFCs

**Issue number:** closes #2410 

## Checklist

### Breaking change checklist

---

### DIFF
--- a/packages/parser/src/schemas/s3.ts
+++ b/packages/parser/src/schemas/s3.ts
@@ -72,8 +72,8 @@ const S3EventNotificationEventBridgeDetailSchema = z.object({
   }),
   object: z.object({
     key: z.string(),
-    size: z.number().nonnegative().optional(),
-    etag: z.string().optional(),
+    size: z.number().nonnegative().optional(), // not present in DeleteObject events
+    etag: z.string().optional(), // not present in DeleteObject events
     'version-id': z.string().optional(),
     sequencer: z.string().optional(),
   }),
--- a/packages/parser/src/schemas/s3.ts
+++ b/packages/parser/src/schemas/s3.ts
@@ -73,7 +73,7 @@ const S3EventNotificationEventBridgeDetailSchema = z.object({
   object: z.object({
     key: z.string(),
     size: z.number().nonnegative().optional(),
-    etag: z.string(),
+    etag: z.string().optional(),
     'version-id': z.string().optional(),
     sequencer: z.string().optional(),
   }),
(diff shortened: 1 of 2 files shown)


dreamorosi (commented):
I thought we discussed that we'd break up the schemas into separate ones for each S3 event notification, why the change?

dreamorosi (dismissed):
Just saw your comment [under the issue](https://github.com/aws-powertools/powertools-lambda-typescript/issues/2410#issuecomment-2074521208), would have been useful to have it as part of the PR body.

---

It would really be helpful, and also set an example for external contributor, if we maintainers took time to write informative descriptions so that the PR are self-contained and can be evaluated as standalone.

Having a "fixes #nnn" doesn't necessarily explain the choices made, and when we go back to the PRs at release or after some time we might not know immediately the why and what behind the change.

dreamorosi (approved):

Issue #2399, opened 2024-04-18
Maintenance: start linting markdown files


dreamorosi:
### Summary

Currently we don't have any linting in place for the documentation.

This leads to inconsistencies when different contributors touch the files, which in some cases lead to increased diff on PRs.

We should take inspiration from the Powertools for AWS Lambda (Python) repo and setup something like (or exactly) [markdownlint](https://github.com/DavidAnson/markdownlint).

### Why is this needed?

So that we can increase consistency and remove cognitive overhead during editing & review.

### Which area does this relate to?

Automation

### Solution

- Explore adding markdownlint to the repo
- Consider adopting most or all the same linting rules that the Python repo uses
- Set up the tool to run during PRs CI on all markdown files
- Set up the tool to run on changed files as `pre-commit` hook

### Acknowledgment

- [X] This request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

PR #2434, opened 2024-04-25
chore(maintenance): add markdownlint

dreamorosi:

This PR introduces markdown linting to the project. This will enforce an objective baseline for style and format of markdown files.

To support this I have added [markdownlint-cli2](https://github.com/DavidAnson/markdownlint-cli2) to the workspace, this is the same CLI used by Powertools for AWS Lambda (Python) from which I also copied the rule set.

Linting is run automatically as `pre-commit` hook via `lint-staged`, which will lint and fix only on staged files that are about to be committed (remember to run `npx husky` to make it take effect), and also as PR check. The latter runs on all markdown files in scope.

The tool supports an ignore file, which I also added to the project, however I was able to make this file work only with the IDE extension but not the CLI, for this reason I replicated the patterns in the command I added to the `package.json`.

For the time being I have ignored the `package/*/README.md` files since there are opened issues already for these, once we tackle them we will remove them from the ignore list.

Finally, if you are using VSCode, I recommend to install the [markdownlint](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) which will surface the issues according to the ruleset defined and can also be configured to format on save.

### Related issues, RFCs

**Issue number:** #2399

## Checklist

### Breaking change checklist

---

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -73,8 +73,6 @@ MD013:
   tables: false
   # Include headings
   headings: true
-  # Include headings
-  headers: true
   # Strict length checking
   strict: false
   # Stern length checking
@@ -107,8 +105,6 @@ MD023: true
 
 # MD024/no-duplicate-heading/no-duplicate-header - Multiple headings with the same content
 MD024:
-  # Only check sibling headings
-  allow_different_nesting: false
   # Only check sibling headings
   siblings_only: false
 
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -1,0 +1,230 @@
+# Rules: https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md
+
+# Default state for all rules
+default: true
+
+# Path to configuration file to extend
+extends: null
+
+# MD001/heading-increment/header-increment - Heading levels should only increment by one level at a time
+MD001: true
+
+# MD002/first-heading-h1/first-header-h1 - First heading should be a top-level heading
+# NOTE: We use h2 due to font size
+MD002: false
+
+# MD003/heading-style/header-style - Heading style
+MD003:
+  # Heading style
+  style: "consistent"
+
+# MD004/ul-style - Unordered list style
+MD004:
+  # List style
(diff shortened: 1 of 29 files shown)


.markdownlint.yaml
```diff
@@ -0,0 +1,230 @@
...
+  # Number of characters for code blocks
+  code_block_line_length: 265
+  # Include code blocks
+  code_blocks: true
+  # Include tables
+  tables: false
+  # Include headings
+  headings: true
+  # Include headings
+  headers: true
```
am29d, line 77: My ide complains that this field is not allowed: 

Schema validation: Property 'headers' is not allowed. Looking at the comment it seems to be redundant with `headings`?

.markdownlint.yaml
```diff
@@ -0,0 +1,230 @@
...
+  # Blank lines below heading
+  lines_below: 1
+
+# MD023/heading-start-left/header-start-left - Headings must start at the beginning of the line
+MD023: true
+
+# MD024/no-duplicate-heading/no-duplicate-header - Multiple headings with the same content
+MD024:
+  # Only check sibling headings
+  allow_different_nesting: false
```
am29d, line 111: This property seems also not valid based on the schema. I have double checked the [json schema](https://github.com/DavidAnson/markdownlint/blob/b2305efafb034b1f328845aec9928b5363ffd646/schema/markdownlint-config-schema.json#L591) and I can't find it there either.

am29d (changes requested):
Great addition! I tested few formatting errors and the error messages are spot on what's wrong, where and what to fix ✅.

Just two comments on the configuration, it seems like they are not valid based on the JSON schema of markdownlint? Maybe I am missing something.

dreamorosi:
No, you're right - I copied the config from the Python repo assuming it was correct but the two settings you pointed out are either missing or have been removed, they don't appear anymore in the [docs for the tool](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md) either.

I have removed them both!

am29d (approved):

Issue #2223, opened 2024-03-13
Maintenance: drop support for Node.js 16


dreamorosi:
### Summary

> [!important]
> Starting from July 1, 2024 Powertools for AWS Lambda (TypeScript) will end support for Node.js 16.

This action follows [the date when AWS Lambda will mark the `nodejs16.x` as deprecated](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-supported) and precedes the date after which you will no longer be able to create new functions using the runtime.

We recommend that you upgrade your existing Node.js 16 functions to Node.js 20 before July 1st, 2024.

End of support does not impact existing functions and functions using versions of Powertools for AWS Lambda (TypeScript) released before the date, however starting from this date we will not guarantee compatibility with Node.js 16.

You can use the following AWS CLI command to list all functions in a specific region using Node.js 16, including published function versions:

> [!Note]
> Make sure to replace the value of the `--region` flag and run this command in each region you have Lambda functions workloads

```sh
aws lambda list-functions --function-version ALL --region <Region> --output text --query "Functions[?Runtime=='nodejs16.x'].FunctionArn"
```

### Why is this needed?

Starting from July 15, 2024 we will no longer be able to deploy new Lambda functions using the `nodejs16.x` runtime, this means we will no longer be able to run integration tests for new code changes.

With this in mind, and considering our iteration cycles last two weeks, we plan on dropping support for the runtime just two weeks before the date.

### Which area does this relate to?

Governance

### Acknowledgment

- [X] This request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

dreamorosi:
Check the `useDefineForClassFields` value in `tsconfig.json` which will flip to `true` once we target `ES2022` ([docs](https://www.typescriptlang.org/tsconfig#useDefineForClassFields)).

dreamorosi:
[...]

PR #2484, opened 2024-05-06
Docs Update for clarify.

ericglasser:
## Summary

When going through the documentation I noticed on inconsistency, and one item that would lead to a failed build.

### Changes

- Changed the version of the lambda layer being used.
- removed `@aws-sdk` from example since this requires adding the sdk layer or the function will fail.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -111,7 +111,6 @@ You can use Powertools for AWS Lambda (TypeScript) by installing it with your fa
               bundling: {
                 externalModules: [
                   '@aws-lambda-powertools/*',
-                  '@aws-sdk/*',
                 ],
               }
             });
--- a/docs/index.md
+++ b/docs/index.md
@@ -89,7 +89,7 @@ You can use Powertools for AWS Lambda (TypeScript) by installing it with your fa
                 const powertoolsLayer = LayerVersion.fromLayerVersionArn(
                   this,
                   'PowertoolsLayer',
-                  `arn:aws:lambda:${Stack.of(this).region}:094274105915:layer:AWSLambdaPowertoolsTypeScriptV2:3`
+                  `arn:aws:lambda:${Stack.of(this).region}:094274105915:layer:AWSLambdaPowertoolsTypeScriptV2:5`
                 );
                 
                 new Function(this, 'Function', {


dreamorosi:
Hi, thank you for opening this PR.

I have tried to reproduce the issue you described on my end but I was not able to.

I have deployed a simple function that accesses a SSM parameter and calls STS. These two calls show that the function is able to use the AWS SDK even though it's being excluded from the bundle. This is because the SDK clients needed by Powertools [are bundled in the layer](https://github.com/aws-powertools/powertools-lambda-typescript/blob/4e76c23186de2b2a7a91e7d75b25a76242d4de57/layers/src/layer-publisher-stack.ts#L109-L115), while those who aren't bundled should be present in the execution environment as long as you are using Node.js 18 or greater. 

I have created a minimal example in an attempt to observe the issue you described, which you can find here: https://github.com/dreamorosi/layertest-2484

The only combination that I can imagine giving issues with that exclusion is if you're using the Node.js 16 managed runtime and you are using another AWS SDK client that is not part of the one 

In that case I would be more inclined to update the section with a callout about that runtime specifically, especially since we will stop supporting it in 2 months (#2223).

--

The layer ARN is however wrong, we'll investigate why it's not being updated and are happy to merge the change.

dreamorosi:
We have merged a change to the automation that updates the layer ARN in the docs (#2487), this should ensure that future releases update the layer ARN correctly.

ericglasser:
> Hi, thank you for opening this PR.
> 
> I have tried to reproduce the issue you described on my end but I was not able to.
> 
> I have deployed a simple function that accesses a SSM parameter and calls STS. These two calls show that the function is able to use the AWS SDK even though it's being excluded from the bundle. This is because the SDK clients needed by Powertools [are bundled in the layer](https://github.com/aws-powertools/powertools-lambda-typescript/blob/4e76c23186de2b2a7a91e7d75b25a76242d4de57/layers/src/layer-publisher-stack.ts#L109-L115), while those who aren't bundled should be present in the execution environment as long as you are using Node.js 18 or greater.
> 
> I have created a minimal example in an attempt to observe the issue you described, which you can find here: https://github.com/dreamorosi/layertest-2484
> 
> The only combination that I can imagine giving issues with that exclusion is if you're using the Node.js 16 managed runtime and you are using another AWS SDK client that is not part of the one
> 
> In that case I would be more inclined to update the section with a callout about that runtime specifically, especially since we will stop supporting it in 2 months (#2223).
> 
> --
> 
> The layer ARN is however wrong, we'll investigate why it's not being updated and are happy to merge the change.

Thank you for this clarification, I was intending to use node 18, so I thought this was still a bug, but looking back at the sample node 16 was being used which explains the discrepancy.

dreamorosi:
Either way, thank you so much for taking the time to open the PR and for using Powertools.

If you find any other are for improvement please don't hesitate to open an issue to discuss it!

PR #2487, opened 2024-05-07
chore(docs): fix update layer arn for cdk

### LINKED ISSUE #2486: Docs: update-layer-arn should also update CDK example

Our [`update_layer_arn.sh`](https://github.com/aws-powertools/powertools-lambda-typescript/blob/4e76c23186de2b2a7a91e7d75b25a76242d4de57/.github/scripts/update_layer_arn.sh) should update the layer version in our docs after a new layer deployment. But it turns out the CDK example using the layer is not updated any more. 

_No response_

### How can we improve?

We need to add `${Stack.of(this).region}` to the [select check](https://github.com/aws-powertools/powertools-lambda-typescript/blob/4e76c23186de2b2a7a91e7d75b25a76242d4de57/.github/scripts/update_layer_arn.sh#L59) to replace the version number. The current implementation has `${cdk.Stack.of(this).region}` where cdk prefix is the causing problem.

### Got a suggestion in mind?

_No response_

### Acknowledgment

am29d:
### Changes

In this PR we remove the `cdk` prefix from the pseudo region check when replacing the layer version in the docs. This will fix the reported issue in #2484 .

**Issue number:**  closes #2486 

---

### DIFF
--- a/.github/scripts/update_layer_arn.sh
+++ b/.github/scripts/update_layer_arn.sh
@@ -56,7 +56,7 @@ for file in $files; do
     # arn:aws:lambda:{env.region}:094274105915:layer:AWSLambdaPowertoolsTypeScript:39
     if [[ "$line" == *"eu-central-1"* ]]; then
       # These are all the framework pseudo parameters currently found in the docs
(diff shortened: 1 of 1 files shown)


dreamorosi (approved):
Thank you for fixing this.

It stopped working because of a change I made. I changed the imports in the example to use `import { Stack } from ...` instead of `import cdk from ...` and broke this in the process.

Issue #2505, opened 2024-05-10
Feature request: For idempotency feature, allow no return value / undefined from lambda handlers. 


kuber-:
### Use case

To make functions with side effects and no return value idempotent, undefined /no return value needs to be handled. Currently, the library fails with **error message**: _Failed to update success record to idempotency store. This error was caused by: Pass options.removeUndefinedValues=true to remove undefined values from map/array/set.._

### Solution/User Experience

Update **DynamoDBPersistenceLayer** to handle undefined with appropriate Marshall options while updating the records in the persistence store. 

### Acknowledgment

- [X] This feature request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [X] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

dreamorosi:
Hi @kuber-, thank you for taking the time to open this issue.

As we discussed on Discord, the short-term solution while we work on a more permanent fix would be to simply return _something_. This will allow the utility to work normally and especially if you don't care about the result, 

I have just tested this with Powertools for AWS Lambda (Python) and they handle this by marshalling the `None` value correctly, for example:

```py
from aws_lambda_powertools.utilities.idempotency import (
    DynamoDBPersistenceLayer,
    idempotent,
)
import boto3
from aws_lambda_powertools.utilities.typing import LambdaContext

persistence_layer = DynamoDBPersistenceLayer(table_name="IdempotencyTable")
dynamodb_local_client = boto3.client("dynamodb", endpoint_url="http://localhost:8000")
persistence_layer.client = dynamodb_local_client


@idempotent(persistence_store=persistence_layer)
def lambda_handler(event: dict, context: LambdaContext):
    print("expensive operation")
    return None

if __name__ == "__main__":
    print(lambda_handler({
        "key": "value"
    }, LambdaContext()) )
```

Results in this item being added to DynamoDB:

```yml
ConsumedCapacity: null
Count: 1
Items:
- data:
    S: 'null'
  expiration:
    N: '1715351592'
  id:
[...]

PR #2521, opened 2024-05-14
improv(idempotency): handle functions with no return value

dreamorosi:
### Changes

This PR improves the Idempotency utility by allowing it to be used on functions that have no return value (i.e. `void`/`undefined`). This is useful when you want to make idempotent a function that causes side effects only, but has no return value.

Before this PR the utility would not work on these functions because it would expect the function being made idempotent to return a value that would then be stored in the persistence layer.

As described in the linked issue, the initial plan was to align with how Powertools for AWS Lambda (Python) handle this edge case and store the value as `'null'` into DynamoDB, however after some initial tests it became clear that this was not a viable option due to the fact that it would have become impossible to covert the value back to the original `undefined` when retrieving the idempotency record.

Likewise, storing the item with the attribute set to `Null` (DynamoDB attribute type) was not an option due to it being converted to JavaScript `null`. 

To preserve the ability for customers to return `null` or `'null'` (string), I decided instead to avoid storing the attribute altogether when the function returns `undefined`, since this is the closest value (or rather lack thereof) to the one being returned.

In order to implement this logic I had to do some light refactoring in the `IdempotencyHandler` class and introduce an intermediary data structure that makes it explicit when the return value is `undefined` vs when `undefined` is used as a mean of control-flow (it'll be clearer when you see the code in the diff I hope - if it's not, please ask).

**Issue number:** #2505

---


packages/idempotency/tests/unit/persistence/DynamoDbPersistenceLayer.test.ts
```diff
@@ -604,7 +604,7 @@ describe('Class: DynamoDBPersistenceLayer', () => {
           id: dummyKey,
         }),
         UpdateExpression:
-          'SET #response_data = :response_data, #expiry = :expiry, #status = :status',
+          'SET #expiry = :expiry, #status = :status, #response_data = :response_data',
```
am29d, line 607: is the reorder intentional?

am29d (changes requested):
Thanks for PR, I see the reasoning behind the change. I was on the fence first, but it's much cleaner for the user to not think about return value, otherwise we'd need to enforce to return something. 

One change request: could we add e2e test to cover this one case?

packages/idempotency/tests/unit/persistence/DynamoDbPersistenceLayer.test.ts
dreamorosi: Yes, it's just in the the tests though. 

The reason why I needed to reorder them is that we are now composing the `UpdateExpression` in a different order and the `response_data` field is now last.

dreamorosi:
@am29d - I have modified one of the e2e test cases to return `undefined` and adjusted the assertions to test that the case is handled correctly.

dreamorosi:
Updated integration tests are passing: https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/9218785428

am29d (approved):

am29d (approved):

Issue #2526, opened 2024-05-15
Bug: API Gateway Console Test button fails validation with @parser envelope ApiGatewayEnvelope because requestContext -> identity -> sourceIp value is not a valid IPv4 or IPv6 network


Cihaan:
### Expected Behavior

We should be able to generate a test form the API Gateway console that allows the ApiGatewayProxyEventModel to be validated and parsed by Zod.

### Current Behavior

When generating a test from the API Gateway console, the value for the sourceIp in requestContext -> identity is "test-invoke-source-ip". Since this value is defined to be an ip [here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/f6ed5a66e486a468d4326849c02f2bf724de47dd/packages/parser/src/schemas/apigw.ts#L25) And so test events from the API Gateway console fail with a 502 error because of the resulting Zod ValidationError.

### Code snippet

```yml
  RestApi:
    Type: AWS::Serverless::Api
    Properties:
      StageName: !Ref Environment
      Name: !Sub '${ProjectName}-${Environment}-api'
      OpenApiVersion: '3.0.1' 
      TracingEnabled: true
      EndpointConfiguration: REGIONAL
      Auth: 
        DefaultAuthorizer: NONE
        AddDefaultAuthorizerToCorsPreflight: false
        ApiKeyRequired: false
      Cors:
        AllowMethods: "'DELETE,GET,HEAD,OPTIONS,POST,PUT'"
        AllowHeaders: "'*'"
        AllowOrigin: "'*'"
      MethodSettings:
        - HttpMethod: '*'
          ResourcePath: '/*'
          LoggingLevel: INFO
          DataTraceEnabled: true
          MetricsEnabled: true
          ThrottlingBurstLimit: 500
          ThrottlingRateLimit: 1000
      AccessLogSetting:
        DestinationArn: !GetAtt ApiAccessLogGroup.Arn
        Format: >-
          {
          "status": "$context.status",
          "traceId": "$context.xrayTraceId",
          "requestId": "$context.requestId",
          "requestTime": "$context.requestTime",
          "httpMethod": "$context.httpMethod",
          "responseLength": "$context.responseLength",
          "resourcePath": "$context.resourcePath",
          "resourceId": "$context.resourceId",
          "sourceIp": "$context.identity.sourceIp",
          "userAgent": "$context.identity.userAgent",
          "accountId": "$context.identity.accountId",
          "caller": "$context.identity.caller",
[...]

PR #2531, opened 2024-05-15
fix(parser): handle API Gateway Test UI sourceIp

dreamorosi:
### Changes

This PR updates the `APIGatewayProxyEvent` schema in the Parser utility to account for an edge case occurring when using the API Gateway Test UI to call an endpoint.

In these cases API Gateway doesn't send a real `sourceIp` but instead sends a string hardcoded to `test-invoke-source-ip`. This caused the schema to fail parsing preventing customers from using the Test UI with endpoints using the Parser utility.

Normally we wouldn't have made changes of this type, but since previous efforts to have the API Gateway service change this behavior have failed, and there's no workaround for customers, we are making an exception until the situation improves.

As of this PR the `sourceIp` field can be either a valid IP address or a string with value `test-invoke-source-ip`.

**Issue number:** #2526

---

### DIFF
--- a/packages/parser/src/schemas/apigw.ts
+++ b/packages/parser/src/schemas/apigw.ts
@@ -22,7 +22,16 @@ const APIGatewayEventIdentity = z.object({
   cognitoIdentityId: z.string().nullish(),
   cognitoIdentityPoolId: z.string().nullish(),
   principalOrgId: z.string().nullish(),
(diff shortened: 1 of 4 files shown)


am29d (approved):
Thanks for adding this edge case, this is definitely save troubles for future UI users!

Issue #2120, opened 2024-02-21
Feature request: deep sort payload hashing


dreamorosi:
### Use case

As a customer using the Idempotency utility I want my payloads to be considered idempotent regardless of the ordering of the keys/items within.

Currently, as discovered [here](https://github.com/aws-powertools/powertools-lambda-python/issues/3781#issuecomment-1954369980), when we hash objects or arrays that have elements ordered differently they result in two different hashes. This causes the requests to be considered unique.

For reference, this is a simplified version of our hashing implementation:
```js
const { createHash } = require("crypto");

/**
 * @param {string} data - The data to be hashed
 * @returns {string} - The hashed data
 */
const hash = (data) => {
  const hash = createHash("sha256");
  hash.update(data);
  return hash.digest("base64");
}
```

Now, let's take two requests:

```js
const requestA = {
  name: "John",
  age: 30,
  city: "New York",
  address: {
    street: "5th Avenue",
    number: 123,
  },
};

const requestB = {
  city: "New York",
  name: "John",
  age: 30,
  address: {
    number: 123,
    street: "5th Avenue",
  },
};
```

These two requests should be considered idempotent despite having the keys ordered differently, however in our current implementation, they are considered as two different requests:

```js
console.log(hash(JSON.stringify(objectA)) === hash(JSON.stringify(objectB))); // false
```

We should implement a function that sorts the objects not only at the top level but also at nested levels.

### Solution/User Experience

The change should be completely transparent for customers and the API/DX of the utility should not change.

In terms of implementation, the one found in [this blog post](https://whitfin.io/blog/sorting-object-recursively-node-jsjavascript/) could be a good starting point:

```js
function sortObject(object) {
  var sortedObj = {},
    keys = Object.keys(object);

  keys.sort(function (key1, key2) {
    (key1 = key1.toLowerCase()), (key2 = key2.toLowerCase());
    if (key1 < key2) return -1;
    if (key1 > key2) return 1;
    return 0;
  });

  for (var index in keys) {
    var key = keys[index];
    if (typeof object[key] == "object" && !(object[key] instanceof Array)) {
      sortedObj[key] = sortObject(object[key]);
    } else {
      sortedObj[key] = object[key];
    }
  }

  return sortedObj;
}

console.log(
  hash(JSON.stringify(sortObject(objectA))) ===
    hash(JSON.stringify(sortObject(objectB)))
); // true
```

[...]

PR #2570, opened 2024-05-22
fix(idempotency): deep sort payload during hashing

arnabrahman:
## Summary

According to the [Issue](https://github.com/aws-powertools/powertools-lambda-typescript/issues/2120), payloads containing the exact same keys should be considered idempotent regardless of their order. 

For example, these two requests should be idempotent.

```
const requestA = {
  name: "John",
  age: 30,
  city: "New York",
  address: {
    street: "5th Avenue",
    number: 123,
  },
};

const requestB = {
  city: "New York",
  name: "John",
  age: 30,
  address: {
    number: 123,
    street: "5th Avenue",
  },
};
```

The same goes for JSON arrays, the following two requests should be idempotent.

```
const requestA = [{
  a: 30,
  c: "John",
  b: "New York",
}]

const requestB = [{
  a: 30,
  b: "New York",
  c: "John"
}];
```
To achieve this, we need to implement a function that deep sorts the payload & use it to sort the payload before generating the hash.

### Changes

Implement a function that recursively sorts the keys of an object or elements of an array. The function is based on [this blog post](https://whitfin.io/blog/sorting-object-recursively-node-jsjavascript/) that was mentioned in the issue description. In addition, added support for sorting objects inside JSON array. 

- Uses `getType` from `@aws-lambda-powertools/commons` to determine the input data type.
- For objects, it sorts the keys alphabetically in a case-insensitive manner using sortObject.
- For arrays, it recursively applies deepSort to each element.
- Primitives are returned unchanged.

Used this function before generating hash inside `BasePersistenceLayer` class. 

Also, written unit tests for the function for different scenarios that I can think of. I couldn't find a method in `jest` that checks if two objects are equal and ensures that the key positions are similar. Therefore, I used `JSON.stringify` to compare the two objects.

I am not sure if I need to update any documentation for this. Please let me know if I need to make any tweaks.

**Issue number:** #2120

---


dreamorosi (changes requested):
The integration tests for this branch [are failing](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/9192311915/job/25280659348#step:8:523) because the sorting changes the hashes/idempotency ID generated for some of the tests.

I have also left a couple comments in the code.

dreamorosi:
Thank you for the PR @arnabrahman, appreciate it!

am29d:
I agree that this is a necessary change and improves idempotency utility and this is a breaking change. As you said, this could disable idempotency for some cases which can lead to severe impact on the customer side.

arnabrahman:
@dreamorosi for the e2e tests, i couldn't run it locally but maybe we need to `deepSort` the record before stringify. https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/idempotency/tests/e2e/makeIdempotent.test.ts#L177

dreamorosi:
@arnabrahman that's right, I believe that if we sort these payloads the tests should pass. 

Please bear with us while we agree on a way forward - I'm unsure whether this should be considered a bug fix or a new feature, in which case it might be a breaking change.

Also what do you think?

arnabrahman:
@dreamorosi , this is a peculiar situation. At a high level, this sounds like a bug fix.

However, from my experience, I don't recall ever sorting a payload of objects before sending it. I might have done so unintentionally, but never intentionally. I believe that in the majority of cases, this change will cause issues. So, from a customer's perspective, this is a `Breaking Change`.

If so, in my opinion, we cannot say it's a bug fix.

dreamorosi:
I see where you're coming from. 

Imo the fact that we were not ordering the payload, especially since Powertools for AWS Lambda (Python) does it, was not intended and even though the change could cause unintended side effects, these are only the results of us not sorting the payload in the first place.

Like you said, sorting is not something customers would (nor should) think about it, but requests with the same payload content should be idempotent, so the fact that this wasn't the case maybe makes it a bug.

dreamorosi:
Putting this on hold until @heitorlessa is back to advise on how to move forward.

heitorlessa:
Take it as a bug fix, as this shouldn't have passed in our tests suite. There are internet proxies and the fact a client or a proxy could interfere with the guaranteed idempotency within a time window makes this a serious bug.

example: If we were changing the idempotency key format, I'd consider as a breaking change. 

Here, it'll cause bigger harm for new customers having mixed results without any way to debug it. In the worst case, we know > 90% of customers don't change the defaults, so idempotency records will expire within an hour.. reducing the impact window for the bugfix.

Issue #1774, opened 2023-10-31
Feature request: time zone aware timestamp in Logger


dreamorosi:
### Use case

By default Logger emits logs with timestamps formatted using the UTC timezone. This is the default behavior in Lambda since the `TZ` value is always set to `UTC` in the execution environment ([source](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime)).

As a customer however I might want to emit logs that contain timestamps that are formatted according to a different timezone (i.e. `2021-12-12T21:21:08.921+01:00` instead of `2021-12-12T21:21:08.921Z`). 

The Logger utility should be able to take in account the value of the `TZ` environment variable, and if it's different than `UTC`, format the timestamps in the logs accordingly.

### Solution/User Experience

By default the Lambda environment has the `TZ` variable set to UTC ([source](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime)), if this is the case logs would be emitted as UTC like they are already. If instead the value of the `TZ` env variable is something else, we format the timestamp taking into account the time zone.

When in UTC, timestamps are formatted like they are today: `2021-12-12T21:21:08.921Z` ([source](https://github.com/aws-powertools/powertools-lambda-typescript/blob/38edc369362030c4608cca75b477998de0f7d22b/packages/logger/src/formatter/LogFormatter.ts#L50-L52)). 

When using a time zone, timestamps should be formatted including the timezone offset: `2021-12-12T21:21:08.921+01:00`. The offset can be positive (`+`) or negative (`-`) and it's followed by `HH:MM` - [see ISO_8601 spec](https://en.wikipedia.org/wiki/ISO_8601).

Below an example implementation that could be used as basis for the implementation:

```ts
const date = new Date(); // Get the current date and time

const dateFormatter = new Intl.DateTimeFormat("en", {
  year: "numeric",
  month: "2-digit",
  day: "2-digit",
  hour: "2-digit",
  minute: "2-digit",
  second: "2-digit",
  hour12: false,
  timeZone: "Europe/Madrid",
});

const parts = dateFormatter.formatToParts(date);

console.log(parts);

const datePart = `${parts[4].value}-${parts[2].value}-${parts[0].value}T${parts[6].value}:${parts[8].value}:${parts[10].value}`;

const offset = -date.getTimezoneOffset();
const offsetSign = offset >= 0 ? "+" : "-";
[...]

PR #2590, opened 2024-05-29
feat(batch): add option to continue processing other group IDs on failure in `SqsFifoPartialProcessor`

arnabrahman:
## Summary
When using SQS FIFO, by default, we will stop processing at the first failure and mark unprocessed messages as failed to preserve ordering. However, this behavior may not be optimal for customers who wish to proceed with processing messages from a different group ID.

Here an option has been given to the customer to skip the default behavior.
### Changes

- Introduced a `skipGroupOnError` option that can be passed to the `processPartialResponseSync` function.
- Made a generic type so that `skipGroupOnError` is accepted only when the processor parameter is of type `SqsFifoPartialProcessor`. Example:
```
         const processor = new SqsFifoPartialProcessor();
         const result = processPartialResponseSync(
            event,
            sqsRecordHandler,
            processor,
            {
              context,
              skipGroupOnError: true,
            }
          );
 ```

- If `skipGroupOnError` is not true, then short-circuit the process on first failure same as before.
- If `skipGroupOnError` is true, then continue processing on the failure. Only if the current group has a previous failure then skip processing the current record and log `SqsFifoMessageGroupShortCircuitError`. Otherwise, process the record.
- Write unit tests for `skipGroupOnError` flag
- Updated the documentation, added examples and added sequence diagrams. I followed the Python powertools documentation for reference.

**Issue number:** #2561 

---

### CODE REVIEW

**Comment by dreamorosi** [comment]
Hi @arnabrahman - thank you for your patience on this.

I've just got back from PTO and will review this in the next 24hrs tops.

File: packages/batch/tests/helpers/factories.ts [hunk 1]
```diff
@@ -15,6 +15,7 @@ const sqsRecordFactory = (body: string): SQSRecord => {
       SentTimestamp: '1545082649183',
       SenderId: 'AIDAIENQZJOLO23YVJ4VO',
       ApproximateFirstReceiveTimestamp: '1545082649185',
+      MessageGroupId: messageGroupId,
```
dreamorosi, line 18:
```suggestion
      ...(messageGroupId ? { MessageGroupId: messageGroupId } : {}),
```

packages/batch/src/SqsFifoPartialProcessor.ts
```diff
@@ -94,16 +153,44 @@ class SqsFifoPartialProcessor extends BatchProcessorSync {
...
+   */
+  private addToFailedGroup(group: string): void {
+    this.failedGroupIds.add(group);
+  }
+
+  /**
+   * Processes a fail record.
+   * @param record - The record that failed.
+   * @param exception - The error that occurred.
+   * @returns The failure response.
```
dreamorosi, line 176: In the docstrings, let's:
- leave 1 space/new-line between description and params
- remove the `@returns` 

for example

```ts
/**
  * Processes a fail record.
  *
  * @param record - The record that failed.
  * @param exception - The error that occurred.
  */
```

packages/batch/src/processPartialResponseSync.ts
```diff
@@ -41,17 +41,37 @@ import type {
  *     context,
  *   });
  * ```
```
dreamorosi, line 43: Let's add a line in between these two examples that says something like: "When working with SQS FIFO queues, we will stop processing at the first failure and mark unprocessed messages as failed to preserve ordering. However, if you want to continue processing messages from different group IDs, you can enable the `skipGroupOnError` option for seamless processing of messages from various group IDs.`

Also let's leave new lines between the example and the text.

packages/batch/src/SqsFifoPartialProcessor.ts
```diff
@@ -94,16 +153,44 @@ class SqsFifoPartialProcessor extends BatchProcessorSync {
...
     this.clean();
 
     return processedRecords;
   }
+
+  /**
+   * Adds the specified group ID to the set of failed group IDs.
+   * @param group - The group ID to be added to the set of failed group IDs.
+   */
+  private addToFailedGroup(group: string): void {
```
dreamorosi, line 168: I think we can convert these to JS private fields, i.e. `private addToFailedGroup` -> `#addToFailedGroup`.

I would default to that and eventually open them up to `protected` if there is demand.

packages/batch/src/types.ts
```diff
@@ -4,18 +4,27 @@ import type {
...
+ * @property skipGroupOnError The option to group on error during processing
  */
-type BatchProcessingOptions = {
+type BatchProcessingOptions<T = BasePartialBatchProcessor> = {
   /**
    * The context object provided by the AWS Lambda runtime. When provided,
    * it's made available to the handler function you specify
    */
   context: Context;
+  /**
```
dreamorosi, line 23: I think we can make this optional, so one doesn't have to pass it whenever they want to specify the other (also this way you can remove it from the unit tests)

docs/utilities/batch.md
```diff
@@ -141,12 +141,23 @@ Processing batches from SQS works in three stages:
...
-```typescript hl_lines="1-4 8 20-22"
---8<-- "examples/snippets/batch/gettingStartedSQSFifo.ts"
-```
+By default, we will stop processing at the first failure and mark unprocessed messages as failed to preserve ordering. However, this behavior may not be optimal for customers who wish to proceed with processing messages from a different group ID.
+
+Enable the `skipGroupOnError` option for seamless processing of messages from various group IDs. This setup ensures that messages from a failed group ID are sent back to SQS, enabling uninterrupted processing of messages from the subsequent group ID.
+
+=== "Recommended"
+
+    ```typescript hl_lines="1-4 13 28-30"
```
dreamorosi, line 152: Here we should change the hl_lines to 1-4 and 8, otherwise it'll look like this

Screenshot 2024-06-03 at 16 56 10

docs/utilities/batch.md
```diff
@@ -141,12 +141,23 @@ Processing batches from SQS works in three stages:
...
+    --8<-- "examples/snippets/batch/gettingStartedSQSFifo.ts"
+    ```
+
+=== "Enabling skipGroupOnError flag"
+
+    ```typescript hl_lines="1-4 13 28-30"
```
dreamorosi, line 158: Here we can remove the `28-` and leave only L30 highlighted

docs/utilities/batch.md
```diff
@@ -141,12 +141,23 @@ Processing batches from SQS works in three stages:
...
+    --8<-- "examples/snippets/batch/gettingStartedSQSFifoSkipGroupOnError.ts"
+    ```
 
 1. **Step 1**. Creates a partial failure batch processor for SQS FIFO queues. See [partial failure mechanics for details](#partial-failure-mechanics)
```
dreamorosi, line 162: We need to indent this at the same level of the code block (one extra indent) otherwise it won't be rendered correctly.

I would also move this to ~L155 so it's only related to the first code block and remove the corresponding `// (1)!` annotation from the new `gettingStartedSQSFifoSkipGroupOnError.ts` code snippet

dreamorosi:
```suggestion
    1. **Step 1**. Creates a partial failure batch processor for SQS FIFO queues. See [partial failure mechanics for details](#partial-failure-mechanics)
```

dreamorosi (changes requested):

dreamorosi (changes requested):
Thank you so much for the work on this PR, the change is super clean and I really appreciate that you went the extra mile and also updated the docs!

I have left only a few minor comments, after which I am looking forward to merge the PR and include the feature in this week's release :)

docs/utilities/batch.md
arnabrahman: One thing moving to `~L155` didn't render properly, i had to put it to `~L156` to make it render correctly.

packages/batch/src/SqsFifoPartialProcessor.ts
arnabrahman: To be clear, should I change it for all private methods/variables of this file? 

I can see a similar pattern is followed in `Parser.ts` from `jmsepath`
https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/jmespath/src/Parser.ts

@dreamorosi

dreamorosi: Only the ones that you're adding in this PR for now

dreamorosi (approved):
Thank you for the thorough work on this PR, I really appreciate the attention to details and openness to work on iterations during the review.

Merging this now 🎉 

PS: I've labelled two more issues ([list](https://github.com/aws-powertools/powertools-lambda-typescript/issues?q=is%3Aopen+is%3Aissue+label%3Abatch+label%3Ahelp-wanted)) related to this utility as help wanted just in case you're interested. They're both in similar complexity and along the lines of this one.

arnabrahman:
Nice, i will take a look after finishing #1774 , @dreamorosi

Issue #2624, opened 2024-06-09
Feature Request: add APIGatewayRequestAuthorizerEventV2 schemas/envelopes


codyfrisch:
### Use case

My application uses Lambda Authorizer functions, and parsing the APIGatewayRequestAuthorizerEventV2 is required. 

### Solution/User Experience

Implement Zod schema for APIGatewayRequestAuthorizerEventV2.

### Acknowledgment

- [X] This feature request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

dreamorosi:
Hi @codyfrisch - thank you for opening this feature request.

The schema is indeed missing and should be added.

I have added this issue to the backlog and also opened it for contributions.

For potential contributors: the goal of this task should be to add new schemas to the `packages/parser/src/schemas` folder that map the payloads described [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-lambda-authorizer.html) and the types [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/trigger/api-gateway-authorizer.d.ts).

If anyone is interested, please leave a comment under this issue so we can assign it to you, and please don't hesitate to ask for questions or clarifications as needed.

dreamorosi:
I have started working on this and in order to do a comprehensive review I have decided to create some API Gateway resources with different configurations, so that I can observe the payloads and write schemas for them.

## REST API (v1)

### No auth

- `POST /no-auth` for unauthenticated requests

Request:

```sh
http POST https://puhdx84jy9.execute-api.eu-west-1.amazonaws.com/prod/no-auth
```

Payload received by handler: 

<details>

```json
{
    "resource": "/no-auth",
    "path": "/no-auth",
    "httpMethod": "POST",
    "headers": {
        "Accept": "*/*",
        "Accept-Encoding": "gzip, deflate",
        "CloudFront-Forwarded-Proto": "https",
        "CloudFront-Is-Desktop-Viewer": "true",
[...]

PR #2665, opened 2024-06-17
feat(parser): enhance API Gateway schemas

dreamorosi:
### Changes

As discussed in the linked issue, the schemas for API Gateway in the Parser utility had some inconsistencies and missing schemas.

In the issue I tested all the integration types and auth that I could think of for API Gateway and observed the payloads for each. I have used these payloads (that are in the issue) to inform the changes in this PR.

This PR fixes this by making the following changes:
- Add `APIGatewayRequestAuthorizerEventSchema` for Lambda Authorizer Request type in API Gateway v1 (REST)
- Add `APIGatewayTokenAuthorizerEventSchema` for Lambda Authorizer Token type in API Gateway v1 (REST)
- Add `APIGatewayRequestAuthorizerEventV2Schema` for Lambda Authorizer in API Gateway v2 (HTTP)
- Reviewed `APIGatewayProxyEventSchema` to conform with the actual event payload (see linked issue)
- Reviewed `APIGatewayProxyEventV2Schema` to conform with the actual event payload (see linked issue)

Additionally, while working on this I have also made three other changes that are not strictly related to the issue at hand but still interest the code I touched
- Extracted some shared schemas between v1 and v2 into a separate file that is now imported by both - before the v2 schemas were importing (and including in the final bundle) all the schemas from v1 only because they were importing a single shared schema
- Added docstrings with description, examples (where relevant), and links to all public schemas related to API Gateway
- Added subpath exports to target import only schemas from API Gateway v1 or v2 - this allows customers to avoid including all the schemas present in Parser when importing one (we'll roll out this change across all schemas)

**Issue number:** #2624

---


dreamorosi:
@am29d - for when you're back.

Many of the test events used in the unit tests are not in line with what I observed and reported in the API Gateway tests discussed in the linked issue.

If possible, I'd ask you to please review [my comment here](https://github.com/aws-powertools/powertools-lambda-typescript/issues/2624#issuecomment-2173779016), and if you agree replace the API Gateway event payloads with the ones I logged from API Gateway.

am29d (changes requested):
Thanks a lot for going through all the requests and adding the docs!  

One minor ask: can we change to kebab-case `apigw-proxy.ts` for consistency?

I agree to change the test events with the ones you have generated. The previous events were afaik from the docs, which we imported from PT python.

dreamorosi:
I have refactored the unit tests for API Gateway to use the events I recorded in the linked issue.

I have put the new test events in dedicated folders and used a slightly different way of reading the files. This new method only reads the files needed for that specific test suite instead of performing i/o to read all the events for each test.

Moving forward we should probably roll out this change across all other schemas/envelopes to improve test runtime.

Finally, I couldn't test the WebSocket API but I suspect there might be an opportunity to split the `APIGatewayProxyEventSchema` further and create a dedicated `APIGatewayProxyWebsocketEventSchema`.

I say this because some of the fields in the `requestContext`, namely: `messageId`, `connectionId`, `routeKey`, and `eventType` only seem to appear in these events. I'll address this in a future issue since I need to first figure out how to deploy a WS API (never done it) and I don't want to block your other PR.

am29d (approved):
Good call on refactoring the test event read operation. I remember having a mental todo but forgot about that. Great improvement!

Issue #2668, opened 2024-06-18
Bug: log sampling is not working in v2


theburningmonk:
### Expected Behavior

When `POWERTOOLS_LOG_SAMPLE_RATE` is set to `0.1` and every invocation includes a call to `logger.refreshSampleRateCalculation()` to recalculate the sampling rate, the debug logs should be recorded for only 10% of invocations.

### Current Behavior

The sampling decision is made at cold start, and then all invocations afterwards will either log all debug logs or none at all. The refreshSampleRateCalculation() method is not giving the correct behaviour.

### Code snippet

This line is included at the start of a handler function: 

```
logger.refreshSampleRateCalculation()
```

The `POWERTOOLS_LOG_LEVEL` env var is set to `ERROR`. 
The `POWERTOOLS_LOGGER_SAMPLE_RATE` env var is set to `0.1`

### Steps to Reproduce

Create a function with:
* `POWERTOOLS_LOGGER_SAMPLE_RATE` env var set to `0.1`
* `POWERTOOLS_LOG_LEVEL` env var set to `ERROR`
* include a call to logger.refreshSampleRateCalculation() at the start of the handler
* include a few debug logs
Invoke the function multiple times to see sampling decision is taken once, at cold start, but then never changed on subsequent invocations on the same worker

### Powertools for AWS Lambda (TypeScript) version

latest

### AWS Lambda function runtime

20.x

### Packaging format used

npm

dreamorosi:
Hi @theburningmonk, thank you for opening the issue.

As I mentioned in [the thread](https://x.com/theburningmonk/status/1803008698618380357) we exchanged I think the report is valid and we have a bug.

I am currently investigating and I will share more info once I have a definitive answer.

dreamorosi:
Hi Yan, just wanted to let you know that I identified the bug and that I have already opened a PR with a proposed fix.

The root cause of the bug was the one that I hinted at yesterday during our exchange on Twitter. Basically we were not restoring the log level to the original one after having sampled a specific invocation.

Once the PR is merged it will be included in the next release.

In the meanwhile, if you want to continue upgrading before the release, a workaround like this would work:

```diff
import { Logger } from '@aws-lambda-powertools/logger';

const logger = new Logger({ sampleRateValue: 0.1, logLevel: 'ERROR' });

export const handler = async () => {
+ logger.setLogLevel('ERROR'); // TODO: remove after next release
  logger.refreshSampleRateCalculation();
  
[...]

PR #2673, opened 2024-06-19
fix(logger): reset log level after sampling refresh

dreamorosi:
### Changes

This PR fixes a bug with the log sampling feature in Logger that was introduced in v2 and that caused log sampling to stay enabled within the same execution environment after the first time an invocation was sampled.

I fixed the bug by storing the initial log level in a private property within the logger instance and then using this initial log level to restore the log level whenever the sampling calculation is not sampled.

**Issue number:** #2668

---


dreamorosi:
Integration tests passing: https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/9566448596

heitorlessa (approved):
suuuperb investigation of one those bugs that look obvious when we find the root cause, but difficult without customer feedback and/or maintainer having quality time to get a feature scrutinized and peer reviewed.

As we briefly discussed on a call, two things we need to keep in mind:

1. This feature will be superseded by a proper well established solution -- buffering debug logs _(have and eat the cake too)_

2. It looks like Logger for TS could beneift from [a "Handler" responsibility](https://docs.python.org/3/howto/logging.html#logging-flow) to eventually keep the Logger interface cleaner

PR #2693, opened 2024-06-26
fix(idempotency): preserve scope of decorated class

### LINKED ISSUE #2692: Bug: idempotent decorator in combination does not maintain class scope
### Expected Behavior

This is probably a follow up to #1481
When the `@idempotent` decorator is used with other powertools decorators such as the `@tracer.captureLambdaHandler()` decorator the lambda function executes without an error with all the features of the decorators.

### Current Behavior

This is probably a follow up to #1481
When the `@idempotent` decorator is used with other powertools decorators such as the `@tracer.captureLambdaHandler()` decorator the lambda function breaks

### Code snippet

```typescript
import type { EventBridgeEvent, EventBridgeHandler } from 'aws-lambda';
import { IdempotencyConfig, idempotent } from '@aws-lambda-powertools/idempotency';

const tracer = new Tracer({ serviceName: 'foobarservice' });
    tableName: 'foo-table',
    keyAttr: 'partitionKey',
    sortKeyAttr: 'sk',
    inProgressExpiryAttr: 'inProgressExpiration',
    staticPkValue: 'Idempotency#pk'
});
const config = new IdempotencyConfig({ expiresAfterSeconds: 60, eventKeyJmesPath: 'detail.deduplicationId' });

export class RequestStatusChangedListener {
    private readonly logger = new MyCustomLogger();

    @tracer.captureLambdaHandler()
    @idempotent({ persistenceStore, config })
    public async handler(event: EventBridgeEvent<string, unknown>): Promise<string> {
        this.logger.debug(`Received event="${JSON.stringify(event)}".`);

        return 'success';
    }
}

const listener = new RequestStatusChangedListener();
export const handler: EventBridgeHandler<string, unknown, string> = listener.handler.bind(listener);
```

### Steps to Reproduce

1. Deploy the code snippet
2. Add some dummy custom logger
3. Invoke the lambda function

### Possible Solution

_No response_

latest

20.x

### Packaging format used

npm

### Execution logs

```Shell
{
    "errorType": "TypeError",
    "errorMessage": "Cannot read properties of undefined (reading 'debug')",
    "stack": [
        "TypeError: Cannot read properties of undefined (reading 'debug')",
        "    at qg.handler (file:///var/task/index.mjs:125:1450)",
        "    at qg.getFunctionResult (file:///var/task/index.mjs:111:22338)",
        "    at qg.handle (file:///var/task/index.mjs:111:22683)",
        "    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)",
        "    at async file:///var/task/index.mjs:111:39254"
    ]
}
```

dreamorosi:
### Changes

This PR addresses the bug in the linked issue that caused class methods decorated with the `@idempotent` decorator to lose the original scope of the class.

In simpler terms, when working with classes it's common to reference other class methods or properties. In JavaScript this is done via the `this` keyword, which in the case of classes is bound to the class instance.

For example, in the code snippet below the `handler` method is able to access the `logger` property by using the `this` keyword, which at runtime is bound to the `fooInstance` instance.

```ts
export class Foo {
    private readonly logger = console;

    public async handler(event: unknown): Promise<string> {
        this.logger.debug('Received event');

        return 'success';
    }
}

const fooInstance = new Foo();
```

When working with class method decorators in TypeScript it's important to keep in mind that the decorator replaces the method implementation with one it defines. In doing so, it's important to pass down the original scope so that the decorated method maintains access to the `this` scope.

This PR makes changes to the `makeIdempotent` function, the `IdempotencyHandler` class, and the `idempotent` decorator implementation so that this happens.

Specifically, the PR:
- converts the `makeIdempotent` function from an arrow function to a regular function so that [it maintains a binding to `this`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions)
- modifies the `idempotent` decorator to explicitly `.bind(this)` on the `makeIdempotent` function when replacing the implementation of the decorated method
- updates the `IdempotencyHandler` to now accept a new `thisArg` parameter and restore it onto the original function when calling it

I have also added a new unit test case and modified one of the integration test cases to confirm that the fix works.

**Issue number:** closes #2692

---

### DIFF
--- a/packages/idempotency/src/IdempotencyHandler.ts
+++ b/packages/idempotency/src/IdempotencyHandler.ts
@@ -1,3 +1,4 @@
+import type { Handler } from 'aws-lambda';
 import type {
   JSONValue,
   MiddyLikeRequest,
@@ -53,6 +54,12 @@ export class IdempotencyHandler<Func extends AnyFunction> {
    * Persistence layer used to store the idempotency records.
    */
   readonly #persistenceStore: BasePersistenceLayer;
+  /**
+   * The `this` context to be used when calling the function.
+   *
+   * When decorating a class method, this will be the instance of the class.
+   */
+  readonly #thisArg?: Handler;
 
   public constructor(options: IdempotencyHandlerOptions) {
     const {
@@ -61,11 +68,13 @@ export class IdempotencyHandler<Func extends AnyFunction> {
       idempotencyConfig,
       functionArguments,
       persistenceStore,
+      thisArg,
     } = options;
     this.#functionToMakeIdempotent = functionToMakeIdempotent;
     this.#functionPayloadToBeHashed = functionPayloadToBeHashed;
     this.#idempotencyConfig = idempotencyConfig;
     this.#functionArguments = functionArguments;
+    this.#thisArg = thisArg;
 
     this.#persistenceStore = persistenceStore;
 
@@ -121,7 +130,10 @@ export class IdempotencyHandler<Func extends AnyFunction> {
   public async getFunctionResult(): Promise<ReturnType<Func>> {
     let result;
     try {
-      result = await this.#functionToMakeIdempotent(...this.#functionArguments);
+      result = await this.#functionToMakeIdempotent.apply(
+        this.#thisArg,
+        this.#functionArguments
+      );
     } catch (error) {
       await this.#deleteInProgressRecord();
       throw error;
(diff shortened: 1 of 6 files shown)


am29d (approved):
Great catch and a quick PR with the fix, Thank you Andrea! 

Thanks @HaaLeo for reporting the issue!

PR #2710, opened 2024-06-29
feat(logger): time zone aware timestamp in Logger

arnabrahman:
## Summary
This PR introduces a new feature allowing customers to emit timestamps with a timezone-aware format in the Logger. By default, Logger emits logs with timestamps formatted using the `UTC` timezone, which is the default behavior in AWS Lambda. 

### Changes

- When a timezone other than UTC is set, timestamps will be formatted to include the relevant timezone offset. 
   Example: If the emitted timestamp format in UTC is `2016-06-20T12:08:10.000Z` and the timezone is set to `America/New_York`, the timestamp will be formatted as `2016-06-20T08:08:10.000-04:00`.
- The timezone offset is formatted according to the ISO_8601 specification, where the offset can be positive (+) or negative (-) and is followed by `HH:MM` - [see ISO_8601 spec](https://en.wikipedia.org/wiki/ISO_8601)
- Unit tests have been written for two different timezones to cover both GMT (+) and GMT (-) timezones.
- While writing the unit tests, I thought if I changed the system timezone by setting the `TZ` variable, it would work in that timezone. But this is not the case. There are some limitations to jest (https://github.com/jestjs/jest/issues/9856, https://github.com/jestjs/jest/issues/9264). I had to `mock` the `getTimezoneOffset` function to make it work. There are some workarounds mentioned about custom environment but I couldn't make those work properly.
- Updated the docs for the new feature, followed the python powertools docs for inspiration.
- But now comes the most surprising and confusing part. All the tests were running fine locally, I wanted to try it inside a Lambda function, so I deployed one with my solution. However, during testing, I got some unexpected results. The solution works fine when I set the TZ environment variable for different timezones. But when I don't set the TZ environment variable, things just don't work. My implementation was something like this:

  ```
  const defaultTimezone = 'UTC';
   if(configuredTimezone && configuredTimezone !== defaultTimezone)
      // Timezone aware implementation
   else
     // UTC timezone format
  ```
  In cases where there is no `TZ` set, the `if` block was getting executed unexpectedly. I spent some time to figure it out what's wrong with my code but I couldn't. 
  
  Then i found that the actual `TZ` value that AWS Lambda sets is not `UTC`,  it's actually `:UTC`. I tried with different runtimes like python/node/ruby but got the same result. This is the most confusing part, I don't know if I am missing something in the [doc](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime) or not.
  
  After the revelation, I changed the implementation & it worked fine in the lambda.
  ```
  const defaultTimezone = 'UTC';
  if (configuredTimezone && !configuredTimezone.includes(defaultTimezone))
    // Timezone aware implementation
  else
    // UTC timezone format
  ```

**Issue number:** #1774 

---


dreamorosi:
Hi @arnabrahman, nice to see you again 😃 

Thank you so much for opening this PR! I should be able to provide a first review by Monday EOD at the latest.

I want to set aside some time to build the package and test it in a real Lambda function to see how it behaves.

docs/core/logger.md
```diff
@@ -519,6 +519,28 @@ We prioritise log level settings in this order:
 
 In the event you have set a log level in Powertools to a level that is lower than the ACL setting, we will output a warning log message informing you that your messages will be discarded by Lambda.
 
+### Setting timestamp to custom Timezone
+
+By default, this Logger and the standard logging library emit records with the default AWS Lambda timestamp in **UTC**.
```
dreamorosi, line 524:
```suggestion
By default, Logger emits records with the default Lambda timestamp in **UTC**, i.e. `2016-06-20T12:08:10.000Z`
```

dreamorosi: Suggesting this change since we don't have a standard logging library in Node.js like Python does.

docs/core/logger.md
```diff
@@ -519,6 +519,28 @@ We prioritise log level settings in this order:
...
+
+<!-- markdownlint-disable MD013 -->
+If you prefer to log in a specific timezone, you can configure it by setting the `TZ` environment variable. You can do this either as an AWS Lambda environment variable or directly within your Lambda function settings. [Click here](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime){target="_blank"} for a comprehensive list of available Lambda environment variables.
```
dreamorosi, line 527:
```suggestion
If you prefer to log in a specific timezone, you can configure it by setting the `TZ` environment variable. You can do this either as an environment variable or directly within your Lambda function settings.

[Click here](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime){target="_blank"} for a comprehensive list of available Lambda environment variables.
```

docs/core/logger.md
```diff
@@ -519,6 +519,28 @@ We prioritise log level settings in this order:
 
 In the event you have set a log level in Powertools to a level that is lower than the ACL setting, we will output a warning log message informing you that your messages will be discarded by Lambda.
 
+### Setting timestamp to custom Timezone
+
+By default, this Logger and the standard logging library emit records with the default AWS Lambda timestamp in **UTC**.
+
+<!-- markdownlint-disable MD013 -->
```
dreamorosi, line 526:
```suggestion
```

docs/core/logger.md
```diff
@@ -519,6 +519,28 @@ We prioritise log level settings in this order:
...
+If you prefer to log in a specific timezone, you can configure it by setting the `TZ` environment variable. You can do this either as an AWS Lambda environment variable or directly within your Lambda function settings. [Click here](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime){target="_blank"} for a comprehensive list of available Lambda environment variables.
+<!-- markdownlint-enable MD013 -->
```
dreamorosi, line 528:
```suggestion
```

examples/snippets/logger/customTimezone.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const handler = async (_event, _context): Promise<void> => {
+  const loggerInUTC = new Logger({ serviceName: 'payment' });
+  loggerInUTC.info('Logging with default AWS Lambda timezone: UTC time');
+
+  process.env.TZ = 'US/Eastern'; // (1)!
+
+  const logger = new Logger({ serviceName: 'order' });
+  logger.info('Logging with US Eastern timezone');
+};
```
dreamorosi, line 11:
```suggestion
import { Logger } from '@aws-lambda-powertools/logger';

const logger = new Logger({ serviceName: 'serverlessAirline' });

export const handler = async (): Promise<void> => {
  logger.info('Hello, World!');

  process.env.TZ = 'Europe/Rome';

  const childLogger = logger.createChild();

  childLogger.info('Ciao, Mondo!');
};
```

examples/snippets/logger/customTimezoneOutput.json
```diff
@@ -0,0 +1,18 @@
...
+    "xray_trace_id": "1-6678152e-4f2ffb716d51f4112ec21b1d"
+  },
+  {
+    "level": "INFO",
+    "message": "Logging with US Eastern timezone",
+    "sampling_rate": 0,
+    "service": "order",
+    "timestamp": "2024-06-23T08:29:34.814-04:00",
+    "xray_trace_id": "1-6678152e-4f2ffb716d51f4112ec21b1d"
+  }
```
dreamorosi, line 17:
```suggestion
{
  "level":"INFO",
  "message":"Hello, World!",
  "sampling_rate":0,
  "service":"serverlessAirline",
  "timestamp":"2024-07-01T11:00:37.886Z",
  "xray_trace_id":"1-66828c55-2bb635c65eb609c820ebe7bc"
}
{
  "level":"INFO",
  "message":"Ciao, Mondo!",
  "sampling_rate":0,
  "service":"serverlessAirline",
  "timestamp":"2024-07-01T13:00:37.934+02:00",
  "xray_trace_id":"1-66828c55-2bb635c65eb609c820ebe7bc"
}
```

docs/core/logger.md
```diff
@@ -519,6 +519,28 @@ We prioritise log level settings in this order:
...
+
+=== "customTimezone.ts"
+
+    ```typescript hl_lines="7"
```
dreamorosi, line 532:
```suggestion
    ```typescript hl_lines="8"
```

docs/core/logger.md
```diff
@@ -519,6 +519,28 @@ We prioritise log level settings in this order:
...
+    --8<-- "examples/snippets/logger/customTimezone.ts"
+    ```
+    
+    1.  You don't need to set the TZ variable again when configuring it as an environment variable in AWS Lambda.
```
dreamorosi, line 536:
```suggestion
```

dreamorosi (commented):
Hi @arnabrahman - thank you for the excellent PR and also for the investigation in the `TZ` value on Lambda.

I have checked internally and I have been told that the `:UTC` value (including the `:`) is intended and that was set like this to be compliant with the [posix standard](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html). 

With that said, I agree that the docs are confusing and I'll be suggesting an update to that page to clarify the actual value. If you want to speed up the process, consider doing the same via the "Feedback" button in the doc page (Thank you!).

Regarding the PR, I have tested a build of this branch on Lambda and was able to confirm that it works as intended.

I have left a few suggestions for changes in the docs, mainly to format the new paragraph and rework the code snippets. This latter one especially is the biggest change from your PR.

In real world usage we don't recommend to instantiate a Logger instance inside the Lambda handler, this is so that the logger can be reused across invocations. However I understand what you were going for, so I have instead changed the sample to create a child logger after changing the `TZ` value.

Looking forward to merge this once the feedbacks are addressed! Thanks again for the great work!

dreamorosi (changes requested):
See comment above

arnabrahman:
I see the doc has been updated,

dreamorosi (approved):
Thank you for addressing the changes and the quick turnaround!

Yes, I noticed as well, well spotted!

I'm marking the PR as approved, however I need to investigate why one of the automation checks is not running - please allow me some time, I'll merge this as soon as it's fixed

Issue #2122, opened 2024-02-21
Feature request: ability to not throw error on full batch failure


dreamorosi:
### Use case

> [!note]
> This feature request comes from #1785

Currently when using the Batch Processor utility, if all the records in a batch are marked as failed the utility throws a `BatchProcessingError`.

Taking into consideration that the utility is supposed to be used with partial failure reporting, a Lambda function that throws an error is functionally equal to a partial failure that reports all items as failed in the sense that all the items in that batch are retried as a result.

While we initially implemented this as an error to reflect the full batch failure in the operational metrics (i.e. function runtime errors), there are cases such as when processing small batches that this behavior can skew the metrics due to higher chances of a full batch to fail.

To accommodate these use cases, as well as those customers who simply want to avoid throwing an error, we should add a new `throwOnFullBatchFailure` option to the utility that allows customers to opt out of the error throwing mechanism.

### Solution/User Experience

I haven't spent a lot of time thinking on where the new option should land, but the two options that come to mind are either when initializing the processor:

```ts
const processor = new BatchProcessor(
  EventType.SQS,
  {
    throwOnFullBatchFailure: false
});
```

or in the process function itself:

```ts
export const handler = async (
  event: SQSEvent,
  context: Context
): Promise<SQSBatchResponse> => {
  return processPartialResponse(event, recordHandler, processor, {
    context,
    throwOnFullBatchFailure: false
  });
};
```

Without looking at the internal implementation - which will inevitably inform the decision - I'm more inclined towards the second option for additional granularity (i.e. I might want to reuse the same `BatchProcessor` across multiple routes) and also because we already have a configuration object parameter that we can extend.

### Acknowledgment

- [X] This feature request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

486:
As discussed with @dreamorosi , I want to support the introduction of a `throwOnFullBatchFailure` option, for two reasons:

[...]

PR #2711, opened 2024-06-30
feat(batch): add option to not throw `FullBatchFailureError` when the entire batch fails

arnabrahman:
## Summary
Currently, when using the Batch Processor utility, if all the records in a batch are marked as failed, the utility throws a `FullBatchFailureError`. However, there can be scenarios where we do not want to throw an error for this condition.

This PR adds a `throwOnFullBatchFailure` option, giving users the ability to bypass the `FullBatchFailureError` when the entire batch fails.
### Changes

- Introduce `throwOnFullBatchFailure` option that can be passed as an option inside `processPartialResponse` or `processPartialResponseSync` functions
- During full batch failure
     - If `throwOnFullBatchFailure` is false,  we will not throw `FullBatchFailureError` exception
     - If `throwOnFullBatchFailure` is not provided or is `true`, we will keep the current implementation of throwing `FullBatchFailureError` exception
- Update the relevant doc blocks  
- Unit tests for full batch failure for processPartialResponse` and `processPartialResponseSync` function
- I believe we have to update the docs but couldn't figure it out in which section. For now, I have added a line for the option.

**Issue number:** #2122 

---


leandrodamascena:
Note to remember: @dreamorosi please let's talk before/after merging this PR to see if it makes sense to add the same in Python and @aws-powertools/lambda-dotnet-core

cc @hjgraca

dreamorosi:
Hi @leandrodamascena, we talked about if a few weeks ago when we did the v3 planning, the consensus was yes, since it impacts the scaling of the poller in some cases.

We can talk about it again also.

dreamorosi:
Hi @arnabrahman - thank you for the patience.

I'll make a couple of commits to the branch to explain the feature in the documentation before we can merge it.

dreamorosi:
> Note to remember: @dreamorosi please let's talk before/after merging this PR to see if it makes sense to add the same in Python and @aws-powertools/lambda-dotnet-core
> 
> cc @hjgraca

And to follow up on this: we went through the feature and we are aligned on the fact that we should move forward. We will also add the same flag on the other versions as time allows it.

docs/utilities/batch.md
```diff
@@ -450,6 +450,18 @@ We can automatically inject the [Lambda context](https://docs.aws.amazon.com/lam
 --8<-- "examples/snippets/batch/accessLambdaContext.ts"
 ```
 
+### Working with full batch failures
+
+By default, the `BatchProcessor` will throw a `FullBatchFailureError` if all records in the batch fail to process, we do this to reflect the failure in your operational metrics.
+
+In some cases, for example such as when working with small batches or when using errors as flow control mechanism, this behavior might not be desired and end up negatively impacting the concurrency of your function.
```
leandrodamascena, line 457: It took me some time to understand that "small batches" refers to the size of the batch and not the amount of messages I'm sending to SQS and processing. I think it makes sense to add "e.g 10 messages at a time".

**My idea (need to improve the english):**
In some cases, for example when working with small batches (e.g 10 message batch size) or when using errors as a flow control mechanism, if your function encounters exceptions (code failure), Lambda will scale down the concurrency, potentially causing performance problems. See the [official documentation ](https://docs.aws.amazon.com/lambda/latest/dg/services-sqs-errorhandling.html#services-sqs-backoff-strategy)for more details.

And what about adding this blog post? https://aws.amazon.com/pt/blogs/compute/introducing-faster-polling-scale-up-for-aws-lambda-functions-configured-with-amazon-sqs/

dreamorosi: OK, I have rephrased the section to make it more clear (I think) and in line with your suggestion.

I included the link to the documentation (thank you!) but opted to not include the one to the blog post, because the details about how errors impact scaling are just in a section that sets the context of the launch - which is itself the focus of the post.

Finally, I used the term `error` instead of `exception` because it's the most idiomatic for Node.js, and also I didn't include the code failure sentence because it kind of narrows down the meaning too much imo.

In my mental model, a code failure is a runtime unhandled exception/error, aka one that comes from a fault or bug in the code. For the type of cases I'm describing there, errors might be thrown intentionally.

leandrodamascena: This is so much better than before!

> In my mental model, a code failure is a runtime unhandled exception/error, aka one that comes from a fault or bug in the code. For the type of cases I'm describing there, errors might be thrown intentionally.

I completely agree with that!

leandrodamascena (approved):
Approved! Amazing work @arnabrahman and @dreamorosi! 

I'm adding this new feature in Python.

am29d (approved):

PR #2717, opened 2024-07-01
feat(maintenance): drop support for Node.js 16.x

### LINKED ISSUE #2223: Maintenance: drop support for Node.js 16

dreamorosi:
### Changes

This PR drops support for Node.js 16 for Powertools for AWS Lambda (TypeScript). As discussed in the linked issue, Node.js 16 has reached EOL and its corresponding AWS Lambda managed runtime has been marked as deprecated.

Below a list of changes included in the PR:
- Remove Node.js 16 from test matrixes, test harnesses, scripts, etc.
- Remove deprecated version from docs, comments, issue/PR templates
- Removed deprecation banner from docs, replaced with v1 EOL
- Raised baseline for Layer canary stack to Node.js 18
- Removed `nodejs16x` from compatible runtimes in Lambda layer
- Replaced usage of `Object(someObject).hasOwnProperty(someProperty)` with `Object.hasOwn(someObject, someProperty)` ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn))
- Removed polyfill for Error.cause ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause)) and 
 - updated `ParseError` in Parser utility to use it, as well as every envelope
 - updated Logger log formatter to assume the option can exist
- Removed conditional behavior around `fetch` in e2e tests since now it's present in all supported runtimes
 
**Issue number:** closes #2223

---

### DIFF
--- a/packages/idempotency/tests/helpers/populateEnvironmentVariables.ts
+++ b/packages/idempotency/tests/helpers/populateEnvironmentVariables.ts
@@ -1,7 +1,7 @@
 // Reserved variables
 process.env._X_AMZN_TRACE_ID = '1-abcdef12-3456abcdef123456abcdef12';
 process.env.AWS_LAMBDA_FUNCTION_NAME = 'my-lambda-function';
-process.env.AWS_EXECUTION_ENV = 'nodejs16.x';
+process.env.AWS_EXECUTION_ENV = 'nodejs20.x';
 process.env.AWS_LAMBDA_FUNCTION_MEMORY_SIZE = '128';
 if (
   process.env.AWS_REGION === undefined &&
--- a/packages/jmespath/src/errors.ts
+++ b/packages/jmespath/src/errors.ts
@@ -108,13 +108,6 @@ class ParseError extends JMESPathError {
  * Error thrown when an incomplete expression is encountered during parsing.
  */
 class IncompleteExpressionError extends ParseError {
-  /**
-   * Expression that was being parsed when the error occurred.
-   *
-   * Can be set by whatever catches the error.
-   */
-  public expression?: string;
-
   public constructor(options: {
     lexPosition: number;
     tokenValue: Token['value'];
--- a/packages/testing/tests/helpers/populateEnvironmentVariables.ts
+++ b/packages/testing/tests/helpers/populateEnvironmentVariables.ts
@@ -1,7 +1,7 @@
 // Reserved variables
 process.env._X_AMZN_TRACE_ID = '1-abcdef12-3456abcdef123456abcdef12';
 process.env.AWS_LAMBDA_FUNCTION_NAME = 'my-lambda-function';
-process.env.AWS_EXECUTION_ENV = 'nodejs16.x';
+process.env.AWS_EXECUTION_ENV = 'nodejs20.x';
 process.env.AWS_LAMBDA_FUNCTION_MEMORY_SIZE = '128';
 if (
   process.env.AWS_REGION === undefined &&
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -69,7 +69,6 @@ body:
       options:
         - 20.x
         - 18.x
-        - 16.x
     validations:
       required: true
   - type: dropdown
--- a/.github/actions/cached-node-modules/action.yml
+++ b/.github/actions/cached-node-modules/action.yml
@@ -14,10 +14,6 @@ outputs:
 runs:
   using: "composite"
   steps:
-    - name: Install npm
-      # We need to keep this npm version until we drop Node.js 16 support because Node.js 16 doesn't support npm 10
-      run: npm i -g npm@next-9
-      shell: bash
     - name: Cache node modules
       id: cache-node-modules
       uses: actions/cache@13aacd865c20de90d75de3b17ebe84f7a17d57d2 # v4.0.0
--- a/.github/workflows/reusable-run-linting-check-and-unit-tests.yml
+++ b/.github/workflows/reusable-run-linting-check-and-unit-tests.yml
@@ -13,7 +13,7 @@ jobs:
       NODE_ENV: dev
     strategy:
       matrix:
-        version: [16, 18, 20]
+        version: [18, 20]
       fail-fast: false
     steps:
       - name: Checkout code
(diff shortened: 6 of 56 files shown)


packages/idempotency/tests/helpers/populateEnvironmentVariables.ts
```diff
@@ -1,7 +1,7 @@
 // Reserved variables
 process.env._X_AMZN_TRACE_ID = '1-abcdef12-3456abcdef123456abcdef12';
 process.env.AWS_LAMBDA_FUNCTION_NAME = 'my-lambda-function';
-process.env.AWS_EXECUTION_ENV = 'nodejs16.x';
+process.env.AWS_EXECUTION_ENV = 'nodejs20.x';
```
am29d, line 4: should we put the default to lowest nodejst version as in other files?

packages/jmespath/src/errors.ts
```diff
@@ -108,13 +108,6 @@ class ParseError extends JMESPathError {
...
  */
 class IncompleteExpressionError extends ParseError {
-  /**
-   * Expression that was being parsed when the error occurred.
-   *
-   * Can be set by whatever catches the error.
-   */
-  public expression?: string;
-
   public constructor(options: {
```
am29d, line 111: Is this related to the PR?

packages/testing/tests/helpers/populateEnvironmentVariables.ts
```diff
@@ -1,7 +1,7 @@
 // Reserved variables
 process.env._X_AMZN_TRACE_ID = '1-abcdef12-3456abcdef123456abcdef12';
 process.env.AWS_LAMBDA_FUNCTION_NAME = 'my-lambda-function';
-process.env.AWS_EXECUTION_ENV = 'nodejs16.x';
+process.env.AWS_EXECUTION_ENV = 'nodejs20.x';
```
am29d, line 4: node18 as default?

am29d (changes requested):
Great work on keeping the track on all the places where node16 had a special treatment! 

Few notes on default node version in various places, so we can settle on either 18 or 20.

packages/jmespath/src/errors.ts
dreamorosi: Changing the `target` in `tsconfig.json` to Node.js 18 caused this issue to pop up.

This error extends another base error that has the same `expression` property, which was redeclared here, thus causing the error.

I should've added a comment in the PR to explain this, my bad.

packages/testing/tests/helpers/populateEnvironmentVariables.ts
dreamorosi: This is just an environment variable for the unit tests, I decided to jump to 20 since it's the same we run on the dev environment when we run the tests.

packages/idempotency/tests/helpers/populateEnvironmentVariables.ts
dreamorosi: Let's discuss it in [this thread](https://github.com/aws-powertools/powertools-lambda-typescript/pull/2717#discussion_r1662133826) and settle both.

packages/testing/tests/helpers/populateEnvironmentVariables.ts
dreamorosi: This environment variable is used only to do two things:
- Know whether we need to append the UA middleware to a client ([here](https://github.com/arnabrahman/aws-lambda-powertools-typescript/blob/8eeb735c8333a5adb649defa47aa5759fbec24ce/packages/commons/src/awsSdkUtils.ts#L4))
- Know whether we need to disable the Tracer ([here](https://github.com/arnabrahman/aws-lambda-powertools-typescript/blob/8eeb735c8333a5adb649defa47aa5759fbec24ce/packages/tracer/src/config/EnvironmentVariablesService.ts#L9))

The variable is used as a proxy for detecting whether we are running within Lambda or elsewhere (i.e. local/CI). In both cases Powertools only cares to know if the variable is present or not, since it's set automatically by Lambda.

For this reason, since we are running Node.js 20x in our dev environment I set it to this version; also so that for when we drop Node.js 18 we have two files less to touch.

However, I can also see an argument for setting it to Node.js 18 for consistency - if we agree on this I'm okay to revert these two files.

am29d: Thanks for the context, this clears things up! Because there is no special case for node18, like we had for node16, I suggest we can move to node20 as a default. What do you think?

dreamorosi: Our development environment is already on Node.js 20, which other default do you mean?

dreamorosi: Talked about this offline & clarified that we are already using Node.js 20 for dev. Resolving.

am29d: dev, layer and target, mostly for consistency. So we keep layer and target at 18 and dev at 20 as it is now.

am29d (approved):

Issue #1568, opened 2023-06-29
Feature request: Change ordering of default log formatter


bilalq:
### Use case

I know that users can provide their own LogFormatter instance, but the I believe the default should still be reasonably well-suited for most people.

The problem today is that the output frontloads a bunch of noise that isn't useful and makes the CloudWatch console log stream view useless without expanding and hard to read/skim even when you do.

At a glance, all you see is whether or not an invocation was a cold start and the ARN. Actually interesting information like log level and message are buried.

<img width="1465" alt="Screenshot 2023-06-29 at 6 58 58 AM" src="https://github.com/aws-powertools/powertools-lambda-typescript/assets/707147/d75c920b-943d-41f5-b049-5e7a0245cea6">

### Solution/User Experience

Technically, V8 and every other JS engine I'm aware of don't guarantee ordering of JS object literals when serializing/deserializing. In practice, when the object key counts are not high, the order as written does get preserved.

If we look here:

https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/logger/src/formatter/PowertoolLogFormatter.ts#L19-L33

Changing:

```typescript
    return {
      cold_start: attributes.lambdaContext?.coldStart,
      function_arn: attributes.lambdaContext?.invokedFunctionArn,
      function_memory_size: attributes.lambdaContext?.memoryLimitInMB,
      function_name: attributes.lambdaContext?.functionName,
      function_request_id: attributes.lambdaContext?.awsRequestId,
      level: attributes.logLevel,
      message: attributes.message,
      sampling_rate: attributes.sampleRateValue,
      service: attributes.serviceName,
      timestamp: this.formatTimestamp(attributes.timestamp),
      xray_trace_id: attributes.xRayTraceId,
    };
```

To something like this:

```typescript
    return {
      level: attributes.logLevel,
      message: attributes.message,
      function_request_id: attributes.lambdaContext?.awsRequestId,
      xray_trace_id: attributes.xRayTraceId,
      service: attributes.serviceName,
      function_arn: attributes.lambdaContext?.invokedFunctionArn,
      function_memory_size: attributes.lambdaContext?.memoryLimitInMB,
      function_name: attributes.lambdaContext?.functionName,
      sampling_rate: attributes.sampleRateValue,
      timestamp: this.formatTimestamp(attributes.timestamp),
      cold_start: attributes.lambdaContext?.coldStart,
    };
```

[...]

PR #2736, opened 2024-07-05
feat(logger): introduce log key reordering functionality

arnabrahman:
## Summary
By default, we can not change the log key positions. This PR gives the user the ability to change log key positions. We can change log order positions for any keys added in runtime
### Changes

- Mostly followed the recommendations from the [comment](https://github.com/aws-powertools/powertools-lambda-typescript/issues/1568#issuecomment-1786186089) from the actual issue.
- A new option ` logRecordOrder` is added for reordering.
- Change `formatAttributes` function for the new option.
- Write unit tests for this.
- Necessary documentation has been added with examples

**Issue number:** #1568 

---


dreamorosi:
Hi @arnabrahman, thank you for opening this PR!

I'll be reviewing it this morning and will get back to you, but overall I think we're going in the right direction.

packages/logger/src/formatter/PowertoolsLogFormatter.ts
dreamorosi, line 1: I've been playing around this, and since we are not doing any kind of deep object merge, I think we can try to implement it without using Lodash.

I've created this test bench suite ([here](https://jsbench.me/kslycmt0hc/1)) to compare the current implementation with some alternatives, and it looks like this one is ~35% faster:

```ts
const attributes = {};

for (const key of logRecordOrder) {
  if (key in baseAttributes && !(key in attributes)) {
    attributes[key] = baseAttributes[key];
  }
}

for (const key in baseAttributes) {
  if (!(key in attributes)) {
		attributes[key] = baseAttributes[key];
	}
}
```

dreamorosi:
> i believe we should type the option for standard structured keys

I was under the same impression, but looking at the docs again I think we're mistaken, since the docs mention this:

> You can change the order of [standard Logger keys](https://docs.powertools.aws.dev/lambda/python/latest/core/logger/#standard-structured-keys) or any keys that will be appended later at runtime

Additionally, in the example they show usage of a key (`request_id`) that is not part of the standard keys and that is added at runtime.

With this in mind I think we should follow the same pattern, if possible.

I haven't tested this in the code, but at first glance if I remember correctly, I think we should be able to include all keys to the new logic by working with the `additionalLogAttributes` (of type `LogAttributes`) object.

Doing this will mean we probably remove this line, which I'm not sure if it'll have any impact on the log output.

https://github.com/aws-powertools/powertools-lambda-typescript/blob/d76eb1d3d1569bd4675f2ff74c909b5b3976e232/packages/logger/src/formatter/PowertoolsLogFormatter.ts#L39

If this is a viable option, we should also modify the types of the new `logRecordOrder` accordingly (maybe an union between the keys of `UnformattedAttributes` and `LogAttributes`?).

---

Let me know if I cleared your questions 😄

arnabrahman:
This summer felt like an eternity but good to be back. 

I was looking at your suggestions @dreamorosi, on a high level it should be possible to include all the `additional` keys although I have to look a bit more to validate this. But first I have to merge the `main` branch but I cannot. I am getting `pre-commit` error for `markdownlint-cli2` in `packages/tracer/README.md`. I ran `setup-local` & tried again but still got the same result. Is there anything that needs to be done?

Screenshot from 2024-08-25 16-19-59

arnabrahman:
@dreamorosi So i have progressed on this. But I have another question. Right now the `logRecordOrder` passes down to child logger. What if we want to override `logRecordOrder` to child instance?


Currently while creating child, `logFormatter` is passed to child instance. https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/logger/src/Logger.ts#L308

And `logFormatter` is not reiniliazed if it's already set https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/logger/src/Logger.ts#L1085


In this case, if we provide `logRecordOrder` to child, it will not be overridden. Correct me if I am wrong

dreamorosi:
> In this case, if we provide logRecordOrder to child, it will not be overridden. Correct me if I am wrong

You are correct, we reuse the same log formatter from the parent because up until now it was reasonable to do so.

For the time being I would opt for keeping the logic as-is and prevent customers from overriding `logRecordOrder` when creating a child. 

This means that child loggers will have the same log formatter and log record order as their parent. If there's demand for this feature we can easily add it in the future, but for now I'd simply go ahead as-is.

dreamorosi:
Hi @arnabrahman, thank you for the PR - I've started reviewing it and should be able to finish by tomorrow morning at the latest.

packages/logger/src/formatter/LogFormatter.ts
```diff
@@ -19,8 +19,14 @@ abstract class LogFormatter implements LogFormatterInterface {
...
   protected envVarsService?: EnvironmentVariablesService;
 
+  /**
+   * An array of keys that defines the order of the log record.
+   */
+  protected logRecordOrder?: LogRecordOrder;
+
   public constructor(options?: LogFormatterOptions) {
     this.envVarsService = options?.envVarsService;
+    this.logRecordOrder = options?.logRecordOrder;
```
dreamorosi, line 29: I've been thinking about this a bit (also why the review took a bit longer), and I think this property - and feature - should be only available in the `PowertoolsLogFormatter` class.

My reasoning is that if customers are creating their own log formatter class by extending `LogFormatter`, then they already have full control over the order of the keys - so having this available doesn't add much and instead pollutes the API.

I think we should position this feature as a way to lightly extend the default log formatter without having to create your own. This also matches what the Python implementation of Powertools does.

To sum up, I would move the `logRecordOrder` property in the `PowertoolsLogFormatter` and use it only there. Additionally, I would also extend the `ConstructorOptions` type to make `logRecordOrder` & `logFormatter` mutually exclusive, [similar to what we do for this other use case](https://github.com/aws-powertools/powertools-lambda-typescript/blob/23a33dc52db9d2bb8cfa8ff043abcc1020ff958e/packages/logger/src/types/Logger.ts#L131).

dreamorosi (changes requested):
Hi @arnabrahman, thank you so much for the patience while I reviewed this and also for the great work!

I think we are getting very closer, I've left a few comments but the main implementation is done I think. 

Let me know if you have any questions or disagree with the comments.

dreamorosi (approved):
Thank you @arnabrahman for another great addition and contribution.

I really appreciate you coming back to this and working on the various iterations until we got here.

Thank you again!

PS: I'll reply to your comment on the other issue later today 😉

Issue #1776, opened 2023-10-31
Feature request: custom function for unserializable values in Logger (JSON replacer)


dreamorosi:
### Use case

Logger emits JSON structured logs, to do so it calls `JSON.stringify()` under the hood. There are several types of objects that cannot natively be serialized or that when serialized using the default behavior loose information. The Logger utility implements a custom replacer function that allows it to apply custom serialization on objects like `Error` or `BigInt` (added in #1117).

Over time, we have received requests for adding custom logic for more types like `Map` or `Set` (#1649) and in the fullness of time we expect to receive more as our customers log more and more types of objects.

To avoid bloating the library in an effort to serialize all kinds of types, we should instead enable customers to provide their own custom function that will be used internally by Logger when serializing logs.

In this way, customers who have edge cases can adapt the logger to their needs and all others can continue enjoying a lightweight logger.

### Solution/User Experience

The Logger utility should have a new constructor option called `jsonReplacerFn` that accepts a function that once set, supersedes the default replacer and gives full control over the serialization to the customer.

```ts
type CustomReplacerFn = (key: string, value: unknown) => void;

const jsonReplacerFn = (): CustomReplacerFn => ((_key, value) => {
  if (typeof value === 'bigint') {
    return value.toString();
  }

  return value;
});

const logger = new Logger({
  jsonReplacerFn
});
```

To implement the feature, the implementer should start with adding a new field of type `CustomReplacerFn` to the [`ConstructorOptions`](https://github.com/aws-powertools/powertools-lambda-typescript/blob/38edc369362030c4608cca75b477998de0f7d22b/packages/logger/src/types/Logger.ts#L24).

This value should be handled in the [`Logger.setOptions()`](https://github.com/aws-powertools/powertools-lambda-typescript/blob/38edc369362030c4608cca75b477998de0f7d22b/packages/logger/src/Logger.ts#L957) method, which should call a new private method called `setJsonReplacerFn()` (to be implemented).

The new `Logger.setJsonReplacerFn()` should handle the option passed during initialization, or default to the [`Logger.getReplacer()`](https://github.com/aws-powertools/powertools-lambda-typescript/blob/38edc369362030c4608cca75b477998de0f7d22b/packages/logger/src/Logger.ts#L698) method, which could also be renamed to `getDefaultJsonReplacerFn()`.

[...]

PR #2739, opened 2024-07-07
feat(logger): custom function for unserializable values (JSON replacer) 

arnabrahman:
## Summary
Logger emits JSON structured logs, to do so it calls JSON.stringify() under the hood. There are several types of objects that cannot natively be serialized or that when serialized using the default behavior loose information. 

Here we are giving customers a way out by providing a custom replacer function which will be used during `JSON.stringify`

### Changes

- Introduce `jsonReplacerFn` property that can be passed into the constructor during initialization.
- Use the custom replacer function if provided during stringify.
- Pass ``jsonReplacerFn` to child instance.
- Unit tests for new functionality
- Update docs for the new feature

Previously:
```

const logger = new Logger({ serviceName: 'serverlessAirline' });

export const handler = async (_event, _context): Promise<void> => {
  logger.info('Serialize with custom serializer', {
    serializedValue: new Set([1, 2, 3]),
  });
};

Output:
{
  "level": "INFO",
  "message": "Serialize with custom serializer",
  "sampling_rate": 0,
  "service": "serverlessAirline",
  "timestamp": "2024-07-07T09:52:14.212Z",
  "xray_trace_id": "1-668a654d-396c646b760ee7d067f32f18",
  "serializedValue": {}
}
```

Now while using custom replacer: 
```
import type { CustomReplacerFn } from '@aws-lambda-powertools/logger/types';

const jsonReplacerFn: CustomReplacerFn = (key: string, value: unknown) =>
  value instanceof Set ? [...value] : value;

const logger = new Logger({ serviceName: 'serverlessAirline', jsonReplacerFn });

export const handler = async (_event, _context): Promise<void> => {
  logger.info('Serialize with custom serializer', {
    serializedValue: new Set([1, 2, 3]),
  });
};

Output:
{
  "level": "INFO",
  "message": "Serialize with custom serializer",
  "sampling_rate": 0,
  "service": "serverlessAirline",
  "timestamp": "2024-07-07T09:52:14.212Z",
  "xray_trace_id": "1-668a654d-396c646b760ee7d067f32f18",
  "serializedValue": [1, 2, 3]
}
```

**Issue number:** #1776 

---


dreamorosi:
Hi @arnabrahman - the PR is looking good!

I'm going to take a few more hours to checkout the repo and review it more in details, but I should be able to post my review today.

dreamorosi (changes requested):
Thank you for this PR, this is a feature I've been wanting to use myself for a while but never got around actually working on it.

I have left a few minor comments related to naming of the new type and examples. Also, after testing the new feature, I realized we might have a chance to make it even better for customers.

As of today, Logger already has its own custom JSON replacer function. This is used to handle some special cases like circular references, `Error`, and `BigInt` types.

With the new feature proposed, customers will be able to provide their own custom JSON replacer function, this is useful in cases where they want to log even more complex types or data structures.

In doing this however they would be losing the goodies that Logger already comes with out-of-the box. For example, by providing a custom replacer, if you still want to log a `BigInt` you now get a runtime error:

```ts
const jsonReplacerFn: CustomReplacerFn = (
  key: string,
  value: unknown
) => {
  if (value instanceof Set || value instanceof Map) {
    return [...value];
  }

  return value;
};

const logger = new Logger({ jsonReplacerFn });

logger.info('many complex types', {
  map: new Map([[1, 2], [3, 4]]),
  set: new Set(['a', 'b', 'c']),
  bigInt: BigInt(42)
});
```

What if instead of replacing the function, we treated the new function as a way to **extend** the built-in JSON replacer? 

The experience would remain the same as the one you're proposing, except that instead of sidestepping the default JSON replacer, we'd call the one provided by the customer (if any) within the method that is currently called `getDefaultReplacer` (which could be renamed to `getJsonReplacer`:

```ts
private getJsonReplacer(): (_: string, value: unknown) => void {
  const references = new WeakSet();

  return (_, value) => {
    let item = value;

    if (this.jsonReplacerFn) item = this.jsonReplacerFn?.(_, item);

    if (item instanceof Error) {
      item = this.getLogFormatter().formatError(item);
    }
    if (typeof item === 'bigint') {
      return item.toString();
    }
    if (typeof item === 'object' && item !== null && value !== null) {
      if (references.has(item)) {
        return;
      }
      references.add(item);
    }

    return item;
  };
}
```

Naturally, to allow for the above we'd have to change/remove the `#setJsonReplacerFn` method to simply assign the `customerReplacerFn` value to the `this.jsonReplacerFn` prop while setting the options, and we'd also have to change the `printLog` method to call the `getDefaultReplacer` now-renamed method when logging.

So, to sum up, the changes needed would be:
- rename the `getDefaultReplacer` method ([here](https://github.com/arnabrahman/aws-lambda-powertools-typescript/blob/87989e4a59a66d92393b9b681569103703d34772/packages/logger/src/Logger.ts#L801)) to `getJsonReplacer`, and update the implementation similar to what suggested above
- since now there's no conditional logic, remove the `#setJsonReplacerFn` method ([here](https://github.com/arnabrahman/aws-lambda-powertools-typescript/blob/87989e4a59a66d92393b9b681569103703d34772/packages/logger/src/Logger.ts#L1192)) and assign the value directly [here](https://github.com/arnabrahman/aws-lambda-powertools-typescript/blob/87989e4a59a66d92393b9b681569103703d34772/packages/logger/src/Logger.ts#L1153)
- call the `getJsonReplacer` method [here](https://github.com/arnabrahman/aws-lambda-powertools-typescript/blob/87989e4a59a66d92393b9b681569103703d34772/packages/logger/src/Logger.ts#L1192)
- add an unit test case similar to the one suggested below, to confirm the change works as intended

```ts
it('should serialize using both the existing replacer and the customer-provided one', () => {
  const jsonReplacerFn: CustomReplacerFn = (
    key: string,
    value: unknown
  ) => {
    if (value instanceof Set || value instanceof Map) {
      return [...value];
    }

    return value;
  };

  const logger = new Logger({ jsonReplacerFn });
  const consoleSpy = jest.spyOn(
    logger['console'],
    getConsoleMethod(methodOfLogger)
  );

  const message = `This is an ${methodOfLogger} log with Set value`;
  const logItem = { value: new Set([1, 2]), number: BigInt(42) };

  // Act
  logger[methodOfLogger](message, logItem);

  // Assess
  expect(consoleSpy).toBeCalledTimes(1);
  expect(consoleSpy).toHaveBeenNthCalledWith(
    1,
    JSON.stringify({
      level: methodOfLogger.toUpperCase(),
      message: message,
      sampling_rate: 0,
      service: 'hello-world',
      timestamp: '2016-06-20T12:08:10.000Z',
      xray_trace_id: '1-5759e988-bd862e3fe1be46a994272793',
      value: [1, 2],
      number: '42',
    })
  );
});
```

Let me know what you think, or if you think this is a bad idea - and thank you for your time and effort as always!

arnabrahman:
First of all, I apologize for the delayed response. I had hoped to review all of your messages from all the threads, but due to work commitments, I haven't been able to. Unfortunately, I won't be able to address this any time soon, as I will be away for the next three weeks. I will get to it during the first week of August. @dreamorosi

dreamorosi:
Hey @arnabrahman, no need to apologize, life happens!

If it's okay for you, I'll push a couple commits to the PR so we can merge it.

In the meantime, enjoy your well deserved vacation!

arnabrahman:
@dreamorosi Sure, please go ahead.

dreamorosi:
@am29d - I took over the PR (see comments above for context).

It's now ready to review whenever you're ready. Thanks!

docs/core/logger.md
```diff
@@ -760,6 +760,26 @@ This is how the printed log would look:
 !!! tip "Custom Log formatter and Child loggers"
     It is not necessary to pass the `LogFormatter` each time a [child logger](#using-multiple-logger-instances-across-your-code) is created. The parent's LogFormatter will be inherited by the child logger.
 
+### Bring your own JSON serializer
+
+You can extend the default JSON serializer by passing a custom serializer function to the `Logger` constructor, using the `jsonReplacerFn` option. This is useful when you need to serialize custom objects or when you want to customize the serialization of specific values.
```
am29d, line 765: > serialize custom objects or when you want to customize the serialization of specific values. 

This needs more clarity, we have just exchanged the words without meaning what the difference is. We could remove the first part to simplify.

docs/core/logger.md
```diff
@@ -760,6 +760,26 @@ This is how the printed log would look:
...
+
+=== "unserializableValues.json"
+
+    ```json hl_lines="8"
+    --8<-- "examples/snippets/logger/unserializableValues.json"
+    ```
+
+By default, Logger uses `JSON.stringify()` to serialize log items and a [custom replacer function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#the_replacer_parameter) to serialize common unserializable values such as `BigInt`, circular references, and `Error` objects.
+
+When you extend the default JSON serializer, your custom serializer function will be called before the default one. This allows you to customize the serialization while still benefiting from the default behavior.
```
am29d, line 781: > your custom serializer function will be called before the default one
Can we reverse the order and make it active. Logger will call your customer serialiser before the default one.

am29d (changes requested):
Great addition, only minor comments on the docs

docs/core/logger.md
dreamorosi: Great feedback - removing the 1st part.

docs/core/logger.md
dreamorosi: Good point, changed to:

> When you extend the default JSON serializer, we will call your custom serializer function before the default one. This allows you to customize the serialization while still benefiting from the default behavior.

dreamorosi (approved):

am29d (approved):
Great collaborative work. Thanks you @arnabrahman and @dreamorosi 👏

Issue #2517, opened 2024-05-13
Bug: `IdempotencyItemAlreadyExistsError` loses referential identity when `IdempotencyConfig` is imported via external pkg


mrerichoffman:
### Expected Behavior

When using the Idempotency utility with ddb persistence, the utility should return the result of a previous operation from cache provided that the request is handled by the same execution environment.

### Current Behavior

When using ddb persistence the utility doesn't return the correct value from the cache, but instead causes a runtime error. If I revert from v2.1.0 to v1.17.0 the bug goes away.

### Code snippet

```typescript
import { IdempotencyConfig } from "@aws-lambda-powertools/idempotency";
import { DynamoDBPersistenceLayer } from "@aws-lambda-powertools/idempotency/dynamodb";
import { makeHandlerIdempotent } from "@aws-lambda-powertools/idempotency/middleware";
import middy from "@middy/core";

export interface ProcessResult {
  processResult: "processed" | "rejected";
}

const getDefaultIdempotencyConfigs = (eventKeyJmesPath = '["correlationId"]') => {
  const persistenceStore = new DynamoDBPersistenceLayer({
    tableName: process.env.IDEMPOTENCY_TABLE_NAME || "IdempotencyTable",
    sortKeyAttr: "sortKey"
  });
  const idempotencyConfig = new IdempotencyConfig({
    throwOnNoIdempotencyKey: true,
    eventKeyJmesPath
  });
  return {
    persistenceStore,
    idempotencyConfig
  };
};

const { persistenceStore, idempotencyConfig } = getDefaultIdempotencyConfigs();

const lambdaHandler = async (event: Record<"string", unknown>): Promise<ProcessResult> => {
  console.info("event", event);
  const processResult: ProcessResult = {
    processResult: "processed"
  };

  return processResult;
};

export const processLambda = middy(lambdaHandler).use(
  makeHandlerIdempotent({
    persistenceStore,
    config: idempotencyConfig
  })
);


``` 

### Steps to Reproduce

Then run the function with the same event payload, { "correlationId": "abc-123-def-456" }, twice and observe the error (shown below) being thrown at the second execution.

### Powertools for AWS Lambda (TypeScript) version

latest

### AWS Lambda function runtime

20.x

### Packaging format used

npm

### Execution logs

```Shell
2024-05-10T22:01:01.282Z	8034ee17-fd14-4244-8603-79500af0f7cb	ERROR	{
    "_logLevel": "error",
[...]

PR #2747, opened 2024-07-08
fix(idempotency): check error identity via names

dreamorosi:
### Changes

This PR makes changes to the custom errors that can be thrown by the Idempotency utility so that their respective [`name` property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/name) is set to the actual name of the error, for example:

```diff
class IdempotencyUnknownError extends Error {
  public constructor(message?: string, options?: ErrorOptions) {
    super(message, options);
+   this.name = 'IdempotencyUnknownError';
  }
}
```

This change allows to handle errors without having to import the error class itself, and without having to rely on a [referential identity](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness) check (i.e. `error instanceof IdempotencyUnknownError`).

Currently, we use `error instanceof ...` in several places within the internals of the utility. This is a result of two factors: **a/** the implementation relies heavily on throwing errors as a way of controlling flow, and **b/** composition and object-oriented programming patterns are prevalent, with errors being thrown in certain code paths and then handled gracefully with `try/catch` blocks in other parts. 

For example, we have a few instances in which we do something [like this](https://github.com/aws-powertools/powertools-lambda-typescript/blob/416093ed0fdc582f2e5e8c3db1662a4498c5f4ed/packages/idempotency/src/IdempotencyHandler.ts#L172C7-L189C8):

```ts
try {
  const { isIdempotent, result } =
    await this.#saveInProgressOrReturnExistingResult();
  if (isIdempotent) return result as ReturnType<Func>;

  return await this.getFunctionResult();
} catch (error) {
  if (
    error instanceof IdempotencyInconsistentStateError &&
    retryNo < MAX_RETRIES
  ) {
    // Retry
    continue;
  }
  // Retries exhausted or other error
  e = error;
  break;
}
```

Without getting too much into the weeds of what the code above does, the important parts to know are that:
1. we run some code in the `try` block - this code might throw either an error that we can recover from, or other types of errors that we can't handle
2. in the `catch` block we check if the error being caught is one that we know how to handle
  - if this is the case, we handle it
  - if not, we re-throw it

Because the code above relies on `instanceof`, JavaScript checks that the object in the left side of the statement is an instance of the class in the right side which might not be the case (for [a number of reasons](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness)). 

One of these reasons can happen because of the famed [dual package hazard](https://nodejs.org/api/packages.html#dual-package-hazard), which in simple terms means when one part of your code imports the CommonJS (CJS) version of a module, and another part of your code imports the same module but using ECMAScript modules (ESM).

This is something that our utilities are naturally exposed to since we started shipping ESM and CJS in v2.0. This is a tradeoff that we accepted at the time to allow customers to use Powertools for AWS Lambda regardless of what they used. 

In practice, this is relatively unlikely to happen to our customers since our utilities have only a few entry points and most of them do a relatively good job [at isolating state](https://nodejs.org/api/packages.html#approach-2-isolate-state). This PR tries to address one of the areas that did not do this propertly.

Specifically, under certain circumstances ([[1]](https://github.com/aws-powertools/powertools-lambda-typescript/issues/2517#issuecomment-2212531459) & [[2]](https://github.com/aws-powertools/powertools-lambda-typescript/issues/2517#issuecomment-2212538881)) customers might end up with multiple copies of the errors defined in the Idempotency utility.

When this happens, the internal flow of the utility breaks down causing requests to be repeated, which is what was happening in the linked issue. Once the changes in this PR are merged, it will be possible the utility should be more resilient to this type of issue because the checks on errors won't rely on referential identity anymore.

For example, the code from above would change to this:

```ts
try {
  const { isIdempotent, result } =
    await this.#saveInProgressOrReturnExistingResult();
  if (isIdempotent) return result as ReturnType<Func>;

  return await this.getFunctionResult();
} catch (error) {
  if (!(error instanceof Error))
    throw new IdempotencyUnknownError(
      'An unknown error occurred while processing the request.',
      { cause: error }
    );
  if (
    error.name === 'IdempotencyInconsistentStateError' && // we now use `error.name` instead of instanceof
    retryNo < MAX_RETRIES
  ) {
    // Retry
    continue;
  }
  // Retries exhausted or other error
  e = error;
  break;
}
```

With the changes above, it won't matter if the objects don't share the same identity, meaning that as long as they share the same implementation the utility will behave as expected.

There are additional areas of improvements that we can address to further decrease the risk of dual-package hazard such as providing even more granular subpath scoped exports (i.e. this customer was importing `IdempotencyConfig` via default export - aka barrel file - which caused the errors to be brought in because they're exported in the `packages/idempotency/src/index.ts` file [here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/idempotency/src/index.ts#L1-L10)). However, in the interest of keeping this essay short (_lol_) and the PR focused I'll look into making these changes in a future PR.

**Issue number:** #2517

---

### DIFF
--- a/packages/idempotency/src/IdempotencyHandler.ts
+++ b/packages/idempotency/src/IdempotencyHandler.ts
@@ -346,8 +346,11 @@ export class IdempotencyHandler<Func extends AnyFunction> {
 
       return returnValue;
     } catch (e) {
+      /* if (!(e instanceof Error)) throw e;
+      if (e.name === 'IdempotencyItemAlreadyExistsError') { */
       if (e instanceof IdempotencyItemAlreadyExistsError) {
-        let idempotencyRecord = e.existingRecord;
+        let idempotencyRecord = (e as IdempotencyItemAlreadyExistsError)
+          .existingRecord;
         if (idempotencyRecord !== undefined) {
           // If the error includes the existing record, we can use it to validate
           // the record being processed and cache it in memory.
--- a/packages/idempotency/src/IdempotencyHandler.ts
+++ b/packages/idempotency/src/IdempotencyHandler.ts
@@ -12,6 +12,7 @@ import {
   IdempotencyInconsistentStateError,
   IdempotencyItemAlreadyExistsError,
   IdempotencyPersistenceLayerError,
+  IdempotencyUnknownError,
 } from './errors.js';
 import { BasePersistenceLayer } from './persistence/BasePersistenceLayer.js';
 import { IdempotencyRecord } from './persistence/IdempotencyRecord.js';
@@ -176,8 +177,13 @@ export class IdempotencyHandler<Func extends AnyFunction> {
 
         return await this.getFunctionResult();
       } catch (error) {
+        if (!(error instanceof Error))
+          throw new IdempotencyUnknownError(
+            'An unknown error occurred while processing the request.',
+            { cause: error }
+          );
         if (
-          error instanceof IdempotencyInconsistentStateError &&
+          error.name === 'IdempotencyInconsistentStateError' &&
           retryNo < MAX_RETRIES
         ) {
           // Retry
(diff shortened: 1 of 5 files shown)


am29d (approved):
Thanks for the PR! It was a deep investigation and great informative summary to understand the problem. 

We could potentially extract the name strings into constants, to avoid any typo mistakes, but the number of errors is not high and we don't have many places where they are used. 

Overall, great work on narrowing down this complex issue 👏

Issue #2049, opened 2024-02-11
Feature request: Support Middy 5.x


saragerion:
### Use case

This feature request is about an upcoming official support of [Middy](https://github.com/middyjs/middy) 5.x. 

The current Powertools version 1.18.0 officially supports Middy up until version 3, as described in the [documentation](https://docs.powertools.aws.dev/lambda/typescript/1.18.0/core/logger/#capturing-lambda-context-info).

Support for Middy v4 is expected to be launched soon as part of [the upcoming major release 2.x](https://github.com/aws-powertools/powertools-lambda-typescript/issues/1694#issuecomment-1790427042).

If and when Middy v5 will be officially supported, it would be good to guide developers on what changes they need to implement in their codebase to successfully leverage the Logger utility together with Middy v5.

### Solution/User Experience

Upgrading to Middy v5 may have a significant impact on the codebase and developer experience.
See Middy's [documentation](https://middy.js.org/docs/upgrade/4-5/) for more info.

### Acknowledgment

- [X] This feature request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

dreamorosi:
Hi @saragerion thank you for opening this issue.

At the moment we don't have concrete plans for officially supporting Middy.js v5, however when the time comes I expect there will be challenges due to Powertools supporting both CJS and ESM and Middy.js having dropped support for the former in their v5 major version.

According to [our versioning policy](https://docs.powertools.aws.dev/lambda/typescript/latest/versioning/#dependency-lifecycle) we are committed to supporting all current AWS Lambda managed runtimes, which at the time of writing are the ones corresponding to Node.js 16 through Node.js 20.

On Middy.js side, [their release cadence](https://middy.js.org/docs/intro/release-cycle) aligns more closely with the Node.js LTS policy. This means that v5 dropped support for Node.js 16.

[...]

PR #2748, opened 2024-07-09
feat(internal): support Middy.js 5.x

dreamorosi:
### Changes

This PR adds support for Middy.js 5.x by including it in some of the integration test cases that are run on the utilities and widening the version range in the peer dependencies from `>= 4.x` to `4.x || 5.x`.

Because we want to continue supporting both versions, I decided to use an npm feature that allows you to install a package under a custom alias. This mechanism allows multiple versions of the same-name package side-by-side, which is what we need in this case.

To modify the `devDependencies` of the project I run these commands:

```sh
npm un @middy/core # remove existing version
npm i -D middy4@npm:@middy/core@4.x
npm i -D middy5@npm:@middy/core@5.x
```

This created two entries in the main `package.json` file, and allows us to import either of the two versions using the name of the alias [like so](https://github.com/aws-powertools/powertools-lambda-typescript/blob/8e6cb10661f1364b31f1f79e4b707644f0297665/packages/tracer/tests/e2e/allFeatures.middy.test.functionCode.ts#L1-L2):

```ts
import middy from 'middy5'; // this will resolve to the node_moduels/middy5/* contents
```

With this, I switched some of the test cases that were already running using ESM output to import v5, and left the others untouched.

I also updated the docs to reflect that we support both versions.

**Issue number:** #2049

---

### DIFF
--- a/docs/core/logger.md
+++ b/docs/core/logger.md
@@ -111,7 +111,7 @@ This functionality will include the following keys in your structured logs:
 === "Middy Middleware"
 
     !!! tip "A note about Middy"
-        We guarantee support only for Middy.js `v4.x`, that you can install it by running `npm i @middy/core@~4`.
+        We guarantee support for both Middy.js `v4.x` & `v5.x` with the latter being available only if you are using ES modules.
         Check their docs to learn more about [Middy and its middleware stack](https://middy.js.org/docs/intro/getting-started){target="_blank"} as well as [best practices when working with Powertools](https://middy.js.org/docs/integrations/lambda-powertools#best-practices){target="_blank"}.
 
     ```typescript hl_lines="2 14"
--- a/docs/core/metrics.md
+++ b/docs/core/metrics.md
@@ -199,7 +199,7 @@ You can add default dimensions to your metrics by passing them as parameters in 
 === "Middy middleware"
 
     !!! tip "A note about Middy"
-        Currently we support up to Middy `v4.x` that you can install it by running `npm i @middy/core@~4`.
+        We guarantee support for both Middy.js `v4.x` & `v5.x` with the latter being available only if you are using ES modules.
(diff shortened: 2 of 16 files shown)


am29d (commented):
Great addition of middy5! Few questions: 

> With this, I switched some of the test cases that were already running using ESM output to import v5, and left the others untouched.

Does this imply that not all e2e tests that can run with middy5 are covered and we need to add them overtime? Will this result in duplication? 

I just realised that we need to fix our peerDeps for parser as well :| will do in a separate PR.

dreamorosi:
> Does this imply that not all e2e tests that can run with middy5 are covered and we need to add them overtime?

Currently we have a number of e2e test cases (i.e. 10 <- fictional number for the sake of the argument). Of these 10, today we are running some of them (i.e. 5) using ESM and some other (i.e. the other 5) with CJS.

Generally speaking, the distribution between the two has been done in a way that each utility has at least 1 or more tests running in each mode.

Because of Middy.js 5.x only supporting ESM, the pool of eligible test cases where to import `middy5` is reduced to these cases only. On the other hand, generally speaking each utility only has one test group (i.e. [this is one](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/logger/tests/e2e/basicFeatures.middy.test.FunctionCode.ts)) that runs tests on functions using Middy.

With all this in mind, the changes in the PR ensure that at least one group of tests for each utility is testing Middy.js 5. I didn't add any new test, and I don't think we should do so, because we are already testing different combinations when looking at the tests all together.

dreamorosi:
> I just realised that we need to fix our peerDeps for parser as well :| will do in a separate PR.

Added in the last commit - good catch!

am29d (approved):
Great work as always, I am sure the middy5 support release will be a surprise for many 😎

PR #2774, opened 2024-07-14
fix(parser): links the error cause with ParseError

### LINKED ISSUE #2773: ParseError no longer use the error.cause property
### Expected Behavior

Earlier the error thrown within the handler would be attached to the ParseError to the .cause property. This no longer is the case. This broke one of our prod services that were relying on this property.

I think it was unintentionally removed as part of dropping Node 16 support.

Here's the link to the change.

https://github.com/aws-powertools/powertools-lambda-typescript/commit/e4eee73770ffccead9212a566335ec256a31af7d#diff-27fac689aa83bd57991ab46c30f4cd72584c529baa433ce899e8b7bd6db7a7deL21

### Current Behavior

ParseError.cause is undefined

### Code snippet

Just throw any validation error in the handler and look for error.cause property.

### Steps to Reproduce

You can run any tests in this file "packages/parser/tests/unit/envelope.test.ts" and check for error.cause property. It will be undefined.

### Possible Solution

I've submitted a PR which reverts the ParseError change that removed this behaviour. 

latest

20.x

### Packaging format used

npm

### Execution logs

_No response_

tusharf5:
### Changes

The ParseError links the passed error cause with the thrown error. 

**Issue number:** fixes #2773

---

### DIFF
--- a/packages/parser/tests/unit/envelope.test.ts
+++ b/packages/parser/tests/unit/envelope.test.ts
@@ -1,4 +1,4 @@
-import { z } from 'zod';
+import { z, ZodError } from 'zod';
 import { Envelope } from '../../src/envelopes/envelope.js';
 import { ParseError } from '../../src/errors.js';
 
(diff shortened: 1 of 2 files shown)


packages/parser/tests/unit/envelope.test.ts
dreamorosi, line 1: Could you please add the lines below at the very top of this file? I just realized this test is actually not being run in the CI:

```ts
/**
 * Test parser envelope
 *
 * @group unit/parser
 */
```

thanks.

dreamorosi (changes requested):
Thank you so much for taking the time to open this PR!

I have left a couple of comments that I think we should address before merging it.

tusharf5:
@dreamorosi  Made the changes based on the feedback, thanks.

dreamorosi (approved):
Thank you for addressing the feedback, merging it!

PR #2786, opened 2024-07-19
feat(parser): allow parser set event type of handler with middy

### LINKED ISSUE #2407: Feature request: allow parser set event type of handler with middy
### Use case

middy can pass event type like as [this](https://middy.js.org/docs/intro/typescript/). So we can pass inferred event type from zod schema to handler.

### Solution/User Experience

```ts
const detailSchema = z.object({
	date: z.string().regex(/^\d{4}-\d{2}-\d{2}$/),
	message: z.string(),
});

	// when using parser
	.use(parser({ schema: detailSchema, envelope: EventBridgeEnvelope }))
	// handler will get event type already inferred without type annotation.
	.handler(async (event) => {
		logger.inf(event.detail.message)
	});
```

### Alternative solutions

_No response_

### Acknowledgment

### Future readers

am29d:
## Summary

This PR introduces type inference for event type of handler when using parser with middy middleware. Based on the combination of provided inputs (schema, envelope, safeParse) the type can change between a single object, an array (i.e. `SqsEnvelope`) and safeParse, where we wrap the result into `ParsedResult`. I have modified the unit tests to use the inference which should have an error, if the type inference breaks.

### Changes

* fixed return types to array for envelopes that return multiple items of provided schema objects
* added `ParseOutput` to narrow down the type of the modified event
* moved `DynamoDBStreamEnvelopeResponse` to types folder
* replaced eslit comments with type casts in unit tests as mentioned in our docs

**Issue number:** closes #2407 

---

### DIFF
--- a/packages/parser/src/envelopes/sqs.ts
+++ b/packages/parser/src/envelopes/sqs.ts
@@ -9,7 +9,7 @@ import { ParseError } from '../errors.js';
  *
  *  The record's body parameter is a string, though it can also be a JSON encoded string.
  *  Regardless of its type it'll be parsed into a BaseModel object.
- *w
+ *
  *  Note: Records will be parsed the same way so if model is str,
  *  all items in the list will be parsed as str and npt as JSON (and vice versa)
  */
--- a/packages/parser/src/envelopes/sqs.ts
+++ b/packages/parser/src/envelopes/sqs.ts
@@ -9,7 +9,7 @@ import { ParseError } from '../errors.js';
  *
  *  The record's body parameter is a string, though it can also be a JSON encoded string.
  *  Regardless of its type it'll be parsed into a BaseModel object.
- *
+ *w
  *  Note: Records will be parsed the same way so if model is str,
  *  all items in the list will be parsed as str and npt as JSON (and vice versa)
  */
@@ -17,7 +17,7 @@ export class SqsEnvelope extends Envelope {
   public static parse<T extends ZodSchema>(
     data: unknown,
     schema: T
-  ): z.infer<T> {
+  ): z.infer<T>[] {
     const parsedEnvelope = SqsSchema.parse(data);
 
     return parsedEnvelope.Records.map((record) => {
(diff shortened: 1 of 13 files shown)


packages/parser/src/envelopes/sqs.ts
```diff
@@ -9,15 +9,15 @@ import { ParseError } from '../errors.js';
  *
  *  The record's body parameter is a string, though it can also be a JSON encoded string.
  *  Regardless of its type it'll be parsed into a BaseModel object.
- *
+ *w
```
dreamorosi, line 12:
```suggestion
 *
```

packages/parser/src/types/parser.ts
```diff
@@ -34,9 +38,46 @@ type ParsedResult<Input = unknown, Output = unknown> =
...
+> = undefined extends TEnvelope
+  ? ParsedResult<unknown, z.infer<TSchema>>
+  : TEnvelope extends EnvelopeArrayReturnType
+    ? ParsedResult<unknown, z.infer<TSchema>>
+    : ParsedResult<unknown, z.infer<TSchema>[]>;
+
+/**
+ * The output of the parser function, can be either schema inferred type or a ParsedResult
+ */
+type ParseOutput<
```
dreamorosi, line 65: I wonder if we should call this `ParserOutput` for consistency with the `ParserOptions`.

So far we seem to have called `Parser*` things that related to the parser function and `Parsed*` things that belong to the parsed object.

Wdyt?

dreamorosi (commented):
Great work on these types, just left one minor comment about naming (& a suggestion to fix a typo).

Looking forward to merge it!

packages/parser/src/types/parser.ts
am29d: Makes sense, I will change that. We have also `ParsedResult`, `ParsedResultSuccess` and `ParsedResultError`, should rename them as well 🤔 ?

dreamorosi (approved):
Thank you for this improvement!

PR #2847, opened 2024-07-26
chore(parser): adds scoped exports for schemas and envelopes

### LINKED ISSUE #2829: Feature request: provide scoped subpath exports for schemas & envelopes
### Use case

The Parser utility has a number of built-in schemas and envelopes that customers can use in their functions. Currently, all of these [except two](https://github.com/aws-powertools/powertools-lambda-typescript/blob/72803e2f99509bca41db1ee0f0523abd93150f48/packages/parser/package.json#L46-L53) are exposed via a barrel file `@aws-lambda-powertools/parser/schemas`.

We should work towards exporting each category of schemas and envelopes on a per-service basis via dedicated exports.

### Solution/User Experience

Customers should be able to import any given schema or envelope using a targeted export, for example, all the API Gateway-related schemas should be available under the `@aws-lambda-powertools/parser/schemas/api-gateway` export and all the S3 ones under `@aws-lambda-powertools/parser/schemas/s3`, the same applies to envelopes which would have, for example, SQS envelopes under `@aws-lambda-powertools/parser/envelopes/sqs`.

With this change, customers would be able to have a slimmer bundle since they'd be importing only the schemas/envelopes of the category they're actually using instead of all of them.

### Alternative solutions

_No response_

### Acknowledgment

### Future readers

daschaa:
### Changes

With this change the exports from schemas and envelopes are also available as scoped exports instead of just the barrel file.

**Steps of testing this change were:**

1. Build the parser utility with `npm run build` insided the parser directory
2. Run the prepack step `npm run prepack` to be able to import this in another project as file import.
3. Set up a testing project. 
4. Install the locally packed parser utility with `npm i $PATH_TO_POWERTOOLS/packages/parser`. This will result in an entry in the package.json like:
     ```
     "dependencies": {
      "@aws-lambda-powertools/parser": "file:../powertools-lambda-typescript-fork/packages/parser",
     }
    ```
5. Try to import schemas with the scoped import e.g.
    ```
    import { KinesisFirehoseSqsRecord } from '@aws-lambda-powertools/parser/schemas';
    ```

**Issue number: closes #2829** 

---


packages/parser/package.json
```diff
@@ -43,18 +43,134 @@
...
     },
+    "./schemas/alb": {
+      "require": "./lib/cjs/schemas/alb.js",
+      "import": "./lib/esm/schemas/alb.js"
+    },
     "./schemas/api-gateway": {
       "require": "./lib/cjs/schemas/apigw.js",
       "import": "./lib/esm/schemas/apigw.js"
     },
+    "./schemas/api-gateway-proxy": {
```
dreamorosi, line 54: I would not expose these at this stage - we extracted these schemas in a shared file so that they could be consumed by both API Gw & API Gw v2, but I don't see an use case for customers wanting to extend them.

For now I would leave them behind the scenes and consider adding them if there's demand.

dreamorosi (changes requested):
Thank you for the PR, and for the test steps in the PR body.

I have a concern only about one group which I would rather exclude from the PR, other than that it looks good.

dreamorosi (approved):
Thank you again for this PR and improving everyone's bundles!

PR #2865, opened 2024-07-31
fix(logger): invalid time zone environment variables leads to error

### LINKED ISSUE #2851: Bug: logger breaks on nodejs Docker lambdas since 2.4.0
The base image used to build Docker images, `public.ecr.aws/lambda/nodejs:20`, sets the environment variable `TZ=:/etc/localtime`. This value is not recognised as valid one and throws an error. I believe this was introduced by [this PR](https://github.com/aws-powertools/powertools-lambda-typescript/pull/2710) that was merged in 2.4.0.

### Expected Behavior

Building a Docker Lambda using the logger library without further configuration, i.e. I'd expect the library to print times in UTC when TZ is `:/etc/localtime`

### Current Behavior

`RangeError: Invalid time zone specified: :/etc/localtime` is raised when logging anything under this condition

### Code snippet

```typescript

const log = new Logger();
log.info("Hello");
```

### Steps to Reproduce

1. Build a Docker Lambda from `public.ecr.aws/lambda/nodejs:20`
2. Install the logger library
3. Log something

### Possible Solution

The library should print times in UTC when TZ is `:/etc/localtime`

latest

20.x

### Packaging format used

npm

### Execution logs

```Shell
RangeError: Invalid time zone specified: :/etc/localtime
    at new DateTimeFormat (<anonymous>)
    at #getDateFormatter (/var/task/node_modules/.pnpm/@aws-lambda-powertools+logger@2.5.0/node_modules/@aws-lambda-powertools/logger/lib/cjs/formatter/LogFormatter.js:85:16)
    at #generateISOTimestampWithOffset (/var/task/node_modules/.pnpm/@aws-lambda-powertools+logger@2.5.0/node_modules/@aws-lambda-powertools/logger/lib/cjs/formatter/LogFormatter.js:103:82)
    at PowertoolsLogFormatter.formatTimestamp (/var/task/node_modules/.pnpm/@aws-lambda-powertools+logger@2.5.0/node_modules/@aws-lambda-powertools/logger/lib/cjs/formatter/LogFormatter.js:53:56)
    at PowertoolsLogFormatter.formatAttributes (/var/task/node_modules/.pnpm/@aws-lambda-powertools+logger@2.5.0/node_modules/@aws-lambda-powertools/logger/lib/cjs/formatter/PowertoolsLogFormatter.js:32:29)
    at Logger.createAndPopulateLogItem (/var/task/node_modules/.pnpm/@aws-lambda-powertools+logger@2.5.0/node_modules/@aws-lambda-powertools/logger/lib/cjs/Logger.js:722:39)
    at Logger.processLogItem (/var/task/node_modules/.pnpm/@aws-lambda-powertools+logger@2.5.0/node_modules/@aws-lambda-powertools/logger/lib/cjs/Logger.js:827:46)
    at Logger.info (/var/task/node_modules/.pnpm/@aws-lambda-powertools+logger@2.5.0/node_modules/@aws-lambda-power
```

daschaa:
### Changes

With this change the time zone argument for the `DateFormat` in the `Logger` utility is set to `UTC` whenever the time zone from the environment variable is not supported (for example `:/etc/localtime`).

**Issue number: closes #2851** 

---


packages/logger/src/formatter/LogFormatter.ts
```diff
@@ -112,6 +112,19 @@ abstract class LogFormatter implements LogFormatterInterface {
   #getDateFormatter = (timeZone: string): Intl.DateTimeFormat => {
     const twoDigitFormatOption = '2-digit';
 
+    if (!Intl.supportedValuesOf('timeZone').includes(timeZone)) {
```
dreamorosi, line 115: Nice one!

I would extract the value in a variable to avoid extra boilerplate and do this:

```ts
const validTimeZone = Intl.supportedValuesOf('timeZone').includes(timeZone) ? timeZone : 'UTC';
```

and then modify the existing `new Intl.DateTimeFormat()` call below to use `validTimeZone`.

dreamorosi (changes requested):
Thank you for the quick PR, and great idea to test the value using existing libraries - I didn't know that method existed.

I've left one minor comment to make the code less verbose, but other than that it's looking good.

dreamorosi (approved):
Thank you for picking up the issue and fixing the bug!

PR #2892, opened 2024-08-07
docs: add parser example for built-in extension using `transform` and `pipe`

### LINKED ISSUE #2891: Docs: add parser example to extend built-in schema with JSON parse

We have an example how to extend a built-in schema for `EventBridgeSchema` which is a trivial case where there is no transformation needed, because the input is an object. 

In cases when we haven a stringified object inside the payload i.e. SQS or API Gateway, we need to use `.transform()` and `.pipe()` functions from zod. A missing example was identified in #2844 . 

_No response_

### How can we improve?

Add an example how to extend a built in schema for `APIGatgewayProxyEvent` which takes a string and uses `transform()` and `pipe()` functions.

### Got a suggestion in mind?

_No response_

### Acknowledgment

am29d:
## Summary

This PR add an example how to extend a built-in schema with `transform` and `pipe` zod functions. This allows more complex transformations and schema chaining. 

I have removed the missing examples in `api-gateway.md`, otherwise you can't build docs.

### Changes

**Issue number:** closes #2891 

---

### DIFF
--- a/examples/snippets/parser/extendAPIGatewaySchema.ts
+++ b/examples/snippets/parser/extendAPIGatewaySchema.ts
@@ -1,7 +1,7 @@
 import type { LambdaInterface } from '@aws-lambda-powertools/commons/types';
 import { Logger } from '@aws-lambda-powertools/logger';
 import { parser } from '@aws-lambda-powertools/parser';
-import { APIGatewayProxyEventSchema } from '@aws-lambda-powertools/parser/schemas';
+import { APIGatewayProxyEventSchema } from '@aws-lambda-powertools/parser/schemas/api-gateway';
 import type { Context } from 'aws-lambda';
 import { z } from 'zod';
 
--- a/examples/snippets/parser/extendAPIGatewaySchema.ts
+++ b/examples/snippets/parser/extendAPIGatewaySchema.ts
@@ -1,0 +1,50 @@
+import type { LambdaInterface } from '@aws-lambda-powertools/commons/types';
+import { Logger } from '@aws-lambda-powertools/logger';
+import { parser } from '@aws-lambda-powertools/parser';
+import { APIGatewayProxyEventSchema } from '@aws-lambda-powertools/parser/schemas';
+import type { Context } from 'aws-lambda';
+import { z } from 'zod';
+
+const logger = new Logger();
+
+const orderSchema = z.object({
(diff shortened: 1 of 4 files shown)


dreamorosi (approved):
Great addition to the docs!

I think we should also add this helper to our utility and export it, so that customers can just use it right away and have less boilerplate code.

For example, we could create a new file under `packages/parser/src/helpers.ts` and add a new schema helper like:

```ts
import { z, type ZodSchema } from 'zod';

const JSONStringified = (schema: ZodSchema) =>
    z
      .string()
      .transform((str, ctx) => {
        try {
          return JSON.parse(str);
        } catch (err) {
          ctx.addIssue({
            code: 'custom',
            message: 'Invalid JSON',
          });
        }
      })
      .pipe(schema);

export { JSONStringified };
```

so that later, customers could import it in their code and use it like:

```ts
import { APIGatewayProxyEventSchema } from '@aws-lambda-powertools/parser/schemas/api-gateway';

const orderSchema = z.object({
  orderId: z.number(),
  order: z.object({
    amount: z.number()
  })
});

const orderEventSchema = APIGatewayProxyEventSchema.extend({
  body: JSONStringified(orderSchema)
});
```

Putting myself into this situation, I would find it very useful to do this rather than going back to the docs every time, also in the future we could add more of these helpers if they come up. But for now at least we have it documented!

I leave it to you if you want to do it now, or open an issue so that someone can pick it up before the next release.

examples/snippets/parser/extendAPIGatewaySchema.ts
```diff
@@ -0,0 +1,50 @@
+import type { LambdaInterface } from '@aws-lambda-powertools/commons/types';
+import { Logger } from '@aws-lambda-powertools/logger';
+import { parser } from '@aws-lambda-powertools/parser';
+import { APIGatewayProxyEventSchema } from '@aws-lambda-powertools/parser/schemas';
```
dreamorosi, line 4:
```suggestion
import { APIGatewayProxyEventSchema } from '@aws-lambda-powertools/parser/schemas/api-gateway';
```

am29d:
You are right, having a small helper function is much better than pointing to the docs. I will create a separate issue for that

am29d:
Closing in favour of new helper function and docs.

PR #2902, opened 2024-08-08
feat(logger): introduce loglevel trace #1589

### LINKED ISSUE #1589: Feature request: add additional log level: TRACE & corresponding `logger.trace()` method
### Use case

Create an additional log level of `TRACE` below `DEBUG`

`DEBUG` contains statements about entering specific functions, iterations of a loop, timing info, etc. as well as actual data such as function parameters, API responses etc.

`INFO` is reserved for normal informational messages, like logging the results of an invocation but not all the intermediate details.

`TRACE` level would allow `DEBUG` to only contain the data, API responses, etc. and `TRACE` then contains the statements for the execution path.

### Solution/User Experience

Add an additional level for `TRACE` as was done for `CRITICAL`. 

### Alternative solutions

_No response_

### Acknowledgment

- [X] This feature request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda-typescript/latest/#tenets)

### Future readers

timo92:
### Changes

This PR introduces a new log level `TRACE` below 'DEBUG' in alignment with the AWS Lambda ALC.

**Issue number:** closes #1589 

---


timo92:
I performed some refactorings of the exisiting code in the last two commits. Let me know if I should move them into a seperate PR.
So far I was unable to run the tests locally due to `Cannot find module '@aws-lambda-powertools/commons'`. Will try again in case some of the merge checks fail.

dreamorosi:
Hi @timo92, thanks for the quick PR.

Looking at the test results it seems that we have two issues:
- Some of the new tests are failing because the `console.trace()` method seems to be `undefined` (see [here](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/10298131939/job/28503374312?pr=2902#step:6:872))
- Test coverage is below 100% (see [here](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/10298131939/job/28503374312?pr=2902#step:6:1018)) - I'd suggest we look at this after resolving the previous one.

The first one is quite interesting, and I think it will require a bit more time on this PR.

Before getting into its details, a bit of context is needed. By default, Logger does not use the global `console` object but instead instantiates its own `Console` (#748). The only case when we use the global one is when customers set the `POWERTOOLS_DEV` environment variable to `true`, which as the name suggests it's a behavior we encourage only during development.

According to the [Node.js docs](https://nodejs.org/docs/latest/api/console.html#new-consoleoptions):
> The global `console` is a special `Console` whose output is sent to [process.stdout](https://nodejs.org/docs/latest/api/process.html#processstdout) and [process.stderr](https://nodejs.org/docs/latest/api/process.html#processstderr).

Even though this is not explicitly called out in the docs, the global `console` has a lot more methods than the parent `Console` - this includes the `trace()` method ([docs](https://nodejs.org/docs/latest/api/console.html#consoletracemessage-args)), **and it's also is why the unit tests are failing.**

My first idea was to suggest we patch the `this.console` object and add a new `trace` method that simply logs whatever it's provided. This could work for when the Logger runs on AWS Lambda, and it's roughly equivalent to what the Lambda service actually does in the Node.js managed runtime (see [here](https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/9af681895d0cdd2094ef452342a604e94dc82cc3/src/LogPatch.js#L164-L166)).

This however requires some extra consideration because in Node.js the `console.trace()` method is supposed to behave differently than the other one, and instead of just logging its arguments it's supposed to log a formatted message and stack trace to the current position in the code, for example:

```js
console.trace('Show me');
// Prints: (stack trace will vary based on where trace is called)
//  Trace: Show me
//    at repl:2:9
//    at REPLServer.defaultEval (repl.js:248:27)
//    at bound (domain.js:287:14)
//    at REPLServer.runBound [as eval] (domain.js:300:12)
//    at REPLServer.<anonymous> (repl.js:412:12)
//    at emitOne (events.js:82:20)
//    at REPLServer.emit (events.js:169:7)
//    at REPLServer.Interface._onLine (readline.js:210:10)
//    at REPLServer.Interface._line (readline.js:549:8)
//    at REPLServer.Interface._ttyWrite (readline.js:826:14)
```

This is clearly not what we want, however it means we will need to also modify the logging behavior when `POWERTOOLS_DEV` is enabled, and document all this so that people are clear on the behavior being aligned with Lambda rather than expecting a stack trace.

--

Side note 1: If you want a faster feedback loop, you can [setup the local dev environment](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/CONTRIBUTING.md#local-environment), by running `npm run setup-local` which will install & build the packages. This is not mandatory, and you can continue pushing the changes here and we'll run them in CI.

Side note 2: The scope of this task unexpectedly grew beyond what was going to be a simple change. We are more than happy to work with you and iterate on the PR, however if you feel like it's not what you thought and want to close the PR we completely understand and will put the issue back in the backlog.

packages/logger/package.json
dreamorosi, line 1: If possible, please revert these changes in formatting so we can keep the PR focused.

dreamorosi (changes requested):
Hi @timo92, apologies for the delayed review.

I have left a couple of minor comments and a suggestion. After they are addressed I think we can merge this right away.

packages/logger/tests/unit/Logger.test.ts
```diff
@@ -57,6 +59,7 @@ describe('Class: Logger', () => {
   beforeEach(() => {
     dateSpy.mockClear();
     process.env = { ...ENVIRONMENT_VARIABLES };
+    process.env.POWERTOOLS_LOG_LEVEL = 'trace';
```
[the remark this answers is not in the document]
timo92: Currently the `POWERTOOLS_LOG_LEVEL` variable is set to `debug` somewhere before tests are executed. I initially added the line to support the new test cases for `trace` as they failed due to being a lower level than `debug.
However I've now updated the tests in the `describe.each` block to explicitly set the logLevel to the logLevel being tested.

packages/logger/tests/unit/Logger.test.ts
```diff
@@ -3595,7 +3608,7 @@ describe('Class: Logger', () => {
       const debugSpy = jest.spyOn(console, 'debug').mockImplementation();
 
       // Act
-      new Logger({ sampleRateValue: 1 });
+      new Logger({ sampleRateValue: 1, logLevel: 'WARN' });
```
timo92, line 3611: Sonarcloud reports this line as being useless however it produces a log message as a side effect.
@dreamorosi Is it possible to suppress the warning in sonarcloud?

dreamorosi: Agreed, I've suppressed it

packages/logger/tests/unit/Logger.test.ts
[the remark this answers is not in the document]
dreamorosi: Makes sense, I just wanted to make sure the PR was focused to the change and avoid having unintended side effects. But I see your point - thanks for reverting it

PR #2911, opened 2024-08-12
chore(maintenance): replace copy command in docs dockerfile

### LINKED ISSUE #2910: Maintenance: replace usage of `ADD` in docs Dockerfile
### Summary

SonarCloud recommends using the `COPY` instruction over the `ADD` for copying local resources. According to `docker:S7029`, the COPY instruction is more straightforward and less prone to errors, making it a better choice for copying local files and directories into a Docker image.

Since it's a one-liner, we should replace the `ADD` usage in the `docks/Dockerfile` file.

### Why is this needed?

Quoting the SonarCloud explanation:

> Using the ADD instruction instead of COPY for local resources in Dockerfiles can lead to several issues, including unexpected behavior, increased complexity, and security risks. The ADD instruction has additional features that can introduce unintended side effects, such as automatically extracting compressed files and fetching remote resources. This can make the behavior of the instruction less predictable and harder to understand or even lead to security issues, if, for example, due to a typo in the source path, the ADD instruction could fetch a remote resource instead of copying a local file.
 If you only need to copy local files or directories into your Docker image, it is recommended to use the COPY instruction instead. Only use the ADD instruction when you need its additional features, such as fetching remote resources or extracting compressed files.

Other

### Solution

_No response_

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR replaces usage of the `ADD` command in the docs Dockerfile with the more straightforward, and according to SonarCloud, better `COPY`.

See the linked issue for full explanation provided by SonarCloud.

**Issue number:** closes #2910

---

### DIFF
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -1,5 +1,5 @@
 # version 9.5.2
 FROM squidfunk/mkdocs-material@sha256:9919d6ee948705ce6cd05e5bc0848af47de4f037afd7c3b91e776e7b119618a4
 
-ADD requirements.txt /tmp/
+COPY requirements.txt /tmp/
 RUN pip install --require-hashes -r /tmp/requirements.txt


docs/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # version 9.5.2
 FROM squidfunk/mkdocs-material@sha256:9919d6ee948705ce6cd05e5bc0848af47de4f037afd7c3b91e776e7b119618a4
 
-ADD requirements.txt /tmp/
+COPY requirements.txt /tmp/
```
leandrodamascena, line 4: To be honest, in this case there is no difference between ADD and COPY, they both work the same way, the difference is that ADD can unpack files and accept remote URLs, while COPY cannot.

To be more "docker-efficient" this should use [Bind mounts](https://docs.docker.com/build/building/best-practices/#add-or-copy) for temporary `RUN` commands, but that's completely irrelevant in this case where it's a single build for something that's ephemeral/temporary/local.

Approving this PR, as this change makes sense to be more compliant with Sonar.

leandrodamascena (approved):
APPROVED!

Issue #2821, opened 2024-07-23
Maintenance: refactor tests (or implementation) so that we can remove usage of literal keys in some tests


daschaa:
### Summary

In the test files we have a lot of accesses that look like `object['property'`. By default, biome throws an error if this kind of access is used and it suggests to use the literal key access instead (`object.property`).
However, in our test cases this can not be done, because the properties are private and therefore this kind of access would fail.

Therefore we need to extend the biome configuration to override the linting for test files, to ignore this rule in test files. I suggest to add the following property to the `biome.json` 

```json
{
"overrides": [
    {
      "include": ["**/*.test.ts"],
      "linter": {
        "rules": {
          "complexity": {
            "useLiteralKeys": "off"
          }
        }
      }
    }
  ],
}
```

### Why is this needed?

We do not want to flood our test files with biome exception comments.

### Which area does this relate to?

Tests

### Acknowledgment

- [X] This request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

daschaa:
@dreamorosi If you agree with the approach, I would like to take the issue :)

dreamorosi:
Hey @daschaa - thank you for opening the issue and proposing a solution.

Before moving forward I'd like us to spend some time evaluating if there's any other option that would allow us to make these properties more accessible.

At least for some of them, namely the `console`, I know some customers might want to access it in their own tests, so perhaps converting them to `protected` could be a good idea (I don't know yet!) as it would allow us to do something like this in the tests:

```ts
class MockMetrics extends Metrics {
  public declare console: Console;
}

// now you can spy on console

it('does stuff', () => {
  const metrics = new MockMetrics();
  const logSpy = vi.spyOn(metrics.console, 'log'); // <- this works
})
```

Alternatively, maybe this is a signal that we are testing the code in the wrong way and we should instead refactor the tests to observe side effects rather than reach into the internals of the class.

For example, let's take this test as example:

```ts
[...]

PR #2942, opened 2024-08-16
test(logger): simplify unit tests structure

dreamorosi:
### Changes

This PR aims at improving maintainability and lower the barrier of entry for working on the unit tests for the Logger feature.

Before working on this PR we had 310 unit test cases, many of which redundant, and most importantly the majority of the tests were organised in a single big file (`Logger.test.ts`) which after 3 yrs of work had grown to over 3K LOC.

As primary objective, the PR breaks down the big test file in smaller and more focused files. As a secondary objective, the PR also in many cases changed the test strategy to prioritize observing side effects (i.e. the log output) versus testing the internals of the Logger or its implementation details.

The result of this work is a new test suite with ~130 test cases and the same coverage level that is instead more interested in asserting that the logger actually logs things in the expect way rather than how its internals work.

With this new testing strategy, we are also enabling `POWERTOOLS_DEV` by default for all tests except one, which allows us to put spies on the global console object, and remove the behavior described in #2821.

In future PRs, once we have migrated this module to Vitest, I'll continue investing some time in further improving the tests by creating custom assertions that will allow us to cut even more boilerplate from the tests.

**Issue number:** #2821

---


packages/logger/tests/unit/configFromEnv.test.ts
```diff
@@ -0,0 +1,198 @@
...
+  it('returns the value of the environment variable AWS_LAMBDA_LOG_LEVEL and aliases it as needed', () => {
+    // Prepare
+    process.env.AWS_LAMBDA_LOG_LEVEL = 'FATAL';
+    const service = new EnvironmentVariablesService();
+
+    // Act
+    const value = service.getAwsLogLevel();
+
+    // Assess
+    expect(value).toEqual('CRITICAL');
```
leandrodamascena, line 26: The only thing that seems odd here is that you SET a `FATAL` level, but compare it to `CRITICAL` and it's true. I know why it's true, but if you add comments in Prepare/Act/Assess, it might help others to understand this test.

packages/logger/tests/unit/logEvent.test.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  it('logs the event with the correct log level and message', () => {
+    // Prepare
+    const logger = new Logger();
+
+    // Act
+    logger.logEventIfEnabled(event);
+
+    // Assess
+    expect(logSpy).toHaveBeenCalledTimes(1);
```
leandrodamascena, line 39: Need this?

packages/logger/tests/unit/logEvent.test.ts
```diff
@@ -0,0 +1,147 @@
...
+    const logger = new Logger();
+    const handler = middy(async () => {}).use(
+      injectLambdaContext(logger, { logEvent: true })
+    );
+
+    // Act
+    await handler(event, context);
+
+    // Assess
+    expect(logSpy).toHaveBeenCalledTimes(1);
```
leandrodamascena, line 79: Need this?

leandrodamascena (approved):
Hello @dreamorosi! Just a few comments that don't impact the tests, but make it easier to read and understand.

APPROVED!

packages/logger/tests/unit/logEvent.test.ts
dreamorosi: Strictly speaking it's not needed, but it aims at giving a better error message in case of failure.

In the next line, we access the content of the first call via `logSpy.mock.calls[0][0]`, which in case of the spy not getting called would fail with an error similar to "unable to access index 0 of undefined" (I'm paraphrasing the error message for clarity).

I found the error message a bit confusing and I would rather it failing earlier at the `expect(logSpy).toHaveBeenCalledTimes(1);` which gives me a better error of "spy hasn't been called at all" (again, paraphrasing).

This is a lot of repetition though, every time I need to check the content of a call I need to first check that it was called, then parse it (next line), etc. I am already planning on creating a custom assertion in the future, which would allow me to replace both lines with this `expect(logSpy).toHaveLoggedObjectContaining({ foo: 'bar' })` but this would be test-runner specific and I'm waiting to migrate this package to vitest before actually doing it.

packages/logger/tests/unit/logEvent.test.ts
dreamorosi: See answer [here](https://github.com/aws-powertools/powertools-lambda-typescript/pull/2942/files/55a2b1e6fee4469c07d280c5d3e9d2df83f836b8#r1722048526).

packages/logger/tests/unit/injectLambdaContext.test.ts
```diff
@@ -0,0 +1,193 @@
...
+    logger.info('Hello, world!');
+
+    // Assess
+    expect(logSpy).toHaveBeenCalledTimes(2);
+    expect(JSON.parse(logSpy.mock.calls[0][0])).toStrictEqual(
+      expect.objectContaining({
+        message: 'Hello, world!',
+        ...getContextLogEntries(),
+      })
+    );
```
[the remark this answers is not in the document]
dreamorosi: You are right, I can remove this first assertion and leave only the second one that tests the updated values - good catch!

packages/logger/tests/unit/configFromEnv.test.ts
dreamorosi: Good point, will add a comment to explain the quirk.

packages/logger/tests/unit/logEvent.test.ts
leandrodamascena: Aaa ok, I see! You are trying to avoid generic errors like undefined/key_errors and focusing on having the utility response, that way if at least 1 call is not made you know that something failed and not a generic key error. Thanks for explaining.

Issue #3011, opened 2024-09-02
Maintenance: improve e2e assertions on traces & idempotency


dreamorosi:
### Summary

Following some changes on AWS Lambda side, we need to improve our integration tests around Idempotency and Tracer so that they correctly assert the shape of traces and logs under certain conditions.

### Why is this needed?

So that we can continue running integration tests on our cade and keep the confidence in our codebase at a reasonable level.

### Which area does this relate to?

Tests

### Acknowledgment

- [X] This request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

PR #3012, opened 2024-09-02
test(maintenance): improve assertions on traces & logs

dreamorosi:
### Changes

This PR improves and simplifies the helpers used to retrieve AWS X-Ray trace data that are used in the integration tests. As a result of this, the integration tests for the utility are also slightly simplified due to the helpers handling more of the data retrieval & validation from the X-Ray API.

The PR also extends some assertions in the Idempotency tests to more flexibly assert timeouts in Amazon CloudWatch Logs.

While working on this, I have also updated the Dependabot config so that it targets the `@aws-cdk/cli-lib-alpha` (dependency of `packages/testing`) as it had fallen behind on updates and was no longer compatible with the more recent AWS CDK installed version.

**Issue number:** #3011

---


dreamorosi:
Passing integration tests on this branch: https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/10671082450

packages/tracer/tests/helpers/invokeAllTests.ts
```diff
@@ -0,0 +1,37 @@
+import { invokeFunction } from '@aws-lambda-powertools/testing-utils';
+
+/**
+ * Invoke function sequentially 3 times with different parameters
+ *
+ * invocation: is just a tracking number (it has to start from 1)
+ * sdkV2: define if we will use `captureAWSClient()` or `captureAWS()` for SDK V2
```
am29d, line 7:
```suggestion
```

am29d, line 7:
```suggestion

```

am29d: Remove unused param

packages/testing/src/types.ts
```diff
@@ -87,6 +87,94 @@ interface TestStackProps {
...
+  startTime: Date;
+  resourceName: string;
+  expectedTracesCount: number;
+};
+
+type XRayTraceDocumentParsed = {
+  name: string;
+  id: string;
+  start_time: number;
+  end_time?: number;
```
am29d, line 102: Why is in snake_case?

packages/idempotency/tests/e2e/makeHandlerIdempotent.test.ts
```diff
@@ -263,9 +263,16 @@ describe('Idempotency E2E tests, middy middleware usage', () => {
...
-      // During the first invocation the function should timeout so the logs should contain 2 logs
-      expect(functionLogs[0]).toHaveLength(2);
-      expect(functionLogs[0][0]).toContain('Task timed out after');
+      try {
+        // During the first invocation the handler should be called, so the logs should contain 1 log
+        expect(functionLogs[0]).toHaveLength(2);
+        expect(functionLogs[0][0]).toContain('Task timed out after');
+      } catch {
+        // During the first invocation the function should timeout so the logs should not contain any log and the report log should contain a timeout message
+        expect(functionLogs[0]).toHaveLength(0);
```
am29d, line 272: I don't fully understand the catch here, can you elaborate?

am29d (approved):
Great work. The code duplication is a bit high for all the tracer tests (middy, decorator, manual), but I think we can address it in a different PR. We have many such cases where we need to check the same outcomes for various setups.

dreamorosi:
> The code duplication is a bit high for all the tracer tests (middy, decorator, manual), but I think we can address it in a different PR

I agree. We have upcoming PRs where we'll work on streamlining similar to what we did in other areas in the past few weeks. For now seeing as the PR was already large I decided to not act on this.

PR #3020, opened 2024-09-03
feat(parameters): adds setParameter function to store SSM parameters

### LINKED ISSUE #2871: Feature request: Put parameter
### Use case

I'd like to be able to create/update a secret parameter's value.

### Solution/User Experience

const ssmClient = new SSMClient({ region: "us-west-2" })

  const putTokenCommand = new PutParameterCommand({
    Name: secretName,
    Type: "SecureString",
    DataType: "text",
    Value: JSON.stringify(secret),
    Overwrite: true,
  });

### Alternative solutions

```Markdown
AWS SSM SDK
```

### Acknowledgment

### Future readers

daschaa:
## Summary

Implements the `setParameter` functionality. Closes #2871 

### Changes

With this change the `SSMProvider` is extended to have a `set` method that allows to pass in a `name` and an `options` object.
In the `options` the `value` property is mandatory to set a value. Along with the value there are multiple properties that were suggested [here](https://github.com/aws-powertools/powertools-lambda-typescript/issues/2871#issuecomment-2268736718). 
The docstring for this method is mainly copy, pasted and adjusted from the `get` method in the same class.

Additional to the implementation in the `SSMProvider`, a `setParameter` function has been implemented and export via the barrel file. It supports the same structure like the `set` method of the `SSMProvider` class.
The docstring is also mainly copy, pasted and adjusted from the `getParameter` function.

There is also a section added to the documentation in the `docs` folder for storing parameters with `setParameter`. I decided to include an basic example and one example for overwriting a parameter value (because this was something I experienced myself in the past, that I need the overwrite flag and I thought this might be a handy example).

Please let me know if i should adjust the documentation and/or add stuff. I tried to keep everything basic and extend later on after the review.

**Open topics**
- [x] Implement end-to-end tests
- [x] Check the `doc` folder if something needs to be adjusted 

**Issue number:** #2871 

---


daschaa:
[FYI] Fixed the sonar cloud issues

dreamorosi:
Hi @daschaa, thank you for the PR!

I'll start reviewing it by tomorrow end of the day at the latest.

packages/parameters/tests/unit/setParameter.test.ts
```diff
@@ -0,0 +1,110 @@
...
+describe('Function: setParameter', () => {
+  const parameterName = '/my-parameter';
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  test('when called and a default provider does not exist, it instantiates one and sets the parameter', async () => {
+    const options: SSMSetOptions = { value: 'my-value' };
+    const client = mockClient(SSMClient)
```
dreamorosi, line 23: In other tests (i.e. `getParameter.test.ts`) we do it like here, but we can improve these tests by creating the client once around L16 and then reusing it across test cases. To do this, we'll need to add an `afterEach` section like described [here](https://docs.powertools.aws.dev/lambda/typescript/latest/utilities/parameters/#handlertestts_2) so we can reset the client between tests.

This should improve performance and reduce some boilerplate.

packages/parameters/tests/unit/setParameter.test.ts
```diff
@@ -0,0 +1,110 @@
...
+
+describe('Function: setParameter', () => {
+  const parameterName = '/my-parameter';
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  test('when called and a default provider does not exist, it instantiates one and sets the parameter', async () => {
+    const options: SSMSetOptions = { value: 'my-value' };
```
dreamorosi, line 22: Ideally we want to follow the convention used throughout tests (see `getParameter.test.ts` for example) and add comments like `// Prepare`, `// Act`, and `// Assess` to ease reading.

packages/parameters/tests/unit/SSMProvider.test.ts
```diff
@@ -1310,4 +1312,92 @@ describe('Class: SSMProvider', () => {
...
+      const version = await provider.set(parameterName, options);
+
+      expect(version).toBe(1);
+      expect(client).toReceiveCommandWith(PutParameterCommand, {
+        Name: parameterName,
+        Value: options.value,
+      });
+    });
+
+    test('returns undefined if version is undefined', async () => {
```
dreamorosi, line 1334: Can the version ever be undefined? I thought that the successful creation of the parameter would always return a version number/id

packages/parameters/src/ssm/SSMProvider.ts
```diff
@@ -790,6 +832,25 @@ class SSMProvider extends BaseProvider {
     return undefined;
   }
 
+  protected async _set(
```
dreamorosi, line 835: I think we can move the implementation in the main `set()` method. At least for now we don't have other interactions with other classes so having it here doesn't necessarily add much.

If we iterate on the inheritance/hierarchy of classes in the future, we can still move it out then.

packages/parameters/src/ssm/SSMProvider.ts
```diff
@@ -790,6 +832,25 @@ class SSMProvider extends BaseProvider {
...
+      Tier: options.tier ?? 'Standard',
+      Type: options.parameterType ?? 'String',
+      Overwrite: options.overwrite ?? false,
+      ...(options.kmsKeyId ? { KeyId: options.kmsKeyId } : {}),
+      ...(options.description ? { Description: options.description } : {}),
+      ...(options?.sdkOptions ?? {}),
+      Name: name,
+      Value: options.value,
+    };
+    const result = await this.client.send(new PutParameterCommand(sdkOptions));
```
dreamorosi, line 849: I would like to consider wrapping this operation in a `try/catch` and throw a custom error here.

For getter methods this is done at the `BaseProvider` level and in case of error we throw a [`GetParameterError`](https://github.com/daschaa/powertools-lambda-typescript-fork/blob/ffddb693e7268b5731bbc195734a232edde976fc/packages/parameters/src/errors.ts#L1-L9).

In this case, customers will call the method directly so if it throws they'll get the error from the AWS SDK directly.

I would suggest we wrap this and throw a new `SetParameterError` (to be created) with cause the error thrown by the SDK, kind of like:
```ts
try {
  const result = await this.client.send(new PutParameterCommand(sdkOptions));
} catch (error) {
  throw new SetParameterError(`Unable to set parameter with name ${name}`, { cause: error });
}
```

dreamorosi (changes requested):
Hi @daschaa, thank you for the patience while I reviewed the PR.

High quality as usual, just left some minor comments - looking forward to merge this!

packages/parameters/tests/unit/SSMProvider.test.ts
daschaa: The `Version` in `PutParameterResult` can be undefined, so I thought i better test it.
Source: https://github.com/aws/aws-sdk-js-v3/blob/39532e4f960bf964f059f0f1d633829675e7f92b/clients/client-ssm/src/models/models_1.ts#L9665

dreamorosi: I see - the type in the SDK does put it as potentially undefined but the [API seems to suggest it's always present](https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_PutParameter.html#API_PutParameter_ResponseSyntax) whenever the request is successful.

I'm ambivalent on this, I think the API is the source of truth but the check is not expensive so I leave it up to you.

daschaa: I also think it would be a better DX if I can assume that the setParameter always returns _something_
How would you deal with this potential undefined value code-wise? Biome prevents non-null assertions and I'm not sure if setting a default value is correct here ( I doubt it )

daschaa: For now I added a biome-ignore

dreamorosi: I think that's sensible, as long as we leave a one-sentence explanation of why for future reference.

daschaa:
@dreamorosi Thanks for the review, I think I have covered everything :)

packages/parameters/src/ssm/setParameter.ts
```diff
@@ -0,0 +1,107 @@
+import { DEFAULT_PROVIDERS } from '../base/DefaultProviders.js';
+import type { SSMSetOptions } from '../types/SSMProvider';
```
dreamorosi, line 2:
```suggestion
import type { SSMSetOptions } from '../types/SSMProvider.js';
```

packages/parameters/tests/unit/setParameter.test.ts
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * @group unit/parameters/ssm/setParameter/function
+ */
+import { PutParameterCommand, SSMClient } from '@aws-sdk/client-ssm';
+import { mockClient } from 'aws-sdk-client-mock';
+import { DEFAULT_PROVIDERS } from '../../src/base';
+import { setParameter } from '../../src/ssm';
+import { SSMProvider } from '../../src/ssm/SSMProvider';
+import 'aws-sdk-client-mock-jest';
+import type { SSMSetOptions } from '../../src/types/SSMProvider';
```
dreamorosi, line 12:
```suggestion
import type { SSMSetOptions } from '../../src/types/SSMProvider.js';
```

packages/parameters/tests/unit/setParameter.test.ts
```diff
@@ -0,0 +1,119 @@
+/**
+ * Test setParameter function
+ *
+ * @group unit/parameters/ssm/setParameter/function
+ */
+import { PutParameterCommand, SSMClient } from '@aws-sdk/client-ssm';
+import { mockClient } from 'aws-sdk-client-mock';
+import { DEFAULT_PROVIDERS } from '../../src/base';
+import { setParameter } from '../../src/ssm';
+import { SSMProvider } from '../../src/ssm/SSMProvider';
```
dreamorosi, line 10:
```suggestion
import { SSMProvider } from '../../src/ssm/SSMProvider.js';
```

packages/parameters/tests/unit/setParameter.test.ts
```diff
@@ -0,0 +1,119 @@
+/**
+ * Test setParameter function
+ *
+ * @group unit/parameters/ssm/setParameter/function
+ */
+import { PutParameterCommand, SSMClient } from '@aws-sdk/client-ssm';
+import { mockClient } from 'aws-sdk-client-mock';
+import { DEFAULT_PROVIDERS } from '../../src/base';
+import { setParameter } from '../../src/ssm';
```
dreamorosi, line 9:
```suggestion
import { setParameter } from '../../src/ssm/index.js';
```

packages/parameters/tests/unit/setParameter.test.ts
```diff
@@ -0,0 +1,119 @@
+/**
+ * Test setParameter function
+ *
+ * @group unit/parameters/ssm/setParameter/function
+ */
+import { PutParameterCommand, SSMClient } from '@aws-sdk/client-ssm';
+import { mockClient } from 'aws-sdk-client-mock';
+import { DEFAULT_PROVIDERS } from '../../src/base';
```
dreamorosi, line 8:
```suggestion
import { DEFAULT_PROVIDERS } from '../../src/base/index.js';
```

dreamorosi (approved):
Thank you for working on this PR, @daschaa!

I appreciate the time you've put into this, both in the initial implementation and subsequent iterations.

I also ran the end to end tests and [they were green](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/10797977963) 🎉

dreamorosi:
@leandrodamascena, @sthulb, @am29d - if either of you see this, please feel free to approve & merge this.

I had to do a couple of touch ups in the form of suggestions and I can't merge it myself because I'm the last one to have committed.

Thanks!

leandrodamascena (approved):

PR #3055, opened 2024-09-12
docs(idempotency): add callout about undefined early return

### LINKED ISSUE #2987: Bug: Idempotency middleware doesn't work for functions without return value
### Expected Behavior

When using the `makeHandlerIdempotent` middleware with a handler function that has no return value(`undefined`), the middleware should only call the handler once.

### Current Behavior

The handler is called multiple times for the same event, even though the idempotency record exists in Dynamo.

The issue is that the package returns the stored return value from previous run from the middleware.
Here: https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/idempotency/src/IdempotencyHandler.ts#L243
And here: https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/idempotency/src/middleware/makeHandlerIdempotent.ts#L140

If the value returned from middleware is truthy, it causes middy to perform an early return(https://middy.js.org/docs/intro/early-interrupt/).
This works for handler that have a return value, but doesn't work when the return value is `undefined` - middy doesn't early return and the handler code is executed again.

### Code snippet

`export handler = middy().use(makeHandlerIdempotent(...)).handler(() => console.log("stuff"))`

### Steps to Reproduce

Use the code snippet, call the lambda multiple times with the same event

### Possible Solution

In order to early return, a truthy value has to be returned from the middleware. 
This kind of sucks, because the first time a lambda runs, it returns `undefined` and we'd have to return something else when idempotency key is present.
In our use case, the return value doesn't matter, but I'm not sure what would make sense as a generic solution.

latest

20.x

### Packaging format used

npm

### Execution logs

_No response_

dreamorosi:
### Changes

This PR adds a callout and link to the Middy.js repo in the docs related to the Idempotency `makeHandlerIdempotent` Middy.js middleware to specify that the AWS Lambda function handler must return a value different than `undefined` for the middleware to work correctly.

**Issue number:** closes #2987

---


docs/utilities/idempotency.md
```diff
@@ -206,6 +206,8 @@ Similar to the `makeIdempotent` function wrapper, you can quickly make your Lamb
     --8<-- "examples/snippets/idempotency/types.ts:3:16"
     ```
 
+Note that for the middleware to work, your Lambda function handler must return a value different from `undefined`. This is a [known limitation of Middy.js early return feature](https://github.com/middyjs/middy/issues/1236). If your use case requires early returns, you can use the `makeIdempotent` function wrapper instead.
```
leandrodamascena, line 209:
```suggestion
For the middleware to work, your Lambda function handler must return a value different from `undefined`. This is a [known limitation of the early return feature in Middy.js](https://github.com/middyjs/middy/issues/1236). If your use case requires early returns, you can use the `makeIdempotent` function wrapper instead.
```

packages/idempotency/src/middleware/makeHandlerIdempotent.ts
```diff
@@ -90,6 +90,11 @@ const shouldSkipIdempotency = (request: MiddyLikeRequest): boolean => {
  * ).use(makeHandlerIdempotent({ persistenceStore: dynamoDBPersistenceLayer }));
  * ```
  *
+ * Note that for the middleware to work, your Lambda function handler must return a value different from `undefined`.
+ * This is a [known limitation of Middy.js early return feature](https://github.com/middyjs/middy/issues/1236).
+ *
+ * If your use case requires early returns, you can use the {@link index.makeIdempotent | makeIdempotent()} function wrapper instead.
```
leandrodamascena, line 96: If you accept the first suggestion, you need to change it here. If not, it's good.

leandrodamascena (approved):
Hey @dreamorosi! I'm approving this PR, feel free to accept or not my suggestions.

leandrodamascena (approved):

PR #3056, opened 2024-09-12
docs(tracer): clarify escape hatch mechanism

### LINKED ISSUE #2943: Bug: Tracer escape hatch does not expose full AWSXRay api
### Expected Behavior

Based on the documentation 
```
Escape hatch mechanism

You can use tracer.provider attribute to access all methods provided by the AWS X-Ray SDK.
```

But this is not true. It only re-exposes a subset of methods.

see https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/tracer/src/provider/ProviderService.ts 

In my case I wanted to overwrite the `setStreamingThreshold()` method which is not implemented. 

### Current Behavior

An error is thrown because the method is not implemented

### Code snippet

```js
const tracer = new Tracer({ serviceName: "serverlessAirline" });
tracer.provider.setStreamingThreshold(1);
```

produces an error.

```json5
{ 
  "errorType":"TypeError",
  "errorMessage":"tracer.provider.setStreamingThreshold is not a function"
 //  ....
}

### Steps to Reproduce

see code snippet

### Possible Solution

1. update docs 
2. implement/proxy all methods

latest

20.x

### Packaging format used

Lambda Layers

### Execution logs

```Shell
{
"errorType":"TypeError","errorMessage":"tracer.provider.setStreamingThreshold is not a function","trace":["TypeError: tracer.provider.setStreamingThreshold is not a function","    at Object.<anonymous> (/var/task/index.js:6:17)","    at Module._compile (node:internal/modules/cjs/loader:1358:14)","    at Module._extensions..js (node:internal/modules/cjs/loader:1416:10)","    at Module.load (node:internal/modules/cjs/loader:1208:32)","    at Module._load (node:internal/modules/cjs/loader:1024:12)","    at Module.require (node:internal/modules/cjs/loader:1233:19)","    at require (node:internal/modules/helpers:179:18)","    at _tryRequireFile (file:///var/runtime/index.mjs:1002:37)","    at _tryRequire (file:///var/runtime/index.mjs:1052:25)","    at _loadUserApp (file:///var/runtime/index.mjs:1081:22)"]}
```

dreamorosi:
### Changes

This PR fixes the documentation around the escape hatch mechanism exposed by the Tracer utility.

In the previous version of the docs we were stating that we expose all the methods from the underlying AWS X-Ray SDK for Node.js - which is not true.

I have updated the docs to clarify that we only expose **some** of the methods and shared a brief explanation of why this is the case, and what to expect from using other methods directly.

To give more visibility over **which** methods are exposed, I have also added the `ProviderService` class from the Tracer package to the API docs. 

**Issue number:** closes #2943

---


docs/core/tracer.md
```diff
@@ -378,7 +378,7 @@ Tracer exposes a `getRootXrayTraceId()` method that allows you to retrieve the [
 
 ### Escape hatch mechanism
 
-You can use `tracer.provider` attribute to access all methods provided by the [AWS X-Ray SDK](https://docs.aws.amazon.com/xray-sdk-for-nodejs/latest/reference/AWSXRay.html).
+You can use `tracer.provider` attribute to access some of the methods provided by the [AWS X-Ray SDK](https://docs.aws.amazon.com/xray-sdk-for-nodejs/latest/reference/AWSXRay.html).
```
leandrodamascena, line 381: You say `some of the methods` but in the line 391 you send a link. I thought about something like:

```suggestion
You can use `tracer.provider` attribute to access [a subset of the methods](https://docs.powertools.aws.dev/lambda/typescript/latest/api/classes/_aws_lambda_powertools_tracer.provider_ProviderService.ProviderService.html) provided by the [AWS X-Ray SDK](https://docs.aws.amazon.com/xray-sdk-for-nodejs/latest/reference/AWSXRay.html).
```

docs/core/tracer.md
```diff
@@ -388,6 +388,10 @@ This is useful when you need a feature available in X-Ray that is not available
     --8<-- "examples/snippets/tracer/escapeHatch.ts"
     ```
 
+We only expose [a subset of the methods available](https://docs.powertools.aws.dev/lambda/typescript/latest/api/classes/_aws_lambda_powertools_tracer.provider_ProviderService.ProviderService.html) in the AWS X-Ray SDK for Node.js based on the most common use cases relevant for AWS Lambda and their compatibility with the Tracer utility.
```
leandrodamascena, line 391: I would remove this paragraph.

1 - You are already explaining it in the first paragraph.
2 - "based on the most common use cases" I think it is vague and doesn't add much.

docs/core/tracer.md
```diff
@@ -388,6 +388,10 @@ This is useful when you need a feature available in X-Ray that is not available
...
+
+If you need to access a method that is not available you can import it directly from the AWS X-Ray SDK for Node.js. Compatibility with the Tracer utility is not guaranteed.
```
leandrodamascena, line 393:
```suggestion
If you need to access a method that is not available, you can import it directly from the AWS X-Ray SDK for Node.js. However, please note that compatibility with the Tracer utility is not guaranteed.
```

docs/core/tracer.md
leandrodamascena: This link will work after merging this PR and refactoring the documentation, right? Because it's broken now. I imagine it's because is missing the comment and not creating the API documentation today.

leandrodamascena (changes requested):
Hi @dreamorosi! I left some suggestions to improve the documentation and make it clearer.

docs/core/tracer.md
dreamorosi: Removed!

leandrodamascena (approved):

PR #3057, opened 2024-09-12
feat(metrics): add ability to pass custom logger

### LINKED ISSUE #2036: Feature request: Option to only publish metrics when there are metrics to publish
### Use case

I'd like to use the `logMetrics` middleware without seeing a warning that I'm not publishing metrics, or writing an empty metrics object to EMF.

I have a Lambda that may emit 0 metrics. I know this is a possibility and don't consider it an issue. 
But I use the `logMetrics` middleware which 1. requires me to create any metrics I _may_ emit at the start of my function, and 2. always calls publishStoredMetrics which logs a warning and emits empty metrics.

```
No application metrics to publish. The cold-start metric may be published if enabled. If application metrics should never be empty, consider using `throwOnEmptyMetrics`
{"_aws":{"Timestamp":1707345436427,"CloudWatchMetrics":[{"Namespace":"ns","Dimensions":[["service"]],"Metrics":[]}]},"service":"service_undefined"}
```

### Solution/User Experience

I see this warning was added in a previous feature https://github.com/aws-powertools/powertools-lambda-typescript/pull/1397 so it's unlikely we would remove it. I'd be happy with a new option `doNotWarnOnEmptyMetrics`, or changing `throwOnEmptyMetrics` to accept `true | false | 'some-option-to-disable-warnings'`.

I'd also be happy if `storedMetrics` was public, or there was a public method `empty(): boolean` so I could write my own middleware that calls `metric.empty() || metric.publishStoredMetrics()`. 

### Alternative solutions

_No response_

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR adds the ability to pass a custom logger object to the Metrics utility. This is useful when you want to enable debug logs or suppress warnings emitted by the utility.

Customers have been asking for a way to silence warnings which is not possible today due to Node.js not having a standard logging library similar to other languages such as Python. To go around the issue, I decided to follow a pattern used by other Node.js libraries in the ecosystem such as AWS SDK for JavaScript v3 and AWS X-Ray SDK for Node.js, which accept a generic "logger" class.

When the logger class is present, these libraries start emitting debug logs and other telemetry. In our case we'll follow the same pattern for `debug` logs but fall back to the global `console` object for warnings since we always want to emit them.

Likewise, in the case of Metrics, the actual EMF blobs will always bypass any custom logger passed by customer and go to stdout. I deliberately choose to go this route to avoid compatibility issues with 3rd party loggers and decrease the risk of data loss.

While I hope customers will use this feature with our Logger, the interface for the logger object is as generic as possible and allows 3rd party loggers or even the Node.js `console` object (aka global console).

For now I opted for documenting this feature only in the API docs and not in the main docs as I want to soft-launch it before publicising it more. In future releases we'll work on adding some debug logs similar to what Powertools for AWS Lambda (Python) does, which was something we couldn't do up until now.

Once that work is done and this pattern is applied across more utilities we'll update the docs as well. This will also give us some time to learn how customers use the feature.

**Example**

This is the DX of how customers would specify a custom logger, when using the Powertools for AWS one this gives the added benefit of controlling the log level and silencing warnings as needed.

```ts

const logger = new Logger({ logLevel: 'error' });
const metrics = new Metrics({ logger });

// ...
```

This pattern is also useful for unit testing, as it essentially allows dependency injection as well.

**Issue number:** closes #2036

---


packages/metrics/tests/unit/repro.test.ts
dreamorosi, line 1: This file was a leftover from a previous PR which wasn't supposed to be committed.

dreamorosi:
Integration tests [are failing](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/11072126830), will look into it tomorrow.

dreamorosi:
Once #3292 is merged, I'll be ready to mark this as ready for review.

dreamorosi:
Integration tests are [finally passing](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/11706553614).

am29d (approved):
Thanks for the PR, LGTM!

I checked the e2e test after main rebase they seem to be flake for the logger. Need to investiage later.

PR #3058, opened 2024-09-12
improv(tracer): set `AWS_XRAY_CONTEXT_MISSING` to `IGNORE_ERROR` when no value is set

### LINKED ISSUE #2406: Feature request: avoid tracing API calls made during init (top-level `await`)
### Use case

Currently when using the Parameters and Tracer utility together, and with top-level `await` (like shown below) used in a certain order an error is logged (**but not thrown**) informing that the X-Ray SDK is unable to find a trace root (expected).

```ts
import { getParameter } from '@aws-lambda-powertools/parameters/ssm';

/**
 * note that tracer is instantiated before the `getParameter` call
 * from this point on, http calls (including the ones with AWS SDK) are traced
 */
const tracer = new Tracer();
const logger = new Logger();
const ssmParameter = await getParameter('/layers/powertools-layer-arn');

  logger.info('param', { ssmParameter });
  
  return {
    statusCode: 200,
    body: JSON.stringify("Hello, World!"),
  };
};
```

This is the log emitted by the AWS X-Ray SDK, which attempts to trace the request made 

```log
2024-04-19T21:34:27.204Z        undefined       ERROR   Error: Missing AWS Lambda trace data for X-Ray. Ensure Active Tracing is enabled and no subsegments are created outside the function handler.
    at Object.contextMissing (file:///var/task/index.mjs:7:16736)
    at e.resolveLambdaTraceData (file:///var/task/index.mjs:7:36770)
    at Object.getSegment (file:///var/task/index.mjs:7:17544)
    at Object.resolveSegment (file:///var/task/index.mjs:7:17251)
    at n (file:///var/task/index.mjs:7:95499)
    at s (file:///var/task/index.mjs:7:97066)
    at /var/runtime/node_modules/@aws-sdk/node_modules/@smithy/node-http-handler/dist-cjs/index.js:233:19
    at new Promise (<anonymous>)
    at _NodeHttpHandler.handle (/var/runtime/node_modules/@aws-sdk/node_modules/@smithy/node-http-handler/dist-cjs/index.js:189:12)
    at async /var/runtime/node_modules/@aws-sdk/node_modules/@smithy/middleware-serde/dist-cjs/index.js:33:24
START RequestId: 1ae7c9e3-37d1-4110-b8c0-cdb63b70b69d Version: $LATEST
{"level":"INFO","message":"param","sampling_rate":0,"service":"service_undefined","timestamp":"2024-04-19T21:34:27.305Z","xray_trace_id":"1-6622e362-563b8ad23b746432303d77c6","ssmParameter":"arn:aws:lambda:eu-west-1:123456789101:layer:AWSLambdaPowertoolsTypeScriptV2:5"}
END RequestId: 1ae7c9e3-37d1-4110-b8c0-cdb63b70b69d
REPORT RequestId: 1ae7c9e3-37d1-4110-b8c0-cdb63b70b69d  Duration: 50.72 ms      Billed Duration: 51 ms  Memory Size: 128 MB     Max Memory Used: 94 MB  Init Duration: 553.91 ms
XRAY TraceId: 1-6622e362-563b8ad23b746432303d77c6       SegmentId: 1e9f991047253160     Sampled: true
```

This is expected in the sense that trace data is available only within the `handler` scope and made available by the AWS Lambda service only when invoking your function.

This is not a runtime error (which is also why this is a feature request and not a bug) but nevertheless it's a nuisance and we should try to find a way to handle this case without making the AWS X-Ray SDK log the error.

Note that swapping the order of the Tracer instantiation and the Parameters call, like below, fixes the issue:

```ts
const ssmParameter = await getParameter('/layers/powertools-layer-arn');
// this is fine because http tracing starts after this point and the call already happened
const tracer = new Tracer();
```

### Solution/User Experience

The experience for customers should stay as-is and the change should be backwards compatible.

### Alternative solutions

_No response_

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR updates the Tracer utility to set the `AWS_XRAY_CONTEXT_MISSING` environment variable to `IGNORE_ERROR` by default.

This change has been made in response to customers adopting features that rely on top-level await and finding the error logs produced by X-Ray SDK not useful.

The X-Ray SDK for Node.js was not designed with this feature in mind, and by default it logs an error when a code path attempts to send trace data outside of the context of a request.

This is because without the request context, there's no X-Ray Trace ID, and so the SDK is unable to establish the lineage of a segment/span.

In many cases this is a desired behavior (although I'd have preferred it was a warning rather than an error log), however now that making async operations during the init phase is possible, it's not uncommon to make outbound http requests or use the AWS SDK to retrieve data or perform setup operations.

In those cases, the auto instrumentation features of Tracer (& the underlying X-Ray SDK) attempt to create a segment for these operations but don't find one because these tasks are run during the init phase, and thus outside of the context of a request.

The change has been made with backwards compatibility in mind, to an extent. Before setting the value we check if the env variable is already set and if so, we leave it as-is. 

I believe this is a good middle ground between furthering our features and maintaining flexibility.

**Issue number:** closes #2406

---

### DIFF
--- a/packages/tracer/src/Tracer.ts
+++ b/packages/tracer/src/Tracer.ts
@@ -1,3 +1,15 @@
+/**
+ * If the AWS_XRAY_CONTEXT_MISSING environment variable is not set, we set it to IGNORE_ERROR.
+ *
+ * This is to prevent the AWS X-Ray SDK from logging errors when using top-level await features that make HTTP requests.
+ * For example, when using the Parameters utility to fetch parameters during the initialization of the Lambda handler - See #2046
+ */
+if (
+  process.env.AWS_XRAY_CONTEXT_MISSING === '' ||
+  process.env.AWS_XRAY_CONTEXT_MISSING === undefined
+) {
+  process.env.AWS_XRAY_CONTEXT_MISSING = 'IGNORE_ERROR';
+}
 import { Utility } from '@aws-lambda-powertools/commons';
 import type {
   AsyncHandler,
--- a/packages/tracer/src/index.ts
+++ b/packages/tracer/src/index.ts
@@ -1,1 +1,13 @@
+/**
+ * If the AWS_XRAY_CONTEXT_MISSING environment variable is not set, we set it to IGNORE_ERROR.
+ *
+ * This is to prevent the AWS X-Ray SDK from logging errors when using top-level await features that make HTTP requests.
+ * For example, when using the Parameters utility to fetch parameters during the initialization of the Lambda handler - See #2046
+ */
+if (
+  process.env.AWS_XRAY_CONTEXT_MISSING === '' ||
+  process.env.AWS_XRAY_CONTEXT_MISSING === undefined
+) {
+  process.env.AWS_XRAY_CONTEXT_MISSING = 'IGNORE_ERROR';
+}
 export { Tracer } from './Tracer.js';
--- a/packages/tracer/src/index.ts
+++ b/packages/tracer/src/index.ts
@@ -1,13 +1,1 @@
-/**
- * If the AWS_XRAY_CONTEXT_MISSING environment variable is not set, we set it to IGNORE_ERROR.
- *
- * This is to prevent the AWS X-Ray SDK from logging errors when using top-level await features that make HTTP requests.
- * For example, when using the Parameters utility to fetch parameters during the initialization of the Lambda handler - See #2046
- */
-if (
-  process.env.AWS_XRAY_CONTEXT_MISSING === '' ||
-  process.env.AWS_XRAY_CONTEXT_MISSING === undefined
-) {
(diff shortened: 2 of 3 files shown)


leandrodamascena (changes requested):
For more coverage and to make sure that if there are any changes in the future, they will be detected, you should add a test ([EnvVarTest](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/tracer/tests/unit/EnvironmentVariablesService.test.ts)) or is this too much? WDYT?

Requesting changes to hear from you.

dreamorosi:
Yea good point - I added it elsewhere because that one you linked is for reading from the env var, but I think it's the same.

leandrodamascena (approved):

Issue #3100, opened 2024-09-22
Feature request: include parser feature in lambda layer


scratchclaggy:
### Use case

I want to use the new parser feature in my lambdas but need to ship it with my code as it is not available in the lambda layer currently.

### Solution/User Experience

We simply need to include the feature when building / publishing the lambda layers.

### Alternative solutions

```Markdown
Keep it out of the layer but update documentation:

It may be that this feature is not used frequently, in which case it makes sense to leave it out of the layers to minimize the layer size. In this case though, it should be noted in the docs that the parser feature is not included in the layer like the other features are.

Ship multiple layers:

If layer size is of concern, there may be a case for shipping a layer that includes the SDKs and the logger / metrics / tracer features, and then shipping individual layers for each of the other features that are used less frequently.
```

### Acknowledgment

- [X] This feature request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

scratchclaggy:
Note that this was also discussed [in the Discord community](https://discord.com/channels/1006478942305263677/1286603199075061801).

dreamorosi:
Hi @scratchclaggy, thank you for opening the issue.

I'll assign the issue to you and will continue the discussion in the PR.

Thank you again!

PR #3101, opened 2024-09-22
chore(layers): add Parser to layer

scratchclaggy:
## Summary

Small PR to add the new parser feature to the published lambda layers.

### Changes

Adds the feature to the layer stacks in the layer package.

**Issue number:** #3100

---

### DIFF
--- a/layers/src/canary-stack.ts
+++ b/layers/src/canary-stack.ts
@@ -54,6 +54,7 @@ export class CanaryStack extends Stack {
           '@aws-lambda-powertools/parameters',
           '@aws-lambda-powertools/idempotency',
           '@aws-lambda-powertools/batch',
+          '@aws-lambda-powertools/parser',
         ],
       },
       environment: {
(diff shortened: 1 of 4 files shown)


scratchclaggy:
This is my first PR against this project, so please let me know if I am missing anything here.

Also, could someone please confirm that the changes from this PR will trigger a new layer to be built, and following this the documentation will be updated automatically.

dreamorosi (changes requested):
Hi @scratchclaggy, thank you for the PR.

The utility doesn't come with Zod by default, so we need to install it.

You should add it to the block around line 110.

am29d:
Hey @scratchclaggy, can you also add the package to our e2e tests. We usually check if the package is present and can be imported. 

While we invoke some of the utilities in our [canary lambda directly](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/layers/tests/e2e/layerPublisher.class.test.functionCode.ts), i.e. logger and metrics. I think we can also run a parse small parse command with a fairly trivial payload, just to be sure that it's safe.

am29d (approved):
Looks good, thanks for adding the tests!

dreamorosi (approved):

Issue #1829, opened 2023-12-21
Feature request: sequential async processing


revmischa:
### Use case

Sometimes I have records that I want processed one at a time, but my processor function happens to be async. 

### Solution/User Experience

It would be nice to request sequential processing of records with async handlers.

### Acknowledgment

- [X] This feature request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

dreamorosi:
Hi @revmischa thank you for taking the time to open this issue.

As we discussed in Discord, I think the request is valid and I remember hearing it from other customers in the past few weeks.

I'm adding this to the backlog so that it can be picked up.

If anyone is interested in contributing, please leave a comment so we dan discuss an implementation.

revmischa:
I also have a use case where I have an async handler but need to process FIFO events sequentially.

I think it's a safe assumption that most every handler anyone will ever write will be async. How else will do you I/O otherwise? And what is the use of a handler that can't do I/O?

dreamorosi:
That's very fair, we are focused on releasing v2 this & next week. 

After that we'll be able to reprise working on new features for the existing utilities. This is one of the issues I'd like to pick up relatively soon.

arnabrahman:
I can work on this next. I can see there is a section in the doc about [Async processing](https://docs.powertools.aws.dev/lambda/typescript/latest/utilities/batch/#async-or-sync-processing), 

```
*If your function is async returning a Promise, use BatchProcessor and processPartialResponse * 
 If your function is not async, use BatchProcessorSync and processPartialResponseSync
```

So, based on the PR description, do we now want to have the option to use `BatchProcessorSync` and `processPartialResponseSync` in an `async` function? t would be helpful if I could have some more context. @dreamorosi, whenever you are free.

revmischa:
Also just curious, what is the use case for sync processing? You can't really do I/O without async right? So what use is a SQS processing function that can't do any I/O?

dreamorosi:
[...]

PR #3109, opened 2024-09-25
feat(batch): sequential async processing of records for `BatchProcessor`

arnabrahman:
## Summary
By default, the `BatchProcessor` processes records in parallel using `Promise.all()`. We want to give users an option to process records sequentially. 
### Changes

- Introduce a new batch processing option, `processInParallel`, which indicates whether records should be processed in parallel or sequentially. 
- I believe this option should only be available for `BatchProcessor` class but I couldn't make the optional property work as the class signature is similar for both `BatchProcessor` & `BatchProcessorSync`. I have omitted this property for `SqsFifoPartialProcessor` though. 
- The records will be processed in parallel or sequentially based on the `processInParallel` option. If `processInParallel` is not provided, the records will default to being processed in parallel. Made the changes in the`BasePartialProcessor` class.
- For the tests, I restructured the async processing test cases. I ran all the tests for both parallel and sequential processing by introducing `describe.each()`. My rationale is that every feature for `BatchProcessor` should behave the same, regardless of whether records are processed in parallel or sequentially. This approach also avoids duplicating test cases for both options.
- Updated the documentation for `processInParallel` and made additional changes where necessary, using my judgment.

**Issue number:** #1829 

---


dreamorosi:
Hi @arnabrahman, thank you for the PR!

I will start reviewing it tomorrow morning - I was on PTO for the first half of the week.

Regarding the SonarCloud findings, I think we can simplify the structure of the tests and remove the findings by just changing it to something like:

```ts
describe('Asynchronously processing', () => {
    const cases = [
      {
        description: 'in parallel',
        options: { processInParallel: true },
      },
      {
        description: 'sequentially',
        options: { processInParallel: false },
      },
    ];

    it.each(cases)('SQS Records $description', ({ options }) => {

    })
})
```

or similar

arnabrahman:
@dreamorosi No worries, i will take a look at this tomorrow.

dreamorosi (approved):
Amazing work on this PR @arnabrahman, appreciate your help!

Thank you also for updating the docs with the new feature 🎉

PR #3195, opened 2024-10-11
feat(logger): include enumerable properties in formatted errors

### LINKED ISSUE #2170: Feature request: include error own properties in logged errors
### Expected Behaviour

I expect powertools to log all enumerable properties on an error. It's common to include a code, and extra information about an error. For example, node's fs errors include `code`, `syscall` and `path`. I'd like powertools to log

```json
{
        "name": "Error",
        "location": "",
        "message": "ENOENT: no such file or directory, open 'not a file'",
        "stack": "Error: ENOENT: no such file or directory, open 'not a file'"
        "errno": -2,
        "code": "ENOENT",
        "syscall": "open",
        "path": "not a file"
}
```

If you log a regular object or classs it's properties will be logged, but if the class extends Error only stack, message, location, name, and cause will be logged.

### Current Behaviour

logger does not log any of the errors enumerable properties

### Code snippet

```typescript
import fs from 'fs/promises'

import { Logger } from '@aws-lambda-powertools/logger'

class MyData {
  constructor(public readonly field: string) { }
}
class MyError extends Error {
  constructor(public readonly field: string) {
    super(field)
  }
}


const logger = new Logger();

// node's built in errors include code, syscall and other properties
// powertools doesn't include these when printing the error object.
await fs.readFile("not a file").catch(err => {
  logger.error("fs error", {
    bare: err,
    spread: { ...err },
    json: JSON.parse(JSON.stringify(err)),
  })
})

const myError = new MyError("field");
logger.error("fs error", {
  bare: myError,
  spread: { ...myError },
  json: JSON.parse(JSON.stringify(myError)),
})

const myData = new MyData("field");
logger.error("fs error", {
  bare: myData,
  spread: { ...myData },
  json: JSON.parse(JSON.stringify(myData)),
})
```

Logging an error thrown by node's built in fs module, see that `bare`, which is how most people would log an error, is missing errno, code, syscall, and path metadata.
spread, and json include the metadata, but not the stack trace, message, or location. spread and json show that these are enumerable properties on the object.

```json
{
    "level": "ERROR",
    "message": "fs error",
    "timestamp": "2024-03-07T00:04:23.697Z",
    "bare": {
        "name": "Error",
        "location": "",
        "message": "ENOENT: no such file or directory, open 'not a file'",
        "stack": "Error: ENOENT: no such file or directory, open 'not a file'"
    },
    "spread": {
        "errno": -2,
        "code": "ENOENT",
        "syscall": "open",
        "path": "not a file"
    },
    "json": {
        "errno": -2,
        "code": "ENOENT",
        "syscall": "open",
        "path": "not a file"
    }
}
```

Here's a simple custom error and it has the same problem.
```json
{
    "level": "ERROR",
    "message": "fs error",
    "timestamp": "2024-03-07T00:04:23.699Z",
    "bare": {
        "name": "Error",
        "location": "test.ts:27",
        "message": "field",
        "stack": "Error: field\n    at <anonymous> (test.ts:27:17)"
    },
    "spread": {
        "field": "field"
    },
    "json": {
        "field": "field"
    }
}
```

This is logging a normal class that doesn't extend Error and you can see that `bare` includes the properties I would expect.
```json
{
    "level": "ERROR",
    "message": "fs error",
    "timestamp": "2024-03-07T00:04:23.701Z",
    "bare": {
        "field": "field"
    },
    "spread": {
        "field": "field"
    },
    "json": {
        "field": "field"
    }
}
```

### Steps to Reproduce

The script included above reproduces with powertools 1.18.1 or 2.0.2 on node 20.

### Possible Solution

Include enumerable properties by spreading the err object.

2.0.2, 1.18.1

20.x

### Packaging format used

npm

### Execution logs

```Shell
See above.
```

dreamorosi:
### Changes

This PR improves the logic used to format errors for Logger to start including all [enumerable properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties) of the error including both own and inherited.

This change was done in response to customer demand and considering it's additive, we don't consider it as a breaking change.

Consider the following error:

```ts
const customSymbol = Symbol('customSymbol');
class CustomError extends Error {
  public otherProperty: string;

  public constructor(
    message: string,
    public readonly code: number
  ) {
    super(message);
    this.name = 'CustomError';
    this.otherProperty = 'otherProperty';
  }

  public [customSymbol] = (): void => {
    // do nothing
  };
}

class SuperCustomError extends CustomError {
  public extraProperty: string;
  public constructor(message: string, code: number) {
    super(message, code);
    this.name = 'SuperCustomError';
    this.extraProperty = 'extraProperty';
  }
}
```

Before this change, calling `logger.error('An error occurred', new SuperCustomError('foo', 303));` generated this log:

```json
{
  "message": "foo",
  "location": "/path/to/my/source-code/my-service/handler.ts:18",
  "name": "SuperCustomError",
  "stack": "Error: Unexpected error #1    at lambdaHandler (/path/to/my/source-code/my-service/handler.ts:18:11) ..." 
}
```

This missed properties that customers might have wanted to see, making the troubleshooting harder than needed.

After the change, the log will instead look like this:

```json
{
  "message": "foo",
  "location": "/path/to/my/source-code/my-service/handler.ts:18",
  "name": "SuperCustomError",
  "stack": "Error: Unexpected error #1    at lambdaHandler (/path/to/my/source-code/my-service/handler.ts:18:11) ...",
  "code": 303,
  "otherProperty": "otherProperty",
  "extraProperty": "extraProperty"
}
```

**Issue number:** closes #2170

---


packages/logger/src/formatter/LogFormatter.ts
```diff
@@ -104,16 +104,30 @@ abstract class LogFormatter {
...
+    for (const key in error) {
+      if (
+        typeof key === 'string' &&
+        key !== 'name' &&
+        key !== 'message' &&
+        key !== 'stack' &&
+        key !== 'cause'
+      ) {
+        formattedError[key] = (errorAttributes as Record<string, unknown>)[key];
+      }
```
am29d, line 127: Can we simplify it a bit? 
```suggestion
      if (typeof key === 'string' && !['name', 'message', 'stack', 'cause'].includes(key)) {
          formattedError[key] = (errorAttributes as Record<string, unknown>)[key];
      }
```

am29d (changes requested):
Looks good, only a small nitpick to reduce repetitive statements

am29d (approved):

PR #3250, opened 2024-10-23
fix(parser): fix cause errors nested structure

### LINKED ISSUE #3208: Bug: ParsedResultError contains nested ParseErrors instead of a ParseError with a cause of ZodError
### Expected Behavior

In the following example:
```typescript
const lambdaHandler = async (
  event: ParsedResult<JobsRequest>,
  context: Context
): Promise<APIGatewayProxyResult> => {

  if (!event.success) {
    return response(400, JSON.stringify((event.error.cause as ZodError).errors));
  }

  return response(200, `Parsed Event: ${JSON.stringify(event)}`);
};

const handler = middy(lambdaHandler).use(
  parser({ schema: jobsRequestSchema, envelope: ApiGatewayEnvelope, safeParse: true })
);
```

`event.error.cause` should contain a ZodError, and calling `event.error.cause.errors` should return something like:

```json
[
    {
        "code": "too_small",
        "minimum": 0,
        "type": "number",
        "inclusive": false,
        "exact": false,
        "message": "ID must be positive",
        "path": [
            "id"
        ]
    }
]
```

### Current Behavior

`event.error.cause` contains a ParseError, which contains another cause which contains the ZodError.

So to get to the ZodError, you have to call `event.error.cause.cause` 

### Code snippet

```typescript
import { APIGatewayProxyResult } from "aws-lambda";
import { ApiGatewayEnvelope } from "@aws-lambda-powertools/parser/envelopes";
import { z } from "zod";
import { parser } from "@aws-lambda-powertools/parser/middleware";
import { ParsedResult } from "@aws-lambda-powertools/parser/types";

const response = (status: number, body: string): APIGatewayProxyResult => ({
  statusCode: status,
  body,
  headers: {
  },
});

const jobsRequestSchema = z.object({
  id: z.number().positive("ID must be positive"),
  from: z.coerce.date(),
  to: z.coerce.date(),
});

type JobsRequest = z.infer<typeof jobsRequestSchema>;

const lambdaHandler = async (event: ParsedResult<JobsRequest>): Promise<APIGatewayProxyResult> => {
  if (!event.success) {
    const errorResponse = {
      error_cause: event.error.cause,
      error_cause_cause: event.error.cause?.cause,
    };
    return response(400, JSON.stringify(errorResponse));
  }

  return response(200, `Hello World, Parsed Event: ${JSON.stringify(event)}`);
};

const handler = middy(lambdaHandler).use(
  parser({ schema: jobsRequestSchema, envelope: ApiGatewayEnvelope, safeParse: true })
);

module.exports = { handler };
```

POSTing the following body:
```json
{
    "id": -1,
    "from": "2024-01-01",
    "to": "2024-01-02"
}
```

returns:
```json
{
    "error_cause": {
        "name": "ParseError"
    },
    "error_cause_cause": {
        "issues": [
            {
                "code": "too_small",
                "minimum": 0,
                "type": "number",
                "inclusive": false,
                "exact": false,
                "message": "ID must be positive",
                "path": [
                    "id"
                ]
            }
        ],
        "name": "ZodError"
    }
}
```

### Steps to Reproduce

Create a lambda with the above code in your handler.ts

### Possible Solution

I haven't tested if this affects other Envelopes as well or not, but the `apigw.ts` `safeParse` function [here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/parser/src/envelopes/apigw.ts)

calls `Envelope.safeParse(...)` on line 30

and then creates a new `ParseError` on line 35, with the `parsedBody.error` as the cause.

`parsedBody.error` is itself already a `ParseError` from [here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/parser/src/envelopes/envelope.ts) on line 64, and that has the `ZodError` as its cause.

So depending on whether it affects all envelopes or not, we could maybe look at changing line 64 in `envelope.ts` to be `error: parsed.error` instead of creating a `new ParseError` with a cause?

latest

18.x

### Packaging format used

npm

### Execution logs

_No response_

am29d:
## Summary

This PR fixes the bug of nested errors inside the `ParsedResultError` structure. 

### Changes

In the generic `Envelope` we have returned a `ParsedResultError` structure 

```ts
{
	success: false,
	error: new ParseError('foo'),
	originalEvent: data
}
```

In all cases the `Envelope` was called from specific envelopes such as `ApiGateway`, `EventBridge` etc. and in all of such cases we repackage the entire response again, which causes `error.cause` to be `ParseError` (from the Envelope call), and the not the expected `ZodError`. 

Therefore, the `Envelope.safeParse` now returns a generic error or whatever we catch within the `safeParse` and no `originalEvent`. The reason for that is that the specific envelope has the entire original payload and should have the responsibility to return this information back to the user. At that point the `originalEvent` is not relevant for the `Envelope` object at all. 

I had to modify the return because TypeScript is not aware of the if-statement before the return: 

```typescript
if (!parsedEnvelope.success) {
      return {

      };
    }
	
    return parsedBody as ParsedResultSuccess<z.infer<T>>;
```

I have also added more specific error assertions in the tests.

**Issue number:** closes #3208 

---


packages/parser/src/envelopes/apigw.ts
```diff
@@ -12,10 +11,7 @@ export const ApiGatewayEnvelope = {
     return Envelope.parse(APIGatewayProxyEventSchema.parse(data).body, schema);
   },
 
-  safeParse<T extends ZodSchema>(
-    data: unknown,
-    schema: T
-  ): ParsedResult<unknown, z.infer<T>> {
+  safeParse<T extends ZodSchema>(data: unknown, schema: T) {
```
dreamorosi, line 14: I'd consider putting back the return type, mostly for consistency with the other envelopes, but please dont' take this as a blocking comment if you don't want to.

dreamorosi (dismissed):
Thank you for fixing the issue and for the PR, Alex.

I have left a minor comment in one of the envelope files, but it's not a blocking one. Feel free to merge as-is if you prefer to leave it.

dreamorosi (approved):

PR #3248, opened 2024-10-23
test(parser): migrate to vitest & simplify structure

dreamorosi:
## Summary

> [!IMPORTANT]
> Please don't update/sync the branch until the PR is marked as ready for review

### Changes

> Please provide a summary of what's being changed

> [!Note]
> This PR touches an unusually high number of files due to it refactoring both the test cases and files containing events for these tests.

This PR migrates the Parser utility tests to use Vitest, and while doing so, also streamlines the test structure to adhere the following guidelines:
- Each test should import/read only files that are relevant to that test
- When testing event payloads for AWS Services, focus on tests that verify the actual behavior of schemas and transforms (i.e. having two identical events for API Gateway where one differs only in the `path` field containing a trailing slash is useless since we don't validate/parse the path beyond ensuring it's a `string`).
- Follow conventions set in the codebase and avoid type casts and suppressions unless strictly necessary

> Please add the issue number below, if no issue is present the PR might get blocked and not be reviewed

**Issue number:** closes #3247

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Disclaimer**: We value your time and bandwidth. As such, any pull requests created on non-triaged issues might not be successful.

PR #3260, opened 2024-10-25
fix(parser): add aws region to kinesis event

### LINKED ISSUE #3227: Bug: Kinesis Data Stream schema missing `awsRegion` field
### Expected Behavior

When parsing a Kinesis Data Stream record, all the fields in the event should be validated and parsed.

### Current Behavior

The `KinesisDataStreamRecord` schema ([here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/a3dd654d8b0e7e054300940586d73b3b23c14b6d/packages/parser/src/schemas/kinesis.ts#L29-L37)) is missing a field that indicates the AWS Region of the record.

### Code snippet

N/A

### Steps to Reproduce

N/A

### Possible Solution

_No response_

latest

20.x

### Packaging format used

npm

### Execution logs

_No response_

### LINKED ISSUE #3261: Bug: type in kinesisRecordMetadata in KinesisFirehoseSchema
### Expected Behavior

[KinesisFirehoseSchema](https://github.com/aws-powertools/powertools-lambda-typescript/blob/a3dd654d8b0e7e054300940586d73b3b23c14b6d/packages/parser/src/schemas/kinesis-firehose.ts#L15) has a field `kinesisRecordMetaData` and the correct name should be `kinesisRecordMetadata`

I have noticed the issue while comparing full objects of the kinesis tests.

### Current Behavior

The field name should be `kinesisRecordMetadata`

### Code snippet

none

### Steps to Reproduce

none

### Possible Solution

_No response_

latest

20.x

### Packaging format used

npm

### Execution logs

_No response_

am29d:
## Summary

This PR adds missing `awsRegion` in the `KinesisDataStreamRecord` schema. I have checked the python version and it is there, so we somehow missed it during implementation. To avoid these cases in the future, I have added a small helper function and tests to make our schemas strict for the test run. This will allow us to find cases where the test event contains properties that we did not include in the schema. 

### Changes

- add helper function to make the provided schema strict with some of the cases for union, arrays, records. There are probably more cases but I did not wanted to implement all spectrum, only what we need right now.
- add tests to verify that there are no properties missing in the schema that are included in the test event

**Issue number:** closes #3227 closes #3261

---


packages/parser/tests/unit/schema/kinesis.test.ts
```diff
@@ -110,4 +116,10 @@ describe('Kinesis ', () => {
       '49640912821178817833517986466168945147170627572855734274000000'
     );
   });
+
+  it('should catch any unknown fields in the example event', () => {
```
dreamorosi, line 120: I see where you're coming from with the addition of `.strict()` to the schemas, but I'm not sure it's the right choice in this situation.

The main concern I have is that we're no longer testing the model we're shipping to customers but instead a more strict version of it. Some customers might end up using `.strict()` on their own, which is actually a good idea, but I think it'd be better to test the schema as-is.

Ultimately, while the `.strict()` method does the job, we're really just observing a side effect rather than checking that the parsed output is what it is.

From a DX perspective, for us maintainers, using `.strict()` gives us an error like this:

image

Whereas checking the object would give us an actual diff of the object, clearly pointing at the missing field:

Screenshot 2024-10-28 at 11 18 46

While I recognise that the output is primarily a style choice, and both are serviceable, I think ultimately testing the output on the same schema we ship is more accurate.

For example, in this case you could modify the first test and check the entire payload instead of only one key, as well as changing the matcher from `toEqual()` to `toStrictEqual()`.

```ts
it('should parse kinesis event', () => {
  // Prepare
  const kinesisStreamEvent = TestEvents.kinesisStreamEvent;
  const expectedPayload = structuredClone(
    TestEvents.kinesisStreamEvent
  ) as KinesisDataStreamEvent;
  expectedPayload.Records[0].kinesis.data = 'Hello, this is a test.';
  expectedPayload.Records[1].kinesis.data = 'This is only a test.';

  // Act
  const parsed = KinesisDataStreamSchema.parse(kinesisStreamEvent);

  // Assess
  expect(parsed).toStrictEqual(expectedPayload);
});
```

This is also how I realized the key was missing while refactoring the tests in #3248.

dreamorosi (commented):
Thanks for the PR.

I have left a comment on the introduction of `.strict()` as I have some concerns on it - but I'm more than open to discuss.

packages/parser/tests/unit/schema/kinesis.test.ts
dreamorosi: That's good, it means the other schemas were correct!

PR #3285, opened 2024-11-04
test(maintenance): switch tracer to vitest

### LINKED ISSUE #3284: Maintenance: migrate test runner to vitest for tracer package
### Summary

To continue the migration from Jest to Vitest we need to move over the tests for the `tracer` package to use the new test runner.

### Why is this needed?

So that we can continue the migration as well as improve the overall performance and consistency of the tests.

Parser

### Solution

_No response_

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR moves the unit and end to end tests for the Tracer package to use vitest as test runner.

This action is part of the ongoing migration from Jest to Vitest, and the vast majority of the changes in the tests focus on changing the imports and using Vitest-specific APIs.

The file with most changes is the `packages/tracer/tests/unit/Provider.test.ts`, which has a module import mock for the `aws-xray-sdk-node` package. Because Jest and Vitest mock imports slightly differently, I had to make more changes in this file than in others.

I also have removed some redundant test cases while maintaining the same level of coverage.

Two extra files from other packages were touched, respectively for:
- `setupEnv.ts` under `packages/testing` was modified to include two environment variables needed for Tracer during tests.
- One Logger test case was modified to take in account the new value of an env variable that was modified as part of the point above.

**Issue number:** closes #3284

---


dreamorosi:
Integration tests passing on this branch: https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/11669742574/job/32492395223

packages/tracer/src/Tracer.ts
```diff
@@ -321,7 +321,7 @@ class Tracer extends Utility implements TracerInterface {
    * @deprecated Use {@link captureAWSv3Client} instead.
    * @param service - AWS SDK v2 client
    */
-  public captureAWSClient<T>(service: T): T {
+  /* v8 ignore start */ public captureAWSClient<T>(service: T): T {
```
am29d, line 324: What is the reason for v8 ignore directive? Deprecation?

packages/tracer/tests/unit/ProviderService.test.ts
```diff
@@ -1,102 +1,94 @@
...
+vi.mock('aws-xray-sdk-core', async (importOriginal) => ({
+  default: {
+    ...(
+      await importOriginal<{
+        default: typeof import('aws-xray-sdk-core');
+      }>()
+    ).default,
+    ...mocks,
+  },
+}));
```
am29d, line 38: wow, this is an interesting level of sorcery

packages/tracer/tests/unit/Tracer.test.ts
```diff
@@ -11,36 +6,29 @@ import {
...
-  [
-    name: string,
-    fcn: (subsegment?: Subsegment) => unknown,
-    parent?: Segment | Subsegment,
-  ]
->;
+/* vi.hoisted(() => {
+  process.env.POWERTOOLS_SERVICE_NAME = 'hello-world';
+  process.env.AWS_XRAY_LOGGING_LEVEL = 'silent';
+}); */
```
am29d, line 18: Forgot to remove?

am29d (changes requested):
Thanks for the PR! I have only few questions for clarification and we are good to merge.

packages/tracer/src/Tracer.ts
dreamorosi: Yes, correct. The method is marked as deprecated (L321) and so I just removed the test that was using it.

packages/tracer/tests/unit/ProviderService.test.ts
dreamorosi: Yes, it's not very intuitive.

Here are [the docs for the Vitest mocking feature](https://vitest.dev/api/vi.html#mock-modules). Basically we are using Vitest to hijack the module resolution for a specific module and returning our own object.

In doing so however, we are not replacing the entire module but only some of it. To do so, we are importing the original module and destructuring its content in the object we return. Next we overwrite some of the properties that we want to provide a mock for (see `mocks` object).

This specific module has also an added complexity which is the fact that it only provides a `default` export, which is why we need to nest the mock object within a `default` property.

dreamorosi:
@am29d thanks for the review, addressed your comments & ready for another round.

PR #3293, opened 2024-11-05
chore(parser): fix type inference for result types

### LINKED ISSUE #3226: Maintenance: Parser inferred result always set event to array type
### Summary

When using the Middy.js middleware for parser, customers can specify envelopes and schemas for the middleware to use to parse the event. To ease usage for customers, we introduced types that are supposed to infer the `event` type based on the arguments passed to the middleware.

Currently when using envelopes, the type of the `event` is always assumed to be an array when it should not be the case.

For example, this is correct:

```ts
import { parser } from '@aws-lambda-powertools/parser/middleware';
import { SqsEnvelope } from '@aws-lambda-powertools/parser/envelopes';
import { z } from 'zod';

  return {
    statusCode: 200,
    body: JSON.stringify('Hello, World!'),
  };
}).use(parser({ schema: z.any(), envelope: SqsEnvelope }));
```

Image

but this is not:

```ts
import { parser } from '@aws-lambda-powertools/parser/middleware';
import { EventBridgeEnvelope } from '@aws-lambda-powertools/parser/envelopes';
import { z } from 'zod';

  return {
    statusCode: 200,
    body: JSON.stringify('Hello, World!'),
  };
}).use(parser({ schema: z.any(), envelope: EventBridgeEnvelope }));
```

Image

In the second case, EventBridge's `detail-type` field is an object and not an array, but the [types here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/a3dd654d8b0e7e054300940586d73b3b23c14b6d/packages/parser/src/types/parser.ts#L53-L60) think it's always an array.

### Why is this needed?

So that the types when using the Middy.js middleware are correct.

Parser

### Solution

_No response_

### Acknowledgment

### Future readers

am29d:
## Summary

This PR fixes inferred for result types. 

### Changes

I have added discriminator field to each envelope to distinguish between array and object return types. We now also have tests to catch any cases where the return type for specific envelope does not match. 

Technically the object-envelopes can also return arrays, but this result will be according to the provided schema, i.e. `z.array` which also matches with current implementation. 

**Issue number:** closes #3226 

---


packages/parser/tests/unit/types.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  KafkaEnvelope,
+  KinesisEnvelope,
+  KinesisFirehoseEnvelope,
+  LambdaFunctionUrlEnvelope,
+  SnsEnvelope,
+  SnsSqsEnvelope,
+  SqsEnvelope,
+  VpcLatticeEnvelope,
+  VpcLatticeV2Envelope,
+} from '../../src/envelopes';
```
dreamorosi, line 17:
```suggestion
} from '../../src/envelopes/index.js';
```

packages/parser/tests/unit/types.test.ts
```diff
@@ -0,0 +1,77 @@
...
+import type { ParserOutput } from '../../src/types/parser';
```
dreamorosi, line 18:
```suggestion
import type { ParserOutput } from '../../src/types/parser.js';
```

dreamorosi (commented):
Thank you for the fix, this will improve the experience for customers.

Besides two minor comments on imports within tests, I have noticed that the new `symbol` property shows up in the IntelliSense autocomplete of the public API:

image

This is not a big deal, but I wonder if we could use alternatives to avoid this. Based on [tweets](https://x.com/colinhacks/status/1816953802446242063) that I read in the past there are at least two options that would avoid this and still serve as discriminator.

The first one is to use a property name that begins with a space, i.e. this:

```diff
export const ApiGatewayEnvelope = {
-  'symbol': 'object' as const,
+  ' symbol': 'object' as const,
```

This works, however according to [some sources](https://x.com/AndaristRake/status/1817090338890150082) the fact that it doesn't show might actually be a bug.

The second is to use an actual symbol, for example, in `packages/parser/src/envelopes/envelope.ts` we could export a symbol like this:

```ts
export const envelopeDiscriminator = Symbol.for('returnType');
```

and then, in each built-in envelope import and use it like this:

```diff
- import { Envelope } from './envelope.js';
+ import { Envelope, envelopeDiscriminator } from './envelope.js';

/**
 * API Gateway envelope to extract data within body key
 */
export const ApiGatewayEnvelope = {
-  'symbol': 'object' as const,
+  [envelopeDiscriminator]: 'object' as const,
```

and same in the `packages/parser/src/types/envelope.ts` file:

```diff
+ import { envelopeDiscriminator } from '../envelopes/envelope.js';

interface ArrayEnvelope {
-  'symbol': 'array';
+  [envelopeDiscriminator]: 'array';
  parse<T extends ZodSchema>(data: unknown, schema: T): z.infer<T>[];
  safeParse<T extends ZodSchema>(
    data: unknown,
    schema: T
  ): ParsedResult<unknown, z.infer<T>[]>;
}
```

with this method the types still work but the discriminator doesn't show up in the public API:

image

Using a symbol vs a string as key would also, afaik, avoid potential collisions in case customers start adding properties to the envelopes since even if they used the same key the reference would be different (but haven't verified this).

Additionally, I think we could consider & test adding the `@internal` annotation in the docstring for the object property so that it doesn't show up in the API docs either.

Finally, just as a comment, I think we might have been able to achieve the same result by adding the discriminator only to one group of the envelopes (i.e. the array ones) and then setting the type of the other as `never` in the interface, so that the absence of the discriminator would have served as discriminator itself. But I actually like being explicit about this, so I'd say let's leave it as-is.

am29d:
Thanks for raising this point, I completely forgot about hiding the property. 

I have picked your second option, and added `@hidden` to remove it from our API Docs. I agree on the last point, let's make it explicit.

dreamorosi (approved):
Thank you for addressing the review comments! Great addition

Issue #3302, opened 2024-11-08
Bug: dimension value not being validated for empty strings


dreamorosi:
### Expected Behavior

When adding a dimension to my metrics, I should get a warning if the dimension value is invalid aka an empty string (`""`) or `undefined`/`null`, and the dimension should not be added to the EMF blobs emitted by the utility.

### Current Behavior

Currently the utility doesn't validate the input but only enforces types. In practice this means it's unlikely customers can pass `undefined` or `null` unless they're suppressing the warning, but can pass an empty string and generate an EMF blob that will have the invalid dimension.

### Code snippet

```typescript
import { Metrics } from '@aws-lambda-powertools/metrics';

const metrics = new Metrics();

export const handler = async () => {
  metrics.addDimension('FunctionName', '');
  metrics.addMetric('CustomMetric', 'Count', 1);
  metrics.publishStoredMetrics();
};
```

### Steps to Reproduce

1. Run the code above
2. Observe that the EMF blob emitted has an empty dimension
3. Verify that the processed metric in CloudWatch does not include any dimensions

### Possible Solution

We should: 1/ validate the input when adding a dimension, 2/ emit a warning if invalid, and 3/ don't add the dimension to the EMF blob

### Powertools for AWS Lambda (TypeScript) version

latest

### AWS Lambda function runtime

20.x

### Packaging format used

npm

### Execution logs

```Shell
{
  "_aws": {
    "Timestamp": 1731086204245,
    "CloudWatchMetrics": [
      {
        "Namespace": "default_namespace",
        "Dimensions": [
          [
            "service",
            "FunctionName"
          ]
        ],
        "Metrics": [
          {
            "Name": "CustomMetric",
            "Unit": "Count"
          }
        ]
      }
    ]
  },
  "service": "service_undefined",
  "FunctionName": "",
  "CustomMetric": 1
}
```

dreamorosi:
[From the BelieveInServerless Discord server](https://discord.com/channels/1182412634423832666/1305505585025908808/1305571709998465165), a customer was having issues while passing a value typed as `any` which at runtime was a `number` and that caused the same behavior described above.

Another +1 for runtime validation.

arnabrahman:
Ok, let's do it 🙂

GusToughSon:
import { Metrics } from '@aws-lambda-powertools/metrics';

class EnhancedMetrics extends Metrics {
  addDimension(name: string, value: string | undefined | null): void {
[...]

PR #3319, opened 2024-11-14
fix(metrics): skip empty string dimension values

arnabrahman:
## Summary
When adding a dimension to my metrics, I should get a warning if the dimension value is invalid aka an empty string ("") or undefined/null, and the dimension should not be added to the EMF blobs emitted by the utility.

### Changes

- Check for `empty`, `undefined` & `null` values for dimension and if found log warning
- Skip invalid value dimensions to EMF, do this for `addDimension` & `addDimensions`
- Unit test the fix

**Issue number:** #3302 

---

### DIFF
--- a/packages/metrics/src/Metrics.ts
+++ b/packages/metrics/src/Metrics.ts
@@ -246,6 +246,7 @@ class Metrics extends Utility implements MetricsInterface {
    * Add multiple dimensions to the metrics.
    *
    * This method is useful when you want to add multiple dimensions to the metrics at once.
+   * Invalid dimension values are skipped and a warning is logged.
    *
    * When calling the {@link Metrics.publishStoredMetrics | `publishStoredMetrics()`} method, the dimensions are cleared. This type of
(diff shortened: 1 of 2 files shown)


dreamorosi (changes requested):
Thank you for the revisions @arnabrahman!

The friends at the Python version told me they already merged the change and used a slightly different wording in the warning log, which you [can see here](https://github.com/aws-powertools/powertools-lambda-python/commit/f8fe2a3c60e132cf211090417b83eb6ff9dffbc2#diff-d4a359e5caba54c548936f54ad216c2b6a81f00929e3dc2b33cba5b43c6b84caR274-R279) and is this:

> The dimension {name} doesn't meet the requirements and won't be added. Ensure the dimension name and value are non empty strings"

As last step before merging, could you please change to this?

I'll merge right after, sorry I missed the change on their part and didn't point it out in my first review of your PR.

arnabrahman:
It's ok, i have updated the warning message.

dreamorosi (approved):
Thanks for addressing the feedback, let's merge it!

Issue #3329, opened 2024-11-18
Feature request: ability to silence / skip publishing metrics via env variable


dreamorosi:
### Discussed in https://github.com/aws-powertools/powertools-lambda-typescript/discussions/3315

<div type='discussions-op-text'>

<sup>Originally posted by **zirkelc** November 14, 2024</sup>
Hi,

I wanted to know if it is possible to _silence metrics_ based on an environment variable in the same way it is possible to [silence logs](https://docs.powertools.aws.dev/lambda/typescript/main/core/logger/#silencing-logs).

Basically, I want to omit publishing metrics if the environment **is not** production. Ideally, that should be controllable with an env var.</div>

### Solution/User Experience

- by default, Metrics is enabled (current behavior)
- if the `POWERTOOLS_METRICS_DISABLED` env variable is present and [truthy](https://github.com/aws-powertools/powertools-lambda-typescript/blob/5350afed92d02b7b8d47f387705f70c59deeeb65/packages/commons/src/config/EnvironmentVariablesService.ts#L90-L94) when the `Metrics` class is instantiated, then it becomes disabled
- if dev mode is enabled by means of `POWERTOOLS_DEV` which can be checked with [this method](https://github.com/aws-powertools/powertools-lambda-typescript/blob/5350afed92d02b7b8d47f387705f70c59deeeb65/packages/commons/src/config/EnvironmentVariablesService.ts#L81-L83), when the `Metrics` class is instantiated, then it also becomes disabled
- If `POWERTOOLS_METRICS_DISABLED` is set to `false`, and dev mode is enabled, then `POWERTOOLS_METRICS_DISABLED` takes precedence and the utility stays enabled.

### Alternative solutions

N/A

### Acknowledgment

- [x] This feature request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [x] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

zirkelc:
I will work on this!

PR #3351, opened 2024-11-24
feat(metrics): disable metrics with `POWERTOOLS_METRICS_DISABLED`

zirkelc:
### Changes

Add a new environment variable `POWERTOOLS_METRICS_DISABLED` to disable emitting metrics.

**Issue number:** #3329 

---


packages/metrics/src/Metrics.ts
```diff
@@ -480,6 +485,13 @@ class Metrics extends Utility implements MetricsInterface {
     return Object.keys(this.storedMetrics).length > 0;
   }
 
+  /**
+   * Whether metrics are disabled.
+   */
+  public isDisabled(): boolean {
+    return this.disabled;
+  }
```
zirkelc, line 493: This method could be omitted, but it followed the style from Logger which also exposes the `getLogEvent` method.
Also, this allows to easily test the effect of different environment variables.

packages/metrics/src/Metrics.ts
```diff
@@ -587,8 +599,14 @@ class Metrics extends Utility implements MetricsInterface {
...
     }
-    const emfOutput = this.serializeMetrics();
-    hasMetrics && this.console.log(JSON.stringify(emfOutput));
+
+    if (!this.disabled) {
+      const emfOutput = this.serializeMetrics();
+      hasMetrics && this.console.log(JSON.stringify(emfOutput));
+    } else {
+      // TODO log warning that metrics are disabled?
+    }
```
zirkelc, line 608: A few notes here:
- added the check on `this.disabled` here instead of in `addMetric` in case it should be possible to "toggle" between enabled and disabled in the future
- added an else-clause in case we need to log a warning if metrics are disabled, like with the empty metrics warning

packages/metrics/src/Metrics.ts
```diff
@@ -921,6 +939,17 @@ class Metrics extends Utility implements MetricsInterface {
...
 
+  /**
+   * Set the disbaled flag based on the environment variables `POWERTOOLS_METRICS_DISABLED` and `POWERTOOLS_DEV`.
+   */
+  private setDisabled(): void {
+    this.disabled =
+      this.getEnvVarsService().isDevMode() ||
+      this.getEnvVarsService().getMetricsDisabled();
+
+    // TODO if POWERTOOLS_METRICS_DISABLED is set to false, and dev mode is enabled, then POWERTOOLS_METRICS_DISABLED takes precedence and the utility stays enabled.
```
zirkelc, line 950: Getting this two work requires some changes to `CommonEnvironmentVariablesService`, so I wanted to align first before I start.

The `CommonEnvironmentVariablesService` only provides a `isValueTrue()` method to check for truthy values. That means `POWERTOOLS_METRICS_DISABLED=false` and `POWERTOOLS_METRICS_DISABLED=undefined` (not set) will both evaluate to `false`. However, in order to give `POWERTOOLS_METRICS_DISABLED` precedence over `POWERTOOLS_DEV`, I would need a method to check if `POWERTOOLS_METRICS_DISABLED` is present and not-truthy.


```ts
class EnvironmentVariablesService {

  public isPresent(name: string): boolean {
    return name in process.env;
  }
}
```

If there's another way that I overlooked, please let me know.

packages/metrics/src/config/EnvironmentVariablesService.ts
```diff
@@ -13,12 +13,23 @@ class EnvironmentVariablesService
...
    * Get the value of the `POWERTOOLS_METRICS_NAMESPACE` environment variable.
    */
   public getNamespace(): string {
     return this.get(this.namespaceVariable);
   }
+
+  /**
+   * Get the value of the `POWERTOOLS_METRICS_DISABLED` environment variable.
+   */
+  public getMetricsDisabled(): boolean {
```
zirkelc, line 28: Naming alternatives: isDisabled, getDisabled

packages/metrics/src/Metrics.ts
dreamorosi: Good point, I think we could add a new method in `CommonEnvironmentVariablesService` called `isValueFalse()` and use that for the precedence logic.

If you go this route and make changes in the `packages/commons/*` directory, make sure to run `npm run build -w packages/commons` to see them reflected in `packages/metrics/*` (& elsewhere).

packages/metrics/src/Metrics.ts
dreamorosi: Thanks for the added details!

I agree with checking for the value here, at least for now this is the least disruptive place to make this change.

Regarding the `if/else` for now I think we can remove it and avoid logging any warning. This is an opt-in behavior that can be enabled only via environment variable, so I think it's fine to stay quiet. 

This is also a feature we recommend explicitly only for development, and for other utilities like Tracer we also don't emit any warning when disabled explicitly.

packages/metrics/src/Metrics.ts
dreamorosi: Agree with making it a method, although I would prefer if we can make it `protected` rather than `public`.

I think it's fair to make it accessible to those who extend the Metrics class, but I wouldn't necessarily open it up for the public API.

packages/metrics/src/Metrics.ts
```diff
@@ -943,6 +972,7 @@ class Metrics extends Utility implements MetricsInterface {
     this.setNamespace(namespace);
     this.setService(serviceName);
     this.setDefaultDimensions(defaultDimensions);
+    this.setDisabled();
```
dreamorosi, line 975: For future usage, I'd move this up after `this.setCustomConfigService(customConfigService);`, so if we want to take in account this value in other setup actions we can do it.

packages/metrics/src/config/EnvironmentVariablesService.ts
```diff
@@ -13,12 +13,23 @@ class EnvironmentVariablesService
 {
   private namespaceVariable = 'POWERTOOLS_METRICS_NAMESPACE';
 
+  private disabledVariable = 'POWERTOOLS_METRICS_DISABLED';
```
dreamorosi, line 16: I know other variables are not declared as such, but since we're adding a new one and it's never reassigned, let's make it `private readonly`.

packages/metrics/src/config/EnvironmentVariablesService.ts
dreamorosi: I think this is ok as it is.

packages/metrics/src/config/EnvironmentVariablesService.ts
```diff
@@ -13,12 +13,27 @@ class EnvironmentVariablesService
...
+   */
+  public getMetricsDisabled(): boolean | undefined {
+    const value = this.get(this.disabledVariable);
+
+    if (this.isValueTrue(value)) return true;
+
+    if (this.isValueFalse(value)) return false;
+
+    return undefined;
+  }
```
zirkelc, line 36: I need to make this return `boolean | undefined` to account for not setting this value.

dreamorosi: What if instead of this, we extract the logic that we currently have in `setDisabled`  from the Metrics class, so that we always return a boolean?

We have precedents (like [this](https://github.com/aws-powertools/powertools-lambda-typescript/blob/12f3e448d8104b08518370a71db12646dededa4f/packages/logger/src/config/EnvironmentVariablesService.ts#L100-L105)) of methods in the env variable config service that look at multiple env variables.

What do you think?

packages/metrics/tests/unit/Metrics.test.ts
```diff
@@ -2343,6 +2366,54 @@ describe('Class: Metrics', () => {
     });
   });
 
+  describe('Method: isDisabled', () => {
+    it('should be enabled by default', () => {
+      const metrics: Metrics = new Metrics({ namespace: TEST_NAMESPACE });
+
+      // Act & Assess
+      // biome-ignore lint/complexity/useLiteralKeys: accessing protected method
+      expect(metrics['isDisabled']()).toBe(false);
```
zirkelc, line 2375: Since `isDisabled` is `protected`, TypeScript is reporting an error if accessed via dot notation.
The bracket notation is allowed by TS as [intentional escape hatch](https://github.com/microsoft/TypeScript/issues/19335). 

I could also refactor this tests to use `publishStoredMetrics()` instead. IMO prefer the cleanliness of the tests, but I'm open to changing it.

packages/metrics/src/config/EnvironmentVariablesService.ts
zirkelc: Like this? 

```suggestion
  public getMetricsDisabled(): boolean {
    const devMode = this.isDevMode();
    const value = this.get(this.disabledVariable);

    if (this.isValueFalse(value)) return false;
    if (this.isValueTrue(value)) return true;
    if(this.isDevMode()) return false;

    return false;
  }
```

It would consolidate the env var logic at one place, which is nice.

packages/metrics/tests/unit/Metrics.test.ts
zirkelc: Probably obsolete if we change `getMetricsDisabled`

dreamorosi: I think it's fine to leave it as you wrote it for now, and is consistent with other tests.

I am also working on another PR that refactors the unit tests for this module to address this (#2821) and remove this type of pattern in favor of the other one you're suggesting. Overall I think observing the actual side effect - in this case the metrics not being emitted - is better and makes the tests less brittle and maintainable over time.

With this type of test that reaches into the implementation, if we for example rename the `isDisabled` property or refactor the internal implementation, the test will fail and also need refactoring while the feature hasn't functionally changed.

If instead we observe the behavior, as long as metrics are not published the refactor was successful.

---

For now however leave it as-is as it'll make the rebasing easier to follow, since using the other pattern would require more changes. If you're curious about how the refactoring looks like, check the PRs in a couple hrs. I was waiting to publish that branch to save you from some merge conflicts.

zirkelc: Okay, let me know if need to change something after your PR is merged! 👍

dreamorosi (approved):
Thank you for the high quality PR @zirkelc - really appreciate it!

And congrats for the first PR merged into this project!

zirkelc:
Thank you @dreamorosi!

PR #3352, opened 2024-11-24
feat(metrics): warn when overwriting dimension

### LINKED ISSUE #3350: Feature request: Warn users when adding multiple values to dimension before flushing
### Use case

In reference to [this post](https://bsky.app/profile/did:plc:aqtn3pn3dk4kd76fleosrrxb/post/3lbkgd7kpik2q), when a user adds multiple values to a single dimension before flushing metrics, only the last value set on the dimension wins.

### Solution/User Experience

It would be nice if the metrics object would do a `console.warn` when multiple distinct values are used on a single dimension before flushing. Current behavior uses the last value added to the dimension.

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

dreamorosi:
### Changes

When working with dimensions with Metrics, the default behavior is to overwrite existing dimensions when a new one with the same name/key is added. Customers have asked us to add a warning when this happens so that they can more easily troubleshoot issues rather than missing data.

#### Before

```ts

const metrics = new Metrics({ singleMetric: true });

metrics.addDimension('test', 'foo');
metrics.addDimension('test', 'bar'); // "bar" silently replaces "foo" as value
```

#### After

```ts

const metrics = new Metrics({ singleMetric: true });

metrics.addDimension('test', 'foo');
metrics.addDimension('test', 'bar'); // "bar" replaces "foo"

// warning: Dimension "test" has already been added. The previous value will be overwritten.
```

The PR also consolidates the implementation of `addDimension()` and `addDimensions()` with the latter now iterating through the dimensions and calling the former for each one, instead of bringing its own separate implementation. In doing so, it also fixes a bug in which the `addDimensions()` method didn't consider default dimensions when counting the total of allowed dimensions in EMF (30).

**Issue number:** closes #3350

---

### DIFF
--- a/packages/metrics/src/Metrics.ts
+++ b/packages/metrics/src/Metrics.ts
@@ -244,7 +244,7 @@ class Metrics extends Utility implements MetricsInterface {
       Object.hasOwn(this.defaultDimensions, name)
     ) {
       this.#logger.warn(
-        `Dimension with "${name}" has already been added. The previous value will be overwritten.`
+        `Dimension "${name}" has already been added. The previous value will be overwritten.`
       );
     }
     this.dimensions[name] = value;
--- a/packages/metrics/src/Metrics.ts
+++ b/packages/metrics/src/Metrics.ts
@@ -239,6 +239,14 @@ class Metrics extends Utility implements MetricsInterface {
         `The number of metric dimensions must be lower than ${MAX_DIMENSION_COUNT}`
       );
     }
+    if (
+      Object.hasOwn(this.dimensions, name) ||
+      Object.hasOwn(this.defaultDimensions, name)
+    ) {
+      this.#logger.warn(
+        `Dimension with "${name}" has already been added. The previous value will be overwritten.`
+      );
+    }
     this.dimensions[name] = value;
   }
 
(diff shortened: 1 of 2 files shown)


leandrodamascena:
I'm adding this one in Python as well. Thanks for working on this @dreamorosi

packages/metrics/src/Metrics.ts
```diff
@@ -255,25 +263,9 @@ class Metrics extends Utility implements MetricsInterface {
...
-      }
-    }
-    if (Object.keys(newDimensions).length > MAX_DIMENSION_COUNT) {
-      throw new RangeError(
-        `Unable to add ${
-          Object.keys(dimensions).length
-        } dimensions: the number of metric dimensions must be lower than ${MAX_DIMENSION_COUNT}`
-      );
+    for (const [name, value] of Object.entries(dimensions)) {
+      this.addDimension(name, value);
```
leandrodamascena, line 267: Just to make sure I understood this change: you are now removing the logic from this method and moving it to `public addDimension(name: string, value: string): void {` because you want to have all this logic in one place, right? With this you iterate through the items and now pass the string values ​​directly to this other method instead of having 2 logics in different places. 
If so, good catch!!

leandrodamascena (approved):
Hi @dreamorosi! Congrats on fixing this so quickly! I'll do the same in Python!

APPROVED!

packages/metrics/src/Metrics.ts
dreamorosi: yea, that's what it does.

Issue #3376, opened 2024-11-27
Feature request: change visibility from `private` to `protected` to allow overrides


zirkelc:
### Use case

As discussed before, I would like to suggest to change the visibility of certain methods of `Logger` from `private` to `protected` to allow overrides by subclassing `Logger`. A use case for such an override would be the buffering of log messages (see #3178). 

If approved, I'm happy to submit a PR.

### Solution/User Experience

The following methods could be made `protected`:
https://github.com/aws-powertools/powertools-lambda-typescript/blob/0dc4de15e0ecb4d39585bd094a2d958be2d06bb2/packages/logger/src/Logger.ts#L703-L707

https://github.com/aws-powertools/powertools-lambda-typescript/blob/0dc4de15e0ecb4d39585bd094a2d958be2d06bb2/packages/logger/src/Logger.ts#L838

https://github.com/aws-powertools/powertools-lambda-typescript/blob/0dc4de15e0ecb4d39585bd094a2d958be2d06bb2/packages/logger/src/Logger.ts#L865-L869

### Alternative solutions

```Markdown

```

### Acknowledgment

- [x] This feature request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

dreamorosi:
Hi @zirkelc, the proposal makes sense.

We discussed this in #3178, so feel free to open a PR to make these changes and I'll help you get it merged before the next release!

PR #3377, opened 2024-11-28
feat(logger): change visibility to protected

zirkelc:
## Summary

Change the visibility of `Logger` methods from `private` to `protected` to allow subclassing of `Logger` to add new functionaility

### Changes

**Issue number:** #3376 

---


packages/logger/src/Logger.ts
```diff
@@ -846,6 +846,7 @@ class Logger extends Utility implements LoggerInterface {
             'critical'
           >);
 
+    // TODO: make this.console protected or create getConsole/getConsoleMethod?
     this.console[consoleMethod](
       JSON.stringify(
         log.getAttributes(),
```
zirkelc, line 852: Ideally, `this.console` should also be protected or someway accessible from subclass. Maybe we could add a `getConsole()` or `getConsoleMethod(logLevel: string)`?

dreamorosi (changes requested):
I see that you changed more methods and properties than the ones we discussed in #3376.

I would be inclined to stick to the initial plan and revert the other changes, or if the initial list of methods discussed are not enough to cover your use case, close the PR and invest time in shipping the feature for everyone within the boundaries of the public API.

The reason why we're reticent with making everything accessible is that we don't want to create the false expectation that all these methods/properties are there to stay.

Our commitment to avoiding breaking changes extends exclusively to public methods. Any other method or property is to be considered internal for all intents and purposes. 

This is also why we're gradually moving away from `private` and have started using actual JavaScript private fields (i.e. `private getLogLevelNameFromNumber()` becomes `#getLogLevelNameFromNumber()`).

zirkelc:
I changed the additional methods/properties back to `private`, please have a look.

There is one optional comment about `this.console`: it would be useful to have this accessible from subclasses, but it's not required. I can work around that.

Issue #3200, opened 2024-10-14
Feature request: publish Lambda layers in Malaysia, Mexico, Thailand


leandrodamascena:
### Use case

AWS recently launched new region:

Asia Pacific (Malaysia) - ap-southeast-5

https://aws.amazon.com/pt/blogs/aws/now-open-aws-asia-pacific-malaysia-region/

### Solution/User Experience

Deploy Lambda Layers to this new region.

### Acknowledgment

- [x] This feature request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [x] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

dreamorosi:
Adding the newer Mexico region to the issue (& renamed title).

We'll aim to release them at the next release which will be in 2 weeks.

dreamorosi:
The regions are all bootstrapped in the beta stage, we still need to:
- bootstrap the prod stages
- verify that the Lambda layers are balanced in the new regons

dreamorosi:
@am29d I see you have bootstrapped the regions but didn't add the new regions to the workflow matrix.

With today's release the layer version in the 3 new regions went out of sync again. Please balance the versions again and add open a PR as soon as you can - so they don't go out of sync again.

am29d:
Yep, wanted to finish the work before noon, but then release was out, fix incoming.

dreamorosi:
Ran the [SSM parameter script](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/13522904825) & verified that the correct layer was published in the new regions.

PR #3438, opened 2024-12-20
feat(ci): Add advanced automation

sthulb:
### Changes

This PR adds the ability to bootstrap new regions and create SSM parameters to alias Lambda Layer ARNs.

**Issue number:** #3200, #3439

---


.github/workflows/bootstrap_region.yml
```diff
@@ -0,0 +1,94 @@
...
+          mask-aws-account-id: true
+      - id: workdir
+        name: Create Workdir
+        run: |
+          mkdir -p build/project
+      - id: cdk-install
+        name: Install CDK
+        working-directory: build
+        run: |
+          npm i aws-cdk
```
dreamorosi, line 60: Can we use the version we already have installed in the workspace to avoid inconsistency issues?

.github/workflows/bootstrap_region.yml
```diff
@@ -0,0 +1,94 @@
...
+      - id: credentials
+        name: AWS Credentials
+        uses: aws-actions/configure-aws-credentials@e3dd6a429d7300a6a4c196c26e071d42e0343502
+        with:
+          aws-region: us-east-1
+          role-to-assume: ${{ secrets.REGION_IAM_ROLE }}
+          mask-aws-account-id: true
+      - id: go-setup
+        name: Setup Go
+        uses: actions/setup-go@3041bf56c941b39c61721a86cd11f3bb1338122a # v5.2.0
```
dreamorosi, line 85: Do we need to pin a Go version?

.github/workflows/bootstrap_region.yml
sthulb: I deliberately didn't do it since I didn't want to get into a situation where we're always running a old version by accident

.github/workflows/bootstrap_region.yml
leandrodamascena: Dependabot will update this, no? I don't know if we need to worry.

.github/workflows/bootstrap_region.yml
dreamorosi: I'd rather run an older version that we know works both because of e2e tests and release process, than a new one that might be broken when running this specific workflow.

Regardless, not pinning any version and pulling from the remote is not a great idea. If we're gonna pin it here, we might as well use the one defined in the lock file, so Dependabot updates it once a week

.github/workflows/bootstrap_region.yml
```diff
@@ -0,0 +1,96 @@
...
+          mask-aws-account-id: true
+      - id: workdir
+        name: Create Workdir
+        run: |
+          mkdir -p build/project
+      - id: cdk-install
+        name: Install CDK
+        working-directory: build
+        run: |
+          npm i aws-cdk@2.178.0
```
dreamorosi, line 60: Can we just use the one that is already present in the project?

I am aware that to do that we need to clone the repository and add a couple steps ([example](https://github.com/aws-powertools/powertools-lambda-typescript/blob/674ba36ac93e7697447d56789998484aad2d7908/.github/workflows/publish_layer.yml#L33-L42)) but at least we are already confident that this version of CDK works since it's being used elsewhere.

This workflow will be run only a handful of times each year and CDK has been the number one issue in our pipelines so far, so having this workflow use a known working one makes more sense to me. 

Plus if we hardcode the version here we'll have to open issues & PRs just to update it. Using the version present in the lock file makes it updatable, but still bound to the rest of the project.

dreamorosi (changes requested):
Thank you for the update - I just want to make sure we're using the CDK version from the project, so we can be sure that it'll work since we're already using it in other workflows.

Other than that, assuming we have no way to test this or it has already been tested elsewhere - is there a way to make the `update_ssm` one part of the release workflow? 

Or should we keep it as a manual step? If yes, why?

.github/workflows/bootstrap_region.yml
sthulb: I want to say no here – I don't want to spend more time testing it to ensure I have all of the IAM permissions included

dreamorosi: What does using the CDK version in the `package-lock.json` file has to do with IAM?

dreamorosi: I'm not talking about OIDC provider nor anything on the AWS account side - this is literally replacing the step where you run `npm i aws-cdk` with the ones I linked above:

```yaml
- name: checkout
  uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
  with:
    ref: ${{ github.sha }}
- name: Setup Node.js
  uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
  with:
    node-version: "22"
- name: Setup dependencies
  uses: aws-powertools/actions/.github/actions/cached-node-modules@29979bc5339bf54f76a11ac36ff67701986bb0f0
```

If the IAM permissions used by the workflow work with that CDK you're pulling from npm it'll work with this one. 

The difference is that we'll know CDK won't choke because we're already using it elsewhere.

dreamorosi: I misunderstood your comment - thanks for explaining offline - you were right about CDK versions requiring different IAM policies because this script is used to bootstrap a new region, which does involve deploying an IAM role.

If I understand correctly what we agreed upon, we're still going to use the CDK version in the monorepo, and when the time to deploy a new region comes we'll run the workflow. If it fails, we'll adjust the permissions in our AWS account and re-run it.

dreamorosi (approved):
I still would have preferred to use the CDK version in the project rather than pinning it, like we discussed, but I also want to move forward and close this PR.

If the OpenSSF Scorecard complains I'll make another commit do it.

Issue #3194, opened 2024-10-11
Feature request: parser helper to work with DynamoDB data structure


dreamorosi:
### Use case

When working with Amazon DynamoDB Stream events, the service sends the DynamoDB items in the database-specific format rather than regular JSON objects.

For example, a `string` might be represented as `{ "S": "foo" }` rather than just `foo`. This makes parsing the actual changes within an event hard to parse and validate since it requires customers writing a Zod schema with this in mind.

We should work on adding a helper that eases the authoring of schemas by abstracting away the transformation (unmarshalling).

### Solution/User Experience

The helper will need to depend on the `@aws-sdk/util-dynamodb` package, which should be an optional peer dependency for the Parser utility.

For this reason, we'll need to be intentional in how we expose the helper to avoid forcing the dependency on every Parser customer.

```ts
import { unmarshall } from '@aws-sdk/util-dynamodb';

const DynamoDBMarshalled = <T extends ZodTypeAny>(schema: T) =>
  z
    .record(z.string(), z.unknown())
    .transform((str, ctx) => {
      try {
        return unmarshall(str);
      } catch (err) {
        ctx.addIssue({
          code: 'custom',
          message: 'Could not unmarshall DynamoDB stream record',
          fatal: true,
        });

        return z.NEVER;
      }
    })
    .pipe(schema);
```

Which would then be used as:

```ts
import { DDBUnmarshalled } from '@aws-lambda-powertools/parser/helpers/dynamodb'; // <- new sub-path export
import { DynamoDBStreamSchema } from '@aws-lambda-powertools/parser/schemas/dynamodb';
import { z } from 'zod';

const MyItemSchema = z.object({
  foo: z.string(),
  bar: z.number().optional()
})

const EventSchema = DynamoDBStreamSchema.extend({
  dynamodb: z.object({
    NewImage: DDBUnmarshalled(MyItemSchema).optional(),
  }),
})
```

Note that we are also introducing a new sub-path export dedicated to this feature. By exporting the helper in its own path rather than in the main `@aws-lambda-powertools/parser/helpers` path we can make sure that only customers who actually want to use the helper need to install the extra peer dependency.

### Acknowledgment

- [x] This feature request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
[...]

PR #3442, opened 2024-12-23
feat(parser): `DynamoDBMarshalled` helper to parse DynamoDB data structure

arnabrahman:
## Summary
When working with Amazon DynamoDB Stream events, the service sends the DynamoDB items in the database-specific format rather than regular JSON objects.

For example, a string might be represented as { "S": "foo" } rather than just foo. This makes parsing the actual changes within an event hard to parse and validate since it requires customers writing a Zod schema with this in mind.

This PR introduces a helper function `DynamoDBMarshalled` to work this structure
### Changes

- Helper function to unmarshall and validate dynamoDB stream schema
- New helper function is exported to a subpath
- Unit test the new helper function
- Write up the docs for the new helper function

**Issue number:** #3194 

---


packages/parser/src/helpers/dynamodb.ts
```diff
@@ -0,0 +1,82 @@
...
+ *     }
+ *   ]
+ * }
+ * ```
+ *
+ * @param schema - The schema to validate the JSON string against
+ */
+const DynamoDBMarshalled = <T extends ZodTypeAny>(schema: T) =>
+  z
+    .record(z.string(), z.custom<AttributeValue>())
```
arnabrahman, line 66: I would appreciate suggestions on using this type.

Setting the type to `unknown` causes a TypeScript error. The `@aws-sdk/util-dynamodb` package has a peer dependency on `@aws-sdk/client-dynamodb` and we are using the `AttributeValue` type from `@aws-sdk/client-dynamodb`.  Should we also declare `@aws-sdk/client-dynamodb` as a peer dependency? 

Alternatively, we could set the type to `unknown` and suppress the TypeScript error using `biome-ignore`, though I would prefer to avoid this approach.

dreamorosi: I was on the fence and I shared the concern, but after thinking about it I think using the type from `@aws-sdk/client-dynamodb` is fine, anyway you'd need these dependencies only if you're actually working with this helper.

I would however change the helper from this:

```ts
z.record(z.string(), z.custom<AttributeValue>())
```

to this:

```ts
z.union([
  z.custom<Attributevalue>(),
  z.record(z.string(), z.custom<AttributeValue>())
])
```

Since the `unmarshall()` function has this signature `Record<string, AttributeValue> | AttributeValue`.

dreamorosi (changes requested):
Thank you for the PR and for your patience!

I have a minor comment in the signature of the helper so that it's aligned with the underlying `unmarshall` function, but other than that I am ready to approve the PR.

packages/parser/src/helpers/dynamodb.ts
arnabrahman: One thing, should we also add `@aws-sdk/client-dynamodb` as a peer dependency? 

The reason is that if someday, `@aws-sdk/util-dynamodb` decides to use its own type for `unmarshall` rather than `AttributeValue` as this is the only use case for `client-dynamodb` inside `@aws-sdk/util-dynamodb` and removes `@aws-sdk/client-dynamodb` from its peer dependencies, we might encounter errors here.

I don't know I might be overthinking here.

dreamorosi (approved):
Thank you for the PR!

dreamorosi:
Just as FYI @arnabrahman - Over the next couple of weeks the team and I will invest time in creating a roadmap and backlog for the next two quarters.

With that, we'll be creating a number of issues and label them accordingly.

arnabrahman:
@dreamorosi Thanks for the update, will be looking forward to it 🚀

PR #3470, opened 2025-01-14
fix(tracer): forward `X-Amzn-Trace-Id` header when instrumenting fetch

### LINKED ISSUE #3467: Bug: fetch requests don't forward X-Ray trace header
### Expected Behavior

When I make requests using the Fetch API and using Tracer, the `X-Amzn-Trace-Id` header should be constructed and forwarded along with the request so that I can get a full trace for a request.

For example, if I have an API Gateway A -> Lambda A -> API Gateway B -> Lambda B, and the first function in the chain is instrumented with Tracer, we should be able to see a trace like this:

Image

This is already supported when using the global `http` and `https` modules and their derivatives (i.e. `axios`) but doesn't work in our implementation when using Tracer with `fetch`.

### Current Behavior

Currently the setup described above generates two disconnected traces, one for each API Gateway -> Lambda pair.

### Code snippet

```ts
import { Tracer } from "@aws-lambda-powertools/tracer";

const logger = new Logger();
const tracer = new Tracer();

export const handler = async (event: { userId: number }) => {
	const parent = tracer.getSegment();
	const segment = parent?.addNewSubsegment("### functionA handler");
	segment && tracer.setSegment(segment);

	await fetch(`${process.env.API_URL}/functionB`, {
		method: "POST",
		body: JSON.stringify({ userId: event.userId }),
	});

	segment?.close();
	parent && tracer.setSegment(parent);

	return {
		statusCode: 200,
	};
};
```

### Steps to Reproduce

1. Deploy the infrastructure below
<details>

```ts
/// <reference path="./.sst/platform/config.d.ts" />

export default $config({
	app(input) {
		return {
			name: "missing-trace",
			removal: input?.stage === "production" ? "retain" : "remove",
			home: "aws",
		};
	},
	async run() {
		$transform(sst.aws.Function, (args) => {
			args.permissions = [
				{
					actions: [
						"xray:PutTraceSegments",
						"xray:PutTelemetryRecords",
						"xray:GetSamplingRules",
						"xray:GetSamplingTargets",
						"xray:GetSamplingStatisticSummaries",
					],
					resources: ["*"],
				},
			];
			args.transform = {
				function(args, opts, name) {
					args.tracingConfig = {
						mode: "Active",
					};
				},
			};
		});

		const apiB = new sst.aws.ApiGatewayV1("ApiB", {
			transform: {
				stage(args, opts, name) {
					args.xrayTracingEnabled = true;
				},
			},
		});

		apiB.route("POST /functionB", {
			handler: "src/functionB.handler",
			runtime: "nodejs22.x",
			timeout: "30 seconds",
			memory: "256 MB",
			logging: {
				retention: "1 day",
			},
		});

		apiB.deploy();

		const apiA = new sst.aws.ApiGatewayV1("ApiA", {
			transform: {
				stage(args, opts, name) {
					args.xrayTracingEnabled = true;
				},
			},
		});

		apiA.route("POST /functionA", {
			handler: "src/functionA.handler",
			runtime: "nodejs22.x",
			timeout: "30 seconds",
			memory: "256 MB",
			logging: {
				retention: "1 day",
			},
			environment: {
				API_URL: apiB.url,
			},
		});

		apiA.deploy();
	},
});
```

</details>

2. Make a request to the first API Gateway endpoint (i.e. `http POST https://api-id.execute-api.eu-west-1.amazonaws.com/stage/functionA userId=1`
3. Observe the traces

### Possible Solution

Our implementation of the `fetch` module diverges significantly from the [one in `aws-xray-sdk-node-fetch`](https://github.com/aws/aws-xray-sdk-node/blob/master/sdk_contrib/fetch/lib/fetch_p.js). Their implementation relies on monkey patching, which as far as I can tell would only work with CJS.

On our side we decided (#1619) to instead [use the `node:diagnostics_channel`](https://github.com/nodejs/undici/blob/main/docs/docs/api/DiagnosticsChannel.md) which is the recommended way to instrument requests made with `fetch`.

During the implementation, we followed the [types](https://github.com/nodejs/undici/blob/main/types/diagnostics-channel.d.ts) present in `@types/node` which suggest it's not possible to modify the `Request` object when instrumenting.

In reality however `request` object in the `message` has a [`addHeader()` method](https://github.com/nodejs/undici/blob/main/docs/docs/api/DiagnosticsChannel.md#undicirequestcreate) that can be used to forward the `X-Amzn-Trace-Id`.

To make this work, we should construct the header to include the `Root`, `Parent`, and `Sampled` fields and add it to the request. I made a PoC of this and I already got it working. The screenshot at the top of the issue comes from a request instrumented with Tracer using Fetch.

latest

22.x

### Packaging format used

npm

### Execution logs

```Shell

```

dreamorosi:
### Changes

This PR updates the implementation of the global fetch instrumentation in Tracer so that the `X-Amzn-Trace-Id` header is automatically populated and added to requests when Tracer is enabled.

This allows X-Ray to attribute the segment/span to the correct parent and build the correct service map for a request.

Prior to this PR the header was not added and this resulted in disjointed traces. This happened only when using the Fetch API, but not when using the `https` module - which is why this is considered a bug.

The linked issue contains more info on why we did not add this in the first place and what was missing.

**Issue number:** fixes #3467

---

### DIFF
--- a/packages/tracer/src/provider/ProviderService.ts
+++ b/packages/tracer/src/provider/ProviderService.ts
@@ -26,6 +26,7 @@ import http from 'node:http';
 import https from 'node:https';
 import { addUserAgentMiddleware } from '@aws-lambda-powertools/commons';
 import type { DiagnosticsChannel } from 'undici-types';
+import { environmentVariablesService } from '../config/EnvironmentVariablesService.js';
 import {
   findHeaderAndDecode,
   getRequestURL,
@@ -94,22 +95,6 @@ class ProviderService implements ProviderServiceInterface {
     return getSegment();
   }
 
-  private getRootTraceId(): Record<string, string> | undefined {
-    const xRayTraceEnv = process.env._X_AMZN_TRACE_ID || '';
-
-    if (!xRayTraceEnv.includes('=')) return { Root: xRayTraceEnv };
-
-    const xRayTraceData: Record<string, string> = {};
-
-    for (const field of xRayTraceEnv.split(';')) {
-      const [key, value] = field.split('=');
-
-      xRayTraceData[key] = value;
-    }
-
-    return xRayTraceData;
-  }
-
   /**
    * Instrument `fetch` requests with AWS X-Ray
    *
@@ -143,14 +128,12 @@ class ProviderService implements ProviderServiceInterface {
         );
         subsegment.addAttribute('namespace', 'remote');
 
+        // addHeader is not part of the type definition but it's available https://github.com/nodejs/undici/blob/main/docs/docs/api/DiagnosticsChannel.md#undicirequestcreate
         // @ts-expect-error
         request.addHeader(
           'X-Amzn-Trace-Id',
-          `Root=${this.getRootTraceId()?.Root};Parent=${subsegment.id};Sampled=${subsegment.notTraced ? '0' : '1'}`
+          `Root=${environmentVariablesService.getXrayTraceId()};Parent=${subsegment.id};Sampled=${subsegment.notTraced ? '0' : '1'}`
         );
-
-        // @ts-expect-error
-        request.addHeader('foo', 'bar');
 
         (subsegment as HttpSubsegment).http = {
           request: {
--- a/packages/tracer/src/provider/ProviderService.ts
+++ b/packages/tracer/src/provider/ProviderService.ts
@@ -94,6 +94,22 @@ class ProviderService implements ProviderServiceInterface {
     return getSegment();
   }
 
+  private getRootTraceId(): Record<string, string> | undefined {
+    const xRayTraceEnv = process.env._X_AMZN_TRACE_ID || '';
+
+    if (!xRayTraceEnv.includes('=')) return { Root: xRayTraceEnv };
+
+    const xRayTraceData: Record<string, string> = {};
+
+    for (const field of xRayTraceEnv.split(';')) {
+      const [key, value] = field.split('=');
+
+      xRayTraceData[key] = value;
+    }
+
+    return xRayTraceData;
+  }
+
   /**
    * Instrument `fetch` requests with AWS X-Ray
    *
@@ -126,6 +142,15 @@ class ProviderService implements ProviderServiceInterface {
           requestURL.hostname
         );
         subsegment.addAttribute('namespace', 'remote');
+
+        // @ts-expect-error
+        request.addHeader(
+          'X-Amzn-Trace-Id',
+          `Root=${this.getRootTraceId()?.Root};Parent=${subsegment.id};Sampled=${subsegment.notTraced ? '0' : '1'}`
+        );
+
+        // @ts-expect-error
+        request.addHeader('foo', 'bar');
 
         (subsegment as HttpSubsegment).http = {
           request: {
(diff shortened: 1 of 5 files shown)


packages/tracer/src/provider/ProviderService.ts
```diff
@@ -127,6 +128,13 @@ class ProviderService implements ProviderServiceInterface {
         );
         subsegment.addAttribute('namespace', 'remote');
 
+        // addHeader is not part of the type definition but it's available https://github.com/nodejs/undici/blob/main/docs/docs/api/DiagnosticsChannel.md#undicirequestcreate
+        // @ts-expect-error
+        request.addHeader(
+          'X-Amzn-Trace-Id',
+          `Root=${environmentVariablesService.getXrayTraceId()};Parent=${subsegment.id};Sampled=${subsegment.notTraced ? '0' : '1'}`
+        );
```
leandrodamascena, line 136: XRay can accept arbitrary fields as part of the header, some examples are `lineage` and `Self` for example, which are inserted as part of the Lambda and ALB mechanism respectively.

I'm not really sure if this is relevant here in most cases where you just want to instrument http calls, I'm also not sure if this will have any impact where you want to do end2end tracing, but I can imagine a scenario where you are invoking an HTTP endpoint in an APIGW and it might be interesting - if possible - to keep these fields.

References: 

https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/models/trace_header.py#L21
https://docs.aws.amazon.com/pdfs/xray/latest/devguide/xray-guide.pdf - page151 (Lineage)

leandrodamascena (changes requested):
Hi @dreamorosi! The TS implementations look good to me, the only thing is that it makes sense for us to check if it makes sense (and if it is possible) to keep the arbitrary fields.

packages/tracer/src/provider/ProviderService.ts
dreamorosi: As far as I can tell, neither the Python nor the Node.js X-Ray SDKs include any data into this header other than the Root trace ID, the entity ID, and the sampling decision.

These are the implementations in Node.js for the base `https` module ([link](https://github.com/aws/aws-xray-sdk-node/blob/d27bfd2fbcf93cac6b80b5490bbef06715f4fde6/packages/core/lib/patchers/http_p.js#L134-L135)) and the `fetch` module ([link](https://github.com/aws/aws-xray-sdk-node/blob/master/sdk_contrib/fetch/lib/fetch_p.js#L119-L122)); and this is the implementation in Python for the `request` module ([link](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/ext/util.py#L14)).

The main goal of this PR is to bring our implementation in line with what is being traced when using `https`, so I'm inclined to leave the header as-is.

If we know that X-Ray supports extra data in this scenario, and we have demand to add it, we can do so in a separate PR.

leandrodamascena: In the Python implementation, it considers the `data` field as an arbitrary field and includes it in the header.

But yes, we can wait for feedback from customers. 

I'm Resolving this conversation and approving the PR.

leandrodamascena (approved):

PR #3482, opened 2025-01-16
fix(parser): DynamoDBStream schema & envelope

### LINKED ISSUE #3481: Bug: DynamoDB Stream schema & envelope should unmarshall payloads
### Expected Behavior

Both the `DynamoDBStreamEnvelope` and the `DynamoDBStreamSchema` schemas should unmarshall the parts of the payload that include DynamoDB `AttributeValue`s into JavaScript objects.

### Current Behavior

Currently they leave the payload as-is, which leaves it to the customer to make the conversion in their code.

### Code snippet

N/A

### Steps to Reproduce

N/A

### Possible Solution

_No response_

latest

22.x

### Packaging format used

npm

### Execution logs

```Shell

```

dreamorosi:
### Changes

This PR updates the logic of the `DynamoDBStreamSchema` and `DynamoDBStreamEnvelope` built-in schemas so that they automatically handle DynamoDB attribute values for the customer and convert them to JavaScript native objects during the parsing and validation.

As part of the PR, I have also improved the error messages in the parser to include the index of the record being processed as well as the full path of the issue when parsing fails. For example, now the error would be something like:

```ts
new ParseError('Failed to parse DynamoDB Stream envelope', {
  cause: new ZodError([
    {
      code: 'invalid_type',
      expected: 'object',
      received: 'undefined',
      path: ['Records', 0, 'dynamodb'],
      message: 'Required',
    }
  ]),
})
```

and like this for failed parsing of `Keys`, `NewImage`, and `OldImage`:

```ts
new ParseError('Failed to parse DynamoDB Stream envelope', {
  cause: new ZodError([
    {
      code: 'invalid_type',
      expected: 'number',
      received: 'object',
      path: ['Records', 0, 'dynamodb', 'NewImage'],
      message: 'Expected number, received object',
    }
  ]),
})
```

Likewise, when parsing an envelope using `safeParse`, the errors can also be combined when multiple records have issues:

```ts
new ParseError('Failed to parse records at indexes 0, 1', {
    cause: new ZodError([
      {
        code: 'invalid_type',
        expected: 'string',
        received: 'undefined',
        path: ['Records', 0, 'dynamodb', 'NewImage', 'Message'],
        message: 'Required',
      },
      {
        code: 'invalid_type',
        expected: 'string',
        received: 'number',
        path: ['Records', 1, 'dynamodb', 'NewImage', 'Message'],
        message: 'Expected string, received number',
      },
    ]),
  }),
  originalEvent: event,
})
```

Note that in all cases the full path of the object is preserved even though we are performing multiple Zod validations under the hood.

While working on this PR, I also brought into the `commons` package an implementation of the unmarshall function inspired by the one in the AWS SDK. This is similar to what Powertools for AWS Lambda (Python) does, and allows us to provide this behavior without forcing an extra dependency. Our implementation also differs from the reference one in a couple minor ways since we don't have to support old Node.js versions.

This new behavior is compatible with the `DynamoDBHelper` introduced in the previous release. This is because in order for customers to use the helper they will have to extend & replace the schema, thus bypassing the transform introduced in this PR.

Finally, we're treating this change as a fix because this should've been the default behavior all along.

**Issue number:** fixes #3481

---

### DIFF
--- a/packages/parser/src/schemas/dynamodb.ts
+++ b/packages/parser/src/schemas/dynamodb.ts
@@ -1,19 +1,67 @@
+import { unmarshallDynamoDB } from '@aws-lambda-powertools/commons/utils/unmarshallDynamoDB';
 import { z } from 'zod';
 
-const DynamoDBStreamChangeRecord = z.object({
-  ApproximateCreationDateTime: z.number().optional(),
-  Keys: z.record(z.string(), z.record(z.string(), z.any())),
-  NewImage: z.record(z.string(), z.any()).optional(),
-  OldImage: z.record(z.string(), z.any()).optional(),
-  SequenceNumber: z.string(),
-  SizeBytes: z.number(),
-  StreamViewType: z.enum([
-    'NEW_IMAGE',
-    'OLD_IMAGE',
-    'NEW_AND_OLD_IMAGES',
-    'KEYS_ONLY',
-  ]),
-});
+const DynamoDBStreamChangeRecord = z
+  .object({
+    ApproximateCreationDateTime: z.number().optional(),
+    Keys: z.record(z.string(), z.record(z.string(), z.any())),
+    NewImage: z.record(z.string(), z.any()).optional(),
+    OldImage: z.record(z.string(), z.any()).optional(),
+    SequenceNumber: z.string(),
+    SizeBytes: z.number(),
+    StreamViewType: z.enum([
+      'NEW_IMAGE',
+      'OLD_IMAGE',
+      'NEW_AND_OLD_IMAGES',
+      'KEYS_ONLY',
+    ]),
+  })
+  .transform((object, ctx) => {
+    const result = { ...object };
+
+    const unmarshallAttributeValue = (
+      imageName: 'NewImage' | 'OldImage' | 'Keys',
+      image: Record<string, unknown>
+    ) => {
+      try {
+        // @ts-expect-error
+        return unmarshallDynamoDB(image) as Record<string, unknown>;
+      } catch (err) {
+        ctx.addIssue({
+          code: 'custom',
+          message: `Could not unmarshall ${imageName} in DynamoDB stream record`,
+          fatal: true,
+          path: [imageName],
+        });
+        return z.NEVER;
+      }
+    };
+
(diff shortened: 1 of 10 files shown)


packages/parser/src/schemas/dynamodb.ts
```diff
@@ -35,10 +83,25 @@ const DynamoDBStreamToKinesisRecord = DynamoDBStreamRecord.extend({
...
+      NewImage: z.record(z.string(), z.any()).optional(),
+      OldImage: z.record(z.string(), z.any()).optional(),
+      SequenceNumber: z.string(),
+      SizeBytes: z.number(),
+      StreamViewType: z.enum([
+        'NEW_IMAGE',
+        'OLD_IMAGE',
+        'NEW_AND_OLD_IMAGES',
+        'KEYS_ONLY',
+      ]),
```
am29d, line 99: Can we factor this duplication?

packages/parser/tests/unit/envelopes/dynamodb.test.ts
```diff
@@ -1,141 +1,206 @@
...
-        DynamoDBStreamEnvelope.parse({ foo: 'bar' }, schema)
-      ).toThrow();
+        DynamoDBStreamEnvelope.parse(
+          event,
+          z
+            .object({
+              Message: z.string(),
+            })
+            .strict()
+        )
```
am29d, line 35: Is that default biome formatting?

am29d (changes requested):
Great PR, thanks for the work bringing the unmarschalling into envelope and schema! 

Just a minor code duplication we'd need to fix, otherwise ready to go.

packages/parser/tests/unit/envelopes/dynamodb.test.ts
dreamorosi: Yep

dreamorosi: I tried again to put it better, and it put it back when formatting.

packages/parser/src/schemas/dynamodb.ts
dreamorosi: Yes good point, I forgot.

am29d (approved):
Beautiful!

PR #3485, opened 2025-01-17
fix(parser): identitySource can be null in APIGatewayRequestAuthorizerEventV2Schema

### LINKED ISSUE #3483: Bug: `APIGatewayRequestAuthorizerEventV2Schema` has invalid definition for the `identitySource` field
### Expected Behavior

The `identitySource` field in `APIGatewayRequestAuthorizerEventV2` can be null when the Lambda authorizer is not configured with an identity source (see screenshot from the AWS Console below).

Image

### Current Behavior

The current definition does not account for this scenario, leading to an error:

https://github.com/aws-powertools/powertools-lambda-typescript/blob/e1bc6c5e324d8e76fdde043d962d9c1c42e8c0c3/packages/parser/src/schemas/apigwv2.ts#L233

Image

Even the `@types/aws-lambda` package has invalid type definition:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/01bf6ea3ae73629f4d488d777df91781e466ba15/types/aws-lambda/trigger/api-gateway-authorizer.d.ts#L38

### Code snippet

```typescript
APIGatewayRequestAuthorizerEventV2Schema.parse(event);
```

### Steps to Reproduce

When creating a Lambda authorizer, do not specify an identity source.

### Possible Solution

add `.nullable()`

latest

20.x

### Packaging format used

npm

### Execution logs

```Shell

```

am29d:
## Summary

This PR allows to parse `identitySource` with null value in `APIGatewayRequestAuthorizerEventV2Schema`. 

### Changes

**Issue number:** closes #3483 

---

### DIFF
--- a/packages/parser/tests/unit/schema/apigwv2.test.ts
+++ b/packages/parser/tests/unit/schema/apigwv2.test.ts
@@ -98,6 +98,21 @@ describe('API Gateway HTTP (v2) Schemas', () => {
       // Assess
       expect(parsedEvent).toEqual(event);
     });
+
+    it('should parse the authorizer event with null identitySource', () => {
+      // Prepare
+      const event = getTestEvent({
+        eventsPath,
+        filename: 'authorizer-request',
+      });
+      event.identitySource = null;
+
+      // Act
+      const parsedEvent = APIGatewayRequestAuthorizerEventV2Schema.parse(event);
+
+      // Assess
+      expect(parsedEvent).toEqual(event);
+    });
   });
 
   describe('APIGatewayRequestContextV2Schema', () => {
--- a/packages/parser/src/schemas/apigwv2.ts
+++ b/packages/parser/src/schemas/apigwv2.ts
@@ -230,7 +230,7 @@ const APIGatewayRequestAuthorizerEventV2Schema = z.object({
   version: z.literal('2.0'),
   type: z.literal('REQUEST'),
   routeArn: z.string(),
-  identitySource: APIGatewayStringArray,
+  identitySource: APIGatewayStringArray.nullish(),
   routeKey: z.string(),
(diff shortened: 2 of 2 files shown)


packages/parser/tests/unit/schema/apigwv2.test.ts
```diff
@@ -98,6 +98,21 @@ describe('API Gateway HTTP (v2) Schemas', () => {
...
       expect(parsedEvent).toEqual(event);
     });
+
+    it('should parse the authorizer event with null identitySource', () => {
+      // Prepare
+      const event = getTestEvent({
+        eventsPath,
+        filename: 'authorizer-request',
+      });
+      event.identitySource = null;
```
dreamorosi, line 108: Since we have established that when not present `identitySource` is actually `null` as opposed to not being present at all, I think we should instead modify the test payload in `packages/parser/tests/events/apigw-rest/authorizer-request.json` and add the field there.

This way we can avoid adding an extra test and use the one at L88 which tests the same payload. With the field set to `.nullish()` both cases are covered by that case.

dreamorosi (changes requested):
I think we can avoid adding a test by modifying an existing one. Other than that, ready to merge.

packages/parser/tests/unit/schema/apigwv2.test.ts
am29d: `authorizer-request.json` has an array of strings as value, so we have two values we need to test, `null` and array of strings.

dreamorosi: You are right, I was looking at the REST folder.

dreamorosi (approved):
Thank you for fixing this!

Issue #3503, opened 2025-01-20
Bug: EventBridge envelope reports incorrect path


dreamorosi:
### Expected Behavior

When using the `EventBridgeEnvelope` to parse an event, the error returned by the envelope should contain the correct path so that customers can easily find the root cause of the issue.

### Current Behavior

This is the error emitted:

```
ParseError: Failed to parse envelope. This error was caused by: [
  {
    "code": "invalid_type",
    "expected": "string",
    "received": "undefined",
    "path": [
      "name"
    ],
    "message": "Required"
  }
].
```

The correct path should be `["detail", "name"]` since the validation applies to the `detail` field.

### Code snippet

```json
{
  "version": "0",
  "id": "6a7e8feb-b491-4cf7-a9f1-bf3703467718",
  "detail-type": "EC2 Instance State-change Notification",
  "source": "aws.ec2",
  "account": "111122223333",
  "time": "2017-12-22T18:43:48Z",
  "region": "us-west-1",
  "resources": [
    "arn:aws:ec2:us-west-1:123456789012:instance/i-1234567890abcdef0"
  ],
  "detail": {
    "instance_id": "i-1234567890abcdef0",
    "state": "terminated"
  },
  "replay-name": "replay_archive"
}
```

and using the following code:

```ts
const schema = z.object({
  'name': z.string()
});

const parsed = EventBridgeEnvelope.parse(event, schema);
```

### Steps to Reproduce

N/A

### Powertools for AWS Lambda (TypeScript) version

latest

### AWS Lambda function runtime

22.x

### Packaging format used

npm

### Execution logs

```Shell

```

PR #3504, opened 2025-01-20
fix(parser): EventBridge envelope uses correct path

dreamorosi:
### Changes

This PR fixes the issue described in the linked issue by making it so that the `EventBridgeEnvelope` throws an error that includes the correct path to the issue. This will allow customers to more quickly identify the validation/parsing issue.

To implement the change, and since we are not doing any special transform or handling of the payload in the `detail` field, I also refactored the envelope implementation so that the schema provided by the customer to validate the payload is used to extend the base `EventBridgeSchema` event. This simplifies the implementation and lets Zod do the heavy lifting.

While working on this PR I also removed some unused test events that were leftover from previous refactors.

**Issue number:** #3503

---


am29d (approved):
This looks clean, I like it! Buckel up abstract `Envelope`, we probably going to remove you entirely.

PR #3511, opened 2025-01-22
fix(parser): API Gateway Envelopes handle non-JSON

### LINKED ISSUE #3512: Bug: API Gateway envelopes assumes message body is always a JSON
### Expected Behavior

When working with the `ApiGatewayEnvelope` and `ApiGatewayV2Envelope` I should be able to extract the message body regardless of whether it's a plain text string, base64-encoded string, or a JSON-stringified one.

### Current Behavior

When the message body is not a JSON-stringified object the envelope throws a runtime error as it's unable to parse it with `JSON.parse()`.

### Code snippet

```ts
import { z } from 'zod';
import { ApiGatewayEnvelope } from '@aws-lambda-powertools/parser/envelopes';

  try {
    ApiGatewayEnvelope.parse(event, z.string());
  } catch (error) {
    console.error(error);
  }
}
```

### Steps to Reproduce

Send a request with a body that is either plain text or an encoded binary.

### Possible Solution

_No response_

latest

22.x

### Packaging format used

npm

### Execution logs

```Shell

```

dreamorosi:
### Changes

This PR refactors the API Gateway (REST) & API Gateway v2 (HTTP) envelopes so that they don't force the transform of the `body` field and thus can handle plain text and base64 binary-encoded bodies.

The body is now left as a string and it's up to the caller to set the appropriate transform in the Zod schema passed to the envelope, for example to 

```ts
import { z } from 'zod';

const schema = z.object({
  message: z.string()
});

ApiGatewayEnvelope.parse(event, JSONStringified(schema));
```

If there's demand for it, in the future we can consider adding an helper for base64 decoding.

Additionally I standardized the error structure by always throwing a `ParseError` with the original error as cause:

```ts
new ParseError('Failed to parse API Gateway body', {
  cause: new ZodError([
    {
      code: 'invalid_type',
      expected: 'string',
      received: 'undefined',
      path: ['path'],
      message: 'Required',
    },
    {
      code: 'invalid_type',
      expected: 'object',
      received: 'null',
      path: ['body'],
      message: 'Expected object, received null',
    },
  ]),
})
```

**Issue number:** closes #3512

---

### DIFF
--- a/packages/parser/src/envelopes/apigw.ts
+++ b/packages/parser/src/envelopes/apigw.ts
@@ -2,7 +2,7 @@ import type { ZodSchema, z } from 'zod';
 import { ParseError } from '../errors.js';
 import { APIGatewayProxyEventSchema } from '../schemas/apigw.js';
 import type { ParsedResult } from '../types/parser.js';
-import { Envelope, envelopeDiscriminator } from './envelope.js';
+import { envelopeDiscriminator } from './envelope.js';
 
 /**
  * API Gateway envelope to extract data within body key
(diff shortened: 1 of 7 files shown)


am29d (approved):

am29d (approved):
Thanks for PR, and making the tests more explicit with ParseErrors.

PR #3513, opened 2025-01-22
fix(parser): allow SQS envelopes to handle non-JSON strings

### LINKED ISSUE #3510: Bug: SQS Envelope assumes message body is always a JSON
### Expected Behavior

When working with the `SqsEnvelope` I should be able to extract the message body regardless of whether it's a plain text string or a JSON-stringified one.

### Current Behavior

When the message body is not a JSON-stringified object the envelope throws a runtime error as it's unable to parse it with `JSON.parse()`.

### Code snippet

```ts
import { z } from 'zod';
import { SqsEnvelope } from '@aws-lambda-powertools/parser/envelopes';

  try {
    SqsEnvelope.parse(event, z.string());
  } catch (error) {
    console.error(error);
  }
}
```

### Steps to Reproduce

Send a message like this:

```
aws sqs send-message \           
  --queue-url https://sqs.eu-west-1.amazonaws.com/123456789012/sqs-parser-MyQueue \
  --message-body 'hello'
```

Observe the logs:

```
 ParseError: Failed to parse envelope. This error was caused by:
 Unexpected token 'h', "hello" is not valid JSON.
```

### Possible Solution

We should avoid parsing by default and instead leave it to the caller to decide.

While working on this, we should also check the SQS Schema and the SNS + SQS Envelope, since they share some of the code.

Additionally, we should ensure the schema fails parsing when the `Records` array is empty, as mentioned in #3475.

latest

22.x

### Packaging format used

npm

### Execution logs

```Shell

```

dreamorosi:
### Changes

This PR updates the `SqsEnvelope` and `SnsSqsEnvelope` so that they can handle messages other than JSON strings, i.e. plain text strings like `"hello world"`.

The PR is still WIP and I will update the description once it's done.

To complete this PR I also had to refactor the Middy.js middleware unit tests since they were using the SQS envelope I was touching. Because of this I took the opportunity to simplify & standardize them so that they follow the same conventions as the rest of the project.

**Issue number:** fixes #3510

---


packages/parser/src/envelopes/snssqs.ts
```diff
@@ -0,0 +1,185 @@
...
+        );
+      } catch (error) {
+        throw new ParseError(
+          `Failed to parse SQS Record at index ${recordIndex}`,
+          {
+            cause: new ZodError(
+              error instanceof ZodError
+                ? (error as ZodError).issues.map((issue) => ({
+                    ...issue,
+                    path: ['Records', recordIndex, 'body', ...issue.path],
```
am29d, line 82: Now that we have `createError` helper, do you think it'd make sense to use it here, and add later to other envelopes? We probably need ot pass the 2nd parameter as input, as it's often different. What do you think?

am29d (commented):
Thanks for the PR! Just a minor question on reusing `createError` helper.

packages/parser/src/envelopes/snssqs.ts
dreamorosi: Yea, possibly.

I think let's go through all the envelopes and then we see where we can simplify / extract further. Using this function removes also the `// @ts-expect-error` since it's now properly typed.

I'd rather complete the entire refactor first though, as I don't want to backtrack before getting to the end. Ideally I'd like to make it to the next release with most/all the envelopes bugs fixed - then we can refactor and optimize.

leandrodamascena (approved):

Issue #3530, opened 2025-01-25
Bug: Kinesis envelopes assume data is always a JSON


dreamorosi:
### Expected Behavior

When working with Kinesis-related envelopes I should be able to extract the data regardless of whether it's a plain text string or a JSON-stringified one. This should work on top of the decoding already happening.

### Current Behavior

Currently both envelopes treat the body as a JSON string and try to parse it.

### Code snippet

N/A

### Steps to Reproduce

N/A

### Powertools for AWS Lambda (TypeScript) version

latest

### AWS Lambda function runtime

22.x

### Packaging format used

npm

### Execution logs

```Shell

```

PR #3531, opened 2025-01-25
fix(parser): allow Kinesis envelopes to handle non-JSON strings

dreamorosi:
### Changes

This PR updates the implementation of both the Kinesis Data Stream and Kinesis Firehose envelopes so that they don't make assumptions about the contents of the data and instead let the customer apply transforms via the schema provided.

The PR also exports the `DynamoDBStreamToKinesisChangeRecord` so that customers can extend the schema for events coming from Kinesis Data Streams that contain DynamoDB Stream events.

**Issue number:** #3530

---


am29d (approved):
Thanks for the PR, another bug closed!

PR #3534, opened 2025-01-27
fix(parser): allow VPC envelopes to handle non-JSON strings

### LINKED ISSUE #3533: Bug: VPC Lattice envelopes assume data is always a JSON
### Expected Behavior

When working with VPC Lattice-related envelopes I should be able to extract the data regardless of whether it's a plain text string or a JSON-stringified one. This should work on top of the decoding already happening.

### Current Behavior

Currently both envelopes treat the body as a JSON string and try to parse it, which means if the body is either a plain text or base64-encoded string the parsing will fail.

### Code snippet

N/A

### Steps to Reproduce

N/A

### Possible Solution

_No response_

latest

22.x

### Packaging format used

npm

### Execution logs

```Shell

```

dreamorosi:
### Changes

This PR updates the implementation of both the VPC Lattice v1 & v2 envelopes so that they don't try transform the body of the request as JSON without the customer explicitly passing a schema that does so.

This allows customers to handle plain text or base64-encoded as well.

In this PR I also removed the `@anatine/zod-mock` dependency, moved the `packages/parser/tests/schemas/util.js` file to its own folder at `packages/parser/tests/helpers/util.js`, and removed the `Envelope` implementation since this was the last set of built-in envelopes using it. These three changes account for 90% of the diff.

Also with this PR we should be done with the major refactoring of the Parser envelopes. There are still a few things we should address like improving the docstrings of some envelopes/schemas, improving the docs of the utility, and creating sub-path exports for each set of envelopes like we did for the built-in schemas - but we'll do this in the next releases.

**Issue number:** fixes #3533

---


packages/parser/src/envelopes/envelope.ts
```diff
@@ -1,82 +1,7 @@
...
-      };
-    }
-  },
-};
-
 /**
  * This is a discriminator to differentiate whether an envelope returns an array or an object
  * @hidden
  */
 const envelopeDiscriminator = Symbol.for('returnType');
```
am29d, line 5: where do put this thing now? utils?

am29d (approved):
Well done, thanks a lot for the moving the bug bash forward so quickly!

PR #3549, opened 2025-02-03
chore: update `actions/cached-node-modules` GitHub Action

### LINKED ISSUE #3548: Maintenance: update `actions/cached-node-modules` GitHub Action
### Summary

GitHub is updating the architecture behind the GitHub Actions cache, in doing so they are also deprecating some versions of the `actions/cache` GitHub Action. One of our custom actions (`aws-powertools/actions/.github/actions/cached-node-modules`) uses a deprecated version which will stop working in the coming weeks.

I have already updated the action in the `aws-powertools/actions` repo, we now need to update the workflows on this side to use the new version.

### Why is this needed?

So that continue using the GitHub Action.

Automation

### Solution

_No response_

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR updates the hash for the `actions/cached-node-modules` GitHub Actions to use the latest version. This will allow us to continue using the composite action, which now uses a more recent version of `actions/cached`.

**Issue number:** closes #3548

---

### DIFF
--- a/.github/workflows/make-release.yml
+++ b/.github/workflows/make-release.yml
@@ -57,7 +57,7 @@ jobs:
         run: |
           npm set "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}"
       - name: Setup dependencies
-        uses: aws-powertools/actions/.github/actions/cached-node-modules@d406bac5563f1d8c793519a3eedfe620f6a14872
+        uses: aws-powertools/actions/.github/actions/cached-node-modules@29979bc5339bf54f76a11ac36ff67701986bb0f0
       - name: Publish to npm
         run: |
           NPM_CONFIG_PROVENANCE=true npx lerna publish from-package --git-head ${{ github.sha }} --yes 
--- a/.github/workflows/make-version.yml
+++ b/.github/workflows/make-version.yml
@@ -28,7 +28,7 @@ jobs:
           node-version: "22"
           cache: "npm"
       - name: Setup dependencies
-        uses: aws-powertools/actions/.github/actions/cached-node-modules@d406bac5563f1d8c793519a3eedfe620f6a14872
+        uses: aws-powertools/actions/.github/actions/cached-node-modules@29979bc5339bf54f76a11ac36ff67701986bb0f0
       - name: Version
         id: bump-version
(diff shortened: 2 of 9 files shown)


am29d (approved):

dreamorosi:
@leandrodamascena - why the `do-not-merge` label?

leandrodamascena:
Before merging, I have a question that I don't know the answer to and I probably need to adopt this in Python as well.

I see that you are adding an action from `aws-powertools/actions` with sha hash, so does this mean that if you update this action in the source repository, dependabot will be able to bump this version here OR this sha if it is from the `main` branch?

dreamorosi:
> Before merging, I have a question that I don't know the answer to and I probably need to adopt this in Python as well.
> 
> I see that you are adding an action from `aws-powertools/actions` with sha hash, so does this mean that if you update this action in the source repository, dependabot will be able to bump this version here OR this sha if it is from the `main` branch?

It's a hash from the `main` branch, making changes to the repo does not change the source of this action, Dependabot won't pick it up even if we make changes because it requires a release hash, which we don't do.

I've changed this action 6mo ago the last time, and I don't expect to change it for a long time.

dreamorosi:
Integration tests on this branch were [passing](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/13109918465).

leandrodamascena:
> > Before merging, I have a question that I don't know the answer to and I probably need to adopt this in Python as well.
> 
> > 
> 
> > I see that you are adding an action from `aws-powertools/actions` with sha hash, so does this mean that if you update this action in the source repository, dependabot will be able to bump this version here OR this sha if it is from the `main` branch?
> 
> 
> 
> It's a hash from the `main` branch, making changes to the repo does not change the source of this action, Dependabot won't pick it up even if we make changes because it requires a release hash, which we don't do.
> 
> 

Thanks for the explanation.

dreamorosi:
If we add more actions to that repo, which was the point of having it in the first place, we should probably set up actual releases and leverage Dependabot.

The thing I'm unsure about is how to release multiple packages from the same repo, but we can look into it if needed.

leandrodamascena:
> 
> The thing I'm unsure about is how to release multiple packages from the same repo, but we can look into it if needed.

I'm not sure I understand the question. But if you're talking about releasing multiple actions, I think if we just create a tag and GH release it, dependabot will work.

PR #3564, opened 2025-02-06
improv(parser): allow extending base provider with no AWS SDK client

### LINKED ISSUE #2400: Feature request: allow `BaseProvider` in Parameters to not have a client
### Use case

When creating a custom Parameters provider I should be able to seamless set my own backend, regardless of whether this might be AWS, a 3rd party, or anything else.

Currently the `BaseProvider` expects a [constructor parameter](https://github.com/aws-powertools/powertools-lambda-typescript/blob/2a69d41157a3c51055f1e2a30a95650c4500d5d1/packages/parameters/src/base/BaseProvider.ts#L51) called `proto` (as-in prototype, bad name) that you're supposed to pass and that will be used whenever you are not passing an AWS SDK client.

This implementation is too tightly coupled with the notion that `BaseProvider` expects to be extended by a provider that relies on an AWS SDK client. This is not always the case, and as a customer implementing my own provider I should not have to work around this limitation.

### Solution/User Experience

For end customers using the Parameters utility as-is, the experience should stay the same.

For those other customers who implement their own provider, the implementation should be simpler and not require to pass a class prototype to be instantiated as client.

### Alternative solutions

_No response_

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR improves slightly the experience in Parameters by allowing customers to create custom providers without having to specify a no-op parameter.

In addition to doing this, I also renamed the parameter to `awsSdkV3ClientPrototype` for added clarity, and improved the documentation around it.

**Issue number:** closes #2400

---

### DIFF
--- a/packages/parameters/src/base/BaseProvider.ts
+++ b/packages/parameters/src/base/BaseProvider.ts
@@ -8,6 +8,7 @@ import {
 import { EnvironmentVariablesService } from '../config/EnvironmentVariablesService.js';
 import { GetParameterError, TransformParameterError } from '../errors.js';
 import type {
+  BaseProviderConstructorOptions,
   BaseProviderInterface,
   GetMultipleOptionsInterface,
   GetOptionsInterface,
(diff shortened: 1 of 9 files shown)


packages/parameters/src/base/BaseProvider.ts
```diff
@@ -44,27 +45,25 @@ abstract class BaseProvider implements BaseProviderInterface {
...
-    clientConfig?: unknown;
-    proto: new (config?: unknown) => unknown;
-  }) {
+    awsSdkV3ClientPrototype,
+  }: BaseProviderConstructorOptions) {
     this.store = new Map();
     this.envVarsService = new EnvironmentVariablesService();
     if (awsSdkV3Client) {
-      if (!isSdkClient(awsSdkV3Client)) {
+      if (!isSdkClient(awsSdkV3Client) && awsSdkV3ClientPrototype) {
```
leandrodamascena, line 53: Is there a chance that `awsSdkV3Client` is v2 and the caller doesn't pass the `awsSdkV3ClientPrototype` parameter to this constructor? If so, this will end the upgrade using a v2 client on line 59, which isn't desired, no?

I can be wrong and this could be being passed somewhere in the code, but I didn't see it.

leandrodamascena (changes requested):
Hi @dreamorosi! Just a comment that I think makes sense to read before approving the PR.

packages/parameters/src/base/BaseProvider.ts
dreamorosi: We don't support v2 client at all

leandrodamascena (approved):

PR #3568, opened 2025-02-07
feat(parser): simplify `ParseResult` and `parse` inference

### LINKED ISSUE #3096: Feature request: infer `ParsedResult` types from inputs of parse function
### Use case

When using the Parser with the `safeParse` option enabled, the result is a `ParsedResult` type that can contain either the parsed payload or an error depending on the result of the parsing.

Today, customers need to specify two types: one for the input type, usually the one of the envelope, and an output type, either the schema or the transformed type.

This is ok when using envelopes, but when parsing a schema directly the two types can be redundant and/or confusing.

I don't know if it's possible, but we could explore the possibility of making the second type optional when typing the `ParsedResult` or even making it optional and infer the types from the decorator usage.

## Before

```ts
import { parser } from '@aws-lambda-powertools/parser';
import { APIGatewayProxyEventV2Schema } from '@aws-lambda-powertools/parser/schemas/api-gatewayv2';
import type {
  ParsedResult,
  APIGatewayProxyEventV2,
} from '@aws-lambda-powertools/parser/types';

class Lambda {
  @parser({ schema: APIGatewayProxyEventV2Schema, safeParse: true })
  public async handler(
    event: ParsedResult<APIGatewayProxyEventV2, APIGatewayProxyEventV2>,
    _context: Context
  ) {
    if (event.success) {
      console.log(event.data); // this is APIGatewayProxyEventV2
    } else {
      // event.error is now strongly typed
      console.error(event.error);
      // event.originalEvent is now strongly typed
      console.error(event.originalEvent);
    }
  }
}

export const handler = new Lambda().handler;
```

## After

```diff
class Lambda {
  @parser({ schema: APIGatewayProxyEventV2Schema, safeParse: true })
  public async handler(
-    event: ParsedResult<unknown, APIGatewayProxyEventV2>,
+    event: ParsedResult<APIGatewayProxyEventV2>,
    _context: Context
  ) {
    if (event.success) {
      console.log(event.data); // this is APIGatewayProxyEventV2
    } else {
      // event.error is now strongly typed
      console.error(event.error);
      // event.originalEvent is now strongly typed
      console.error(event.originalEvent);
    }
  }
}
```

Especially in cases where we don't use an envelope, the `Output` type is always the `z.infer<typeof XXXSchema>`.

### Alternative solutions

```Markdown
If this is not feasible, I would suggest refactoring the `ParsedResult` type which today looks like this:


type ParsedResult<Input = unknown, Output = unknown> =
  | ParsedResultSuccess<Output>
  | ParsedResultError<Input>;


To switch the order and make it more obvious what is what, i.e.

```ts
type ParsedResult<PayloadSchema = unknown, EnvelopeSchema = unknown> =
  | ParsedResultSuccess<PayloadSchema>
  | ParsedResultError< EnvelopeSchema>;
```

So that one can just specify the first one when not using envelopes. 

This solution however might be counterintuitive because we are now reverting the order of the parameters. The good thing about the current solution is that logically input comes before the output.
```


### Acknowledgment

### Future readers

am29d:
### Changes

This PR simplifies `ParseResult` type usage. Previously you would need to specify input and output types, which were redundant when not using envelopes. 

This change also required type narrowing for the `parse` function. 

**Issue number:** closes #3096 

---


dreamorosi:
@am29d can I convert it back to draft until the CI is green?

am29d:
> @am29d can I convert it back to draft until the CI is green?

Let me fix it.

packages/parser/src/types/index.ts
dreamorosi, line 1: What happened here?

am29d: Looks like a weird sort

dreamorosi (approved):
Great work, I tested this locally and it works perfectly 🔥 

Thank you for fixing this!

PR #3576, opened 2025-02-10
chore(ci): remove pinned actions & pkg versions

### LINKED ISSUE #3573: Maintenance: update bootstrap region automation to remove pinned/aliased dependencies
### Summary

Some of the changes introduced in #3438 lowered our OpenSSF Scorecard due to both pinned dependencies and actions using the `latest` tag instead of a commit hash.

Specifically, the two are:
- the shared GitHub Action [here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/4e9ff0717223f8305eda684608ce1435489b9749/.github/workflows/bootstrap_region.yml#L93)
- the CDK version [here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/4e9ff0717223f8305eda684608ce1435489b9749/.github/workflows/bootstrap_region.yml#L60-L60)

Which correspond to these two alerts:
- https://github.com/aws-powertools/powertools-lambda-typescript/security/code-scanning/77
- https://github.com/aws-powertools/powertools-lambda-typescript/security/code-scanning/76

Note that neither of them involves code that we ship to customers, this is only automation used for us to bootstrap new regions we use to deploy resources.

### Why is this needed?

So we can restore our OpenSSF Scorecard score.

Other

### Solution

The first should use a commit hash, similar to what we do here.

The second one should use the CDK version already present in the monorepo rather than install a different one. 

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR updates the `bootstrap_region.yml` workflow with three changes:
- The `copy_layers` step is now dependant on the `bootstrap` one
- The  `bootstrap` step checks out the repo and uses the CDK version present - this avoids pinning the CDK version
- The `copy_layers` step uses a specific version of the shared action identified by its commit hash, rather than the `latest` one

**Issue number:** closes #3573

---

### DIFF
--- a/.github/workflows/bootstrap_region.yml
+++ b/.github/workflows/bootstrap_region.yml
@@ -63,12 +63,12 @@ jobs:
         name: Create Workdir
         run: |
           mkdir -p build/project
-          - id: cdk-project
-          name: CDK Project
-          working-directory: build/project
-          run: |
-            npx cdk init app --language=typescript
-            AWS_REGION="${{ inputs.region }}" npx cdk bootstrap
+      - id: cdk-project
+        name: CDK Project
+        working-directory: build/project
+        run: |
+          npx cdk init app --language=typescript
+          AWS_REGION="${{ inputs.region }}" npx cdk bootstrap
 
   copy_layers:
     name: Copy Layers
--- a/.github/workflows/bootstrap_region.yml
+++ b/.github/workflows/bootstrap_region.yml
@@ -38,7 +38,6 @@ jobs:
     name: Bootstrap Region
     runs-on: ubuntu-latest
     permissions:
-      contents: write
       id-token: write
     environment: layer-${{ inputs.environment }}
     steps:
@@ -75,7 +74,6 @@ jobs:
     runs-on: ubuntu-latest
     needs: bootstrap
     permissions:
-      contents: write
       id-token: write
     environment: layer-${{ inputs.environment }}
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/bootstrap_region.yml
```diff
@@ -34,14 +34,24 @@ permissions:
   contents: read
 
 jobs:
-  cdk:
-    name: Install CDK
+  bootstrap:
+    name: Bootstrap Region
     runs-on: ubuntu-latest
     permissions:
       contents: write
```
leandrodamascena, line 41: Please correct if I am wrong, but you are not pushing commits or creating PR in this action, why do you need `contents: write` here? It will lower openssf scorecard.

.github/workflows/bootstrap_region.yml
```diff
@@ -68,6 +73,7 @@ jobs:
   copy_layers:
     name: Copy Layers
     runs-on: ubuntu-latest
+    needs: bootstrap
     permissions:
       contents: write
```
leandrodamascena, line 78: Same here.

leandrodamascena (changes requested):
Please check my comments.

.github/workflows/bootstrap_region.yml
dreamorosi: Good point - @sthulb can we remove it?

dreamorosi: Removed in both

dreamorosi:
Removed the `content: write` from both jobs, thank you for noticing!

leandrodamascena (approved):

PR #3593, opened 2025-02-12
feat(logger): add circular buffer

### LINKED ISSUE #3589: Feature request: add circular buffer to hold logs
### Use case

As part of the implementation of the log buffering feature for Logger we need to implement logic that allows the Logger to add log entries to a buffer and later retrieve them. This buffer logic should be able to hold one or more buffers of logs with each buffer being tied to a request.

For example, assuming the following code:

```ts
const logger = new Logger({
  buffering: {
    maxBytes: 1024
  }
});

export const handler = (event: { userId: string }) => {
  logger.debug('hello', { userId: event.userId });
}
```

The following two requests:
1. `_X_AMZN_TRACE_ID=abcd`; event `{ userId: 'Alice' }`
2. `_X_AMZN_TRACE_ID=efgh`; event `{ userId: 'Bob' }`

Should allow the Logger class to create two buffers, one for each request, in which to store the respective log of each request. Logs of different requests should never end up in the same buffer:

1. `['abcd', [{ userId: 'Alice' }]]`
2. `['efgh', [{ userId: 'Bob' }]]`

For this part of the implementation we will focus on the data structures that will allow us to create and hold these buffers, how the logs are added to the buffers and what identifies requests are points that will be tackled in separate issues.

For a full overview of the feature, please refer to this RFC (#3410) and specifically to [this comment](https://github.com/aws-powertools/powertools-lambda-typescript/discussions/3410#discussioncomment-12097189) for the final spec.

### Solution/User Experience

This should be a standalone module under `packages/logger/src/` in which we'll add the logic needed to create a circular buffer.

To keep the buffer decoupled from the rest, consumers of this buffer should only: 1/ set a max size in bytes, 2/ optionally provide an `onBufferOverflow()` callback function that is called by the buffer when the buffer was full, and 3/ be able to set, get, delete items at a specific key.

Since we need to create one buffer for each AWS Lambda invocation, my current line of thinking is to use a `Set` to hold the logs of a request (aka the buffer), and a `Map` to hold these buffers. Since these two data structures don't provide a way to keep track of size in bytes nor an eviction mechanism, I was thinking we could extend them rather than starting from scratch.

The first of the two pieces is the buffer itself, below is some pseudo code of how I think it could work:

```ts
class SizedItem<V> {
  public value: V;
  public logLevel: number;
  public bytesSize: number;

  public constructor(value: V, logLevel: number) {
    this.value = value;
    this.logLevel = logLevel;
    this.bytesSize = (...) // TODO: actually calculate the size in bytes
  }
}

class SizedSet<V> extends Set<SizedItem<V>> {
  public currentBytesSize: number;

  public constructor() {
    super();
    this.currentBytesSize = 0;
  }

  public add(item: SizedItem<V>): this {
    this.currentBytesSize += item.byteSize;
    super.add(item);
    return this;
  }

  public delete(item: SizedItem<V>): boolean {
    const byteSize = item.byteSize;
    const isDeleted = super.delete(item);
    if (isDeleted) {
      this.currentBytesSize -= byteSize;
    }
    return isDeleted;
  }

  public clear(): void {
    super.clear();
    this.currentBytesSize = 0;
  }

  public shift(): V | undefined {
    const firstElement = this.values().next().value;
    if (!firstElement) {
      return undefined;
    }
    this.delete(firstElement);
    return firstElement;
  }
}
```

The key features of this extended set are:
- it starts with a size of `0` and every time an item is added, the item size is added to the tally
- when an item is deleted, its size in bytes is subtracted from the total
- we provide a custom method called `shift` (name borrowed from [the analogous array method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/shift)) that gets the first element in the set and deletes it

I also added a custom data structure called `SizedItem`, its main purpose is to centralize the logic to calculate the size of an item, and reuse it as much as possible rather than recalculate it.

The second part is the map of buffers that holds all the request buffers, again below is some pseudo code:

```ts
class CircularMap<V> extends Map<
  string,
  SizedSet<V>
> {
  #maxBytesSize: number;
  #onBufferOverflow?: () => void;

  public constructor({
    maxBytesSize,
    onBufferOverflow,
  }: {
    maxBytesSize: number;
    onBufferOverflow?: () => void;
  }) {
    super();
    this.#maxBytesSize = maxBytesSize;
    this.#onBufferOverflow = onBufferOverflow;
  }

  public set(key: string, value: V, logLevel: number): this {
    const item = new SizedItem<V>(value, logLevel);

    if (item.size > this.#maxBytesSize) {
      throw Error('Item too big');
    }

    const buffer = this.get(key) || new SizedSet();
    if (buffer.currentBytesSize === 0) {
      super.set(key, buffer.add(item));

      return this;
    }

    if (buffer.currentBytesSize + item.size >= this.#maxBytesSize) {
      this.#deleteFromBufferUntilSizeIsLessThanMax(buffer, item);
      // inform the callback that the buffer was full and an element was removed
      this.#onBufferOverflow && this.#onBufferOverflow();
    }
    super.set(key, buffer.add(item));
    return this;
  }

  #deleteFromBufferUntilSizeIsLessThanMax = (
    buffer: SizedSet<V>,
    item: SizedItem<V>,
  ) => {
    while (buffer.currentBytesSize + item.size >= this.#maxBytesSize) {
      buffer.shift();
    }
  };
}
```

The key features of this second data structure are:
- it accepts a `maxBytesSize` parameter that represents the max size a buffer can reach
- it's responsible for orchestrating the addition and removal of items based on the max size
- it calls the optional callback whenever a buffer reaches the max size, but it's not concerned with what the callback does
- when attempting to add an item that is larger than the buffer max size, it throws an error

I have not tested these implementations at all, so they might be broken or I might be missing something, so whoever implements this please feel free to deviate from it. If you do, I'd still be curious to discuss the reasoning and benefits, but other than that it's fine.

Also, I imagine that calculating the size of items might require different treatments depending on the type of the item stored. Ideally we should be able to store objects (i.e. `{ message: 'hello world', age: 42 }`), but if that complicates things too much we can also just assume everything is a string and move on (i.e. `'{"message":"hello world","age"42}'`). The decision for this topic is left to the implementer, but I'd say let's prioritize completion over versatility at this stage, since we already know we'll be storing logs - the question is just whether we're storing them serialized or not.

In terms of unit tests for these new data structures, you can use [these as reference](https://github.com/aws-powertools/powertools-lambda-python/blob/d24a25896895e11608aba5bdc70171603f34687e/tests/unit/logger/required_dependencies/test_logger_buffer_cache.py) for an idea of which cases we should handle. 

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

VatsalGoel3:
## Summary  
This PR adds a new log buffering module for the logger package.

## Changes  
- Implemented a **log buffering module** with the following components:  
  - **`SizedItem`**: Encapsulates a log entry and calculates its byte size using JSON serialization.  
  - **`SizedSet`**: Extends `Set` to track the cumulative size of log entries and provides a FIFO removal method via `shift()`.  
  - **`CircularMap`**: Maps request identifiers to `SizedSet` buffers, enforces a maximum buffer size (in bytes), and evicts the oldest entries when necessary. Includes an optional `onBufferOverflow` callback.  
- Added **unit tests** (`packages/logger/tests/unit/logBuffer.test.ts`) to validate the log buffering module.  

**Issue number:** closes #3589   

---

By submitting this pull request, I confirm that I have read and followed the **Powertools for AWS Lambda (TypeScript)** contributing guidelines, and I confirm that you can use, modify, copy, and redistribute this contribution under the terms of your choice.


dreamorosi:
Hi @VatsalGoel3, thank you for the PR.

Could you please restore the PR template instead of rewriting it from scratch?

packages/logger/tests/unit/logBuffer.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    const circularMap = new CircularMap<object>({
+      maxBytesSize: maxBytes,
+    });
+    // Act & Assess
+    expect(() => {
+      circularMap.setItem('trace-1', { message: 'a very long message' }, 1);
+    }).toThrowError('Item too big');
+  });
+
+  it('should evict items when buffer overflows and call the overflow callback', () => {
```
dreamorosi, line 100: Let's rewrite this test this way, so it's less ambiguous:

```ts
it('evicts items when the buffer overflows and calls the overflow callback', () => {
  // Prepare
  const options = {
    maxBytesSize: 35,
    onBufferOverflow: vi.fn(),
  };
  const circularMap = new CircularMap<object>(options);
  const smallEntry = { message: 'log' };

  // Act
  circularMap.setItem('trace-1', smallEntry, 1);
  circularMap.setItem('trace-1', smallEntry, 1);
  circularMap.setItem('trace-1', smallEntry, 1);

  // Assess
  expect(options.onBufferOverflow).toHaveBeenCalledTimes(1);
  expect(circularMap.get('trace-1')?.currentBytesSize).toBeLessThan(
    options.maxBytesSize
  );
});
```

packages/logger/tests/unit/logBuffer.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    // Assess
+    const expectedByteSize = Buffer.byteLength(JSON.stringify(logEntry));
+    expect(item.byteSize).toBe(expectedByteSize);
+  });
+});
+
+describe('SizedSet', () => {
+  it('should add an item and update currentBytesSize correctly', () => {
+    // Prepare
+    const set = new SizedSet<object>();
```
dreamorosi, line 19: Let's change all the `new SizedSet<object>();` to `new SizedSet<Record<string, unknown>>();` so the type is more accurate

packages/logger/tests/unit/logBuffer.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    // Act
+    const item = new SizedItem(logEntry, 1);
+    // Assess
+    const expectedByteSize = Buffer.byteLength(JSON.stringify(logEntry));
+    expect(item.byteSize).toBe(expectedByteSize);
+  });
+});
+
+describe('SizedSet', () => {
+  it('should add an item and update currentBytesSize correctly', () => {
```
dreamorosi, line 17: Let's change the tense of the test cases so they read in active voice, i.e. this `should add an item and update currentBytesSize correctly` becomes this `adds an item and updates the currentBytesSize correctly`

packages/logger/tests/unit/logBuffer.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    const item = new SizedItem({ test: 'value' }, 1);
+    // Act
+    set.add(item);
+    // Assess
+    expect(set.currentBytesSize).toBe(item.byteSize);
+    expect(set.has(item)).toBe(true);
+  });
+
+  it('should delete an item and update currentBytesSize correctly', () => {
+    // Prepare
```
dreamorosi, line 29: In all test cases, let's add space between each section, i.e. add a new line before each `// Act` and `// Assess`

packages/logger/src/logBuffer.ts
```diff
@@ -0,0 +1,92 @@
+import { buffer } from 'node:stream/consumers';
```
dreamorosi, line 1: Let's remove these unused imports

packages/logger/src/logBuffer.ts
```diff
@@ -0,0 +1,92 @@
...
+import { it } from 'node:test';
+
+export class SizedItem<V> {
+  public value: V;
+  public logLevel: number;
+  public byteSize: number;
+
+  constructor(value: V, logLevel: number) {
+    this.value = value;
```
dreamorosi, line 10: At least for now, even if we made it generic, let's enforce that `value` should be a string at runtime and if it's not, throw an error. We'll use `import { isString } from '@aws-lambda-powertools/commons/typeutils';` to test it.

In the constructor, as first thing, we do this:

```ts
if (!isString(value)) {
  throw new Error('Value should be a string');
}
```

and then remove the `JSON.stringify()` from the line below.

packages/logger/tests/unit/logBuffer.test.ts
```diff
@@ -0,0 +1,123 @@
+import { describe, expect, it, vi } from 'vitest';
+import { CircularMap, SizedItem, SizedSet } from '../../src/logBuffer';
+
+describe('SizedItem', () => {
+  it('should calculate the byteSize based on JSON serialization', () => {
+    // Prepare
+    const logEntry = { message: 'hello world' };
```
dreamorosi, line 7: Let's update all the tests to use strings, basically we run `JSON.stringify()` on the log item, before adding it.

dreamorosi (changes requested):
Hi @VatsalGoel3, great PR so far!

Left a few notes that I'd like us to address before moving forward, let me know if you have any questions.

VatsalGoel3:
@dreamorosi 

I've incorporated the requested changes and made a few refinements:

Refactored the test for buffer overflow eviction

Removed JSON.stringify() in SizedItem

1. Now enforcing string values at runtime using isString()

2. Switched <object> to <string> to reflect the intended usage better.
Let me know if we should revisit this for potential flexibility down the road!

Updated test cases

1. Added type adjustments in SizedSet and CircularMap to align with enforcing string values.

2. Reworked the eviction test case to ensure correct behavior.

packages/logger/tests/unit/logBuffer.test.ts
```diff
@@ -0,0 +1,146 @@
+import { describe, expect, it, vi } from 'vitest';
+import { CircularMap, SizedItem, SizedSet } from '../../src/logBuffer';
```
dreamorosi, line 2:
```suggestion
import { CircularMap, SizedItem, SizedSet } from '../../src/logBuffer.js';
```

packages/logger/src/logBuffer.ts
```diff
@@ -0,0 +1,93 @@
...
+    const firstElement = this.values().next().value;
+    if (firstElement) {
+      this.delete(firstElement);
+    }
+    return firstElement;
+  }
+}
+
+export class CircularMap<V> extends Map<string, SizedSet<V>> {
+  #maxBytesSize: number;
```
dreamorosi, line 49:
```suggestion
  readonly #maxBytesSize: number;
```

packages/logger/src/logBuffer.ts
```diff
@@ -0,0 +1,93 @@
...
+  #onBufferOverflow?: () => void;
```
dreamorosi, line 50:
```suggestion
  readonly #onBufferOverflow?: () => void;
```

packages/logger/src/logBuffer.ts
```diff
@@ -0,0 +1,93 @@
...
+        this.#onBufferOverflow();
+      }
+    }
+
+    buffer.add(item);
+    super.set(key, buffer);
+    return this;
+  }
+
+  #deleteFromBufferUntilSizeIsLessThanMax = (
```
dreamorosi, line 85:
```suggestion
  readonly #deleteFromBufferUntilSizeIsLessThanMax = (
```

packages/logger/src/logBuffer.ts
```diff
@@ -0,0 +1,93 @@
+import { isString } from '@aws-lambda-powertools/commons/typeutils';
```
dreamorosi, line 1:
```suggestion
import { isString } from '@aws-lambda-powertools/commons/typeutils';

```

dreamorosi (approved):
Thank you for the PR and for addressing my comments!

I'm going to be creating several more issues that open for contributions in the next day or so, if you're interested keep an eye out.

leandrodamascena (approved):

Issue #3625, opened 2025-02-19
Feature request: log a warning when buffer overflows


dreamorosi:
### Use case

In #3617 we have implemented a mechanism to buffer logs. The log buffer is designed to be a circular buffer that evicts from the buffer the oldest items when adding a log that is too big to fit.

As a customer I want to have an indication that this eviction happened when I flush the logs from the buffer. This is because when reading the buffer I might notice missing log entries, so to avoid confusion there should be a warning log that informs me of the eviction.

### Solution/User Experience

At first, when designing the feature, we considered emitting a warning every time a log is removed from the buffer because of an overflow. At the end we opted to not go in this direction because we realized that this could cause a lot of warning logs in certain noisy systems and also, if the buffering threshold was set to `warn` this log could've gone missing.

To get around both issues we instead decided to emit this log only when the log buffer is flushed. This is where we think customers will expect to see this log. If the buffer is never flushed, then there's no reason for customers to care that the log buffer overflowed. If instead the log is flushed, we think that this is when customers might go look at logs, and at that point knowing why some of the logs might be missing would be the most helpful.

To implement this mechanism we'll have to introduce a new internal flag in the CircularMap class, perhaps a `public hasBufferOverflown` (feel free to find a better name) that starts out as `false` and is then flipped to `true` when the buffer has actually overflown.

To know when this has happened, we will likely have to use the [`#deleteFromBufferUntilSizeIsLessThanMax`](https://github.com/aws-powertools/powertools-lambda-typescript/blob/6968ca87f55ef7574c7904e268ba0604d4b591b7/packages/logger/src/logBuffer.ts#L86-L94) (what a mouthful aha) method in  `CircularMap` to flip the `hasBufferOverflown` flag to `true`.

Then, we could use this `hasBufferOverflown` in the `flush()` method of Logger and in case it's `true`, emit a warning **after** having flushed the buffer that says something along the lines of "Some of the log entries might be missing because the buffer overflew and some got removed"  (feel free to reword this or check how Powertools for AWS Lambda (Python) worded it).

So to recap:
- add a new property to `CircularMap` that keeps a state about whether the buffer has ever overflowed
- if the buffer ever deletes a record from the buffer because it was full (but not when it's emptied), then flip the state to `true`
[...]

PR #3639, opened 2025-02-20
feat(logger): Emit a warning on buffer flush

ConnorKirk:
## Summary
This PR adds a warning to the `flushBuffer` method when the buffer being flushed has had logs evicted.

### Changes

* Add a new property, `evictionCount` to the `SizedSet` class. It is incremented when a log is evicted from the set.
* Update `flushBuffer` method to emit a warning if there any logs have been evicted from the buffer being flushed
* Add a test to verify that the warning is emitted when a buffer that contained an evicted log is flushed.

Things to note:
1. Rather than track the presence/absence of an eviction, I thought it might be useful to track the count. It's trivial to swap back if you prefer. This is a slight deviation from the design. Might it be useful to see "100 logs have been evicted" rather than just "Logs have been evicted"?
2. Each trace(/buffer) may or may not have evicted logs. Rather than storing a boolean on the CircularMap, I think it would be more accurate to store whether each buffer has had an eviction. I chose to add this to the `SizedSet` class, but another Map on the `CircularMap` mapping `traceId` to eviction status could also work. 

**Issue number:** #3625 

---


packages/logger/tests/unit/logBuffer.test.ts
```diff
@@ -107,4 +107,34 @@ describe('flushBuffer', () => {
...
+    // Act
+    const longMessage = 'blah'.repeat(10);
+
+    let i = 0;
+    while (i < 4) {
+      logger.info(
+        `${i} This is a really long log message intended to exceed the buffer ${longMessage}`
+      );
+      i++;
+    }
```
ConnorKirk, line 126: This was a bit fiddly to configure correctly so that we'd overflow the buffer. If we change the default size of the buffer in future this might break.

packages/logger/src/logBuffer.ts
```diff
@@ -17,6 +17,7 @@ export class SizedItem<V> {
 
 export class SizedSet<V> extends Set<SizedItem<V>> {
   public currentBytesSize = 0;
+  public evictedLogCount = 0;
```
ConnorKirk, line 20: I dislike that this is converting a more generic class to something more specific, but extending the SizedSet in a new class (`Buffer`?) seemed like overkill.

packages/logger/src/Logger.ts
```diff
@@ -1266,6 +1269,16 @@ class Logger extends Utility implements LoggerInterface {
             ).toLowerCase() as keyof Omit<LogFunction, 'critical'>);
       this.console[consoleMethod](item.value);
     }
+    if (buffer.evictedLogCount > 0) {
+      this.printLog(
+        LogLevelThreshold.WARN,
+        this.createAndPopulateLogItem(
+          LogLevelThreshold.WARN,
+          `${buffer.evictedLogCount} were evicted from the buffer and are not shown. Increase buffer size to store more logs in the buffer`,
```
ConnorKirk, line 1277: I think this message could do with some word smithing. I wanted to convey:
1. Logs have been evicted
2. The quantity of logs that have been evicted. Is it meaningful to know 100 logs were evicted vs 1? 
3. What the user can do to change this (i.e. increase buffer size). This is a bit hard to specify as we haven't added the config option yet

packages/logger/src/logBuffer.ts
dreamorosi: Why are we keeping the count and incrementing it instead of just having a `boolean`? What's the advantage?

packages/logger/tests/unit/logBuffer.test.ts
dreamorosi: Agreed, but there's another issue open in which we'll make the buffer size configurable, so once that's done we can modify this test to set a fixed one.

packages/logger/src/Logger.ts
ConnorKirk: It might also be good to check what Python does - I will have a look.

dreamorosi: I think keeping the exact count makes things a bit too brittle. I would just say "some logs might be missing" and leave it at that. Especially because I don't think in most cases you know how big is a log in practice, especially when dealing with input data (i.e. logging a request payload).

This is something we can add in the future if there's demand, but for now I'd just go with the simple option.

dreamorosi (approved):
Thank you for this PR!

Feel free to go on the issue list and leave a comment under the next one, so I can assign it to you.

Well done!

PR #3641, opened 2025-02-21
feat(logger): Enable log buffering feature

### LINKED ISSUE #3634: Feature request: allow enabling log buffering via constructor
### Use case

In #3617 we have implemented a mechanism to buffer logs. The next step in the implementation is to make the feature configurable via constructor options.

### Solution/User Experience

Customers should be able to configure log buffering via constructor options.

The `ConstructorOptions` type/object is already pretty crowded ([link](https://github.com/aws-powertools/powertools-lambda-typescript/blob/a629010a5b81c44d9229f19fbaee163bdb6e6c59/packages/logger/src/types/Logger.ts#L179-L181)), so we should add the new settings as a nested object.

In practice, we should add a new optional `bufferConfig` object to `ConstructorOptions` and this object should have the following keys all optional:
- `enabled`: boolean (defaults to `true`)
- `maxBytes` - this allows customers to configure the max size of the buffer in bytes (optional defaults to `1024`)
- `flushOnErrorLog` - this allows customers to toggle whether the buffer is flushed when calling `logger.error()` (optional defaults to `true`)
- `bufferAtVerbosity` - this allows customers to configure the threshold at which we'll buffer logs (optional defaults to `DEBUG` | `debug`, accepts `TRACE`, `DEBUG`, `INFO`, and `WARN` plus their corresponding small case versions - you can never buffer `ERROR` logs)

The configuration logic of the Logger can be found in the [`setOptions()`](https://github.com/aws-powertools/powertools-lambda-typescript/blob/a629010a5b81c44d9229f19fbaee163bdb6e6c59/packages/logger/src/Logger.ts#L1157-L1159) method. The order in which things are initialized in this method matters, but for this feature I can't think of any dependency so I'd just add it at the [end here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/a629010a5b81c44d9229f19fbaee163bdb6e6c59/packages/logger/src/Logger.ts#L1195).

To follow the existing pattern, we should add a new `#setLogBuffering()` method (feel free to use a better name as long as it starts with the `#set` prefix). This method:
- is called only if `ConstructorOptions['bufferConfig']` is present
- accepts the `ConstructorOptions['bufferConfig']` object
- configures the appropriate fields that were added in #3617 (we might need to move the instantiation of the `#buffer` here, since the max size is know only at this point)

Then, after this is added, we should also:
- make the [`flushBuffer()`](https://github.com/aws-powertools/powertools-lambda-typescript/blob/a629010a5b81c44d9229f19fbaee163bdb6e6c59/packages/logger/src/Logger.ts#L1252) method public
- add the newly public method to the [`LoggerInterface`](https://github.com/aws-powertools/powertools-lambda-typescript/blob/a629010a5b81c44d9229f19fbaee163bdb6e6c59/packages/logger/src/types/Logger.ts#L189)
- modify the [`error()` method](https://github.com/aws-powertools/powertools-lambda-typescript/blob/a629010a5b81c44d9229f19fbaee163bdb6e6c59/packages/logger/src/Logger.ts#L332-L334) to flush the buffer if log buffering is enabled **&&** `flushOnErrorLog` is `true`

Besides the implementation, we should update the unit tests that use the feature to now use both the public config and new public methods as well as add tests for the new config + flush-on-error behavior. Likewise, we should make sure the new types are added and documented properly (see other types in the same file for reference).

### Alternative solutions

```Markdown
N/A
```

### Acknowledgment

### Future readers

ConnorKirk:
## Summary
This PR enables the buffering feature for the logger utility.
It allows the user to configure the buffering functionality in the logger constructor.

The core of this PR is adding allow the behaviour of log buffering to be configured.

### Changes

1. Add  `logBufferOptions` object to the Logger config
2. Enable log buffering by default
3. Make `flushBuffer` method public

**Issue number:** closes #3634 

---

### DIFF
--- a/packages/logger/src/Logger.ts
+++ b/packages/logger/src/Logger.ts
@@ -188,7 +188,7 @@ class Logger extends Utility implements LoggerInterface {
    * Max size of the buffer. Additions to the buffer beyond this size will
    * cause older logs to be evicted from the buffer
    */
-  #maxBufferBytesSize = 1024;
+  #maxBufferBytesSize = 20480;
 
   /**
    * Contains buffered logs, grouped by _X_AMZN_TRACE_ID, each group with a max size of `maxBufferBytesSize`
@@ -1284,7 +1284,7 @@ class Logger extends Utility implements LoggerInterface {
    * Flushes all items of the respective _X_AMZN_TRACE_ID within
    * the buffer.
    */
-  flushBuffer(): void {
+  public flushBuffer(): void {
     const traceId = this.envVarsService.getXrayTraceId();
     if (traceId === undefined) {
       return;
--- a/packages/logger/src/Logger.ts
+++ b/packages/logger/src/Logger.ts
@@ -1245,10 +1245,11 @@ class Logger extends Utility implements LoggerInterface {
       maxBytesSize: this.#maxBufferBytesSize,
     });
 
-    if (options.enabled !== undefined) {
-      this.isBufferEnabled = options.enabled;
+    if (options.enabled === false) {
+      this.isBufferEnabled = false;
+    } else {
+      this.isBufferEnabled = true;
     }
-
     const bufferAtLogLevel = options.bufferAtVerbosity?.toUpperCase();
 
     if (this.isValidLogLevel(bufferAtLogLevel)) {
--- a/packages/logger/src/Logger.ts
+++ b/packages/logger/src/Logger.ts
@@ -172,7 +172,7 @@ class Logger extends Utility implements LoggerInterface {
   /**
    * Represents whether the buffering functionality is enabled in the logger
    */
-  protected isBufferEnabled = true;
+  protected isBufferEnabled = false;
 
   /**
    * Whether the buffer should be flushed when an error is logged
--- a/packages/logger/src/Logger.ts
+++ b/packages/logger/src/Logger.ts
@@ -172,25 +172,28 @@ class Logger extends Utility implements LoggerInterface {
   /**
    * Represents whether the buffering functionality is enabled in the logger
    */
-  protected isBufferEnabled = false;
+  protected isBufferEnabled = true;
 
+  /**
+   * Whether the buffer should be flushed when an error is logged
+   */
+  protected flushOnErrorLog = true;
   /**
    * Log level threshold for the buffer
    * Logs with a level lower than this threshold will be buffered
+   * Default is DEBUG
    */
-  protected bufferLogThreshold: number = LogLevelThreshold.DEBUG;
+  protected bufferAtVerbosity: number = LogLevelThreshold.DEBUG;
   /**
    * Max size of the buffer. Additions to the buffer beyond this size will
    * cause older logs to be evicted from the buffer
    */
-  readonly #maxBufferBytesSize = 1024;
+  #maxBufferBytesSize = 1024;
 
   /**
    * Contains buffered logs, grouped by _X_AMZN_TRACE_ID, each group with a max size of `maxBufferBytesSize`
    */
-  readonly #buffer: CircularMap<string> = new CircularMap({
-    maxBytesSize: this.#maxBufferBytesSize,
-  });
+  #buffer?: CircularMap<string>;
 
   /**
    * Log level used by the current instance of Logger.
@@ -330,6 +333,9 @@ class Logger extends Utility implements LoggerInterface {
    * @param extraInput - The extra input to log.
    */
   public error(input: LogItemMessage, ...extraInput: LogItemExtraInput): void {
+    if (this.isBufferEnabled && this.flushOnErrorLog) {
+      this.flushBuffer();
+    }
     this.processLogItem(LogLevelThreshold.ERROR, input, extraInput);
   }
 
@@ -1167,6 +1173,7 @@ class Logger extends Utility implements LoggerInterface {
       environment,
       jsonReplacerFn,
       logRecordOrder,
+      logBufferOptions,
     } = options;
 
     if (persistentLogAttributes && persistentKeys) {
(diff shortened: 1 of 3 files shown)


packages/logger/src/types/Logger.ts
```diff
@@ -173,12 +173,44 @@ type LogRecordOrderOption = {
...
+     */
+    enabled?: boolean;
+
+    /**
+     * Maximum size in bytes of each buffer.
+     */
+    maxBytes?: number;
+    /**
+     * Flush the buffer when an error is logged
+     */
```
dreamorosi, line 194:
```suggestion
     * @default `true`
     */
```

packages/logger/src/types/Logger.ts
```diff
@@ -173,12 +173,44 @@ type LogRecordOrderOption = {
...
+ *
+ * Used to configure the log buffer functionality.
+ */
+type LogBufferOption = {
+  logBufferOptions?: {
+    /**
+     * Whether logs should be buffered
+     */
+    enabled?: boolean;
+
```
dreamorosi, line 187:
```suggestion
```

packages/logger/src/types/Logger.ts
```diff
@@ -173,12 +173,44 @@ type LogRecordOrderOption = {
...
+    /**
+     * Maximum size in bytes of each buffer.
+     */
```
dreamorosi, line 190:
```suggestion
     * @default `20480`
     */
```

packages/logger/src/types/Logger.ts
```diff
@@ -173,12 +173,44 @@ type LogRecordOrderOption = {
...
+ */
+type LogBufferOption = {
+  logBufferOptions?: {
+    /**
+     * Whether logs should be buffered
+     */
+    enabled?: boolean;
+
+    /**
+     * Maximum size in bytes of each buffer.
```
dreamorosi, line 189:
```suggestion
     * Maximum size of the buffer in bytes
```

packages/logger/src/Logger.ts
```diff
@@ -172,25 +172,28 @@ class Logger extends Utility implements LoggerInterface {
...
+   * Default is DEBUG
    */
-  protected bufferLogThreshold: number = LogLevelThreshold.DEBUG;
+  protected bufferAtVerbosity: number = LogLevelThreshold.DEBUG;
   /**
    * Max size of the buffer. Additions to the buffer beyond this size will
    * cause older logs to be evicted from the buffer
    */
-  readonly #maxBufferBytesSize = 1024;
+  #maxBufferBytesSize = 1024;
```
dreamorosi, line 191: Let's change it to `20480` as discussed [here](https://github.com/aws-powertools/powertools-lambda-typescript/issues/3634#issuecomment-2669420699).

packages/logger/src/Logger.ts
```diff
@@ -172,25 +172,28 @@ class Logger extends Utility implements LoggerInterface {
   /**
    * Represents whether the buffering functionality is enabled in the logger
    */
-  protected isBufferEnabled = false;
+  protected isBufferEnabled = true;
```
dreamorosi, line 175: This should stay `false`, the feature will be opt in.

dreamorosi (changes requested):
The PR is going well, thank you for working on it.

I've left a handful of comments, but we're definitely in the right direction.

dreamorosi:
> ## [Quality Gate Passed](https://sonarcloud.io/dashboard?id=aws-powertools_powertools-lambda-typescript&pullRequest=3641) **Quality Gate passed**
> Issues  [1 New issue](https://sonarcloud.io/project/issues?id=aws-powertools_powertools-lambda-typescript&pullRequest=3641&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [0 Accepted issues](https://sonarcloud.io/project/issues?id=aws-powertools_powertools-lambda-typescript&pullRequest=3641&issueStatuses=ACCEPTED)
> 
> Measures  [0 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=aws-powertools_powertools-lambda-typescript&pullRequest=3641&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [0.0% Coverage on New Code](https://sonarcloud.io/component_measures?id=aws-powertools_powertools-lambda-typescript&pullRequest=3641&metric=new_coverage&view=list)  [0.0% Duplication on New Code](https://sonarcloud.io/component_measures?id=aws-powertools_powertools-lambda-typescript&pullRequest=3641&metric=new_duplicated_lines_density&view=list)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=aws-powertools_powertools-lambda-typescript&pullRequest=3641)

@ConnorKirk - please take a look at this also

ConnorKirk:
One clarifying question - If a user provides the `logBufferOptions` object in the Logger config, but does not specify `"enabled": true`, should buffering be enabled?

At the moment, buffering is disabled unless explicitly enabled, even if the rest of the log buffering object is provided.

dreamorosi:
> One clarifying question - If a user provides the `logBufferOptions` object in the Logger config, but does not specify `"enabled": true`, should buffering be enabled?
> 
> At the moment, buffering is disabled unless explicitly enabled, even if the rest of the log buffering object is provided.

Yes, it should be enabled whenever `logBufferOptions` is provided, and disabled when `logBufferOptions` is `undefined` or defined and `logBufferOptions.enabled` is `false`.

packages/logger/src/Logger.ts
```diff
@@ -174,23 +174,26 @@ class Logger extends Utility implements LoggerInterface {
    */
   protected isBufferEnabled = false;
 
+  /**
+   * Whether the buffer should be flushed when an error is logged
+   */
+  protected flushOnErrorLog = true;
```
dreamorosi, line 180: I don't think we have made this configurable, the constructor type has it but we are never updating it even when passed.

dreamorosi (changes requested):

packages/logger/src/Logger.ts
ConnorKirk: Whoops. Sorry

Issue #3650, opened 2025-02-24
Maintenance: add `NO-OP` user agent


dreamorosi:
### Summary

The AWS SDK offers the possibility to appending a user agent to requests sourcing the value from a known env variable (`AWS_SDK_UA_APP_ID`). We should set the environment variable when one is not set to include our ua to requests.

### Why is this needed?

So that we can append the Powertools user agent to AWS SDK clients more effectively.

### Which area does this relate to?

Governance

### Acknowledgment

- [x] This request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

PR #3651, opened 2025-02-24
chore(maintenance): set app sdk ua env variable when none is set

dreamorosi:
### Changes

This PR adds a mechanism to set the `AWS_SDK_UA_APP_ID` env variable when one is set so that AWS SDK clients include the value when making requests. The env variable is set when importing Powertools and allows us to send our user agent along with requests. 

This new mechanism appends a `NO-OP` user agent, and the PR also implements new logic to add feature-specific user agents only when the `NO-OP` is not present. The UA attached to requests looks like this `app/PT/NO-OP/2.14.0`.

**Issue number:** #3650

---


am29d (approved):
Wunderbar!

PR #3665, opened 2025-02-26
chore(ci): adopt new CDK Toolkit Library

### LINKED ISSUE #3117: Maintenance: manually bump `@aws-cdk/cli-lib-alpha`
### Summary

While we use Dependabot for automatically updating our dependencies, the `@aws-cdk/cli-lib-alpha` is being excluded by the updates ([example](https://github.com/aws-powertools/powertools-lambda-typescript/pull/3110/files)) despite being included in the [config](https://github.com/aws-powertools/powertools-lambda-typescript/blob/b3903a07bffd8caf0822447dcc8233b9df9af847/.github/dependabot.yml#L45-L50).

We should update the dependency manually until we find a solution.

### Why is this needed?

With the other AWS CDK related dependencies moving forward with versions and this one being left behind, the integration [tests fail](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/11031051029/job/30636984524).

We need to manually update so that we can continue running the tests.

Other

### Solution

_No response_

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR replaces the usage of the `@aws-cdk/cli-lib-alpha` with the newer `@aws-cdk/toolkit-lib` that was announced in developer preview [earlier today](https://github.com/aws/aws-cdk-cli/discussions/154).

This new toolkit is expected to be the way of interacting with CDK programmatically moving forward, so it makes sense for us to adopt it and remove the older `@aws-cdk/cli-lib-alpha`.

The switch is pretty much a drop-in replacement for us, since we were already abstracting away the CLI from the tests. The bulk of the changes are in the `TestStack` class in the internal `testing` package, which now uses the new module.

With this new toolkit we unlocked also two new benefits:
- since the new package uses the `latest` dist tag on npmjs.com it can now be picked up by Dependabot update, this was not the case with the `@aws-cdk/cli-lib-alpha` which was using a pre-release tag.
- the new module gives more control to how logs are emitted by CDK. Because of this, we can remove a lot of noise from most runs. As part of this PR I implemented a rudimentary log filtering that silences most of the logs except warnings and errors. This happens during regular runs, but when workflows are rerun with [debug logging enabled](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/troubleshooting-workflows/enabling-debug-logging) everything is logged. With this change we can keep the output clean, but still have visibility when something breaks.

Finally, while not related to the new toolkit, I also changed the stack generation to include a `Service` tag to all resources generated as part unit test runs. This was something that I had in the backlog for a while and that will unlock some additional automation down the road. For example, now that things are tagged, we can potentially automate the removal of CloudWatch LogGroup resources, which currently are just piling up. More info on this front in the near future.

**Issue number:** closes #3117

---

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -44,7 +44,7 @@ updates:
         - "aws-sdk-client-mock-vitest"
       aws-cdk:
         patterns:
-        - "@aws-cdk/cli-lib-alpha"
+        - "@aws-cdk/toolkit-lib"
         - "aws-cdk-lib"
         - "aws-cdk"
       typescript:
--- a/.github/workflows/run-e2e-tests.yml
+++ b/.github/workflows/run-e2e-tests.yml
@@ -17,7 +17,6 @@ jobs:
     env:
       NODE_ENV: dev
       PR_NUMBER: ${{ inputs.prNumber }}
-      RUNNER_DEBUG: ${{ runner.debug }}
     permissions:
       id-token: write # needed to interact with GitHub's OIDC Token endpoint.
       contents: read
@@ -75,4 +74,5 @@ jobs:
           CI: true
           ARCH: ${{ matrix.arch }}
           JSII_SILENCE_WARNING_DEPRECATED_NODE_VERSION: true
+          RUNNER_DEBUG: ${{ env.RUNNER_DEBUG }}
         run: npm run test:e2e -w ${{ matrix.package }}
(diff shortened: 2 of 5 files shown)


am29d (approved):
Great work, and super fast adoption! E2E tests are green locally and in CICD. 🚀

PR #3672, opened 2025-02-27
feat(logger): refresh sample rate calculation before each invocation

### LINKED ISSUE #3278: Feature request: Calculate Log Sampling Per Request
### Use case

The log sampling concept is great but not very useful in practice if you have provisioned concurrency turned on due to the decision happening at initialization. 

### Solution/User Experience

Per request sampling would be ideal, as opposed to per log sampling. That would give the user the ability to see the entirety of a request that was sampled rather than a mix of logs for different requests.

### Alternative solutions

_No response_

### Acknowledgment

### Future readers

am29d:
…in logger context## Summary

### Changes

This PR modifies the `sampleRate` behaviour for the logger and will now sample per invocation, instead of execution environment. When using `injectLambdaContext` either as decorator or as middleware, we now automatically refresh sample rate.

As discussed in #3278 we avoid duplicate sample decision during cold starts, as it might happen in the Logger constructor. To make this possible, I have exposed `coldStart` field from the `Utility` class.

**Issue number:** closes #3278 

---


packages/logger/src/Logger.ts
```diff
@@ -489,7 +490,13 @@ class Logger extends Utility implements LoggerInterface {
    * potentially yielding a different outcome.
    */
   public refreshSampleRateCalculation(): void {
-    this.setInitialSampleRate(this.powertoolsLogData.sampleRateValue);
+    /**
+     * we don't want to refresh the sample rate for cold starts
+     * because it might lead to double sampling, one in constructor and one here
```
dreamorosi, line 501: I would move this comment in the main docstring above explaining the same thing: when called during a cold start it results in a no-op.

This will be useful for those customers who use the manual method and call this directly.

dreamorosi (commented):
Left one comment about a docstring, other than that I have built and tested the package and it works, same for e2e tests [which are green](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/13567960070).

packages/logger/src/Logger.ts
am29d: Agree!

dreamorosi (dismissed):
Thanks for the PR!

dreamorosi (approved):
Good to go

PR #3682, opened 2025-03-03
chore(maintenance): simplify Utility class

### LINKED ISSUE #3668: Feature request: add access method to `coldStart` field in the `Utility` class
### Use case

I'd like to have a method that will return the current value of the `coldStart` field of the `Utility` class without changing the value. Right now both `isColdStart` and `getColdStart` changes the value when called. As mentioned in #3278 we can't access the flag without changing it.

### Solution/User Experience

I propose to change `isColdStart` to return the `coldStart` value without changing it which would also reflect its meaning.

### Alternative solutions

```Markdown
Change `coldstart` field to protected.
```

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR updates the implementation of the shared `Utility` class to streamline it and remove redundant methods.

**Issue number:** closes #3668

---

### DIFF
--- a/packages/commons/src/Utility.ts
+++ b/packages/commons/src/Utility.ts
@@ -52,7 +52,7 @@
  */
 export class Utility {
   protected coldStart = true;
-  private readonly defaultServiceName: string = 'service_undefined';
+  protected readonly defaultServiceName: string = 'service_undefined';
 
   /**
    * Get the cold start status of the current execution environment.
@@ -68,7 +68,7 @@ export class Utility {
    *
    * The method also flips the cold start status to `false` after the first invocation.
    */
-  public getColdStart(): boolean {
+  protected getColdStart(): boolean {
     if (this.coldStart) {
       this.coldStart = false;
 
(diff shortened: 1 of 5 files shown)


dreamorosi:
End to end tests for this branch are green: https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/13627970971

packages/commons/src/Utility.ts
```diff
@@ -68,7 +68,7 @@ export class Utility {
    *
    * The method also flips the cold start status to `false` after the first invocation.
    */
-  public getColdStart(): boolean {
+  protected getColdStart(): boolean {
```
leandrodamascena, line 71: I'm writing this here because these are the lines that changed, but you need to update the docstring for this utility.

Also, I probably don't know where this is in the codebase, and I'm 100% sure you have a way to do this, but it seems like you're missing a method to just get the current status of the `coldStart` variable without changing it, no?

leandrodamascena (changes requested):
Hi @dreamorosi! I think after updating the docstring it's good to merge.

packages/commons/src/Utility.ts
leandrodamascena: Ohh you're right! I saw now this is protected. Great.

leandrodamascena (approved):

Petwag:
Hi, not sure where to post but version 2.16.0 is introducing a breaking change by remove the **protected getDefaultServiceName()** function from Utility.

Because the logger library is using the common package with the ^ version allowing for minor upgrade, all installations of the powertool/logger pre 2.16.0 will now fail to run.

dreamorosi:
Hi @Petwag, sorry for the late reply.

You are right and this should not have happened, I have opened an issue (#3879) to fix this and remove the `^` sign from the version ranges, so it should not happen again in future releases.

Issue #3633, opened 2025-02-19
Feature request: flush buffer on uncaught error - Middy.js middleware


dreamorosi:
### Use case

In #3617 we have implemented a mechanism to buffer logs. The next step in the implementation is to allow customers to flush the buffer whenever there's an uncaught error thrown in the handler scope. This will be an opt-in feature and this item focuses on implementing this behavior for the Middy.js middleware.

### Solution/User Experience

As mentioned, this will be an opt-in feature. Customers will be able to enable it by passing a new option to the middleware called `flushBufferOnUncaughtError`:

```ts
const logger = new Logger({
  logLevel: 'WARN',
  bufferConfig: {
    maxBytes: 1024,
  }
});

export const handler = middy()
  .use(injectLambdaContext(logger, {
    flushBufferOnUncaughtError: true, // default is false
  })
  .handler(async (event: unknown) => {
    logger.debug('I am a debug'); // buffered
    logger.info('I am an info'); // NOT buffered
  })
```

In terms of implementation, we'll have to refactor the current middleware which is implemented [here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/logger/src/middleware/middy.ts). To help with the implementation, let me give you a quick primer on Middy.js

Middy.js allows middleware authors to set callback functions for 3 events:
- one that is called before the handler is called (aka `onBefore`)
- one that is called after the handler returns (aka `onAfter`) but only if the handler does not throw
- one that is called when the handler throws an error (aka `onError`)

Currently, since we don't have any special treatment for errors, we implemented it in a way that the same callback is shared between the `onAfter` and `onError` phases. As part of this item we'll have to refactor this.

The first step would be to change the current `injectLambdaContextAfterOrOnError` so that it's called only for `onAfter`. Then, the second would be to define a new callback function specific to the `onError` callback. While we are at it, let's also rename all the callbacks to something shorter like `onBefore`, `onAfter`, and `onError`.

The `onAfter` callback will continue to do what it does today, but will also empty (note not flush) the current buffer. This is a proactive measure to free up memory since after the request has finished there's no reason to keep the logs from that request buffered. 

Focusing on the `onError` one, we'll have to create a callback that simply calls the `logger.flushBuffer()` method when `flushBufferOnUncaughtError` was set to `true`. If log buffering was not enabled in the Logger this should result in a no-op so it's safe to call.

[...]

PR #3690, opened 2025-03-03
feat(logger): Flush buffer on uncaught error in Middy middleware

ConnorKirk:
## Summary
This PR adds support for flushing the log buffer when an uncaught error is encountered using middy middleware.

### Changes

* Add `flushBufferOnUncaughtError` option to middy `injectLambdaContext` options
* Add `clearBuffer` method to the logger. This is needed to clear the buffer in the `onAfter` handler. An alternative I considered was changing the logic of the buffer to only handle one trace id at a time. If a log is buffered with a new trace ID, then the buffer could be reset. 
* Split the `injectLambdaContextAfterOrOnError` into separate handlers . Simplify the names all the handlers.

### Notes

* The `target` parameter of `injectLambdaContext` can be an array of loggers. I wasn't clear on the correct behaviour if there are multiple loggers provided. At the moment it will flush each logger.

**Issue number:** #3633 

---

### DIFF
--- a/packages/logger/src/Logger.ts
+++ b/packages/logger/src/Logger.ts
@@ -449,13 +449,12 @@ class Logger extends Utility implements LoggerInterface {
               message: UncaughtErrorLogMessage,
               error,
             });
-          } else {
-            loggerRef.clearBuffer();
           }
           throw error;
           /* v8 ignore next */
         } finally {
           if (options?.clearState || options?.resetKeys) loggerRef.resetKeys();
+          loggerRef.clearBuffer();
         }
       };
     };
--- a/packages/logger/src/Logger.ts
+++ b/packages/logger/src/Logger.ts
@@ -449,6 +449,8 @@ class Logger extends Utility implements LoggerInterface {
               message: UncaughtErrorLogMessage,
               error,
             });
+          } else {
+            loggerRef.clearBuffer();
           }
           throw error;
           /* v8 ignore next */
--- a/packages/logger/src/Logger.ts
+++ b/packages/logger/src/Logger.ts
@@ -1361,6 +1361,21 @@ class Logger extends Utility implements LoggerInterface {
 
     this.#buffer?.delete(traceId);
   }
+
+  /**
+   * Empties the buffer for the current request
+   *
+   */
+  public clearBuffer(): void {
+    const traceId = this.envVarsService.getXrayTraceId();
+
+    if (traceId === undefined) {
+      return;
+    }
+
+    this.#buffer?.delete(traceId);
+  }
+
   /**
    * Test if the log meets the criteria to be buffered.
    *
--- a/packages/logger/src/middleware/middy.ts
+++ b/packages/logger/src/middleware/middy.ts
@@ -77,11 +77,11 @@ const injectLambdaContext = (
   const setCleanupFunction = (request: MiddyLikeRequest): void => {
     request.internal = {
       ...request.internal,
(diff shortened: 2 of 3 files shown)


packages/logger/src/middleware/middy.ts
```diff
@@ -99,18 +98,33 @@ const injectLambdaContext = (
...
       for (const logger of loggers) {
         logger.resetKeys();
       }
     }
+
+    for (const logger of loggers) {
+      logger.clearBuffer();
+    }
+  };
+  const onError = async ({ error }): Promise<void> => {
```
ConnorKirk, line 112: Typescript complains this error doesn't have a type. What the type would be. I couldn't see a clear answer in the middy docs.

dreamorosi (changes requested):
Thank you for the PR Connor! Left a couple comments here & there but I think we are very close

packages/logger/src/Logger.ts
```diff
@@ -1361,6 +1361,21 @@ class Logger extends Utility implements LoggerInterface {
 
     this.#buffer?.delete(traceId);
   }
+
+  /**
+   * Empties the buffer for the current request
+   *
+   */
+  public clearBuffer(): void {
```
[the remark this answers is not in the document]
ConnorKirk: Sorry, I added an update to my PR description with an explanation. 

I needed to be able to clear the buffer from the middy `after` handler. Since the buffer is private, I added a method to clear it. I think it would be better to not have a public method. There might be a better way I hadn't considered.

I was wondering whether we could change the implementation of the buffer so that it has just one buffer at a time. This would avoid the need to clear it, as it would clear itself when a new trace Id is used.

packages/logger/src/middleware/middy.ts
```diff
@@ -99,18 +98,32 @@ const injectLambdaContext = (
...
+  const onError = async ({ error }: { error: unknown }): Promise<void> => {
+    if (options?.flushBufferOnUncaughtError) {
+      for (const logger of loggers) {
+        logger.flushBuffer();
+        logger.error({
+          message: UncaughtErrorLogMessage,
+          error,
+        });
+      }
+    }
```
dreamorosi, line 120: I suppose we should clear the buffer when we're not flushing it here.

packages/logger/src/Logger.ts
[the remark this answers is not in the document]
dreamorosi: Good catch - I don't think we accounted for this in our original design.

Initially, during an early discussion with @leandrodamascena, I thought we could clear previous buffers before starting a new one but then the design evolved into this multi-request buffer that at this point I'd rather keeping.

I think overall having a public API to clear the buffer without flushing it is not a bad thing. I could see customers wanting to use it in certain cases where they want to free up memory but not necessarily log the items within.

Let's keep the public method, again good catch!

packages/logger/src/Logger.ts
```diff
@@ -449,6 +449,8 @@ class Logger extends Utility implements LoggerInterface {
               message: UncaughtErrorLogMessage,
               error,
             });
+          } else {
+            loggerRef.clearBuffer();
```
dreamorosi, line 453: I think we could just call `loggerRef.clearBuffer()` in the `finally` block. This way it runs both on errors & regular requests.

dreamorosi (approved):
Thank you for the work on this feature, Connor!

This should be the last PR for this feature - we'll write the docs in the next days and expect to ship it on Friday.

Issue #3642, opened 2025-02-22
Feature request: Support setting functionName for cold-starts via middy


steven10172:
### Use case

I would like to use the `logMetrics` middleware and specify an alternative function name. This would allow for names to be more static since CDK may sometimes replaces functions and the trailing characters are random.

### Solution/User Experience

I would recommend updating the middy `ExtraOptions` interface to accept another property called `functionName` that if defined is used instead of `request.context.functionName`.

Example usage:
```
export const handler = middy()
  .use(injectLambdaContext(logger, { resetKeys: true }))
  .use(
    logMetrics(metrics, {
      captureColdStartMetric: true,
      defaultDimensions: { functionName: LambdaNames.GET_ASSET },
    }),
  )
  .handler(getAssetHandler);
```

### Alternative solutions

Another solution would be to expose an object to the `ExtraOptions` interface that would allow attaching any dimensions to the cold start metric. This can then be used to override the `function_name` dimension, though caution would need to be exercised to ensure the value is not overwritten within the cold start check ([ref](https://github.com/aws-powertools/powertools-lambda-typescript/blob/50c912cdc4c35a5268556028e5785a763c6179a8/packages/metrics/src/Metrics.ts#L271-L273))

Example usage:

```
export const handler = middy()
  .use(injectLambdaContext(logger, { resetKeys: true }))
  .use(
    logMetrics(metrics, {
      captureColdStartMetric: true,
      coldStartDimensions: { function_name: LambdaNames.GET_ASSET },
      defaultDimensions: { functionName: LambdaNames.GET_ASSET },
    }),
  )
  .handler(getAssetHandler);
```

### Acknowledgment

- [x] This feature request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [x] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

am29d:
Hey @steven10172 

thanks for opening the issue. After a bit of testing I realise that having this option would reduce the number of metrics for a cold start. It is possible with the right query, but the number of metric will still increase with each Lambda resource replacement. 

[...]

PR #3696, opened 2025-03-04
feat(metrics): allow setting functionName via constructor parameter and environment variable

steven10172:
## Summary

This PR ensures that a manually set `functionName` is not overwritten by middy and class decorators. If `functionName` is not already defined a default value of `context.functionName` will be used with middy and class decorators.

### Changes

- Add `hasFunctionName()` to metrics class to support checking if the `functionName` is already defined
- Add `functionName` to metric constructor to support setting the `functionName` used for cold starts
- Add `POWERTOOLS_METRICS_FUNCTION_NAME` environment variable to support setting the `functionName` via IaC
- Do not call `setFunctionName` with default `context.functionName` in middy and decorator if `functionName` is already defined

This PR supports setting a custom `functionName` to be used with cold starts, that is more stable as `context.functionName` will change often due to the lambda being replaced when the infrastructure is updated.

**Issue number:**  #3642 

---


steven10172:
@dreamorosi it looks like one of the tests failed with a network error. Would you mind rerunning it?

leandrodamascena:
I'm starting the review now.

examples/snippets/metrics/setFunctionName.ts
```diff
@@ -0,0 +1,19 @@
...
+const metrics = new Metrics({
+  namespace: 'serverlessAirline',
+  serviceName: 'orders',
+});
+
+// Setting function name must come before calling `captureColdStartMetric`
+metrics.setFunctionName('my-function-name');
+
+// Ensure we emit the cold start
+metrics.captureColdStartMetric();
```
dreamorosi, line 12: This should be within the function handler

examples/snippets/metrics/setFunctionName.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const handler = async (
+  _event: unknown,
+  _context: unknown
+): Promise<void> => {
+  metrics.addMetric('successfulBooking', MetricUnit.Count, 1);
```
dreamorosi, line 18: We should have a `metrics.publishStoredMetrics()` call below

examples/snippets/metrics/functionNameMiddy.ts
```diff
@@ -0,0 +1,25 @@
...
+
+// Wrap the handler with middy
+export const handler = middy(lambdaHandler)
+  // Use the middleware by passing the Metrics instance as a parameter
+  .use(
+    logMetrics(metrics, {
+      functionName: 'my-function-name',
+      captureColdStartMetric: true,
+    })
+  );
```
dreamorosi, line 25: Let's rewrite this in a more concise way:

```ts
  .use(
    logMetrics(metrics, {
      functionName: 'my-function-name',
      captureColdStartMetric: true,
    })
  )
  .handler(async () => {
    metrics.addMetric('successfulBooking', MetricUnit.Count, 1);
  })
```

examples/snippets/metrics/functionNameDecorator.ts
```diff
@@ -0,0 +1,21 @@
...
+    functionName: 'my-function-name',
+    captureColdStartMetric: true,
+  })
+  public async handler(_event: unknown, _context: unknown): Promise<void> {
+    metrics.addMetric('successfulBooking', MetricUnit.Count, 1);
+  }
+}
+
+const handlerClass = new Lambda();
+export const handler = handlerClass.handler.bind(handlerClass); // (1)
```
dreamorosi, line 21:
```suggestion
export const handler = handlerClass.handler.bind(handlerClass);
```

docs/core/metrics.md
```diff
@@ -230,6 +230,36 @@ You can add default dimensions to your metrics by passing them as parameters in
...
 
+### Setting function name
+
+When emitting cold start metrics, we use the `context.functionName` as the `function_name`
+ dimension. If you want to change the function name you can set the `functionName` by
+ passing it as a parameter in 3 ways:
+
+* in the [Middy-compatible](https://github.com/middyjs/middy){target=_blank} middleware
+* using the `setFunctionName` method
+* in the decorator
```
dreamorosi, line 241:
```suggestion
* in the class method decorator
```

packages/metrics/src/types/Metrics.ts
```diff
@@ -112,6 +112,13 @@ type ExtraOptions = {
    * @see {@link MetricsInterface.captureColdStartMetric | `captureColdStartMetric()`}
    */
   captureColdStartMetric?: boolean;
+  /**
+   * Set the metric instances function name for `ColdStart` metric.
```
dreamorosi, line 116:
```suggestion
   * Function name to use as dimension for the `ColdStart` metric.
```

dreamorosi (changes requested):
Hi @steven10172 - thank you for the PR!

I have left a few comments on the implementation, overall I'd rather not create any new public APIs other than what's necessary.

Let me know if you have any questions about the comments.

dreamorosi (changes requested):
Please ignore my previous review comment.

After looking at the implementation I think we should go back to the drawing board and avoid overcomplicating things.

The function name here refers to the Lambda function resource, not the decorated/wrapped function, so there's no reason for adding a property in every place like we are doing.

I know the discussion in the linked issue went into that direction, but I still stand by my my first proposal in [this comment](https://github.com/aws-powertools/powertools-lambda-typescript/issues/3642#issuecomment-2679107772).

With the current implementation, we're introducing some unnecessary complexity, for example what happens if we do this?

```ts
const metrics = new Metrics({});
metrics.setFunctionName('foo');

export const handler = middy().use(logMetrics(metrics, { functionName: 'bar' )).handler(async () => {})
```

Right now `bar` would silently take precedence. We could add a warning signaling that this is happening, but I think the entire issue would be bypassed by just doing this:

```ts
const metrics = new Metrics({
  functionName: 'foo'
});
```

and potentially deprecating the `setFunctionName()` method.

am29d:
I think the idea with the middy and decorator option is to have one `new Metrics()` objects that is consumed by many functions, so you have the same configuration for all Lambdas, which have different resource names. If we don't allow middy/decorator function configuration, every function would use the same `functionName` or you'd need to create 100s of `new Metric()`. 

I agree that we need strive for simplicity, and having ENV variable to control this parameter is enough, since it is 1:1 mapping between resource name and environment variable. This means we keep three additions, except the middy/decorator option?

* `.setFunctionName()` remains as-is
* `functionName` is a Metric constructor parameter
* ~~The middy handler only sets `functionName` if it's not defined~~
 * Support `POWERTOOLS_FUNCTION_NAME` for better CDK static definition of the value as it's likely a more stable infra value compared to function code

dreamorosi:
> I think the idea with the middy and decorator option is to have one new Metrics() objects that is consumed by many functions

How does that work? This dimension is specifically for the `ColdStart` metric and there can only ever be 1 handler for each sandbox. Here we are not talking about decorating/wrapping arbitrary functions because this doesn't apply there. So in practice there's only ever a 1:1 mapping between Metrics instance and actual Lambda function handler.

With that said, if you really want to do this, you can use the `setFunctionName()` method I guess, but I don't think we should add yet another env variable for this. Both with the constructor parameter and the method are not opinionated about where the actual runtime value comes from, and customers can easily do this:

```ts
const metrics = new Metrics({
  functionName: process.env.FOO
});

// or

metrics.functionName(process.env.FOO);
```

packages/metrics/src/Metrics.ts
```diff
@@ -476,6 +476,17 @@ class Metrics extends Utility implements MetricsInterface {
...
+  /**
+   * Check if a function name has been defined.
+   *
+   * This is useful when you want to only set a function name if it is not already set.
+   *
+   * The method is primarily intended for internal use, but it is exposed for advanced use cases.
+   */
+  public hasFunctionName(): boolean {
+    return this.functionName != null;
+  }
```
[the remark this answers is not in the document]
steven10172: `functionName` cannot be accessed from outside of the class instance as it's `private`. Would you prefer that I mark this as `public` to allow for that pattern?

docs/core/metrics.md
```diff
@@ -230,6 +230,36 @@ You can add default dimensions to your metrics by passing them as parameters in
 
 If you'd like to remove them at some point, you can use the `clearDefaultDimensions` method.
 
+### Setting function name
+
+When emitting cold start metrics, we use the `context.functionName` as the `function_name`
+ dimension. If you want to change the function name you can set the `functionName` by
+ passing it as a parameter in 3 ways:
+
+* in the [Middy-compatible](https://github.com/middyjs/middy){target=_blank} middleware
```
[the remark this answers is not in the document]
steven10172: This and the other doc changes are mostly copied from the [doc section above - Adding default dimensions](https://docs.powertools.aws.dev/lambda/typescript/latest/core/metrics/#adding-default-dimensions). I can make those edits, but it might be worth taking an action item to follow up and fix those as well.

steven10172:
> How does that work? This dimension is specifically for the ColdStart metric and there can only ever be 1 handler for each sandbox. Here we are not talking about decorating/wrapping arbitrary functions because this doesn't apply there. So in practice there's only ever a 1:1 mapping between Metrics instance and actual Lambda function handler.

While that is true, my team bundles the entire codebase and then ships it to the 200+ lambdas that we own. Since the only difference is the handler that we map to the function, we either need a way to configure this via the decorator / middy parameters, or my preferred route an ENV so that we can define this as part of those lambda definitions.

---

I'm ok dropping the `functionName` parameter from `ExtraOptions`, and relying on only:
1. `functionName` constructor parameter
2. `setFunctionName()` until deprecated, if it is going to be deprecated
3. An ENV such as `POWERTOOLS_METRICS_FUNCTION_NAME`


> but I don't think we should add yet another env variable for this

Do you have more details on why? There are a few ENV variables already, and it might be weird recommending that teams solve this via an ENV that they choose the naming for and manually pipe through. Having it as a standard name would allow documentation to point to a recommended path.

dreamorosi:
> Do you have more details on why? There are a few ENV variables already, and it might be weird recommending that teams solve this via an ENV that they choose the naming for and manually pipe through. Having it as a standard name would allow documentation to point to a recommended path.

To be 100% honest I don't have a better argument besides not wanting to take on another configuration option while having a sample of exactly one in terms of customer demand, as per [our tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets) (80/20 rule).

phipag:
I agree with @hjgraca. The function name is only automatically appended to the cold start metric by Powertools which is a very specific feature of the metrics module rather than a module wide or library wide configuration. 

I haven't work on this feature in Java yet but I am inclined to make it part of the decorator as mentioned in my [comment on the issue](https://github.com/aws-powertools/powertools-lambda-typescript/issues/3642#issuecomment-2679066607).

steven10172:
With .NET and Java both planning to implement this via the constructor and decorators I feel my code still stands as-is since it adds decorator (and middy) support. I will also update the code to add constructor support, but table the ENV support until there is a more specific need.

@am29d @dreamorosi Do you agree, or would you prefer that we diverge slightly due to the complexity of how the JS metric class is passed in causing the code to be a bit more complex than other languages.

dreamorosi:
Hi @steven10172 - sorry for the delay, I was off for a few days. And thanks everyone for their inputs.

After thinking about this some more, I think in this case there's more value in being idiomatic and it's ok to slightly diverge.

For this runtime specifically, I think it's best to go with: **1/** the constructor option, and **2/** the environment variable as you suggested, but remove the decorator/middy parameter.

Ultimately even though I was not in favor, the usage you shared for the env variable makes sense in a IaC context, so I trust your experience and concede that it makes sense.

If you agree with this plan, I'm happy to help you get this merged as soon as the changes are in and the review is complete.

steven10172:
I agree with that. My primary use would be through the environment variable so I don't care much about all the other options.

packages/metrics/src/Metrics.ts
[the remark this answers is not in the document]
steven10172: Let me know if you still want this removed, but Middy doesn't have access to `functionName` as it's private.

packages/metrics/src/Metrics.ts
```diff
@@ -768,8 +781,8 @@ class Metrics extends Utility implements MetricsInterface {
    *
    * @param name - The function name
    */
-  public setFunctionName(name: string): void {
-    this.functionName = name;
+  public setFunctionName(name?: string): void {
+    this.functionName = name || this.getEnvVarsService().getFunctionName();
```
steven10172, line 785: @dreamorosi  What are your thoughts on `||` vs `??`. Using `??` would allow values of `""` to be used to ensure no name is set, but using `||` would ensure we always fall back to the environment variable.

This discussion then extends to
- `hasFunctionName()` line 487
- `captureColdStartMetric` function name check (line 384)

As those impact if we set `context.functionName`, which we currently do if set to `""`. Or if we emit the dimension `function_name`, which we do not if set to `""`.

packages/metrics/src/Metrics.ts
```diff
@@ -476,6 +476,17 @@ class Metrics extends Utility implements MetricsInterface {
...
 
+  /**
+   * Check if a function name has been defined.
+   *
+   * This is useful when you want to only set a function name if it is not already set.
+   *
+   * The method is primarily intended for internal use, but it is exposed for advanced use cases.
+   */
+  public hasFunctionName(): boolean {
+    return Boolean(this.functionName);
```
steven10172, line 487: See comment for `setFunctionName`, but we might want to consider changing this as it impacts if setting `""` will be overwritten by `context.functionName` or if this is a pattern to ensure no `function_name` dimension is emitted on cold starts.

packages/metrics/src/Metrics.ts
dreamorosi: See my main review comment.

dreamorosi (changes requested):
Thanks for progressing the PR, we're getting close to merge it.

I think there're some additional simplifications we can do to make everything clearer, specifically we should define a strict order of priority and just apply that. Right now I find it hard to reason about the feature.

Since it's easier to show it than to explain, and to make things faster, I opened a branch with the implementation I think we should adopt. You can find it [here](https://github.com/aws-powertools/powertools-lambda-typescript/compare/feat/metrics_fn_name_override?expand=1).

The branch supports the use case you are proposing and largely builds on top of your implementation.

It's important to keep in mind the environment in which we operate, Lambda, so my mental model for this function name thing would be (in order of priority):
- `functionName` constructor option
- env variable
- context OR override function name in `captureColdStartMetric` call (more on this in a sec)

Each one of them overrides the previous one (aka the one above in the list) only if: 1/ it's a string & defined and 2/ it's a non-empty string.

If none of them is set or satisfies the requirements, the `function_name` dimension is omitted in the `ColdStart` metric.

Having the `setFunctionName()` method complicates things unnecessarily, so we'll mark it as deprecated, remove it from any docs example, and eventually remove it from the lib in the next major version.

To still accommodate customers who don't set the value in the constructor or env variable, don't use middleware/constructor, or even the internal implementation of the middleware - we'll extend the signature of the `captureColdStartMetric(functionName?: string)` method to accept an optional `functionName` that takes effect only if not set anywhere else, i.e.

```ts
class metrics = new Metrics({
  namespace: 'foo',
  functionName: 'bar'
});

  metrics.captureColdStartMetric('lol'); // uses 'bar' from constructor
};
```

```ts
process.env.POWERTOOLS_METRICS_FUNCTION_NAME = 'bar';
class metrics = new Metrics({
  namespace: 'foo',
});

  metrics.captureColdStartMetric('lol'); // uses 'bar' from env
};
```

Since this pattern is used internally in the decorator & middleware, the fn name in the context never overrides what was set in the constructor / env.

---

Please feel free to incorporate the changes from the branch in yours.

dreamorosi:
Hi @steven10172, just wanted to check in to see if you were still interested in completing the PR.

If not, or no longer available (it's fine, things change!), let us know so we can take over the implementation.

steven10172:
I intend to still work on this.

I have no objections to your recommendations and example code. I was not intending to deprecate the other method, but with your buy-in I'm all for it. As I was working through my implementation I recognized there was a non ideal pattern for the non decorator/middy path, thanks for the suggestion of updating the coldstart method.

dreamorosi:
Thanks for updating the PR, I gave it a quick look and I don't see any major issue.

I'll pull the branch and make a final review tomorrow morning.

dreamorosi (dismissed):
Thank you for working on this and addressing all our comments throughout the past couple weeks.

I have made a couple small commits to align the PR with the suggested branch, but the logic stays the same.

packages/metrics/src/Metrics.ts
```diff
@@ -381,8 +383,11 @@ class Metrics extends Utility implements MetricsInterface {
         service: this.defaultDimensions.service,
       });
     }
-    if (this.functionName != null) {
-      singleMetric.addDimension('function_name', this.functionName);
+    for (const value of [this.functionName, functionName]) {
+      if (value && value.trim().length > 0) {
+        singleMetric.addDimension('function_name', value);
+        break;
+      }
```
leandrodamascena, line 390:
```suggestion
	const value = this.functionName?.trim() || functionName?.trim();
	if (value && value.length > 0) {
	  singleMetric.addDimension('function_name', value);
	}
```

Isn't this easier to read and understand?

packages/metrics/src/Metrics.ts
```diff
@@ -917,6 +909,22 @@ class Metrics extends Utility implements MetricsInterface {
...
+   * @param functionName - The function name to be used for the cold start metric set in the constructor
+   */
+  protected setFunctionNameForColdStartMetric(functionName?: string): void {
+    const constructorFunctionName = functionName;
+    const envFunctionName = this.getEnvVarsService().getFunctionName();
+    for (const value of [constructorFunctionName, envFunctionName]) {
+      if (value && value.trim().length > 0) {
+        this.functionName = value;
+        break;
+      }
```
leandrodamascena, line 924: Same case I mentioned before. If the condition is directly related to the value of the variable being present or not, why not if?

packages/metrics/tests/unit/coldStartMetric.test.ts
```diff
@@ -65,27 +65,62 @@ describe('ColdStart metric', () => {
     );
   });
 
-  it('includes the function name in the cold start metric', () => {
+  it('does not override the function name from constructor in the cold start metric', () => {
```
leandrodamascena, line 68: In general, I don't like to change existing tests with new behavior, but rather preserve the existing one and add a new one with new behavior. I read the tests and I don't know if you have a test explicitly covering this.

leandrodamascena (changes requested):
Hey @steven10172 and @dreamorosi! Great job you guys did here! I just left 3 comments that at least are worth checking out, so I'm going to request changes.

packages/metrics/tests/unit/coldStartMetric.test.ts
dreamorosi: The diff is a bit confusing because we renamed the test description and added more tests.

The difference in the test is that before we were using the now marked as deprecated method, while now we use the new one.

Either way point taken.

leandrodamascena: Thanks for explaining.

leandrodamascena (approved):

PR #3713, opened 2025-03-11
docs(validation): api docs & minor fixes

### LINKED ISSUE #3712: Docs: add API Docs & README for Validation utility

We have recently completed the implementation for the upcoming Validation utility, but haven't written any API docs nor README yet.

_No response_

### How can we improve?

We should write API docs in each public method/function/type/property, so that customers can have an idea of how to use the utility.

The actual utility docs will be reviewed in a separate issue.

### Got a suggestion in mind?

_No response_

### Acknowledgment

dreamorosi:
### Changes

This PR adds docstrings to all the types/functions/parameters/errors for the upcoming Validation utility and enables Typedoc for the package.

While documenting the code I noticed a few inconsistencies or redundant code paths, which are refactored in this PR along with their respective tests. I also added the build command and removed the `private` marker from the `package.json` file, since we're gonna start publishing in the next release.

**Issue number:** closes #3712

---


packages/validation/src/types.ts
```diff
@@ -2,35 +2,79 @@ import type {
...
+   */
   envelope?: string;
+  /**
+   * Optional formats for validation.
+   */
   formats?: Record<
     string,
-    | string
+    Format
+    /* | string
```
am29d: huh? what's that?

am29d (commented):
Looks good, just need to clarify accidental comment.

packages/validation/src/types.ts
dreamorosi: removed it, it was a leftover from a previous iteration. The final type is `Format`.

am29d (approved):

PR #3715, opened 2025-03-12
chore: add api docs in mkdocs build

### LINKED ISSUE #3714: Maintenance: build API Docs as part of main docs
### Summary

Our API Docs are built separately from our main docs, this requires two steps both on local envs during development and on CI.

While moving the API Docs away from TypeDoc is currently not an option, we should try to unify the two so that they're built together.

### Why is this needed?

So that we can simplify our local and CI/CD setups.

Other

### Solution

There's a 3rd party MKDocs plugin that runs TypeDoc and builds the API reference as part of the same build.

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR adds a new `mkdocs-typedoc` plugin to our docs so that when running `mkdocs serve` (local dev) or `mkdocs build` (CI) it builds the API reference using TypeDoc in addition to the main docs.

This will simplify the CI setup (I updated the workflow) but most importantly it also improves the authoring experience which now requires a single command to run everything.

As part of the PR, I also moved the API reference link in the main header, so it's more prominent.

image

Finally, I have tweaked the `mkdocs.yml` configuration so that whenever we update any (and `mkdocs serve` is running), it'll trigger an automatic reload - this was not the case for snippets and API reference before.

**Issue number:** closes #3714

---


docs/Dockerfile
```diff
@@ -1,5 +1,8 @@
 # version 9.5.35
 FROM squidfunk/mkdocs-material@sha256:047452c6641137c9caa3647d050ddb7fa67b59ed48cc67ec3a4995f3d360ab32
 
+# Install Node.js
+RUN apk add --no-cache nodejs=20.15.1-r0 npm
```
dreamorosi, line 5: I am aware that we're running v22 elsewhere, but this is the latest version available in the image we're using and I didn't want to invest time in building it from source.

Either way this is used only for local development, while in the CI we don't use this docker file at all - so it's fine.

mkdocs.yml
```diff
@@ -4,12 +4,35 @@ site_author: Amazon Web Services
 repo_url: https://github.com/aws-powertools/powertools-lambda-typescript
 edit_uri: edit/main/docs
 site_url: https://docs.powertools.aws.dev/lambda/typescript
+watch: [
+  docs,
+  packages/batch/src,
```
dreamorosi, line 9: I'm adding each directory one by one because I don't want to add too many files to the watch list. We only care to reload the docs when the source changes, but we want to exclude tests, build files, etc.

I tried to use globs (i.e. `packages/*/src`) or find a way too exclude, but couldn't make it work.

am29d (approved):

PR #3722, opened 2025-03-13
fix(logger): correctly refresh sample rate

### LINKED ISSUE #3718: Bug: manually calling refreshSampleRateCalculation does not refresh the sampling decision
### Expected Behavior

Invoking a hot Lambda function refreshes the sampling decision. This results in that log level being sometimes lowered to `DEBUG` based on my set `sampleRateValue`.

Image

### Current Behavior

Invoking a hot Lambda function does not refresh the sampling decision. The sampling decision seems to only happen on Logger initialization.

Image

### Code snippet

The Lambda function uses the `POWERTOOLS_LOGGER_SAMPLE_RATE` environment variable with a value of `0.5`.

```typescript

const logger = new Logger({
  serviceName: 'powertools-logger-test',
});

  logger.refreshSampleRateCalculation();

  logger.debug('visible sometimes');
  logger.info('@aws-lambda-powertools/logger" = 2.xx');
};
```

### Steps to Reproduce

1. Create a Lambda function according to the provided code snippet.
2. Ensure that the Lambda function uses the `POWERTOOLS_LOGGER_SAMPLE_RATE` environment variable and that it has a high percentage to lower the log level.
3. Install version `2.15.0` of the `@aws-lambda-powertools/logger`, deploy the function, and invoke it a few times to confirm that the `DEBUG` log level is sometimes used.
4. Install version `2.16.0` of the `@aws-lambda-powertools/logger`, deploy the function, and invoke it a few times. The `DEBUG` log level will be used for all invocations or none of them (`refreshSampleRateCalculation` function has no effect).

### Possible Solution

_No response_

latest

20.x

### Packaging format used

npm

### Execution logs

```Shell

```

dreamorosi:
### Changes

This PR refactors the internal logic to calculate and refresh log sampling so that it works for any usage pattern. In #3278 we introduced changes that automatically refresh the sample rate calculation when using the `injectLambdaContext()` class method decorator and Middy.js middleware, but broke the manual method in the process.

The PR fixes the bug reported by a customer in the linked issue.

**Issue number:** fixes #3718

---


packages/logger/src/Logger.ts
```diff
@@ -1142,30 +1180,24 @@ class Logger extends Utility implements LoggerInterface {
...
           this.debug('Setting log level to DEBUG due to sampling rate');
-        } else {
-          this.setLogLevel(
-            this.getLogLevelNameFromNumber(this.#initialLogLevel)
-          );
         }
 
-        return;
+        break;
       }
```
am29d, line 1201: Just curious why break instead of return, it has the same effect here

am29d (commented):
Thanks for the quick fix of bug. I am still a bit puzzled about the cause and also the solution. I see that we have now two fields to capture sampleRate value but use only one of the for the refresh decision. 

You have also decoupled the coldStart into a dedicated counter to check if we are in a warm state or now. 

Can you elaborate on the solution?

dreamorosi:
The previous implementation was checking the cold start in two places:
- when adding context
- when refreshing the cold start


Only the first one flipped the state, but when refreshing the sample rate manually and not using the context (as the example in the linked issue) the state never got flipped and so the method always returned early.

You can verify this by pulling the main branch and adding a test that does this:
- initialize logger w sample rate
- call refresh sample rate manually
- call refresh sample rate manually

If you debug the test and step through (or just put a breakpoint in the method) you'll see it never calls the sample rate refresh internally.

Either way, in hindsight, we shouldn't have relied on the set initial sample rate method to refresh it either, since there's a lot of logic that we don't need to run every time.

Now it's the other way around, and we keep a dedicated state for the sample rate config: one to keep track of the actual rate (which was stored in a nested object that didn't belong to it) and another to keep track of how many times it's been called.

packages/logger/src/Logger.ts
dreamorosi: Intent is clearer with a break, if we ever add some more code below it'll run as we expected, rather than being skipped as it would with a return.

PR #3724, opened 2025-03-13
feat(commons): make utilities aware of provisioned concurrency

### LINKED ISSUE #3723: Feature request: make cold start aware of Provisioned Concurrency
### Use case

When using Provisioned Concurrency (PC) in AWS Lambda, the service initializes and keeps active a number of execution environments so that a portion of the requests can be served by these already-warm environments.

As a Powertools for AWS customer using PC, I'd like the cold start metric, log key, and tracer annotation to correctly reflect whether a request is a cold start or not - which is not currently the case.

### Solution/User Experience

Currently Logger, Metrics, and Tracer are not aware of Provisioned Concurrency (PC) because the state required to report cold starts in all of them is maintained internally and is dependent on the amount of requests _seen_ by the utility.

Lambda provides an environment variable `AWS_LAMBDA_INITIALIZATION_TYPE` that can have two values in the Node.js runtime: `provisioned-concurrency` or `on-demand`.

We could use this value to tweak our cold start detection logic to report a non-cold start (aka warm start) whenever PC is enabled.

This should not require any code changes from customers besides upgrading the packages to the latest version.

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR modifies the logic used to detect cold starts by making it take into account the value of the `AWS_LAMBDA_INITIALIZATION_TYPE` env variable, which [can have `on-demand` or `provisioned-concurrency` as values](https://docs.aws.amazon.com/lambda/latest/dg/provisioned-concurrency.html#pc-environment-variables) according to the environment in which the function is being run.

Depending on the value, we either use the original logic (`on-demand`) or always default to `false`, meaning not a cold start, when running under provisioned concurrency.

This feature allows customers to have traces, logs, and metrics that are more accurate to the environment they're running their functions within.

**Issue number:** closes #3723

---


dreamorosi:
Integration tests are passing: https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/13944894973

packages/commons/src/Utility.ts
```diff
@@ -51,15 +51,46 @@
...
+    }
+  }
+
+  /**
+   * Get the value of the `AWS_LAMBDA_INITIALIZATION_TYPE` environment variable.
+   */
+  protected getInitializationType():
+    | 'unknown'
+    | 'on-demand'
+    | 'provisioned-concurrency' {
```
leandrodamascena, line 74: NodeJS doesn't support snap-start currently, but we never know if it will in the future. I don't think it hurts to add the `snap-start` value here and it doesn't need to change in the future. Snap-start behaves the same as on-demand when it comes to ColdStart.

But its up to you.

leandrodamascena (approved):
Just a small comment, but feel free to ignore! Approved.

PR #3726, opened 2025-03-14
feat(logger): set correlation ID in Logger

### LINKED ISSUE #2863: Feature request: set correlation ID in Logger
### Use case

When working with logs in distributed systems like AWS Lambda, customers want to be able to set arbitrary correlation IDs for their workloads.

Today this is possible using existing Logger features (see "Alternative solutions" section), however the experience is somewhat verbose and could be improved.

Powertools for AWS Lambda (Python) [has a feature](https://docs.powertools.aws.dev/lambda/python/latest/core/logger/#setting-a-correlation-id) that allows customers to include a `correlation_id` key to each log. This feature request proposes to implement the same feature in this version of Powertools for AWS.

This would be a first, low hanging, effort for allowing customers to more easily work with correlation IDs with Powertools for AWS (see #129).

> [!important]
> If you're interested in this feature, please consider leaving a 👍 under this post, or leave a comment - measuring demand and having customer names helps us prioritize features in our backlog.

### Solution/User Experience

The new feature is primarily aimed at customers using the injectLambdaContext class method decorator and Middy.js middleware.

Given an event with this type:

```ts
type MyEvent = {
  headers: {
    my_request_id_header: string
  }
}
```

When using decorators, the experience would look like this:

```ts
import type { LambdaInterface } from '@aws-lambda-powertools/commons/types';

const logger = new Logger();

class Lambda extends LambdaInterface {
  @logger.injectLambdaContext({ correlationIdPath: 'headers.my_request_id_header' })
  public async handler(event: MyEvent, context: Context) {
    // ...
  }
}
```

When using the Middy.js middleware, the experience would instead look like this:

```ts
import { injectLambdaContext } from '@aws-lambda-powertools/logger/middleware';

const logger = new Logger();

export const handler = middy(async (event: MyEvent) => {
  // ...
}).use(injectLambdaContext(logger, {
  correlationIdPath: 'headers.my_request_id_header'
});
```

For the manual instrumentation the method is less compelling, but also a very low hanging fruit as it would just be an alias for `logger.appendKeys({ correlation_id: ... })` (see [docs](https://docs.powertools.aws.dev/lambda/typescript/latest/core/logger/#temporary-keys)):

```ts

const logger = new Logger();

export const handler = async (event: MyEvent) => {
  logger.setCorrelationId(event.headers?.headers.my_request_id_heeader);
  
  // ...
}
```

Each one of these snippets would generate a log entry similar to this:

```json
{
    "level": "INFO",
    "message": "Collecting payment",
    "timestamp": "2021-05-03 11:47:12,494+0000",
    "service": "payment",
    "cold_start": true,
    "function_name": "test",
    "function_memory_size": 128,
    "function_arn": "arn:aws:lambda:eu-west-1:12345678910:function:test",
    "function_request_id": "52fdfc07-2182-154f-163f-5f0f9a621d72",
    "correlation_id": "correlation_id_value" // <- here
}
```

The feature would also come with a handful built-in correlation ID expressions, for example:
- `API_GATEWAY_REST` => `"requestContext.requestId"`
- `APPSYNC_RESOLVER` => `'request.headers."x-amzn-trace-id"'`
- full list [here](https://docs.powertools.aws.dev/lambda/python/latest/core/logger/#built-in-correlation-id-expressions)

These would be used like this:

```ts
import { API_GATEWAY_REST } from '@aws-lambda-powertools/logger/correlation-paths';
import type { LambdaInterface } from '@aws-lambda-powertools/commons/types';

const logger = new Logger();

class Lambda extends LambdaInterface {
  @logger.injectLambdaContext({ correlationIdPath: API_GATEWAY_REST })
  public async handler(event: MyEvent, context: Context) {
    // ...
  }
}
```

### Open questions

> [!Note]
> The points below should be addressed before the feature is ready to be implemented, if you want to contribute please help us instead of opening a PR.

The current implementation in Powertools for AWS Lambda (Python) relies on JMESPath expressions for extracting the correlation ID from the payload. This choice was sensible for them both because of their ecosystem but also because they already have the `jmespath` module in their dependency tree because of other features.

For us this is still a decision point to be addressed. Taking a dependency on `@aws-lambda-powertools/jmespath` that we released for the Idempotency utility would be easy in terms of complexity, however before moving forward with this I’d like us to explore **1/** whether it’s possible to set the dependency as optional `peerDependency` and load it only when using the feature, and if not, **2/** whether the impact on the bundle is acceptable.

I am not particularly inclined towards other JSON-query languages at this stage, but I am open to consider proposals if any. However implementing a function-based query feature (i.e. `correlationId: (event) => (event.headers.my_request_id_header)`) is out of question because of the reasons described in aws-powertools/powertools-lambda-typescript#1644 (unless anyone has compelling arguments).

### Alternative solutions

```Markdown

const logger = new Logger();

export const handler = async (event: MyEvent) => {
  logger.setCorrelationId(event.headers?.headers.my_request_id_heeader);
  
  // ...
}
```

### Acknowledgment

### Future readers

am29d:
### Changes

This PR adds a new feature to add correlation ID to the logger. As proposed in the issue, customers who don't use this feature won't need to import additional 70kb of code they will not use. Therefore for using this feature, customers need to install our jmespath utility manually and inject it into the logger. 

**Issue number:** closes #2863 

---


packages/logger/tests/unit/injectLambdaContext.test.ts
```diff
@@ -239,4 +240,339 @@ describe('Inject Lambda Context', () => {
...
+      id: 'eventbridge-id',
+    };
+
+    beforeEach(() => {
+      process.env = {
+        ...ENVIRONMENT_VARIABLES,
+        POWERTOOLS_DEV: 'true',
+      };
+      vi.clearAllMocks();
+    });
```
dreamorosi, line 261: This is the same we already have at L25

packages/logger/tests/unit/injectLambdaContext.test.ts
```diff
@@ -239,4 +240,339 @@ describe('Inject Lambda Context', () => {
       );
     }
   );
+
+  describe('Correlation ID', () => {
```
dreamorosi, line 244: Let's not go back to nested tests and keep them all flat - if they're many and we think need grouping, it's ok to create a separate file.

packages/logger/src/types/index.ts
```diff
@@ -10,3 +10,9 @@ export type {
   LoggerInterface,
 } from './Logger.js';
 export type { UnformattedAttributes } from './logKeys.js';
+export type {
+  JSONPrimitive,
+  JSONObject,
+  JSONArray,
+  JSONValue,
+} from './json.js';
```
dreamorosi: File doesn't exist, and we should not re-export types we already have in `@aws-lambda-powertools/commons/types`.

docs/core/logger.md
```diff
@@ -207,6 +207,72 @@ When debugging in non-production environments, you can log the incoming event us
...
+
+To gest started, install `@aws-lambda-powertools/jmespath` package and configure the search function for the correlationId:
+
+=== "Setup the Logger to use JMESPath search"
+
+    ```typescript
+    --8<-- "examples/snippets/logger/correlationIdLogger.ts"
+    ```
+
+    ???+ tip You can retrieve correlation IDs via `getCorrelationId` method.
```
dreamorosi, line 220: This is not rendering correctly.

packages/logger/src/types/Logger.ts
```diff
@@ -260,14 +276,14 @@ type LoggerInterface = {
 };
```
dreamorosi, line 276: The new public methods should be added here

packages/logger/package.json
```diff
@@ -47,6 +47,10 @@
     "./types": {
       "import": "./lib/esm/types/index.js",
       "require": "./lib/cjs/types/index.js"
+    },
+    "./correlationId": {
+      "import": "./lib/esm/correlationId.js",
+      "require": "./lib/cjs//correlationId.js"
```
dreamorosi, line 53: Double `//` int the path, plus you're missing the types in the `typesVersions` object below at L56 - if you build the pkg and import it, it'll not find this sub-path export.

packages/logger/src/Logger.ts
```diff
@@ -1439,6 +1450,44 @@ class Logger extends Utility implements LoggerInterface {
...
+   * @param event - The event object
+   */
+  public setCorrelationIdFromPath(
+    correlationIdPath: string,
+    event: JSONObject
+  ): void {
+    if (!this.#correlationIdSearchFn) {
+      this.warn(
+        'correlationIdPath is set but no search function was provided. The correlation ID will not be added to the log attributes.'
+      );
```
dreamorosi, line 1466: Let's just early return and remove the nested `else` branch.

packages/logger/src/Logger.ts
```diff
@@ -1439,6 +1450,44 @@ class Logger extends Utility implements LoggerInterface {
...
+      if (correlationId) this.setCorrelationId(correlationId);
+    }
+  }
+
+  /**
+   * Set the correlation ID in the log attributes.
+   * @param value - The value to set as the correlation ID
+   */
+  public setCorrelationId(value: unknown): void {
```
dreamorosi, line 1480: I know the Python version has it this way, so the implementation that was ported here is in line with that - but I think this could be improved.

With this implementation, customers who are not using class method decorators nor Middy.js middleware won't have access to the feature, but can only set a correlation ID "manually".

It's true that at that point the customer has access to the `event` object, but I feel like this method & the `setCorrelationIdFromPath` could've been merged.

We could even overload it to accept both signatures/usages:

```ts
public setCorrelationId(value: unknown): void // current usage
public setCorrelationId(value: unknown, correlationIdPath: string): void // version that accepts a JMESpath expr
public setCorrelationId(value: unknown, correlationIdPath?: string): void { // actual implementation
  if (typeof correlationIdPath === 'string') {
    // implementation of setCorrelationIdFromPath() goes here
    return;
  }

  this.appendKeys({ correlation_id: value });
}
```

This way we'd also solve the protected method access issue in the Middy.js middleware, that for now it's unable to use the search fn.

packages/logger/tests/unit/injectLambdaContext.test.ts
```diff
@@ -239,4 +240,339 @@ describe('Inject Lambda Context', () => {
...
+      expect(console.info).toHaveLoggedNth(
+        1,
+        expect.objectContaining({
+          message: 'Using API Gateway request ID',
+          correlation_id: 'api-gateway-request-id',
+        })
+      );
+    });
+
+    it('should handle undefined correlation ID gracefully', async () => {
```
dreamorosi, line 545: please change test names to active voice for brevity (and consistency)

packages/logger/tests/unit/injectLambdaContext.test.ts
```diff
@@ -239,4 +240,339 @@ describe('Inject Lambda Context', () => {
...
+      expect(console.info).toHaveLoggedNth(
+        2,
+        expect.objectContaining({
+          message: 'Log from second logger',
+          correlation_id: '12345-test-id',
+        })
+      );
+    });
+
+    it('should work with both correlationIdPath and logEvent options together', async () => {
```
dreamorosi, line 482: This test is redundant

packages/logger/tests/unit/injectLambdaContext.test.ts
```diff
@@ -239,4 +240,339 @@ describe('Inject Lambda Context', () => {
...
+      expect(logger.getCorrelationId()).toBe(numericId);
+    });
+
+    it('should handle complex objects as correlation ID', () => {
+      // Prepare
+      const logger = new Logger();
+      const complexId = { requestId: 'abc123', timestamp: Date.now() };
+
+      // Act
+      logger.setCorrelationId(complexId);
```
dreamorosi, line 432: At least with the current implementation - this is just an alias of set attribute, the test does nothing.

packages/logger/tests/unit/injectLambdaContext.test.ts
```diff
@@ -239,4 +240,333 @@ describe('Inject Lambda Context', () => {
...
+    it('should propagate correlation ID to child loggers', () => {
+      // Prepare
+      const logger = new Logger();
+      logger.setCorrelationId('parent-correlation-id');
+
+      // Act
+      const childLogger = logger.createChild();
+      childLogger.info('Hello from child!');
+
+      // Assess
```
dreamorosi, line 394: Correlation ID is stored in temporary attributes, so if there's a test that assesses they're propagated this will propagated too.

A more useful test for this feature would be to check that the search function is propagated to children instead.

packages/logger/tests/unit/injectLambdaContext.test.ts
```diff
@@ -239,4 +240,333 @@ describe('Inject Lambda Context', () => {
...
+
+      // Act
+      logger.addContext(context);
+      logger.setCorrelationId('12345-test-id');
+      logger.info('Hello, world!');
+
+      // Assess
+      expect(console.info).toHaveBeenCalledTimes(1);
+      expect(console.info).toHaveLoggedNth(
+        1,
```
dreamorosi, line 326: With the current implementation, this is the same test as a couple down below - they're just aliases to `setAttribute()`

packages/logger/src/correlationId.ts
```diff
@@ -0,0 +1,23 @@
+import type { JSONObject } from '@aws-lambda-powertools/commons/types';
+import { search as JMESPathSearch } from '@aws-lambda-powertools/jmespath';
+import { PowertoolsFunctions } from '@aws-lambda-powertools/jmespath/functions';
+
+const search = (expression: string, data: JSONObject) => {
```
dreamorosi, line 5: Please add a docstring to this function to explain usage

packages/logger/src/correlationId.ts
```diff
@@ -0,0 +1,23 @@
...
+  return JMESPathSearch(expression, data, {
+    customFunctions: new PowertoolsFunctions(),
+  });
+};
+
+const correlationPaths = {
```
dreamorosi, line 11: Please add a docstring here, as well as each key

packages/logger/src/correlationId.ts
```diff
@@ -0,0 +1,23 @@
...
+  API_GATEWAY_REST: 'requestContext.requestId',
+  API_GATEWAY_HTTP: 'requestContext.requestId',
+  APPSYNC_AUTHORIZER: 'requestContext.requestId',
+  APPSYNC_RESOLVER: 'request.headers."x-amzn-trace-id"',
+  APPLICATION_LOAD_BALANCER: 'headers."x-amzn-trace-id"',
+  EVENT_BRIDGE: 'id',
+  LAMBDA_FUNCTION_URL: 'requestContext.requestId',
+  S3_OBJECT_LAMBDA: 'xAmzRequestId',
+  VPC_LATTICE: 'headers."x-amzn-trace-id"',
+};
```
dreamorosi, line 21:
```suggestion
} as const;
```

packages/logger/src/correlationId.ts
dreamorosi: Unsure because there's a lot of review comments, but `data` here could probably be `unknown` and then casted within the usage.

In a future PR, we might also want to relax the type of `search` in `@aws-lambda-powertools/jmespath` but not now.

dreamorosi (changes requested):
Left a bit of comments here and there - I see you've made another review in the meanwhile, so a couple of them might now be outdated. If they are you can resolve them.

packages/logger/tests/unit/injectLambdaContext.test.ts
am29d: we can do both

packages/logger/src/Logger.ts
am29d: I was about to make it public, but this might get things a bit easier.

packages/logger/src/correlationId.ts
```diff
@@ -0,0 +1,61 @@
...
+import { search as JMESPathSearch } from '@aws-lambda-powertools/jmespath';
+import { PowertoolsFunctions } from '@aws-lambda-powertools/jmespath/functions';
+
+/**
+ * This function is used to search for a correlation ID in the event data and is a wrapper
+ * around the JMESPath search function. It allows you to specify a JMESPath expression
+ * to extract the correlation ID from the event data.
+ * @param expression - The JMESPath expression to use for searching the correlation ID.
+ * @param data - The event data to search in.
+ * @returns - The correlation ID extracted from the event data.
```
dreamorosi, line 11: Despite VSCode suggesting them (I don't know how to turn them off either, we have removed all of them from the docs).

am29d: do you mean the return statements or the params?

docs/core/logger.md
```diff
@@ -207,6 +207,74 @@ When debugging in non-production environments, you can log the incoming event us
...
+???+ tip
+    You can retrieve correlation IDs via `getCorrelationId` method.
+
+You can set a correlation ID using `correlationIdPath` parameter by passing a JMESPath expression, including our custom JMESPath functions or set it manually by calling `setCorrelationId` function.
+
+=== "Setting correlation ID manually"
+
+    ```typescript
+    --8<-- "examples/snippets/logger/correlationIdManual.ts"
+    ```
```
dreamorosi, line 229:
```suggestion
    ```

    1. Alternatively, if the payload is more complex you can use a JMESPath expression as second parameter when prividing a search function in the constructor.
```

docs/core/logger.md
```diff
@@ -207,6 +207,74 @@ When debugging in non-production environments, you can log the incoming event us
...
+
+???+ tip
+    You can retrieve correlation IDs via `getCorrelationId` method.
+
+You can set a correlation ID using `correlationIdPath` parameter by passing a JMESPath expression, including our custom JMESPath functions or set it manually by calling `setCorrelationId` function.
+
+=== "Setting correlation ID manually"
+
+    ```typescript
```
dreamorosi, line 227:
```suggestion
    ```typescript hl_lines="10"
```

examples/snippets/logger/correlationIdOutput.json
dreamorosi, line 1: This file should go in the `examples/snippets/logger/samples` directory

examples/snippets/logger/correlationIdPayload.json
dreamorosi, line 1: This file should go in the `examples/snippets/logger/samples` directory

dreamorosi (commented):
Thank you for addressing the previous comments, the implementation looks good now.

I've left a few minor suggestions on the docs/examples but other than that we're good.

dreamorosi (approved):
Thank you for working on this, it was one I was personally looking forward to a lot.

Thanks also for the several iterations, appreciate your patience.

am29d:
Thanks for taking the time and your attention to detail  to improve the docs and polish the implementation 💛

thiagomeireless:
Thank you for managing that issue @dreamorosi and thank you for working on it so quickly @am29d, great job guys, this project is amazing!

PR #3736, opened 2025-03-17
test: fix transient fails in log sampling tests

### LINKED ISSUE #3735: Maintenance: fix e2e test for log sampling
### Summary

The e2e test for log sampling in the Logger utility is failing during some test runs and passing during others. Usually for each run it'll fail, and then succeed at the next retry.

### Why is this needed?

We should look into it and fix it, so that we can have tests passing more consistently.

Tests

### Solution

_No response_

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR updates the e2e test for Logger, related to log sampling, to account for the new mechanism introduced in #3278. The test was using the decorator - which now refreshes the sampling calculation at each request - but was also calling the refresh method manually, which caused double sampling in some cases and causing the test to fail.

As a side quest, I also slightly improved the logs for the e2e tests. In #3665 we adopted a new CLI toolkit and back then I silenced most of the logs except when running in debug mode. Over time I realized that I'd have preferred to have some indication that the stacks were being created/destroyed, so I improved the logs to do that. Now they look like this:

image

Enabling debug mode when re-running still prints out everything.

**Issue number:** closes #3735

---

### DIFF
--- a/packages/logger/tests/e2e/sampleRate.decorator.test.FunctionCode.ts
+++ b/packages/logger/tests/e2e/sampleRate.decorator.test.FunctionCode.ts
@@ -9,23 +9,12 @@ const LOG_MSG = process.env.LOG_MSG || 'Hello World';
 const logger = new Logger({
   sampleRateValue: SAMPLE_RATE,
 });
-let firstInvocation = true;
 
 class Lambda implements LambdaInterface {
-  private readonly logMsg: string;
+  private readonly logMsg = LOG_MSG;
 
-  public constructor() {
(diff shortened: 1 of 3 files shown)


dreamorosi:
Here are 2 consecutive runs where the tests are passing at the first try, after the fix:
- https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/13901877421
- https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/13899288815

leandrodamascena (approved):

PR #3737, opened 2025-03-17
chore: make tests run in multiple regions

### LINKED ISSUE #3730: Maintenance: run e2e tests in on bigger runner
### Summary

Our e2e test suite is growing and is expected to grow even more as we add multiple features and utilities.

Currently we have ~35 parallel jobs that run in a chunks of 30 at the time. This means we're not effectively parallelizing as much as we could.

Additionally, we're starting to experience transient failures due to timeouts from CloudFormation due to the high number of operations resulting from dozens of stacks being deployed at once, since each job can have more than one stack in it.

Before investing in any specific optimization of our test suite, a potentially low hanging fruit would be to expand the test suite to run into multiple AWS regions rather than on a single one.

### Why is this needed?

This way, the operations would be split into more regions thus lowering the chance of transient issues derived from timeouts.

Automation

### Solution

Pick a region randomly between a set of available target regions and run the test for a job there.

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR updates the workflow for e2e tests so that it distributes test suite deployments across multiple regions within the same test run. This should allow us to parallelize the tests further without incurring into timeouts or slowdowns.

**Issue number:** closes #3730

---


leandrodamascena (approved):
Hi @dreamorosi! I didn't focus on the tests themselves while reviewing because you have more context on that! But the timeout configuration and pipeline look good to me.

PR #3741, opened 2025-03-19
fix(parser): ddb base schema + other exports

### LINKED ISSUE #3740: Bug: DynamoDBStreamSchema can't be extended properly
### Expected Behavior

I should be able to extend the Zod schema `DynamoDBStreamSchema` and appropriately access values at `Records[*].dynamodb` after parsing and in inferred types.

This should work when using the new-ish `DynamoDBMarshalled` helper which removes the need for the built-in transform that is breaking this currently.

### Current Behavior

If you extend `DynamoDBStreamSchema` as in [this](https://docs.powertools.aws.dev/lambda/typescript/latest/utilities/parser/#dynamodbstreamschema-with-dynamodbmarshalled) example, any other values/types from the `DynamoDBStreamChangeRecord` (i.e. `Records[*].dynamodb.*`) are lost.

Ideally the [`DynamoDBStreamChangeRecord`](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/parser/src/schemas/dynamodb.ts#L90) could be extended as well, but because this schema currently has a Zod transform in it, it cannot be extended.

### Code snippet

See Discord thread: https://discord.com/channels/1006478942305263677/1006527385409179678/1350204854148010048

### Steps to Reproduce

See Discord thread: https://discord.com/channels/1006478942305263677/1006527385409179678/1350204854148010048

### Possible Solution

Export a base schema for DynamoDBStream that doesn't have a transform as part of it so it can be extended

latest

22.x

### Packaging format used

npm

### Execution logs

```Shell

```

dreamorosi:
### Changes

This PR fixes the missing export discussed in the linked issue and also takes the opportunity to review the overall structure of the Parser exports to simplify it and provide a better IDE experience for customers.

While doing so, I also updated the structure of the files within the package to more closely align with the exports, which allows us to use star exports more effectively (i.e. `@aws-lambda-powertools/parser/schemas/<module-name>`). This required some import changes, mainly in tests, which made the diff big.

Finally, I also reviewed all the example snippets and made sure they all use scoped sub-path exports rather than the barrel file export. This way customers copying the snippet from the code will have a smaller bundle size, since they'll be importing only the schemas/envelopes that they're actually using rather than all of them. This also increased the diff considerably.

**Issue number:** fixes #3740

---


examples/snippets/parser/envelopeDecorator.ts
```diff
@@ -1,7 +1,7 @@
 import type { LambdaInterface } from '@aws-lambda-powertools/commons/types';
 import { Logger } from '@aws-lambda-powertools/logger';
 import { parser } from '@aws-lambda-powertools/parser';
-import { EventBridgeEnvelope } from '@aws-lambda-powertools/parser/envelopes';
+import { EventBridgeEnvelope } from '@aws-lambda-powertools/parser/envelopes/event-bridge';
```
am29d, line 4:
```suggestion
import { EventBridgeEnvelope } from '@aws-lambda-powertools/parser/envelopes/eventbridge';
```

should be one word

am29d (changes requested):
One minor typo, otherwise good to merge

am29d (approved):

PR #3747, opened 2025-03-19
chore(layers): add workflows for govcloud layers

### LINKED ISSUE #3423: Feature request: Make layer available in GovCloud
### Use case

I would like to use Powertools for my project as a Lambda layer. One of the regions I have to support is in GovCloud, which means I cannot use this layer as it currently stands.

Powertools for Python is available in GovCloud ([source](https://docs.powertools.aws.dev/lambda/python/latest/#lambda-layer-govcloud)), so I'd like to request similar functionality here. Thanks for your consideration.

### Solution/User Experience

Powertools available as a Lambda layer in GovCloud.

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

am29d:
### Changes

This PR adds two workflows to publish layers to govcloud. The documentation update will come in a separate PR.

**Issue number:** closes #3423 

---


.github/workflows/layers_govcloud.yml
```diff
@@ -0,0 +1,200 @@
...
+        type: string
+        required: true
+      version:
+        description: Layer version to duplicate
+        type: string
+        required: true
+
+run-name: Layer Deployment (GovCloud) - ${{ inputs.environment }}
+
+permissions:
```
dreamorosi, line 39: This should be added to the other workflow as well, or it'll raise an OSSF scorecard issue

.github/workflows/layer_govcloud_verify.yml
```diff
@@ -0,0 +1,91 @@
...
+    steps:
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@ececac1a45f3b08a01d2dd070d28d111c5fe6722 # v4.1.0
+        with:
+          role-to-assume: ${{ secrets.AWS_IAM_ROLE }}
+          aws-region: us-east-1
+          mask-aws-account-id: true
+      - name: Output ${{ matrix.layer }}
+        run: |
+          aws --region us-east-1 lambda get-layer-version-by-arn --arn 'arn:aws:lambda:us-east-1:094274105915:layer:${{ matrix.layer }}:${{ inputs.version }}' | jq -r '{"Layer Version Arn": .LayerVersionArn, "Version": .Version, "Description": .Description, "Compatible Runtimes": .CompatibleRuntimes[0], "Compatible Architectures": .CompatibleArchitectures[0], "SHA": .Content.CodeSha256} | keys[] as $k | [$k, .[$k]] | @tsv' | column -t -s $'\t'
```
dreamorosi, line 42: Can we add a comment above this that explains what this incantation does?

sthulb (dismissed):

dreamorosi (changes requested):
Added two comments, one about docs and another about a missing permission.

I don't have a lot of context on what the workflow is supposed to do, and I haven't tested it so I assume the secrets and permissions needed by the workflows are already (or will be) in place.

Besides that, if possible, please update the sections [here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/d13f3c6236b853dd1800409bd4942680ab2abfd0/docs/maintainers.md?plain=1#L164) to make sure it's clear at which point of the release we need to run these workflows.

.github/workflows/layer_govcloud_verify.yml
```diff
@@ -0,0 +1,91 @@
...
+      id-token: write
+      contents: read
+    strategy:
+      matrix:
+        layer:
+          - AWSLambdaPowertoolsTypeScriptV2
+    environment: Prod (Readonly)
+    steps:
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@ececac1a45f3b08a01d2dd070d28d111c5fe6722 # v4.1.0
```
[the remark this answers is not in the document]
am29d: 4.1.0 is latest

.github/workflows/layer_govcloud_verify.yml
am29d: Ah, noble practitioner of the cloud arts! Let me explain this mystical incantation that you seek to understand.

  This spell calls upon the ancient AWS powers in the realm of us-east-1, seeking knowledge of a magical layer stored in the great Lambda archives. The sacred ARN (Arcane Resource Name) points to a powerful artifact known as 'AWSLambdaPowertoolsTypeScriptV2'.

  The spell is woven in three parts:
  1. First, it summons the layer's knowledge using the aws lambda get-layer-version-by-arn enchantment
  2. Then, through the mystical 'jq' runes, it transforms the raw ethereal data into readable scrolls
  3. Finally, the 'column' inscription arranges these revelations in neat columns for mortal eyes

  The spell reveals sacred attributes of the layer:
  - Its true Version number
  - The LayerVersionArn (its full true name)
  - The Description of its powers
  - The Compatible Runtimes it may bond with
  - The Compatible Architectures it can inhabit
  - The SHA, a mystical hash that ensures its purity

  This knowledge is vital for those who would verify that the same magical construct exists across the different realms of AWS, from the commercial lands to the governmental territories.

am29d: I have added a comment, fellow wizard.

.github/workflows/layers_govcloud.yml
am29d: why? the other workflow does not read antyhing, it just calls get-layer-version-by-arn

.github/workflows/layer_govcloud_verify.yml
```diff
@@ -0,0 +1,91 @@
...
+  commercial:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    strategy:
+      matrix:
+        layer:
+          - AWSLambdaPowertoolsTypeScriptV2
+    environment: Prod (Readonly)
```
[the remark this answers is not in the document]
am29d: we don't have readonly for prod, might be good to add it as well similar to python.

.github/workflows/layers_govcloud.yml
dreamorosi: Read the third paragraph here: https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

Each workflow must set the default least level of permissions at the top level if it's also setting other permissions at the job level, which you're doing.

.github/workflows/layer_govcloud_verify.yml
dreamorosi: love this, thank you 😄

.github/workflows/layers_govcloud.yml
am29d: if you don't set permission it's defaults to none, but I see the point of OSSF, not ideal. I will set it explicitly. 

We'd need to add it to python as well, since we copied the workflows.

am29d:
Added the workflow trigger to the layer deployment pipeline and updated the maintainers doc diagram. The deployment numbers are estimated, I will change them once it has been deployed. 

We still need to add readonly env for the prod, I don't want to pass write IAM to readonly workflow, @sthulb . 

Before the next release we need to deploy and bump the versions once, like we do for new region deployments.

.github/workflows/layers_govcloud.yml
dreamorosi: Yea

.github/workflows/layer_govcloud_verify.yml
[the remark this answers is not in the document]
dreamorosi: Since we're reading a public layer, can we use any of the other existing environments like `layer-beta`, `layer-prod`, or even `e2e-tests`?

Ultimately I think it's actually better to assume another unrelated IAM role, so we also make sure the layer is public as we expect it to be.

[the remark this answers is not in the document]
am29d: Prod (Readonly) is present now, resolving.

dreamorosi (approved):
Thank you both, if possible let's test the workflows in beta right after merging

PR #3768, opened 2025-03-24
chore(parameters): update AWS SDK client type

### LINKED ISSUE #2400: Feature request: allow `BaseProvider` in Parameters to not have a client

dreamorosi:
### Changes

This PR is a second attempt at fixing the issue described in the linked issue, which is to make the usage of a client in the `BaseProvider` class optional.

The first PR updated the logic but omitted making the parameter optional in the types. This PR fixes that.

The PR also removes a 3rd party dependency on `hashi-vault-js` which was used in the examples, in favor of a much simpler and dependency free direct API call using `fetch`.

**Issue number:** closes #2400

---


leandrodamascena:
I'm going to review this tomorrow morning.

leandrodamascena (approved):
This PR looks good to me! Even though this is a simple example to show customers how to extend the parameter provider, I wouldn't show customers how to do this via HTTP call instead of [Vault Client](https://github.com/nodevault/node-vault) because customers need to deal with unseal, retry, certificate and so on, and using HTTP for that is super hard.

But it's just a comment! Approving.

dreamorosi:
I hear you, but that was an unofficial client and it was bringing some transitive dependencies that were causing GitHub alerts, so for this I'll leave it as is for now.

PR #3808, opened 2025-04-05
docs(parameters): clarify caching behavior of `getAppConfig`

### LINKED ISSUE #3801: Docs: Document that getAppConfig can only be reused for the same AppConfig application & environment

Subsequent calls to `getAppConfig` for different applications/environments fail. Looking at [its implementation](https://github.com/aws-powertools/powertools-lambda-typescript/blob/55905275d25a402acf81f8dbe7b9ac4e2cf038f5/packages/parameters/src/appconfig/getAppConfig.ts#L144) it make sense because the `environment` as well as the `application` arguments are cached. 

Therefore, something like this fails with some unintuitive "ConfigurationNotFound" error:

```js
import { getAppConfig } from '@aws-lambda-powertools/parameters/appconfig';

// This works just fine
let config = await getAppConfig('config1', {
    environment: 'first',
    application: 'myApp',
    transform: 'json'
});
console.log(config);

// This will throw an error
config = await getAppConfig('config2', {
    environment: 'second',
    application: 'myApp',
    transform: 'json'
});
console.log(config);
```

https://docs.powertools.aws.dev/lambda/typescript/latest/utilities/parameters/#fetching-app-configurations

### How can we improve?

Just state that the applications/environment arguments are cached and one has to go with the `AppConfigProvider` class if one wants to fetch multiple configurations from multiple apps/environments within the same lambda function.

### Got a suggestion in mind?

N/A

### Acknowledgment

dreamorosi:
### Changes

This PR adds a callout to the [`getAppConfig()` docs section](https://docs.powertools.aws.dev/lambda/typescript/latest/utilities/parameters/#fetching-app-configurations) to make it explicit that the underlying AppConfig provider is cached.

image

This means customers can only retrieve configurations within the same application/environment combination.

This is generally ok, since configurations should be mapped to a specific application/environment, however based on the linked issue it appears that some customers might be organizing configs differently.

--

The PR also makes two minor, unrelated, fixes to other broken references that I noticed in the docs build logs while testing this.

**Issue number:** closes #3801

---


leandrodamascena (approved):
I probably need to add something like this in Python too!

Approved!

PR #3816, opened 2025-04-08
fix(logger): warn only once on ALC log level mismatch

### LINKED ISSUE #3815: Bug: Log pollution with the same warning on every warm execution when using injectLambdaContext middleware
### Expected Behavior

No warning logged that the current log level does not match AWS Lambda Advanced Logging Controls minimum log level when `injectLambdaContext` Middy middleware is used.

### Current Behavior

On every warm execution with `injectLambdaContext` Middy middleware a warning is logged that the current log level does not match AWS Lambda Advanced Logging Controls minimum log level. 

Problem:
-  `injectLambdaContext` calls `refreshSampleRateCalculation` on every execution;
- `refreshSampleRateCalculation` calls`this.setLogLevel` with the `#initialLogLevel` on every warm call when debug sampling is not enabled;
- #initialLogLevel contains the default value INFO, because AWS Lambda Advanced Logging Controls minimum log level is set (to a higher value than INFO). 

Since the following commit https://github.com/aws-powertools/powertools-lambda-typescript/commit/8c8d6b2ea4ccd473f56b05913169cc5995765562, because `refreshSampleRateCalculation` was added to `middy.ts`.

### Code snippet

```typescript

const logger = new Logger();

  .use(injectLambdaContext(logger))
  .handler(async () => ({}));
```

### Steps to Reproduce

1. Set application log level to WARN (AWS Lambda Advanced Logging Controls | `systemLogLevelV2`);
2. Set environment variable `POWERTOOLS_LOG_LEVEL=WARN`;
3. Don't configure any debug sampling;
4. Use `injectLambdaContext` Middy middleware;
5. From a warm execution (coldStart=false) onwards every time a WARN is logged.

### Possible Solution

`setInitialLogLevel` in `Logger.ts` assigns the `#initialLogLevel` regardless if awsLogLevelShortCircuit is true and another check is added (or replaced for the `awsLogLevelShortCircuit` condition in `setInitialLogLevel`) if AWS Lambda Advanced Logging Controls is configured and sets the initial log level with this value.

Another solution would be documenting that Powertool's log level should be combined with AWS Lambda's application log level. The downside of this would be possible log pollution coming from libraries that don't use the Powertool logger, but something else (console eg).

V2.16.0 - V2.18.0 (latest currently)

22.x

### Packaging format used

npm

### Execution logs

```Shell
{
    "level": "WARN",
    "message": "Current log level (INFO) does not match AWS Lambda Advanced Logging Controls minimum log level (WARN). This can lead to data loss, consider adjusting them.",
    "timestamp": "X",
    "service": "X",
    "cold_start": false,
    "function_arn": "X",
    "function_memory_size": "128",
    "function_name": "X",
    "function_request_id": "X",
    "sampling_rate": 0,
    "xray_trace_id": "X"
}
```

dreamorosi:
### Changes

This PR implements a new internal behavior for Logger that adds the ability to log a warning only once per logger instance.

Besides the new behavior described above, the PR also addresses another bug described in the linked issue which caused the Logger to not save the Advanced Logging Controls log level upon initialization, which in turns caused the warning to be emitted when log sampling was refreshed.

**Issue number:** fixes #3815

---

### DIFF
--- a/packages/logger/src/Logger.ts
+++ b/packages/logger/src/Logger.ts
@@ -237,6 +237,11 @@ class Logger extends Utility implements LoggerInterface {
   };
 
   /**
+   * Map used to store the warning messages that have already been logged.
+   */
+  readonly #warnOnceMap = new Map<string, boolean>();
+
+  /**
    * Log level used by the current instance of Logger.
    *
    * Returns the log level as a number. The higher the number, the less verbose the logs.
@@ -715,6 +720,17 @@ class Logger extends Utility implements LoggerInterface {
   }
 
   /**
+   * Log a warning message once per unique message.
+   *
+   * @param message - The log message.
+   */
+  #warnOnce(message: string): void {
+    if (this.#warnOnceMap.has(message)) return;
+    this.#warnOnceMap.set(message, true);
+    this.warn(message);
+  }
+
+  /**
    * Factory method for instantiating logger instances. Used by `createChild` method.
    * Important for customization and subclassing. It allows subclasses, like `MyOwnLogger`,
    * to override its behavior while keeping the main business logic in `createChild` intact.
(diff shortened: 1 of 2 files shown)


am29d (approved):
Thanks for the quick fix! 

✅ e2e tests: https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/14333635580

Issue #3821, opened 2025-04-10
Feature request: Log warning message if  Advanced Logging Controls log level is less verbose than Logger level


tknisch:
### Use case

If I use the Logger for example at `Debug` level and the Advanced Logging Controls are set to `Info` I will not get the log messages printed.
It would be nice, the a warning message could be printed if this constellation is given.

### Solution/User Experience

Print a warning message (configurable) if log messages would be suppressed by Advanced Logging Controls.

### Alternative solutions

```Markdown

```

### Acknowledgment

- [x] This feature request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [x] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

dreamorosi:
Hi @tknisch, thank you for opening this issue.

As we [discussed on Discord](https://discord.com/channels/1006478942305263677/1006527385409179678/1359801819857027262), I think this is something we missed and should address.

We already have a way to detect whether Advanced Logging Controls is enabled and its level, so I think with some light refactoring we should be able to check that value and emit a warning if the log level is less verbose and log buffering is enabled.

We should also add a callout to the docs as additional measure.

Thank you for bringing this up!

ConnorKirk:
I can take a look

leandrodamascena:
Hi, @dreamorosi. @anafalcao is working on this implementation in Python, just to make sure the idea of ​​the implementation here is clear.

We already issue a warning if the Powertools LogLevel is more verbose than ALC and this works as expected. But now we need to issue a new warning when the customer flushes the buffer so that the customer knows that the `flush_buffer` will have no effect.

If the idea is to enable based on the buffer level log I think we need to review this. I think it makes more sense when flushing, since that is where the data loss will occur.

WDYT?

dreamorosi:
Imo we should just log a warning when you initialize the logger and these conditions happen:
- you enable log buffering
- the ALC log level is less verbose than log buffering

[...]

PR #3834, opened 2025-04-15
fix(logger): warn customers when the ALC log level is less verbose than log buffer

ConnorKirk:
## Summary
When the Advanced Logging Control is less verbose than the Logger Log buffer, buffered logs will not be visible. This PR adds a warning that is emitted when this scenario is detected.

There's an existing warning message if the ALC Log Level is less verbose than the Logger Log Level.

### Changes

* Emit a warning when ALC Log level is less verbose than buffer level
* Add a test for this behaviour

**Issue number:**  #3821 

---

### DIFF
--- a/packages/logger/src/Logger.ts
+++ b/packages/logger/src/Logger.ts
@@ -133,6 +133,13 @@ class Logger extends Utility implements LoggerInterface {
    * Log level used internally by the current instance of Logger.
    */
   private logLevel: number = LogLevelThreshold.INFO;
+
+  /**
+   * Advanced Logging Control Log Level
+   * If not a valid value this will be left undefined, even if the
+   * environment variable AWS_LAMBDA_LOG_LEVEL is set
+   */
+  #alcLogLevel?: Uppercase<LogLevel>;
   /**
    * Persistent log attributes that will be logged in all log items.
    */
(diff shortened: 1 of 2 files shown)


packages/logger/src/Logger.ts
```diff
@@ -1378,6 +1378,14 @@ class Logger extends Utility implements LoggerInterface {
       this.#bufferConfig.bufferAtVerbosity =
         LogLevelThreshold[bufferAtLogLevel];
     }
+    const AlcLogLevel =
+      this.getEnvVarsService().getAwsLogLevel() as keyof typeof LogLevelThreshold;
```
ConnorKirk, line 1382: `getAwsLogLevel` returns the string of the `AWS_LAMBDA_LOG_LEVEL` env var. I needed to cast it to a LogLevelThreshold enum. I considered whether this logic should be part of the `getAwsLogLevel` function.

packages/logger/src/Logger.ts
```diff
@@ -1378,6 +1378,14 @@ class Logger extends Utility implements LoggerInterface {
...
+
+    if (LogLevelThreshold[AlcLogLevel] > this.#bufferConfig.bufferAtVerbosity) {
+      this.#warnOnce(
+        'Advanced Loggging Controls (ALC) Log Level is higher than Log Buffering Log Level. Buffered logs will be filtered by ALC'
+      );
```
ConnorKirk, line 1387: This is quite a verbose message. There might be a better way to phrase it

leandrodamascena:
I know this PR is in draft stage, but please do not merge it before we have further discussion in issue #3821. I hope to write new comments by tomorrow.

packages/logger/src/Logger.ts
dreamorosi: I don't mind the verbosity of the message, but I would reword the "is higher" to "is less verbose" to make it extra clear of what's happening.

I'm pretty sure if we ask 2 people which one is higher/lower in the context of log verbosity they'll give you two different answers.

ConnorKirk: Agreed. I got confused whilst making the PR

packages/logger/src/Logger.ts
dreamorosi: Agree - I'm inclined to suggest a light refactor.

I think we should:
- add a private `#alcLogLevel` prop to the class that starts as `undefined`
- during the `setOptions()` method, read the value with `this.getEnvVarsService().getAwsLogLevel()` and check it with `this.isValidLogLevel()`
- if valid, update `#alcLogLevel`, if not leave it as is

Then use that `#alcLogLevel` here, as well as in `awsLogLevelShortCircuit()` (also to refactor).

What do you think?

dreamorosi:
I had a chat with @leandrodamascena and we agreed that we should add another log warning when we flush the buffer that says something along the lines of (feel free to reword it):

```
'Advanced Loggging Controls (ALC) Log Level is less verbose than Log Buffering Log Level. Some logs might be missing.'
```

Longer-ish explanation as for why is in the linked issue.

packages/logger/src/Logger.ts
ConnorKirk: Makes sense. Will do

dreamorosi (approved):
Thank you for the PR!

Issue #1362, opened 2023-03-09
Feature request: POWERTOOLS_DEV=1 should pretty print stack traces


everett1992:
### Use case

```
import { Logger } from '@aws-lambda-powertools/logger'
const err = new Error('message')
console.log(err)
new Logger().error("err", err)
```
Compare native node formatting, with clear new line and indentation, to powertools
```
POWERTOOLS_DEV=1 node test.mjs
Error: message
    at file://test.mjs:2:13
    at ModuleJob.run (node:internal/modules/esm/module_job:193:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:530:24)
    at async loadESM (node:internal/process/esm_loader:91:5)
    at async handleMainPromise (node:internal/modules/run_main:65:12)
```

```
{
    "level": "ERROR",
    "message": "err",
    "service": "service_undefined",
    "timestamp": "2023-03-09T19:29:41.087Z",
    "error": {
        "name": "Error",
        "location": "node:internal/modules/esm/module_job:193",
        "message": "message",
        "stack": "Error: message\n    at file://test.mjs:2:13\n    at ModuleJob.run (node:internal/modules/esm/module_job:193:25)\n    at async Promise.all (index 0)\n    at async ESMLoader.import (node:internal/modules/esm/loader:530:24)\n    at async loadESM (node:internal/process/esm_loader:91:5)\n    at async handleMainPromise (node:internal/modules/run_main:65:12)"
    }
}
```

### Solution/User Experience

Powertools should print the stack trace similarly to native node.

### Alternative solutions

```Markdown
I understand that may not be possible with JSON output. As an alternative powertools could provide a non-json, terminal-focused, formatter for development.

Im not sure if that meets the tenents

> AWS Lambda only. We optimise for AWS Lambda function environments and supported runtimes only. Utilities might work with web frameworks and non-Lambda environments, though they are not officially supported.
```

### Acknowledgment

- [X] This feature request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/#tenets)
- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [Java](https://github.com/awslabs/aws-lambda-powertools-java/), and [.NET](https://github.com/awslabs/aws-lambda-powertools-dotnet/)

### Future readers

[...]

PR #3852, opened 2025-04-23
improv(logger): Format the stack trace as an array of strings in dev mode

ConnorKirk:
## Summary

When POWERTOOLS_DEV is enabled, the stack trace will be split on newline characters to create an array. This improves readability compared to a single long string

### Changes

* Update the `formatError` method of the default Powertools formatter. When POWERTOOLS_DEV is enabled, the `stack` property will be split into an array
* Add a test for this new behaviour
* Update some existing tests affected by this behaviour change

**Issue number:** #1362 

---


packages/logger/src/formatter/LogFormatter.ts
```diff
@@ -131,6 +131,13 @@ abstract class LogFormatter {
       }
     }
 
+    if (
```
dreamorosi, line 134: Can we move this to L119 (similar to what we do for `cause`) so we avoid a reassignment?

dreamorosi (approved):
Thanks for the PR - I'll merge it after tomorrow's release so it's included in the following release.

Other than that's it's approved so from your side you can consider it done.

dreamorosi:
Sorry, I had to make a small change in the CI because some e2e tests were failing.

Whenever you can (no urgency whatsoever) could you please rebase your branch so they're incorporated?

PR #3896, opened 2025-05-06
feat(idempotency): support for `Redis` as idempotency backend 

arnabrahman:
## Summary
This Pr adds support for using `Redis` as an idempotency backend.  The implementation follows the same convention as the Python implementation of the same feature for powertools

### Changes

- Add `Redis` support for idempotency backend
-  New layer class created for redis, `RedisPersistanceLayer`, this more or less follows the python [implementation](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/utilities/idempotency/persistence/redis.py)
- Added some `console` logging similar to the Python implementation, but I think in this project, there is no precedent for doing that. So not sure if it should stay or not. 
- `RedisConnection` class to handle Redis client creation
- Using`@redis/client` package as default redis client, but the user can bring their own client to use it.
- Unit tests written for the code
- This Pr only has the code & unit tests, documentation/integration tests would go inside a different PR

Example:

When using default redis client
```
import { randomUUID } from 'node:crypto';
import { makeIdempotent } from '@aws-lambda-powertools/idempotency';
import type { Request, Response, SubscriptionResult } from './types.js';
import { RedisPersistenceLayer } from '@aws-lambda-powertools/idempotency/redis';

// Need to call the `init` method to connect to redis
const persistenceStore = await new RedisPersistenceLayer({
  url: 'redis://localhost:6379',
}).init();

const createSubscriptionPayment = async (
  event: Request
): Promise<SubscriptionResult> => {
  // ... create payment
  return {
    id: randomUUID(),
    productId: event.productId,
  };
};

export const handler = makeIdempotent(
  async (event: Request, _context: Context): Promise<Response> => {
    try {
      const payment = await createSubscriptionPayment(event);

      return {
        paymentId: payment.id,
        message: 'success',
        statusCode: 200,
      };
    } catch (error) {
      throw new Error('Error creating payment');
    }
  },
  {
    persistenceStore,
  }
);

```

When using your own redis client

```
import { createClient } from '@redis/client';
import { randomUUID } from 'node:crypto';
import { makeIdempotent } from '@aws-lambda-powertools/idempotency';
import type { Request, Response, SubscriptionResult } from './types.js';
import { RedisPersistenceLayer } from '@aws-lambda-powertools/idempotency/redis';
import type { RedisClientProtocol } from '@aws-lambda-powertools/idempotency/redis/types';

const redisClient = createClient({ url: 'redis://localhost:6379' });
await redisClient.connect();

const persistenceStore = new RedisPersistenceLayer({
  client: redisClient as RedisClientProtocol,
});

const createSubscriptionPayment = async (
  event: Request
): Promise<SubscriptionResult> => {
  // ... create payment
  return {
    id: randomUUID(),
    productId: event.productId,
  };
};

export const handler = makeIdempotent(
  async (event: Request, _context: Context): Promise<Response> => {
    try {
      const payment = await createSubscriptionPayment(event);

      return {
        paymentId: payment.id,
        message: 'success',
        statusCode: 200,
      };
    } catch (error) {
      throw new Error('Error creating payment');
    }
  },
  {
    persistenceStore,
  }
);

 ```

**Issue number:**  #3183 

---

### CODE REVIEW

File: packages/idempotency/src/persistence/RedisPersistenceLayer.ts [hunk 1]
```diff
@@ -0,0 +1,368 @@
...
+       *
+       * Conditions to successfully save a record:
+       *
+       * The idempotency key does not exist:
+       *   - first time that this invocation key is used
+       *   - previous invocation with the same key was deleted due to TTL
+       *   - SET see https://redis.io/commands/set/
+       */
+
+      console.debug(
```
dreamorosi, line 239: Let's remove all instances of `console.debug` for now - I know they're present in the Python implementation but we haven't added them in this version yet.

The reason is that right now there's no way to turn them off, so we avoided them and left only warnings.

In the medium term (perhaps right after this PR) we could do the same thing that we did for Metrics and the new Event Handler, and accept a `logger` class instance - so that the customer controls the level.

But for now better to remove it.

packages/idempotency/src/persistence/RedisPersistenceLayer.ts
```diff
@@ -0,0 +1,368 @@
...
+    super();
+
+    this.#statusAttr = options.statusAttr ?? 'status';
+    this.#expiryAttr = options.expiryAttr ?? 'expiration';
+    this.#inProgressExpiryAttr =
+      options.inProgressExpiryAttr ?? 'in_progress_expiration';
+    this.#dataAttr = options.dataAttr ?? 'data';
+    this.#validationKeyAttr = options.validationKeyAttr ?? 'validation';
+    this.#orphanLockTimeout = Math.min(10, this.expiresAfterSeconds);
+    if (options.client) {
```
dreamorosi, line 80: After talking with the team I think we settled on deviating from the Python implementation, and instead have customers pass us an instance of any Redis-compatible client.

The main reason is that we want to make sure to not force any choice on customers. The current implementation still brings in the `@redis/client` dependency even if you're using another client (i.e. Glide). Even if that can be fixed with conditional asynchronous import in the `init()` method or by having another dedicated sub-path export for Redis only - I think the core of the matter is that we want customers to decide.

Because of this, I'd say we can remove all the Redis-specific options from the constructor options and instead make a generic `client` parameter mandatory. We'll also assume that the client is already connected.

This way customers can use Redis, Valkey, Momento, Upstash, or whatever they like as long as it satisfies the `RedisClientProtocol` (which we might want to rename to something like `RedisCompatibleClient`) interface.

Then in our documentation we'll provide at least examples for `@redis/client` which stays a peer dependency and Valkey Glide.

dreamorosi (dismissed):
Thank you @arnabrahman for the amazing PR.

I have tested a build of this implementation in Lambda with a Serverless Valkey cluster and it works both with the Redis and Valkey Glide clients - which is great to see!

I have left a couple comments but the main area of change is around how we deal with the client. There are more details below, but the gist of it is that I think we should just accept a connected `client` instance and let the customer decide 100% - this means removing the `RedisConnection` part where we instantiate a client for them.

Besides leaving the vendor/flavor choice to the customer, we also know that there are a lot of ways to connect to a Redis-compatible cluster: for example types of certificates, auth methods, and other configurations. Because of this, it's highly likely that customers will want to bring their own configured client anyway - so let's go all in with this idea and only provide the Idempotency logic.

packages/idempotency/src/persistence/RedisPersistenceLayer.ts
leandrodamascena: I agree with that! TypeScript has a different ecosystem than Python - where Rust is everywhere - and it's normal to use libraries that are Rust-Python bindings. I confess that if I had to implement Python today, I wouldn't assume a standard library to create the instance, but rather I would leave it up to whoever wants to bring their own pre-built Redis client.

arnabrahman: Yeah, this definitely simplifies things on our end and eliminates a lot of the complexity we were debating, especially around handling different Redis connection patterns.

arnabrahman:
@dreamorosi I have addressed the review comments. 

Quick thought: do you think it makes sense to introduce a `BasePersistenceOptions` that both Redis and DynamoDB can extend? Most of the options are pretty similar, and the only thing that seems Redis-specific is the `client`.

dreamorosi:
Hi @arnabrahman, thank you!

Yes feel free to streamline wherever applicable, especially when it comes to types. Good idea!

leandrodamascena (approved):
Hi @arnabrahman and @dreamorosi, I have nothing to add here, the implementation is following all the standards we observed when doing this in Python. To be honest, it's a good amount of great work, congrats.

BTW, you'll open another PR to update the documentation @dreamorosi ?

arnabrahman:
@leandrodamascena Yes, we agreed to create separate Pr for documentation. If this pr is merged, i can start on the documentation part.

dreamorosi:
@leandrodamascena I need your review/approval again. 

As discussed I have renamed the new persistence layer to `CachePersistenceLayer` to be more generic, I have also changed all mentions in types/comments/code to reflect this change. The logic hasn't changed.

In terms of how to describe this new persistence layer, I have copied the wording from [this page](https://aws.amazon.com/elasticache/), so from now on I think we'll address this as "Valkey-, and Redis OSS-compatible".

The API reference now looks like this, and shows both Glide and Redis clients: 

image

packages/idempotency/src/persistence/CachePersistenceLayer.ts
```diff
@@ -0,0 +1,340 @@
...
+import type {
+  CacheClient,
+  CachePersistenceOptions,
+} from '../types/CachePersistence.js';
+import type { IdempotencyRecordStatusValue } from '../types/IdempotencyRecord.js';
+import { BasePersistenceLayer } from './BasePersistenceLayer.js';
+import { IdempotencyRecord } from './IdempotencyRecord.js';
+
+/**
+ * Valey- and Redis OOS-compatible persistence layer for idempotency records.
```
leandrodamascena, line 20:
```suggestion
 * Valkey and Redis OOS-compatible persistence layer for idempotency records.
```

leandrodamascena (approved):
If you want, just fix the typo. Approving.

leandrodamascena (approved):

packages/idempotency/src/persistence/CachePersistenceLayer.ts
```diff
@@ -0,0 +1,340 @@
...
+ * for more details on the configuration and usage patterns.
+ *
+ * **Using Valkey Glide Client**
+ *
+ * @example
+ * ```ts
+ * import { GlideClient } from '@valkey/valkey-glide';
+ * import { CachePersistenceLayer } from '@aws-lambda-powertools/idempotency/cache';
+ *
+ * const client = await GlideClient.createClient({
```
avifenesh, line 40: Recommended adding request timeout base on user preferences, this is the only config we recommend to users to set.

packages/idempotency/src/persistence/CachePersistenceLayer.ts
```diff
@@ -0,0 +1,340 @@
...
+ *   addresses: [{
+ *     host: process.env.CACHE_ENDPOINT,
```
avifenesh, line 42: Important to use the config endpoint if using EC, and not node addresses, it let us the ability to tolerate IP changes even if the nodes are not responding.

packages/idempotency/src/persistence/CachePersistenceLayer.ts
```diff
@@ -0,0 +1,340 @@
...
+  }
+
+  protected async _updateRecord(record: IdempotencyRecord): Promise<void> {
+    const item: Record<string, unknown> = {
+      [this.#statusAttr]: record.getStatus(),
+      [this.#expiryAttr]: record.expiryTimestamp,
+      [this.#dataAttr]: record.responseData,
+    };
+
+    const encodedItem = JSON.stringify(item);
```
avifenesh, line 167: You can also make it a buffer, check what costs less.

packages/idempotency/src/persistence/IdempotencyRecord.ts
```diff
@@ -13,7 +13,7 @@ import type { DynamoDBPersistenceLayer } from './DynamoDBPersistenceLayer.js';
  */
 class IdempotencyRecord {
   /**
-   * The expiry timestamp of the record in milliseconds UTC.
+   * The expiry timestamp of the record in seconds UTC.
```
avifenesh, line 16: generally ms are better for this kind of operations

packages/idempotency/src/persistence/CachePersistenceLayer.ts
dreamorosi: What's the default timeout in the client?

avifenesh: 250 ms.

packages/idempotency/src/persistence/CachePersistenceLayer.ts
dreamorosi: We are using JSON mainly for consistency with other persistence layers - we might evaluate buffers in the future but no commitment unless customers ask for it.

Issue #3777, opened 2025-03-25
Bug: `addDimensions()` should create a set


dreamorosi:
### Expected Behavior

When calling the `metrics.addDimensions()` method of the Metrics utility, the dimensions should create a new dimension set rather than concatenating the dimensions into the existing set.

Basically this:

```ts
const metrics = new Metrics({
  namespace: 'serverlessAirline',
  serviceName: 'orders',
});

export const handler = async () => {
  metrics.addDimension('environment', 'prod');

  metrics.addDimensions({
    dimension1: "1",
    dimension2: "2"
  });

  // continue emitting & flushing metrics as today
  metrics.addMetric('foo', MetricUnit.Count, 1);
  metrics.publishStoredMetrics();
};
```

should generate this:

```json
{
   "_aws":{
      "CloudWatchMetrics":[
         {
            "Namespace":"A",
            "Dimensions":[
               ["dimension1"],
               ["dimension1", "dimension2"],
            ],
            "Metrics":[
               {
                  "Name":"foo",
                  "Unit":"Bytes"
               }
            ]
         }
      ]
   },
   "dimension1":"1",
   "dimension2":"2",
   "foo":[1.0]
}
```

### Current Behavior

Currently this:

```ts
const metrics = new Metrics({
  namespace: 'serverlessAirline',
  serviceName: 'orders',
});

export const handler = async () => {
  metrics.addDimension('environment', 'prod');

  metrics.addDimensions({
    dimension1: "1",
    dimension2: "2"
  });

  // continue emitting & flushing metrics as today
  metrics.addMetric('foo', MetricUnit.Count, 1);
  metrics.publishStoredMetrics();
};
```

instead generates this:

```json
{
   "_aws":{
      "CloudWatchMetrics":[
         {
            "Namespace":"A",
            "Dimensions":[
               ["dimension1", "dimension2"],
            ],
            "Metrics":[
               {
                  "Name":"foo",
                  "Unit":"Bytes"
               }
            ]
         }
      ]
   },
   "dimension1":"1",
   "dimension2":"2",
   "foo":[1.0]
}
```

### Code snippet

See above.

### Steps to Reproduce

See above.

### Possible Solution

The EMF specification already supports this type of use case, and specifically, it represents a dimension set as an array/list of dimensions within the `Dimensions` array/list of a metric, for example:

```json
{
   "_aws":{
      "CloudWatchMetrics":[
         {
            "Namespace":"A",
[...]

PR #3964, opened 2025-05-22
fix(metrics): addDimensions() documentation and tests

matteofigus:
# Fix: `addDimensions()` should create a set

### Changes

This PR fixes issue #3777 where `addDimensions()` was incorrectly adding dimensions to the existing set instead of creating a new dimension set.

### Changes made:

1. Fixed the `addDimensions()` method in `Metrics.ts` to create a new dimension set rather than adding to existing dimensions
2. Added a `dimensionSets` array to store multiple dimension sets
3. Updated the metrics serialization logic to handle multiple dimension sets
4. Added unit tests to verify the correct behavior of dimension sets
5. Updated documentation to explain the correct behavior of `addDimensions()`

### Documentation updates:

- Added a new section in the metrics documentation explaining how to use dimension sets
- Created a new code example showing how to use `addDimensions()` to create multiple dimension sets
- Clarified that `addDimensions()` creates a new set rather than modifying existing dimensions

**Issue number:** #3777

---


packages/metrics/src/types/index.ts
dreamorosi, line 1: I'm a bit unsure about the changes in this entire file. It seems that we re-declared all the types that were previously in the `packages/metrics/src/types/Metrics.ts` file for no apparent reason.

This seems to break the compilation (see failed CI) - can we revert?

matteofigus:
Thanks for catching this! I've reverted the types/index.ts file to its original state. This was an unintended change that happened during the implementation. The PR now only contains documentation updates and test fixes.

packages/metrics/src/Metrics.ts
```diff
@@ -267,9 +273,19 @@ class Metrics extends Utility implements MetricsInterface {
    * @param dimensions - An object with key-value pairs of dimensions
    */
   public addDimensions(dimensions: Dimensions): void {
-    for (const [name, value] of Object.entries(dimensions)) {
-      this.addDimension(name, value);
+    const newDimensionKeys = Object.keys(dimensions).filter(
```
dreamorosi: This implementation doesn't match the `addDimension` method (nor the docstring above).

We are not checking if the dimension value is valid like we do in the other method, and when overwriting a dimension we don't let customers know.

Other than that I think it could work. Here's a suggestion:

```ts
const newDimensionSet: Dimensions = {};
for (const [key, value] of Object.entries(dimensions)) {
  if (!value) {
    this.#logger.warn(
      `The dimension ${key} doesn't meet the requirements and won't be added. Ensure the dimension name and value are non empty strings`
    );
    continue;
  }
  if (
    Object.hasOwn(this.dimensions, key) ||
    Object.hasOwn(this.defaultDimensions, key) ||
    Object.hasOwn(newDimensionSet, key)
  ) {
    this.#logger.warn(
      `Dimension "${key}" has already been added. The previous value will be overwritten.`
    );
  }
  newDimensionSet[key] = value;
  if (MAX_DIMENSION_COUNT <= this.getCurrentDimensionsCount()) {
    throw new RangeError(
      `The number of metric dimensions must be lower than ${MAX_DIMENSION_COUNT}`
    );
  }
}
this.dimensionSets.push(dimensions);
```

packages/metrics/src/Metrics.ts
```diff
@@ -267,9 +273,19 @@ class Metrics extends Utility implements MetricsInterface {
...
+      (key) => !this.defaultDimensions[key] && !this.dimensions[key]
+    );
+
+    if (
+      newDimensionKeys.length + this.getCurrentDimensionsCount() >=
```
dreamorosi: We should also update the `this.getCurrentDimensionsCount()` method so that it factors in the `this.dimensionSets` - even if in a set, these dimensions still count against the total 30 allowed.

packages/metrics/src/Metrics.ts
```diff
@@ -692,28 +709,51 @@ class Metrics extends Utility implements MetricsInterface {
...
+    }
 
     return {
       _aws: {
         Timestamp: this.#timestamp ?? new Date().getTime(),
         CloudWatchMetrics: [
           {
             Namespace: this.namespace || DEFAULT_NAMESPACE,
-            Dimensions: [dimensionNames],
+            Dimensions: dimensionNames as [string[]],
```
dreamorosi, line 760: Unsure about this change, was it intentional?

packages/metrics/src/Metrics.ts
```diff
@@ -692,28 +709,51 @@ class Metrics extends Utility implements MetricsInterface {
...
             Metrics: metricDefinitions,
           },
         ],
       },
-      ...this.defaultDimensions,
-      ...this.dimensions,
-      ...metricValues,
-      ...this.metadata,
+      ...Object.assign(
```
dreamorosi: Can we put back the spread operator (`...`)? With current implementation we're instantiating a new object (with `Object.assing()`) only to immediately spread it.

dreamorosi (changes requested):
Thanks for the changes, and apologies for the delayed review. Got caught between PTO and other launches.

I've left a few comments, I think we're getting closer but we need to tweak the logic to align with the `addDimension` (singular) method.

I've also answered to your questions in the linked issue, I hope it'll clarify the doubts - apologies for the confusion there.

packages/metrics/src/Metrics.ts
matteofigus: Done

packages/metrics/src/Metrics.ts
matteofigus: Yes, I think it thinks it's intentional because we used to return dimensionNames an array and now we return an array of arrays instead

packages/metrics/src/Metrics.ts
matteofigus: Done

packages/metrics/src/Metrics.ts
matteofigus: done

PR #3945, opened 2025-05-20
feat(commons): environment variable helpers

dreamorosi:
### Changes

> Please provide a summary of what's being changed

This PR adds new zero-dependency functional utilities to read and parse environment variables.

Customers can read and cast environment variables into `string`, `number`, and `boolean` using the respective helpers.

By default, the environment variable is considered required, and the function will throw an error if not present.

```ts
import { getStringFromEnv } from '@aws-lambda-powertools/commons/utils/env';

const myEnvVar = getStringFromEnv({
  key: 'MY_ENV_VAR',
});
```

Customers can customize the error message thrown by the helper by passing an optional `errorMessage` parameter.

```ts
import { getBooleanFromEnv } from '@aws-lambda-powertools/commons/utils/env';

const myEnvVar = getBooleanFromEnv({
  key: 'MY_ENV_VAR',
  errorMessage: 'MY_ENV_VAR is required for this function',
});
```

Alternatively, if the env variable is optional, they can also provide a default fallback value that will be used if the env variable is not present.

```ts
import { getNumberFromEnv } from '@aws-lambda-powertools/commons/utils/env';

const myEnvVar = getNumberFromEnv({
  key: 'MY_ENV_VAR',
  defaultValue: 42,
});
```

The PR also adds a few higher-level utilities to read slightly more complex values from environment variables.

For example, `getTruthyBooleanFromEnv` and `getFalsyBooleanFromEnv` respectively allow customers to read values like `1`, `y`, `yes`, `on`, `true`, `0`, `n`, `no`, `off`, `false` and cast them to their corresponding `boolean` values.

```ts
import {
  getTruthyBooleanFromEnv,
  getFalsyBooleanFromEnv
} from '@aws-lambda-powertools/commons/utils/env';

process.env.MY_ENV_VAR = 'on';
const myEnvVar = getTruthyBooleanFromEnv({
  key: 'MY_ENV_VAR',
});
// ^ true

process.env.MY_ENV_VAR = 'no';
const myEnvVar = getFalsyBooleanFromEnv({
  key: 'MY_ENV_VAR',
});
// ^ true
```

Likewise, the `isDevMode` helper allows customers to detect if their application is running in a development environment based on the `POWERTOOLS_DEV` env variable, and the `getServiceName` one retrieves the value of the `POWERTOOLS_SERVICE_NAME` one.

[...]

PR #3970, opened 2025-05-23
chore(commons): consolidate boolean parsing

### LINKED ISSUE #3967: Maintenance: consolidate env variable helpers for booleans
### Summary

In #3945 we added some environment variable helpers to the commons packages. A community member @benjymoses provided feedback that some of them were unclear and could be improved.

For example we had a base helper to retrieve and cast boolean values from the environment, but we also had more specific helpers like `getTruthyBooleanFromEnv` and `getFalsyBooleanFromEnv`.

These two specifically represented a state rather than an actual value so at minimum they should've been called `isTruthyBooleanFromEnv` and `isFalsyBooleanFromEnv` since they returned whether the value was truth-y or fals-y rather than the actual value.

On top of that, they were redundant in the sense that the original `getBooleanFromEnv` could've been extended to handle both regular `false`/`true` strings as well as additional values.

### Why is this needed?

So that we can consolidate the implementations and disambiguate the meaning/behavior of these helpers.

_No response_

### Solution

_No response_

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR removes the `getTruthyBooleanFromEnv` and `getFalsyBooleanFromEnv` that were added in #3945 in favor of a more powerful and extended `getBooleanFromEnv` helper that can now handle both regular `true`/`false` values but also optionally other falsy/truthy values like `on`, `off`, `yes`, `no`, and so on.

A longer explanation of the change can be found in the linked issue. Thank you @benjymoses for flagging this and proposing this implementation.

**Issue number:** closes #3967

---


am29d (approved):
Much cleaner version now, thanks for the PR!

PR #3974, opened 2025-05-24
fix(event-handler): fix decorated scope in appsync events

### LINKED ISSUE #3973: Bug: `AppSyncEventsResolver` decorators break scope
### Expected Behavior

When using the `@onPublish` and `@onSubscribed` decorator of the `AppSyncEventsResolver` the decorated methods should preserve the `this` scope of the class they belong to.

This way, these methods can continue accessing class properties and methods.

### Current Behavior

When applying the decorators the decorated method is no longer bound and `this` goes to either the instance of `AppSyncEventsResolver` or `undefined`, thus preventing from using other class methods/properties.

### Code snippet

```ts
const app = new AppSyncEventsResolver({ logger: console });

class Lambda {
  public scope = 'scoped';

  @app.onPublish('/foo')
  public async handleFoo(payload: string) {
    return `${this.scope} ${payload}`;
  }

  public async handler(event: unknown, context: Context) {
    return this.stuff(event, context);
  }

  async stuff(event: unknown, context: Context) {
  }
}
const lambda = new Lambda();
const handler = lambda.handler.bind(lambda);
```

### Steps to Reproduce

N/A

### Possible Solution

_No response_

latest

22.x

### Packaging format used

npm

### Execution logs

```Shell

```

dreamorosi:
### Changes

This PR improves the decorator logic of the `AppSyncEventsResolver` by fixing a fundamental issue that prevented it from being useful.

Prior to this fix the decorators would consistently cause the decorated methods to become unbound and lose the reference to `this`, preventing customers to access other class methods or properties from within the decorated methods.

This was happening because unlike other decorator implementations in the codebase, the one in this area of the codebase works in two steps:
1. at init time the decorator saves a reference of the decorated method for later usage
2. at runtime the `app.resolve` method can call a decorated method depending on routing decisions

By the time the reference is called, the initial scope is lost causing the issue. At the same time, the `this` scope is not yet available at init time - this is because decorators are evaluated before class init - so it's not possible to save a reference or bind `this` along with the the decorated method.

Because of this, we need to get around this decorator design aspect.

Specifically, at runtime (step 2) - which is when `app.resolve()` is called - we introspect the class methods, find their definition, and when we encounter the `.resolve` keyword (which represents the method), we save a reference to `this` for later use. This reference is then used and passed to the decorated method to apply the correct `this` value.

While this is relatively unorthodox, as long as we find one instance of `this` that belongs to the class to which the decorated method belongs it should be fine.

**Issue number:** fixes #3973

---

### DIFF
--- a/packages/event-handler/src/appsync-events/AppSyncEventsResolver.ts
+++ b/packages/event-handler/src/appsync-events/AppSyncEventsResolver.ts
@@ -114,11 +114,11 @@ class AppSyncEventsResolver extends Router {
     if (aggregate) {
       try {
         return {
-          events: await (handler as OnPublishHandlerAggregateFn).apply(this, [
+          events: await (handler as OnPublishHandlerAggregateFn)(
             event.events,
             event,
-            context,
-          ]),
+            context
+          ),
         };
       } catch (error) {
         this.logger.error(`An error occurred in handler ${path}`, error);
@@ -131,11 +131,11 @@ class AppSyncEventsResolver extends Router {
         event.events.map(async (message) => {
           const { id, payload } = message;
           try {
-            const result = await (handler as OnPublishHandlerFn).apply(this, [
+            const result = await (handler as OnPublishHandlerFn)(
               payload,
               event,
-              context,
-            ]);
+              context
+            );
             return {
               id,
               payload: result,
@@ -173,7 +173,7 @@ class AppSyncEventsResolver extends Router {
     }
     const { handler } = routeHandlerOptions;
     try {
-      await (handler as OnSubscribeHandler).apply(this, [event, context]);
+      await (handler as OnSubscribeHandler)(event, context);
     } catch (error) {
       this.logger.error(`An error occurred in handler ${path}`, error);
       if (error instanceof UnauthorizedException) throw error;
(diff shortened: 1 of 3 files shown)


packages/event-handler/src/appsync-events/AppSyncEventsResolver.ts
```diff
@@ -87,11 +94,12 @@ class AppSyncEventsResolver extends Router {
     }
 
     if (isAppSyncEventsPublishEvent(event)) {
-      return await this.handleOnPublish(event, context);
+      return await this.handleOnPublish(event, context, options);
```
svozza, line 97: Is there a reason why we're calling `await` in the return statement? Is it because of this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#improving_stack_trace?

svozza (approved):

packages/event-handler/src/appsync-events/AppSyncEventsResolver.ts
dreamorosi: To be honest I wasn't aware of that, but it's a neat benefit!

I think I added it during dev/testing to be able to inspect the return payload.

PR #3991, opened 2025-05-28
docs(event-handler): add docs page for Bedrock Agent Function

### LINKED ISSUE #3990: Docs: add docs for Bedrock Agents feature

We are getting close to launching a new Event Handler for Bedrock Agent Function calls and we don't have a doc page for it.

_No response_

### How can we improve?

We should add a dedicated documentation page for the new feature prior to launch.

### Got a suggestion in mind?

_No response_

### Acknowledgment

dreamorosi:
### Changes

This PR adds a page to the docs dedicated to the new Event Handler for Bedrock Agent Functions. 

This page borrows from the one being added in aws-powertools/powertools-lambda-python#6564 but deviates a bit due to the fact that we don't support OpenAPI specs for this feature.

This page follows the same structure as similar pages and includes:
- a mermaid diagram to explain the flow (moved the Action Group in the same "box" as the agent)
- a getting started section with SAM & CDK templates to create an agent
- a section that explains how to create a tool, how parameters are handled
- a section about error handling & custom responses
- a section about working with session attributes
- a section about how to access the current event and context

**Issue number:** closes #3990

---


docs/features/event-handler/bedrock-agents.md
```diff
@@ -0,0 +1,142 @@
...
+
+    === "AWS CDK"
+
+        ```typescript
+        --8<-- "examples/snippets/event-handler/bedrock-agents/templates/gettingStartedCdk.ts"
+        ```
+
+### Usage
+
+To create an agent, use the `BedrockAgentFunctionResolver` to register your tools and handle the requests. The resolver will automatically parse the request, route it to the appropriate function, and return a well-formed response that includes the tool's output and any existing session attributes.
```
leandrodamascena:
```suggestion
Use the `BedrockAgentFunctionResolver` to register your tools and handle the requests to your Lambda function. The resolver will automatically parse the request, route it to the appropriate function, and return a well-formed response that includes the tool's output and any existing session attributes.
```

docs/features/event-handler/bedrock-agents.md
```diff
@@ -0,0 +1,142 @@
...
+
+    ```typescript hl_lines="24-27"
+    --8<-- "examples/snippets/event-handler/bedrock-agents/sessionAttributes.ts"
+    ```
+
+### Logging
+
+By default, the `BedrockAgentFunctionResolver` uses the global `console` logger and emits only warnings and errors.
+
+You can change this behavior by passing a custom logger instance to the `BedrockAgentFunctionResolver` constructor and setting its log level. Alternatively, you can also enable [Lambda Advanced Logging Controls](https://docs.aws.amazon.com/lambda/latest/dg/monitoring-cloudwatchlogs-advanced.html) and setting the log level to `DEBUG`.
```
leandrodamascena, line 132: Why not add the Powertools log level explanation here instead of ALC? Setting the Powertools log level will solve the problem both in local environment (testing, mock API and other things) and in production, running on Lambda. ALC will solve it only in production.

I would not mention ALC in this doc.

docs/features/event-handler/bedrock-agents.md
```diff
@@ -0,0 +1,142 @@
...
+
+When debug logging is enabled, the resolver will emit logs that show the underlying handler registration and the routing process. This is useful for understanding how the agent resolves the tools and routes the requests.
+
+For example, when using the [Powertools for AWS Lambda logger](../logger.md), you can set the `LOG_LEVEL` to `DEBUG` in your environment variables or at the logger level and pass the logger instance to the `BedrockAgentFunctionResolver` constructor to enable debug logging.
```
leandrodamascena, line 136: If you accept the previous suggestion, you can remove it.

docs/features/event-handler/bedrock-agents.md
```diff
@@ -0,0 +1,142 @@
...
+
+Currently, we don't support parsing `array` types, so you will receive them as strings.
+
+=== "Define a tool"
+
+  ```typescript hl_lines="4 6 20"
+  --8<-- "examples/snippets/event-handler/bedrock-agents/gettingStartedFunctionsTool.ts"
+  ```
+  
+  1. The `description` field is optional, but highly recommended in the action group definition.
```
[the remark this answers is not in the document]
dreamorosi: Yea, good point. Can't add it until tomorrow when I get on a laptop. But if you wanna suggest a line like you did with the previous ones I can accept it.

The gist of why we need it is that it gives the LLM better context on the tool and how to use it.

leandrodamascena (changes requested):
Hey @dreamorosi! i just left some comments that probably make sense you review it.

docs/features/event-handler/bedrock-agents.md
dreamorosi: In [this implementation](https://github.com/aws-powertools/powertools-lambda-typescript/blob/583e3ae9a5095a1185b2bd9dff31e6f3f50ff577/packages/event-handler/src/bedrock-agent/BedrockAgentFunctionResolver.ts#L58), setting the ALC env variable (`AWS_LAMBDA_LOG_LEVEL`) would also work locally.

Contrary to Python, there's no way for customers to set a global debug log level. One way to do it for TS customers in this utility, is to use a logger library (i.e. PT Logger, which is shown below). Another, if they don't want to add a logging library is to enable ALC. I've seen many TS customers using and being confused about how ALC interacts with Powertools, I would prefer to leave the mention unless you feel strongly against it.

Powertools logger is shown prominently right after together with the only code example about logging.

leandrodamascena: Thanks for detailed explanation @dreamorosi! Keep it!!

docs/features/event-handler/bedrock-agents.md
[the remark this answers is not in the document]
dreamorosi: Ok, added.

Issue #1166, opened 2022-11-14
Feature request: GraphQL API Event Handler


FilipPyrek:
### Use case

> We discussed it here on [Twitter](https://twitter.com/dreamorosi/status/1590685049015308288) with @dreamorosi  and the outcome was that Lambda Powertools for Python already has this feature and this feature request is to re-implement similar feature also in Lambda Powertools for TypeScript.

The problem in our case is that when we use lambda in combination with GraphQL's nested resolvers feature (and mainly AppSync's implementation of nested resolvers) the lambda cold starts begin to stack-up which leads to couple of seconds of aggregated cold starts of all nested resolvers which is very unpleasant.

### Solution/User Experience

GraphQL API handler in Lambda Powertools for Python has a resolver library which does the nested resolution inside a single lambda function and instead of creating the abstraction on infrastructure level, as AppSync does, it creates it on source code level. This resolution can be done thanks to informations from [`$ctx.info`](https://docs.aws.amazon.com/appsync/latest/devguide/resolver-context-reference.html).

This way lambda resolvers would be only for top-level Mutation/Query fields and the rest would be done inside lambda function(s).

And it will reduce any cold starts to minimum. (= cold start of single lambda function)

### Alternative solutions

```Markdown
1. Query all the data and return it all inside Query/Mutation top level field resolver lambda functions - this way we loose advantage of GraphQL where clients can select what part of data they are interested in, so we can query only the relevant data on backend. 

2. Create simple custom nasty resolver based on `if` statements to distinguish what fields the client is requesting - very nasty source code.

3. Try all possible optimisations to reduce the cold starts. That can be very hard and even in some cases teams can hit a limit where it's not possible to optimize any further..
```

### Acknowledgment

- [X] This feature request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/#tenets)
- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [Java](https://github.com/awslabs/aws-lambda-powertools-java/)

dreamorosi:
Hi @FilipPyrek thank you for opening this feature request!

As I said on Twitter, I think this is an interesting feature that we should consider adding to this library at some point.

At the moment we are focused on implementing Parameters and Idempotency, as well as finishing Lambda Layers and investigating ESM support.

[...]

PR #3999, opened 2025-06-01
feat(event-handler): add single resolver functionality for AppSync GraphQL API

arnabrahman:
## Summary
This PR will add the ability to register single resolvers for AppSync GraphQL API. 
### Changes

- Add new `appsync-graphql` utility inside `event-handler`
- For this PR, we are only adding the single resolver functionality for AppSync GraphQL API
- Followed this [suggestion](https://github.com/aws-powertools/powertools-lambda-typescript/issues/1166#issuecomment-2883461970) and same coding structure for `appsync-events`
- TypeScript does not natively support unpacking an object as named arguments like Python’s `kwargs`. We must match the function signature accordingly. So [this](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/event_handler/appsync.py#L190) is not possible, I resolved the arguments as an object.
- 
Example usage of the utility: Followed [this](https://docs.aws.amazon.com/appsync/latest/devguide/tutorial-lambda-resolvers-js.html)
```ts

const logger = new Logger();
const app = new AppSyncGraphQLResolver({
  logger: logger
});

const posts: Record<number, Record<string, unknown>> = {
  1: { id: '1', title: 'First book', author: 'Author1', url: 'https://amazon.com/', content: 'SAMPLE TEXT AUTHOR 1 SAMPLE TEXT AUTHOR 1 SAMPLE TEXT AUTHOR 1 SAMPLE TEXT AUTHOR 1 SAMPLE TEXT AUTHOR 1 SAMPLE TEXT AUTHOR 1', ups: '100', downs: '10', },
  2: { id: '2', title: 'Second book', author: 'Author2', url: 'https://amazon.com', content: 'SAMPLE TEXT AUTHOR 2 SAMPLE TEXT AUTHOR 2 SAMPLE TEXT', ups: '100', downs: '10', },
  3: { id: '3', title: 'Third book', author: 'Author3', url: null, content: null, ups: null, downs: null },
  4: { id: '4', title: 'Fourth book', author: 'Author4', url: 'https://www.amazon.com/', content: 'SAMPLE TEXT AUTHOR 4 SAMPLE TEXT AUTHOR 4 SAMPLE TEXT AUTHOR 4 SAMPLE TEXT AUTHOR 4 SAMPLE TEXT AUTHOR 4 SAMPLE TEXT AUTHOR 4 SAMPLE TEXT AUTHOR 4 SAMPLE TEXT AUTHOR 4', ups: '1000', downs: '0', },
  5: { id: '5', title: 'Fifth book', author: 'Author5', url: 'https://www.amazon.com/', content: 'SAMPLE TEXT AUTHOR 5 SAMPLE TEXT AUTHOR 5 SAMPLE TEXT AUTHOR 5 SAMPLE TEXT AUTHOR 5 SAMPLE TEXT', ups: '50', downs: '0', },
};


class Lambda {
  @app.resolver({ fieldName: 'getPost' })
  async getPost({ id }: { id: number; }) {
    return posts[id];

  }
  @app.resolver({ fieldName: 'addPost', typeName: 'Mutation' })
  async addPost({ id, title, author, url, content }: { id:number, title: string, author: string, url: string, content: string, }) {
    posts[id] = { id: String(id), title, author, url, content };
    return posts[id];
  }


  async handler(event: unknown, context: Context) {
    return app.resolve(event, context, {
      scope: this
    });
  }
}

const lambda = new Lambda();
```

```gql
type Post {
	id: ID!
	author: String!
	title: String
	content: String
	url: String
	ups: Int
	downs: Int
	relatedPosts: [Post]
}

type Mutation {
	addPost(
		id: ID!,
		author: String!,
		title: String,
		content: String,
		url: String
	): Post!
}

type Query {
	getPost(id: ID!): Post
	allPosts: [Post]
}

schema {
	query: Query
	mutation: Mutation
}
```

**Issue number:** #1166 

---


arnabrahman:
I have also begun exploring the implementation of the batch resolver. I am trying the powertools-python batch resolver [example](https://docs.powertools.aws.dev/lambda/python/3.13.0/core/event_handler/appsync/#batch-resolvers). 

Schema:
```gql
type Post {
	post_id: ID!
	title: String
	author: String
	relatedPosts: [Post]
}

type Query {
	getPost(post_id: ID!): Post
}

schema {
	query: Query
}
```

```python
from __future__ import annotations

from typing import Any, TypedDict

from aws_lambda_powertools import Logger
from aws_lambda_powertools.event_handler import AppSyncResolver
from aws_lambda_powertools.utilities.data_classes import AppSyncResolverEvent
from aws_lambda_powertools.utilities.typing import LambdaContext

app = AppSyncResolver()
logger = Logger()

posts = {
    "1": { "post_id": "1", "title": "First book", "author": "Author1", },
    "2": { "post_id": "2", "title": "Second book", "author": "Author2", },
    "3": { "post_id": "3", "title": "Third book", "author": "Author3", },
    "4": { "post_id": "4", "title": "Fourth book", "author": "Author4", },
    "5": { "post_id": "5", "title": "Fifth book", "author": "Author5", }
}

posts_related = {
    "1": [posts["4"]],
    "2": [posts["3"], posts["5"]],
    "3": [posts["2"], posts["1"]],
    "4": [posts["2"], posts["1"]],
    "5": [],
}


def search_batch_posts(posts: list) -> dict[str, Any]:
    return {post_id: posts_related.get(post_id) for post_id in posts}

class Post(TypedDict, total=False):
    post_id: str
    title: str
    author: str

@app.resolver(type_name="Query", field_name="getPost")
def get_post(
    post_id: str = ""
) -> Post:
    return posts.get(post_id, {})


@app.batch_resolver(type_name="Query", field_name="relatedPosts")
def related_posts(event: list[AppSyncResolverEvent]) -> list[Any]:  
    # Extract all post_ids in order
    post_ids: list = [record.source.get("post_id") for record in event]  

    # Get unique post_ids while preserving order
    unique_post_ids = list(dict.fromkeys(post_ids))
    # Fetch posts in a single batch operation
    fetched_posts = search_batch_posts(unique_post_ids)

    # Return results in original order
    return [fetched_posts.get(post_id) for post_id in post_ids]


def lambda_handler(event, context: LambdaContext) -> dict:
    return app.resolve(event, context)
```

If I try this query

```gql
query MyQuery {
  getPost(post_id: "2") {
    relatedPosts {
      post_id
      author
      relatedPosts {
        post_id
        author
      }
    }
  }
}
```
There is an error: `[ERROR] ResolverNotFoundError: No resolver found for 'Post.relatedPosts' Traceback (most recent call last):`

If I change the typeName to `Post`, it works -> `@app.batch_resolver(type_name="Post", field_name="relatedPosts")`

Result:
```json
{
  "data": {
    "getPost": {
      "relatedPosts": [
        {
          "post_id": "3",
          "author": "Author3",
          "relatedPosts": [
            {
              "post_id": "2",
              "author": "Author2"
            },
            {
              "post_id": "1",
              "author": "Author1"
            }
          ]
        },
        {
          "post_id": "5",
          "author": "Author5",
          "relatedPosts": []
        }
      ]
    }
  }
}
```
The reason is probably because [parentTypeName](https://docs.powertools.aws.dev/lambda/python/3.13.0/core/event_handler/appsync/#advanced_batch_resolver_payloadjson) value is `Post`.

 Not sure what I am doing wrong here. @dreamorosi @leandrodamascena

dreamorosi:
Hi @arnabrahman, thank you so much for the PR.

I'm going to need a bit more time to review this, I was out of office yesterday and catching up with things today.

I expect to provide a first review tomorrow morning.

packages/event-handler/src/types/appsync-graphql.ts
```diff
@@ -0,0 +1,139 @@
+import type { RouteHandlerRegistry } from '../appsync-graphql/RouteHandlerRegistry.js';
+
+// biome-ignore lint/suspicious/noExplicitAny: We intentionally use `any` here to represent any type of data and keep the logger is as flexible as possible.
+type Anything = any;
+
+/**
+ * Interface for a generic logger object.
+ */
+type GenericLogger = {
```
dreamorosi: You might've branched out before this was merged, but we moved this type and the one above (`Anything`) to a separate file so we can share it (see [here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/event-handler/src/types/common.ts)).

If possible let's use them instead.

packages/event-handler/src/appsync-graphql/Router.ts
```diff
@@ -0,0 +1,229 @@
...
+   */
+  protected readonly isDev: boolean = false;
+  /**
+   * The environment variables service instance.
+   */
+  protected readonly envService: EnvironmentVariablesService;
+
+  public constructor(options?: GraphQlRouterOptions) {
+    this.envService = new EnvironmentVariablesService();
+    const alcLogLevel = this.envService.get('AWS_LAMBDA_LOG_LEVEL');
```
dreamorosi: We are in the process of simplifying the way we handle environment variables, if possible let's switch to [this pattern](https://github.com/aws-powertools/powertools-lambda-typescript/blob/351e88e34fb5650239a97c400e3a339c2c60f2a1/packages/event-handler/src/bedrock-agent/BedrockAgentFunctionResolver.ts#L54-L57) instead (you might have to rebase/rebuild the commons package for it to be available).

dreamorosi (commented):
Amazing work as usual @arnabrahman - I haven't tested this on an AppSync API (will try tomorrow) but the implementation looks good. I've only left a couple minor comments.

I'll request also @leandrodamascena's review (tomorrow is ok), since he knows more about this resolver than me.

Good idea also splitting the PR for single vs batch resolvers.

leandrodamascena:
> I have also begun exploring the implementation of the batch resolver. I am trying the powertools-python batch resolver [example](https://docs.powertools.aws.dev/lambda/python/3.13.0/core/event_handler/appsync/#batch-resolvers).
> 
> Schema:
> 
> ```graphql
> type Post {
> 	post_id: ID!
> 	title: String
> 	author: String
> 	relatedPosts: [Post]
> }
> 
> type Query {
> 	getPost(post_id: ID!): Post
> }
> 
> schema {
> 	query: Query
> }
> ```
> 
> ```python
> from __future__ import annotations
> 
> from typing import Any, TypedDict
> 
> from aws_lambda_powertools import Logger
> from aws_lambda_powertools.event_handler import AppSyncResolver
> from aws_lambda_powertools.utilities.data_classes import AppSyncResolverEvent
> from aws_lambda_powertools.utilities.typing import LambdaContext
> 
> app = AppSyncResolver()
> logger = Logger()
> 
> posts = {
>     "1": { "post_id": "1", "title": "First book", "author": "Author1", },
>     "2": { "post_id": "2", "title": "Second book", "author": "Author2", },
>     "3": { "post_id": "3", "title": "Third book", "author": "Author3", },
>     "4": { "post_id": "4", "title": "Fourth book", "author": "Author4", },
>     "5": { "post_id": "5", "title": "Fifth book", "author": "Author5", }
> }
> 
> posts_related = {
>     "1": [posts["4"]],
>     "2": [posts["3"], posts["5"]],
>     "3": [posts["2"], posts["1"]],
>     "4": [posts["2"], posts["1"]],
>     "5": [],
> }
> 
> 
> def search_batch_posts(posts: list) -> dict[str, Any]:
>     return {post_id: posts_related.get(post_id) for post_id in posts}
> 
> class Post(TypedDict, total=False):
>     post_id: str
>     title: str
>     author: str
> 
> @app.resolver(type_name="Query", field_name="getPost")
> def get_post(
>     post_id: str = ""
> ) -> Post:
>     return posts.get(post_id, {})
> 
> 
> @app.batch_resolver(type_name="Query", field_name="relatedPosts")
> def related_posts(event: list[AppSyncResolverEvent]) -> list[Any]:  
>     # Extract all post_ids in order
>     post_ids: list = [record.source.get("post_id") for record in event]  
> 
>     # Get unique post_ids while preserving order
>     unique_post_ids = list(dict.fromkeys(post_ids))
>     # Fetch posts in a single batch operation
>     fetched_posts = search_batch_posts(unique_post_ids)
> 
>     # Return results in original order
>     return [fetched_posts.get(post_id) for post_id in post_ids]
> 
> 
> def lambda_handler(event, context: LambdaContext) -> dict:
>     return app.resolve(event, context)
> ```
> 
> If I try this query
> 
> ```graphql
> query MyQuery {
>   getPost(post_id: "2") {
>     relatedPosts {
>       post_id
>       author
>       relatedPosts {
>         post_id
>         author
>       }
>     }
>   }
> }
> ```
> 
> There is an error: `[ERROR] ResolverNotFoundError: No resolver found for 'Post.relatedPosts' Traceback (most recent call last):`
> 
> If I change the typeName to `Post`, it works -> `@app.batch_resolver(type_name="Post", field_name="relatedPosts")`
> 
> Result:
> 
> ```json
> {
>   "data": {
>     "getPost": {
>       "relatedPosts": [
>         {
>           "post_id": "3",
>           "author": "Author3",
>           "relatedPosts": [
>             {
>               "post_id": "2",
>               "author": "Author2"
>             },
>             {
>               "post_id": "1",
>               "author": "Author1"
>             }
>           ]
>         },
>         {
>           "post_id": "5",
>           "author": "Author5",
>           "relatedPosts": []
>         }
>       ]
>     }
>   }
> }
> ```
> 
> The reason is probably because [parentTypeName](https://docs.powertools.aws.dev/lambda/python/3.13.0/core/event_handler/appsync/#advanced_batch_resolver_payloadjson) value is `Post`.
> 
> Not sure what I am doing wrong here. @dreamorosi @leandrodamascena

Thanks for finding a bug in the Python documentation, @arnabrahman! You always exceed our standards with these catches ❤️ ! Yeah, the valid value is `Post` and not `Query` because when resolving a batch schema the initial `Query` (getPost) make a `Post` to `relatedPosts`. Tbh this should be `Query` as well, but in this case I think we need to pass the parent post as parameter and would difficult the example.

Thanks

packages/event-handler/src/appsync-graphql/Router.ts
```diff
@@ -0,0 +1,233 @@
...
+   * const lambda = new Lambda();
+   * export const handler = lambda.handler.bind(lambda);
+   * ```
+   *
+   * @param fieldName - The name of the Query field to register the handler for.
+   * @param handler - The handler function to be called when the event is received.
+   * @param options - Optional route options.
+   * @param options.typeName - The name of the GraphQL type to use for the resolver (defaults to 'Query').
+   */
+  public onQuery(
```
dreamorosi: Can we make this and the other generic so that users can pass the type for the arguments?

For example:

```ts
app.onQuery<{ postId: string }>('getPost', async ({ postId }) => {
                                                // ^ this is now type string
});
```

We did something similar for this other [handler](https://github.com/aws-powertools/powertools-lambda-typescript/blob/c74a67c0271147ef59c758c02372e1cb671adf80/packages/event-handler/src/bedrock-agent/BedrockAgentFunctionResolver.ts#L146).

It'll work only for this notation and not work with decorators, but that's ok.

packages/event-handler/src/appsync-graphql/Router.ts
```diff
@@ -0,0 +1,233 @@
...
+    const alcLogLevel = getStringFromEnv({
+      key: 'AWS_LAMBDA_LOG_LEVEL',
+      defaultValue: '',
+    });
+    this.logger = options?.logger ?? {
+      debug: alcLogLevel === 'DEBUG' ? console.debug : () => undefined,
+      error: console.error,
+      warn: console.warn,
+    };
+    this.onQueryRegistry = new RouteHandlerRegistry({
```
dreamorosi: I've been looking [at the implementation in Pytho](https://github.com/aws-powertools/powertools-lambda-python/blob/14054ab50e281b59c8e813fc1eb52f986888a578/aws_lambda_powertools/event_handler/graphql_appsync/_registry.py#L47-L51)n and I think we might need to also use a single registry rather than two.

If we keep in mind only handlers for specific fields in queries and mutations this implementation works - and is actually better - but in practice customers can also set handlers on custom types or not use a type at all.

For example, given this schema:

```graphql
type Merchant {
    id: String!
    name: String!
    description: String
    locations: [Location]
}
```

customers should be able to set a resolver like:

```ts
app.resolver(async () => {
  // ...
}, {
  typeName: 'Merchant',
  fieldName: 'location',
});
```

as well as one like:

```ts
app.resolver(async () => {
  // ...
}, {
  fieldName: 'location',
});
```

The last one is referred to as ["nested mappings"](https://docs.powertools.aws.dev/lambda/python/latest/core/event_handler/appsync/#nested-mappings) in the Python implementation.

Because of this, I think it makes sense to use a single registry and use field and type names as keys - I don't have a preference around keeping them separate keys in the registry like we're doing now, or concatenating them like Python does - whatever is faster / easier.

dreamorosi (changes requested):
Thanks for your patience.

I ran some tests deploying this with an AppSync API and it works well, good work here!

I have left a couple comments:
- one around making the methods that register handlers generic to improve experience
- another to create a more generic `resolver` method that can handle other cases covered by the Python implementation

Besides this, I'd also see if we can add some helper functions like [these](https://docs.powertools.aws.dev/lambda/python/latest/core/event_handler/appsync/#scalar-functions) for scalar types in this PR.

We can address the `batchResolver` method in a separate PR.

Thanks again!

packages/event-handler/src/appsync-graphql/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * We use this function to ensure that the event is an object and has the required properties
+ * without adding any dependency.
+ *
+ * @param event - The incoming event to check
+ */
+const isAppSyncGraphQLEvent = (
+  event: unknown
+): event is AppSyncGraphQLEvent => {
+  if (typeof event !== 'object' || event === null || !isRecord(event)) {
```
dreamorosi, line 15: This validation doesn't work now - I wasn't able to see the `selectionSetList` field in my tests.

This did:
```ts
if (!isRecord(event)) {
    return false;
}
return (
    'identity' in event &&
    'result' in event &&
    isRecord(event.request) &&
    isRecord(event.request.headers) &&
    'domainName' in event.request &&
    isRecord(event.info) &&
    isString(event.info.fieldName) &&
    isString(event.info.parentTypeName) &&
    isRecord(event.info.variables) &&
    'error' in event &&
    'prev' in event &&
    isRecord(event.stash) &&
    Array.isArray(event.outErrors) &&
    isRecord(event.arguments) &&
    'source' in event
);
```

Note that `isRecord` and `isString` already imply that the value is present (aka not `undefined`) so we can probably skip the `'field' in event && isRecord(event.field)` and just use `isRecord(event.field)` unless TypeScript complains.

arnabrahman:
@dreamorosi Thanks for taking the time to review this. I will look at the suggestions and get back on this.

packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
```diff
@@ -0,0 +1,161 @@
...
+   * @throws {ResolverNotFoundException} If no resolver is registered for the given field and type.
+   */
+  async #executeSingleResolver(event: AppSyncGraphQLEvent): Promise<unknown> {
+    const { fieldName, parentTypeName: typeName } = event.info;
+    const queryHandlerOptions = this.onQueryRegistry.resolve(
+      typeName,
+      fieldName
+    );
+    if (queryHandlerOptions) {
+      return await queryHandlerOptions.handler.apply(this, [event.arguments]);
```
svozza: The caller of this function does the awaiting so there's no need to await here. Likewise on L136

packages/event-handler/tests/unit/appsync-graphql/AppSyncGraphQLResolver.test.ts
```diff
@@ -0,0 +1,174 @@
...
+    const result = await app.resolve(null, context);
+
+    // Assess
+    expect(console.warn).toHaveBeenCalledWith(
+      'Received an event that is not compatible with this resolver'
+    );
+    expect(result).toBeUndefined();
+  });
+
+  it('throw error if there are no onQuery handlers', async () => {
```
svozza: Nit: should be 'throws`.

packages/event-handler/tests/unit/appsync-graphql/AppSyncGraphQLResolver.test.ts
```diff
@@ -0,0 +1,174 @@
...
+    // Act && Assess
+    await expect(
+      app.resolve(onQueryEventFactory('getPost'), context)
+    ).rejects.toThrow(
+      new ResolverNotFoundException('No resolver found for Query-getPost')
+    );
+    expect(console.error).toHaveBeenCalled();
+  });
+
+  it('throw error if there are no onMutation handlers', async () => {
```
svozza: s/throw/throws

packages/event-handler/tests/unit/appsync-graphql/AppSyncGraphQLResolver.test.ts
```diff
@@ -0,0 +1,174 @@
+import context from '@aws-lambda-powertools/testing-utils/context';
+import {
+  onMutationEventFactory,
+  onQueryEventFactory,
+} from 'tests/helpers/factories.js';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { AppSyncGraphQLResolver } from '../../../src/appsync-graphql/AppSyncGraphQLResolver.js';
+import { ResolverNotFoundException } from '../../../src/appsync-graphql/errors.js';
+
+describe('Class: AppSyncGraphQLResolver', () => {
```
svozza, line 8: We should add a test where we don't provide a logger, just to ensure that our fallback works:

```ts
it('logs only warnings and errors using global console object if no logger supplied', async () => {

  // rest of test

  // Assess
  expect(console.debug).not.toHaveBeenCalled();
})

```

packages/event-handler/tests/unit/appsync-graphql/Router.test.ts
```diff
@@ -0,0 +1,144 @@
+import { Router } from 'src/appsync-graphql/Router.js';
```
svozza, line 1: Similar to previous comment, we should test cresting the Router without supplying a `logger` object to the constructor.

packages/event-handler/src/appsync-graphql/Router.ts
```diff
@@ -0,0 +1,233 @@
...
+   *
+   * As a decorator:
+   *
+   * @example
+   * ```ts
+   * import { AppSyncGraphQLResolver } from '@aws-lambda-powertools/event-handler/appsync-graphql';
+   *
+   * const app = new AppSyncGraphQLResolver();
+   *
+   * class Lambda {
```
svozza, line 108: Given https://github.com/aws-powertools/powertools-lambda-typescript/issues/3973, do we want to have decorator support in this PR or wait until we have a generic solution to the `this` binding problem that we can use across all these event-handler types.

svozza:
This is great work. One question I have: both the AppSync events and Bedrock agents event handlers pass the `event` and the `context` objects into their resolvers and I think we should do the same here. Here's an example of what I mean:

```ts
it('tool function has access to the event variable', async () => {
    // Prepare
    const app = new BedrockAgentFunctionResolver();

    app.tool(
      async (_params, options) => {
        return options?.event;
      },
      {
        name: 'event-accessor',
        description: 'Accesses the event object',
      }
    );

    const event = createEvent('event-accessor');

    // Act

    // Assess
    expect(result.response.function).toEqual('event-accessor');
    expect(result.response.functionResponse.responseBody.TEXT.body).toEqual(
      JSON.stringify(event)
    );
  });
  ```

File: packages/event-handler/src/appsync-graphql/Router.ts (hunk 11 above)
Comment (dreamorosi) [reply packages/event-handler/src/appsync-graphql/Router.ts:108]: Good point.

We're not adding this resolver to the exports in the `package.json` yet, so we still have some time to decide while we close this PR and work on the batch resolver one (follow up to this).

I was hoping to work on #3973 sometime next week and decide if we want to move forward there or not, and then either leave this as-is, or remove it like we did with Bedrock Agents.

Comment (svozza) [reply packages/event-handler/src/appsync-graphql/Router.ts:108]: Sounds good.

File: packages/event-handler/tests/unit/appsync-graphql/AppSyncGraphQLResolver.test.ts (hunk 9 above)
Comment (svozza) [reply packages/event-handler/tests/unit/appsync-graphql/AppSyncGraphQLResolver.test.ts:8]: Also we should have a test to verify we respect the environment variable `AWS_LAMBDA_LOG_LEVEL` being set to `DEBUG`. Example here: https://github.com/aws-powertools/powertools-lambda-typescript/blob/d124369db6ec63c9610ea4a01c30bee5560f4d8d/packages/event-handler/tests/unit/bedrock-agent/BedrockAgentFunctionResolver.test.ts#L98

File: packages/event-handler/tests/unit/appsync-graphql/AppSyncGraphQLResolver.test.ts [hunk 12]
```diff
@@ -0,0 +1,283 @@
...
+      onGraphqlEventFactory('addPost', 'Mutation', {
+        title: 'Post Title',
+        content: 'Post Content',
+      }),
+      context
+    );
+
+    // Assess
+    expect(console.debug).not.toHaveBeenCalledWith();
+    expect(console.debug).not.toHaveBeenCalledWith();
```
svozza: How come we're calling this twice?

packages/event-handler/src/types/appsync-graphql.ts
```diff
@@ -0,0 +1,127 @@
...
+// #region Events
+
+/**
+ * Event type for AppSync GraphQL.
+ *
+ * https://docs.aws.amazon.com/appsync/latest/devguide/resolver-context-reference.html
+ *
+ * For strongly typed validation and parsing at runtime, check out the `@aws-lambda-powertools/parser` package.
+ */
+type AppSyncGraphQLEvent = {
```
svozza: This event type is already defined in the `aws-lambda` types package: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/trigger/appsync-resolver.d.ts

packages/event-handler/src/appsync-graphql/Router.ts
```diff
@@ -0,0 +1,169 @@
...
+   * Whether the router is running in development mode.
+   */
+  protected readonly isDev: boolean = false;
+  /**
+   * The environment variables service instance.
+   */
+  protected readonly envService: EnvironmentVariablesService;
+
+  public constructor(options?: GraphQlRouterOptions) {
+    this.envService = new EnvironmentVariablesService();
```
svozza: Now we're using the utilities in `@aws-lambda-powertools/commons/utils/env` we should be able to remove this.

packages/event-handler/src/appsync-graphql/Router.ts
```diff
@@ -0,0 +1,169 @@
...
+    });
+    this.logger = options?.logger ?? {
+      debug: alcLogLevel === 'DEBUG' ? console.debug : () => undefined,
+      error: console.error,
+      warn: console.warn,
+    };
+    this.resolverRegistry = new RouteHandlerRegistry({
+      logger: this.logger,
+    });
+    this.isDev = this.envService.isDevMode();
```
svozza: Let's use the standalone utility function not the method from envService: https://github.com/aws-powertools/powertools-lambda-typescript/blob/66fb73eab20f46b6deddc8a7037d370595733125/packages/commons/src/envUtils.ts#L221C7-L221C16

packages/event-handler/src/appsync-graphql/Router.ts
```diff
@@ -0,0 +1,169 @@
...
+        fieldName,
+        handler: handler as ResolverHandler,
+        typeName,
+      });
+
+      return;
+    }
+
+    const resolverOptions = handler;
+    return (_target, _propertyKey, descriptor: PropertyDescriptor) => {
```
svozza: The decorator implementation will likely have to change to reflect this fix: https://github.com/aws-powertools/powertools-lambda-typescript/pull/3974.

packages/event-handler/src/appsync-graphql/utils.ts
arnabrahman: Hey, can you explain this a bit? I didn't quite understand what you meant by not seeing `selectionSetList` in your tests. Apart from that, I’ve addressed all the other feedback.

svozza (approved):
This is fantastic work, all my comments have been addressed. Really looking forward to using this feature myself!

packages/event-handler/src/appsync-graphql/utils.ts
dreamorosi: Hi Sorry I wasn't clear in my explanation.

I meant to convey two separate points, let me try again. The `isAppSyncGraphQLEvent` function has two separate checks:
- whether the `event` is defined, not `null`, and a record
- whether certain fields within the `event` payload are present and have specific shapes

The first part of the comment was referring to the first check. Rather than using this statement `if (typeof event !== 'object' || event === null || !isRecord(event)) {`, it should be enough to just use this `if (!isRecord(event)) {`.

This is because the `isRecord` already asserts the type and checks that it's not `null`.

The second part of the comment was instead focused on the content of the `event` payload. Specifically, in the checks we're checking for the presence of a field named `selectionSetList` and asserting that each one of its items is a string.

This field however is not always present, or at least it doesn't always contain strings. I am not 100% sure of the specifics, but in my brief test it didn't contain any string (aka field). This might be due to what's described [here](https://docs.aws.amazon.com/appsync/latest/devguide/resolver-context-reference-js.html#aws-appsync-resolver-context-reference-info-js):
> `selectionSetList` exposes only fields that belong to the current type. If the current type is an interface or union, only selected fields that belong to the interface are exposed.

---

Since the first one is a nitpick, and the second requires more testing - I'm inclined to merge the PR and potentially amend the `selectionSetList` check after I've done more test in the next couple days.

dreamorosi (approved):
Thank you for the amazing work on this utility, I really appreciate the effort and the level of quality.

I'm aiming at writing the docs for this feature tomorrow and release this on Wednesday 🎉 

PS: I left [one last comment](https://github.com/aws-powertools/powertools-lambda-typescript/pull/3999#discussion_r2175695706) to clarify what I meant in one of the review items, but I don't think we should block the PR over it.

PR #4068, opened 2025-06-19
feat(kafka): lazily deserialize key/value/headers

### LINKED ISSUE #4067: Feature request: decode key/val lazily to allow error handling
### Use case

When working with the recently merged `kafkaConsumer`, as a customer I want to be able to handle decoding errors and potentially take additional steps to log, persist, or redirect messages that failed to decode.

The current implementation eagerly decodes keys and values before the customer handler runs, which makes it impossible for customers to handle errors, short of wrapping it in a global `try/catch`. 

### Solution/User Experience

Customers should be able to do this:

```ts
const handler = kafkaConsumer(
  async (event) => {
    for (const record of event.records) {
      try {
        const { value, key } = record; // decoding & parsing now happens here when accessing
        // ... your business logic
      } catch (error) {
        // handle error, maybe send to a DLQ
      }
    }
  },
  {
    value: {
      type: SchemaType.JSON,
      parserSchema: valueZodSchema,
    },
    key: {
      type: SchemaType.JSON,
      parserSchema: keyZodSchema,
    },
  }
);
```

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR refactors the way keys, values, and headers are deserialized and parsed so that they are processed lazily when the customer accesses them, and not eagerly before they're passed to the AWS Lambda handler.

This approach allows customers to write code like the one below, and introduce custom error handling for different types of errors:

```ts
const handler = kafkaConsumer(
  async (event) => {
    for (const record of event.records) {
      try {
        const { value, key } = record; // decoding & parsing now happens here when accessing
        // ... your business logic
      } catch (error) {
        // handle error types, maybe send to a DLQ
      }
    }
  },
  {
    value: {
      type: SchemaType.JSON,
      parserSchema: valueZodSchema,
    },
    key: {
      type: SchemaType.JSON,
      parserSchema: keyZodSchema,
    },
  }
);
```

On a more technical note, the lazy processing relies on custom `get`ter functions for objects. This allows us to run custom logic just in time when the customer accesses a property.

The downside of this approach is that these `get` functions can only be synchronous. This is not an issue when it comes to deserializing since all these operations are synchronous, however since we were dynamically importing some of the deserializers to avoid introducing extra deps for customers I had to do a significant refactor in the logic that chooses and loads the deserializers to load the extra modules asynchronously in [closures](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Closures) while staying sync.

As part of the refactor I also improved the error handling to make it consistent and added some unit test cases. 

**Issue number:** closes #4067

---

### DIFF
--- a/packages/kafka/src/consumer.ts
+++ b/packages/kafka/src/consumer.ts
@@ -2,14 +2,19 @@ import type { AsyncHandler } from '@aws-lambda-powertools/commons/types';
 import { isNull, isRecord } from '@aws-lambda-powertools/commons/typeutils';
 import type { StandardSchemaV1 } from '@standard-schema/spec';
 import type { Context, Handler } from 'aws-lambda';
+import { deserialize as deserializeJson } from './deserializer/json.js';
+import { deserialize as deserializePrimitive } from './deserializer/primitive.js';
+import { deserialize as deserializeProtobuf } from './deserializer/protobuf.js';
 import {
   KafkaConsumerAvroMissingSchemaError,
+  KafkaConsumerDeserializationError,
   KafkaConsumerParserError,
   KafkaConsumerProtobufMissingSchemaError,
 } from './errors.js';
 import type {
   ConsumerRecord,
   ConsumerRecords,
+  Deserializer,
   Record as KafkaRecord,
   MSKEvent,
   SchemaConfig,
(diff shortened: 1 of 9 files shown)


leandrodamascena:
Taking a look.

leandrodamascena (approved):
Great improve @dreamorosi! APPROVED!

PR #4111, opened 2025-07-01
feat(event-handler): support `onQuery` and `onMutation` handlers

### LINKED ISSUE #4110: Feature request: support `onQuery` and `onMutation` resolvers in AppSync GraphQL
### Use case

In addition to registering resolvers with `app.resolver()` customers should be able to use a short-hand syntax specific to Mutation and Query operations as described in the initial spec (#1166).

### Solution/User Experience

Similar to what was implemented in an earlier version of #3999, we should allow customers to register resolvers like this:

```ts
import {
  AppSyncGraphQLResolver,
  makeId,
} from '@aws-lambda-powertools/event-handler/appsync-graphql';

app.onQuery('getPost', async ({ id }) => {
  logger.debug('getPost called', { id });
  // Simulate fetching a post from a database or other source
  return {
    id,
    title: `Post Title ${id}`,
    content: `Content for post ${id}`,
  };
});


app.onMutation('addPost', async (args) => {
  logger.debug('addPost called', { args });
  const { author, title, content } = args;
  // Simulate creating a post in a database or other source
  return {
    id: makeId(), // Generate a unique ID if not provided
    author,
    title,
    content,
  };
});
```

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR adds two new methods to the `AppSyncGraphQLResolver` to allow customers to easily register route handlers for queries and mutations respectively via `onQuery()` and `onMutation()`.

These methods are mainly syntactic sugar as they use the more versatile `resolver()` method under the hood, and share the same route registry.

The PR also makes targeted updates to the types and folder structure to streamline and reduce duplication. Unit tests were also updated to maintain 100% coverage.

**Issue number:** closes #4110 

---


packages/event-handler/src/appsync-graphql/RouteHandlerRegistry.ts
```diff
@@ -1,14 +1,15 @@
...
   RouteHandlerRegistryOptions,
 } from '../types/appsync-graphql.js';
+import type { AppSyncGraphQLResolver } from './AppSyncGraphQLResolver.js';
 
 /**
  * Registry for storing route handlers for GraphQL resolvers in AWS AppSync GraphQL API's.
  *
  * This class should not be used directly unless you are implementing a custom router.
- * Instead, use the {@link Router} class, which is the recommended way to register routes.
+ * Instead, use the {@link AppSyncGraphQLResolver | `AppSyncGraphQLResolver`} class, which is the recommended way to register routes.
```
svozza, line 12: Just for my own information, why is `AppSyncGraphQLResolver` repeated here? I see a similar pattern in a few other places in this PR and was just curious.

svozza (approved):

packages/event-handler/src/appsync-graphql/RouteHandlerRegistry.ts
dreamorosi: The first argument after `@link` is the actual link/reference, the second is the label/description.

This is purely cosmetic, but writing it like that renders the link as `AppSyncGraphQLResolver` (monospaced) rather than plain text.

svozza: Ah gotcha!

Issue #4034, opened 2025-06-11
Maintenance: remove some issue & PR automation workflows


dreamorosi:
### Summary

Over the past week we've been running some tests to replace some of the workflows to manage issues & PRs current running in GitHub Actions  with a dedicated GitHub App.

So far things have worked out fine and we have four features working, that can replace or complement existing workflows:
- close issue comment ([example](https://github.com/aws-powertools/powertools-lambda-typescript/issues/3920#issuecomment-2961666538))
- PR acknowledgement section check ([example](https://github.com/aws-powertools/powertools-lambda-typescript/runs/43870923565))
- PR related issue check ([example](https://github.com/aws-powertools/powertools-lambda-typescript/runs/43870922755))
- `help-wanted` label on issue (it posts a comment encouraging contributors to leave a comment to claim the issue)

Because of this, we can phase out some of the workflows that were doing these activities.

### Why is this needed?

So that we can streamline our maintenance work.

### Acknowledgment

- [x] This request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

PR #4145, opened 2025-07-07
chore: bump biome formatter to next major & reformat

### LINKED ISSUE #4144: Maintenance: bump biome formatter to next major
### Summary

Biome has released a new major version a few weeks ago, we've been putting off the upgrade because it introduces a couple new rules that require manual attention but at this point enough weeks have passed and we should address the upgrade.

### Why is this needed?

So that we can keep up to date with the latest version of the package.

Other

### Solution

_No response_

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR manually bumps `@biomejs/biome` to the latest version (`2.x`) which constitutes a major release upgrade for us since we were stuck at `1.x`.

The upgrade introduces a couple new rules that are a good thing for the project, but that required some manual attention. The changes focus almost exclusively on import sorting and unused variables.

Since the PR diff is very big, I tried to address each package in a dedicated commit and I suggest reviewers to use the dropdown at the top-left of the PR view to select the commits while reviewing to limit the number of files.

Aside from that, apologies in advance for the big PR, splitting the work in multiple PRs would've required changes to the CI to support both versions at the same time and it wasn't worth it.

I also don't expect reviewers to look at every single file, but rather get a gist of the changes and review it as a whole, knowing that the tests are still passing.

**Issue number:** closes #4144

---


packages/logger/tests/e2e/advancedUses.test.ts
```diff
@@ -104,7 +104,7 @@ describe('Logger E2E - Advanced uses', () => {
         try {
           JSON.parse(log);
           return true;
-        } catch (error) {
+        } catch (_error) {
```
svozza: How come this isn't just using a bare `catch` like the others?

```ts
} catch {
  return false;
}
```

packages/parser/tests/unit/schema/kinesis.test.ts
```diff
@@ -187,7 +187,7 @@ describe('Schema: Kinesis', () => {
 
     const transformedInput = {
       ...testEvent,
-      Records: testEvent.Records.map((record, index) => {
+      Records: testEvent.Records.map((record, _index) => {
```
svozza: Why include `index` at all here, `testEvent.Records.map((record) => {` would suffice.

.github/scripts/download_pr_artifact.js
```diff
@@ -1,6 +1,6 @@
-module.exports = async ({github, context, core}) => {
```
svozza, line 1: Not for this PR but at some point we should move these scripts to user ESM rather than CommonJS.

examples/snippets/event-handler/bedrock-agents/stopConversation.ts
```diff
@@ -10,7 +10,7 @@ app.tool<{ city: string }>(
   async ({ city }, { event }) => {
     try {
       throw new Error('Simulated error for demonstration purposes');
-    } catch (error) {
+    } catch (_error) {
```
svozza: As mentioned, is there a reaosn we're not doing `} catch {` like we are in other places in the code? Same for all the other instances of this pattern in this commit (1f780c5).

svozza:
>  I tried to address each package in a dedicated commit and I suggest reviewers to use the dropdown at the top-left of the PR view to select the commits while reviewing to limit the number of files

Thanks for this, made reviewing the changes much easier.

.github/scripts/download_pr_artifact.js
dreamorosi: The medium term plan is removing all these altogether. I was actually undecided whether to format them in the first place.

I've been working on a GitHub App that should centralize all these actions across repos in an event driven way (repo is private but will give you access).

In the past month we moved a few things #4034 and I'm planning on doing more this quarter.

packages/logger/tests/e2e/advancedUses.test.ts
dreamorosi: I wasn't very consistent with this - I'll make another pass and remove them, good point!

.github/scripts/download_pr_artifact.js
svozza: Oh nice that would be great.

Issue #4109, opened 2025-07-01
Feature request: Ensure parser schema types are compatible with @types/aws-lambda


bilalq:
### Use case

Hello,

For a long time now, we, like many others, have been using `@types/aws-lambda` as our reference types for Lambda events. For safety, we used our own typeguard functions. Now that Powertools vends a parser, we'd like to start using it in our middleware. However,  when working with the SQS event, I saw a type incompatibility around the `md5OfMessageAttributes` field.

The [@types/aws-lambda package](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/trigger/sqs.d.ts) defines it as such:
```ts
// SQS
// https://docs.aws.amazon.com/lambda/latest/dg/invoking-lambda-function.html#supported-event-source-sqs
export interface SQSRecord {
    messageId: string;
    receiptHandle: string;
    body: string;
    attributes: SQSRecordAttributes;
    messageAttributes: SQSMessageAttributes;
    md5OfBody: string;
    md5OfMessageAttributes?: string;
    eventSource: string;
    eventSourceARN: string;
    awsRegion: string;
}
```

The Powertools parser schema has it modeled like this:
```ts
/**
 * Zod schema for a SQS record inside an SQS event.
 */
const SqsRecordSchema = z.object({
  messageId: z.string(),
  receiptHandle: z.string(),
  body: z.string(),
  attributes: SqsAttributesSchema,
  messageAttributes: z.record(z.string(), SqsMsgAttributeSchema),
  md5OfBody: z.string(),
  md5OfMessageAttributes: z.string().optional().nullable(),
  eventSource: z.literal('aws:sqs'),
  eventSourceARN: z.string(),
  awsRegion: z.string(),
});
```

Note the difference of `md5OfMessageAttributes?: string` vs `md5OfMessageAttributes: z.string().optional().nullable()`.

### Solution/User Experience

The SQS or Lambda teams can probably best speak to whether `null` is possible or it really should be `string | undefined`. Depending on that answer, either the zod schema here should update or an update should be made to `@types/aws-lambda`.

Going forward though, it would be nice if there were unit tests that compared the Zod schema parsed shapes against the types from `@types/aws-lambda`. There are many libraries and utilities that have been written around the types from that package, so i t would be great if the parsed events from Powertools were compatible with those.

### Alternative solutions

```Markdown

```

### Acknowledgment

[...]

PR #4165, opened 2025-07-14
fix(parser): Removed the nullable type from the md5OfMessageAttributes in SqsRecordSchema

sdangol:
## Summary

This PR removes the `nullable` type from the `md5OfMessageAttributes` in the `SqsRecordSchema` aligning with the [@types/aws-lambda package](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/trigger/sqs.d.ts)

### Changes

1. Removed the `nullable` type from the `md5OfMessageAttributes` property in the `SqsRecordSchema`
2. Updated the data in `firehose-sqs.json` test event to remove the `null` value of `md5OfMessageAttributes`

**Issue number:** #4109 

---


bboure:
Hi.

After this "fix", I am getting this error.

```
{
    "errorType": "ParseError",
    "errorMessage": "Failed to parse SQS Envelope. This error was caused by: [\n  {\n    \"expected\": \"string\",\n    \"code\": \"invalid_type\",\n    \"path\": [\n      \"Records\",\n      0,\n      \"md5OfMessageAttributes\"\n    ],\n    \"message\": \"Invalid input: expected string, received null\"\n  }\n].",
    "name": "ParseError",
    "stack": [
        "ParseError: Failed to parse SQS Envelope. This error was caused by: [",
        "  {",
        "    \"expected\": \"string\",",
        "    \"code\": \"invalid_type\",",
        "    \"path\": [",
        "      \"Records\",",
        "      0,",
        "      \"md5OfMessageAttributes\"",
        "    ],",
        "    \"message\": \"Invalid input: expected string, received null\"",
        "  }",
        "].",
        "    at Object.parse (/node_modules/@aws-lambda-powertools/parser/lib/esm/envelopes/sqs.js:36:19)",
    ]
}

```

Logging the event shows that I do get this attribute as `null`

From the logs (strigified payload)
```
"md5OfMessageAttributes":null
```

I am not quite sure in what circumstances this occurs, but we should probably revert this change.

Thanks

dreamorosi:
Hi @bboure, thanks for reporting.

> I am not quite sure in what circumstances this occurs

I think this would be helpful to understand if the type and schemas are indeed off on not. For now, as discussed in the linked issue we assumed that the type was correct and we weren't but we also weren't able to reproduce the field being `null` in a function so we just aligned ourselves with the type.

If you could share steps to reproduce the field being `null` in Lambda, we'll make sure to verify it and revert the change right away.

bboure:
Thanks @dreamorosi 

So, here is a snippet of my code that reproduces this issue. 

From another Lambda function:

```typescript
export const sendMessageBatch = async (
  queueUrl: string,
  messages: Array<{
    id: string;
    body: string;
    delayInSeconds?: number;
  }>,
) => {
  const entries: SendMessageBatchRequestEntry[] = messages.map((message) => ({
    Id: message.id,
    MessageBody: message.body,
    DelaySeconds: message.delayInSeconds,
  }));

  const command = new SendMessageBatchCommand({
    QueueUrl: queueUrl,
    Entries: entries,
  });

  await sqsClient.send(command);
};
```

Usage

```typescript
const events = [
  { id: '123', body: JSON.stirngify({ foo: 'bar' }) }
];
await sendMessageBatch(NOTIFICATION_QUEUE_URL, events);
```

After some research, I figured that adding `MessageAttributes` in `SendMessageBatchCommand` forces this value to be present and it makes the validation pass.


```typescript
    Id: message.id,
    MessageBody: message.body,
    // Adding this works  👇
    MessageAttributes: {
      ContentType: { DataType: 'String', StringValue: 'application/json' },
    },
    DelaySeconds: message.delayInSeconds,
````

**Comment by bboure** [comment]
And just in case it's relevant:

NodeJS version: 22
architecture: ARM
Lambda is in a VPC

**Comment by dreamorosi** [comment]
Ok, thanks - we'll look into it on Monday.

**Comment by dreamorosi** [comment]
I've looked into it and I'm unable to reproduce the "bug".

I have created a minimal CDK stack with 2 functions and 1 queue that you can find **[here](https://github.com/dreamorosi/repro-4165)**. You're welcome to clone the repo, deploy the stack, and verify the statements below in your own account.

The producer function uses the code you shared above to publish messages to the queue, while the consumer one is triggered by the queue and uses various combinations of built-in SQS schemas and envelopes to parse the event.

After deploying the stack and triggering the producer (instructions are in the README), you can see from the logs of the consumer function that all parse operations succeed despite the message attributes being empty and the `md5OfMessageAttributes` field not being present.

After re-reading the PR and remembering what it was about, this makes sense to me, since the PR marks the field as `optional()` (previously was `.optional().nullable()`).

As you can see from the git history, since it was introduced, the `md5OfMessageAttributes` has never been required ([link](https://github.com/aws-powertools/powertools-lambda-typescript/blob/4f526ecda845c5f2586463e3edbe8d1b39f20f5b/packages/parser/src/schemas/sqs.ts#L29)), so the error message you're seeing is probably due to either an older version of Parser (unsure how `.optional().nullable()` interact with each other) or an override to the schema being applied in your code.

Either way based on the above and the information you provided, I'm not inclined to revert this PR.

**Comment by bboure** [comment]
Thanks @dreamorosi 

After doing some more tests, comparing with our solution, I finally found how to reproduce. It happens when enabling `maxBatchingWindow`. See this PR https://github.com/dreamorosi/repro-4165/pull/1/files

```
{
    "level": "ERROR",
    "message": "Failed to parse SQS event",
    "timestamp": "2025-09-13T13:47:34.199Z",
    "service": "sqs-consumer",
    "sampling_rate": 0,
    "xray_trace_id": "1-68c575f5-0c82ef561607ff44f3e38830",
    "error": {
        "name": "ZodError",
        "location": "file:///var/task/index.mjs:3",
        "message": "[\n  {\n    \"expected\": \"string\",\n    \"code\": \"invalid_type\",\n    \"path\": [\n      \"Records\",\n      0,\n      \"md5OfMessageAttributes\"\n    ],\n    \"message\": \"Invalid input: expected string, received null\"\n  }\n]",
        "stack": "ZodError: [\n  {\n    \"expected\": \"string\",\n    \"code\": \"invalid_type\",\n    \"path\": [\n      \"Records\",\n      0,\n      \"md5OfMessageAttributes\"\n    ],\n    \"message\": \"Invalid input: expected string, received null\"\n  }\n]\n    at new ZodError (file:///var/task/index.mjs:3:23677)\n    at file:///var/task/index.mjs:4:687\n    at e.safeParse (file:///var/task/index.mjs:25:4634)\n    at Runtime.hb [as handler] (file:///var/task/index.mjs:25:22959)\n    at Runtime.handleOnceNonStreaming (file:///var/runtime/index.mjs:1206:29)"
    }
}
```

dreamorosi:
Interesting!

Before we fix this, let me ask some questions internally. I find it weird that enabling a seemingly unrelated setting in the Event Source Mapping (ESM) causes the field to be omitted or set to `null` when there're no attributes.

In the meantime I suggest to manually override the field in the schema, or patch the producer like you suggested earlier.

ghost:
@dreamorosi I can confirm that `md5OfMessageAttributes` is null when `maxBatchingWindow` is specified. I did not test without `maxBatchingWindow`, however.

dreamorosi:
We have reverted the change and will release the updated schema in the next release.

Issue #4135, opened 2025-07-05
Maintenance: improve runtime validations in Metrics utility


dreamorosi:
### Summary

After comparing the Python and TypeScript implementations of AWS Lambda Powertools Metrics, I found a number of runtime validations that are present in the Python version but missing in the TypeScript version. These missing validations could lead to potential issues in production if not addressed.

Overall TypeScript version relies heavily on compile-time type safety without adequate runtime safeguards, and in certain cases, it allows invalid data to be passed at runtime.

### Why is this needed?

The TypeScript implementation is missing several critical runtime validations that are present in the Python version:

1. **Metric Value Validation**: No runtime check for numeric values
2. **Metric Unit Validation**: No runtime validation against valid CloudWatch units  
3. **Metric Resolution Validation**: No runtime validation for resolution values (1 or 60)
4. **Dimension Name Validation**: Only validates dimension values, not names, and allows whitespace-only strings

These missing validations could lead to:
- Invalid metrics being sent to CloudWatch
- Runtime errors when invalid data is passed
- Silent failures where metrics are dropped by CloudWatch
- Inconsistent behavior between Python and TypeScript implementations

### Which area does this relate to?

Metrics

### Solution

**1. Runtime Metric Value Type Validation**

The Python implementation includes a runtime check to ensure that metric values are valid numbers. If a non-numeric value is passed, it raises a `MetricValueError`.

Python code:
```python
if not isinstance(value, numbers.Number):
    raise MetricValueError(f"{value} is not a valid number")
```

[Python Source](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/metrics/provider/cloudwatch_emf/cloudwatch.py#L140)

The TypeScript implementation has no runtime validation if types are bypassed or incorrect values passed.

[TypeScript Source - addMetric method](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/metrics/src/Metrics.ts#L372-L379)

**2. Runtime Metric Unit Validation**

The Python implementation validates that the metric unit is one of the valid CloudWatch units at runtime. If an invalid unit is passed, it raises a `MetricUnitError`.

Python code:
```python
def _extract_metric_unit_value(self, unit: str | MetricUnit) -> str:
    if isinstance(unit, str):
        if unit in self._metric_unit_valid_options:
            unit = MetricUnit[unit].value
        if unit not in self._metric_units:
            raise MetricUnitError(
[...]

PR #4181, opened 2025-07-16
improv(metrics): Added runtime validations for the metrics utility functions

sdangol:
## Summary

This PR adds runtime validations for the metrics utility functions. It ensures that metrics and dimensions meet the required constraints before they are added, providing better error handling and feedback to users. This was done to prevent potential issues that could have come up due to invalid data being passed to Cloudwatch.

### Changes

- Added runtime validation for the following:
  - Metric Name Validation: Validated according to the constraints in https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_MetricDatum.html
  - Metric Value Validation: Validated for a valid numeric value
  - Metric Unit Validation: Validated for allowed units specified in the [constants](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/metrics/src/constants.ts#L38-L66) file
  - Metric Resolution Validation: Validated for allowed resolution specified in the [constants](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/metrics/src/constants.ts#L73-L76) file
  - Dimension Name Validation: Both Dimension name and value are validated for non empty, non null defined value.
- Added a new utility function `isStringUndefinedNullEmpty` in the commons package to validate string inputs
- Added constants for metric name length constraints (`MIN_METRIC_NAME_LENGTH` and `MAX_METRIC_NAME_LENGTH`)
- Added tests for the new validations

**Issue number:** 
#4135 

---


packages/metrics/src/Metrics.ts
```diff
@@ -1067,6 +1078,24 @@ class Metrics extends Utility implements MetricsInterface {
     value: number,
     resolution: MetricResolution
   ): void {
+    if (!isString(name)) throw new Error(`${name} is not a valid string`);
+    if (
+      name.length < MIN_METRIC_NAME_LENGTH ||
+      name.length > MAX_METRIC_NAME_LENGTH
+    )
+      throw new Error(
```
dreamorosi: Can/should we improve the error here by either throwing a [`RangeError`](https://nodejs.org/api/errors.html#class-rangeerror) or a custom error we define? If we go the second route please look at how we define errors in other packages of the project.

packages/commons/tests/unit/typeUtils.test.ts
```diff
@@ -119,6 +120,32 @@ describe('Functions: typeUtils', () => {
     });
   });
 
+  describe('Function: isStringUndefinedNullEmpty', () => {
+    it('should return true if input is undefined', () => {
```
dreamorosi:
```suggestion
    it('returns true if input is undefined', () => {
```

dreamorosi: Can we apply this to all the tests, and also add a comment `// Act & Assess` on top of each call?

packages/commons/src/typeUtils.ts
```diff
@@ -154,6 +154,28 @@ const isNullOrUndefined = (value: unknown): value is null | undefined => {
...
+ * ```typescript
+ * import { isStringUndefinedNullEmpty } from '@aws-lambda-powertools/commons/typeUtils';
+ *
+ * const value = 'foo';
+ * if (isStringUndefinedNullEmpty(value)) {
+ *   // value is either undefined, null, or an empty string
+ * }
+ * ```
+ *
+ * @param value The value to check
```
dreamorosi:
```suggestion
 * @param - value The value to check
```

dreamorosi (changes requested):
Great work on this PR Swopnil! 

Left 3 minor comments but other than it's almost good to go!

Regarding the discussion we were having about other Powertools for AWS runtimes in the linked issue, did you see if they are implementing the same check you brought up?

packages/commons/src/typeUtils.ts
sdangol: The other JSDoc comments on the functions have the same structure without the dash in the `@param`

packages/commons/tests/unit/typeUtils.test.ts
sdangol: I've applied this to all the tests and added the comment for each call

packages/metrics/src/Metrics.ts
sdangol: Updated to use `RangeError`

sdangol:
> Regarding the discussion we were having about other Powertools for AWS runtimes in the linked issue, did you see if they are implementing the same check you brought up?

Yes, I looked into the Python and .NET and the metric name constraint validation is missing. I tested the python version and the metrics were getting silently dropped there as well. I will create the issues for this in their repos.

packages/commons/src/typeUtils.ts
dreamorosi: Ok, other packages do, but we're not as consistent I agree.

If you can, add the dash to this file and whenever you find an occurrence without in future PRs, please feel free to add it.

sdangol: Hadn't looked at the other files :) I've now updated these occurrences with the dashed format.

packages/metrics/src/Metrics.ts
dreamorosi: Sorry I wasn't clear - I think we should change all the ones in this function (aka below this line) to `RangeError`

sdangol: Updated all the ones below that to use `RangeError`

dreamorosi (approved):

dreamorosi:
Thank you for the work on this PR @swopnildangol - great work and turnaround :)

Issue #4171, opened 2025-07-15
Maintenance: move away from `lerna publish` command


dreamorosi:
### Summary

As part of the migration off `lerna` we should remove our usage of `lerna publish`, which as the name suggests is currently used to publish the packages to the npm registry during the release workflow.

This functionality can most likely be assumed by `npm` workspaces without any additional dependencies, but we'll have to do some tests to verify.

### Why is this needed?

See reasoning of why we're migrating off `lerna` in the parent issue, but ultimately removing it would allow us to do this:

```sh
 npm un lerna

removed 554 packages, changed 2 packages, and audited 1088 packages in 5s

found 0 vulnerabilities
```

### Which area does this relate to?

Automation

### Acknowledgment

- [x] This request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

dreamorosi:
Hi @sdangol, before you start working on the other issue we spoke about, we could really use your help with this one.

Before we can move off `lerna publish` we need to make a quick test to ensure we're publishing correctly at the first try during the next release.

Below is a plan I came up with, if you have any suggestions to simplify it while maintaining the same test fidelity please let me know:
- [ ] Set up [verdaccio](https://verdaccio.org) in a EC2 instance
- [ ] Create a sample private repo with npm 3 workspaces using `@test-pkg/a`, `@test-pkg/b`, and `@test-pkg/c` - **one of them should be marked as private, the other two as public** (the idea here is to mimic the setup we have in this repo in a very minimal way)
- [ ] Make mock release in sample repo targeting verdaccio using `npm publish` in a GitHub Action (you can copy almost exactly the `make-release.yaml` file in this repo)
- [ ] Clean up sample repo and verdaccio server
- [ ] Update `make-release.yaml` to use `npm publish` as tested instead of `lerna publish`

[...]

Issue #4170, opened 2025-07-15
Maintenance: move away from `lerna version` command


dreamorosi:
### Summary

As part of the migration off `lerna` we should remove our usage of `lerna version` and bump our project's version + generate the CHANGELOG in other ways.

Ideally whatever we'll end up using should live as an action under `aws-powertools/actions` so that it can be reused by `aws-powertools/powertools-mcp` and other Node.js-based repositories under this org.

### Why is this needed?

See reasoning in the parent issue, but ultimately removing it would allow us to do this:

```sh
 npm un lerna

removed 554 packages, changed 2 packages, and audited 1088 packages in 5s

found 0 vulnerabilities
```

### Which area does this relate to?

Automation

### Acknowledgment

- [x] This request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

sdangol:
@dreamorosi Do we want to have the release type determined by us? Or rely on the automatic determination from `bump-n-go`.

dreamorosi:
Automatic, but we can take an optional input that will be passed to the `bump-n-go` action we have in the actions repo.

The general steps to do are:
- make a release in the actions repo and get the commit hash of that release
- update the make version workflow in this repo to use the action from the actions repo pinning the release hash (we should replace the lerna command basically)

I know it's a bit vague but I can provide more details when I get to my laptop tomorrow if this is not clear.

PR #4195, opened 2025-07-22
refactor(ci): Updated the publish command in the CI to use `npm publish` instead of `lerna publish`

sdangol:
## Summary

This PR updates the `make-release` and `publish-package` Github workflows to switch from using publish commands from `lerna` to the native `npm` publish command to publish the packages to the npm registry.

### Changes

- Updated the `lerna publish` command to `npm publish` in the `make-release.yml` and `publish-package.yml` Github workflows
- For generating provenance statements, the `--provenance` flag was used
-  The `from-package` from lerna was removed as `npm publish` uses the `--workspaces` flag to publish all the workspaces which are not marked as `private`. If a package with the same version is tried to be published, it will throw an error
- To publish an individual package through `publish-package.yml`, the `--workspace <workspace-path>` was used

**Issue number:** 
#4171 

---

### DIFF
--- a/.github/workflows/make-release.yml
+++ b/.github/workflows/make-release.yml
@@ -65,7 +65,7 @@ jobs:
         uses: aws-powertools/actions/.github/actions/cached-node-modules@29979bc5339bf54f76a11ac36ff67701986bb0f0
       - name: Publish to npm
         run: |
-          NPM_CONFIG_PROVENANCE=true npx lerna publish from-package --git-head ${{ github.sha }} --yes 
+          npm publish --workspaces --provenance
       - name: Set release version
         id: set-release-version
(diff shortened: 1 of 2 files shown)


dreamorosi (approved):
Thank you for the work on this PR and for the tests in the issue!

We can address this either in this PR or in the PR we'll open for #4170, but in L72 of the `make-release` workflow we're reading the version number from the `lerna.json` file, which will be removed/no longer updated.

To replace that line I'd consider reading the version from `packages/commons/package.json` since it'll always be updated.

sdangol:
Yeah, let's do that in the PR that we open next which probably would remove all the `lerna` dependencies as well.

Issue #4140, opened 2025-07-06
Feature request: Implement Route Matching & Resolution System for Event Handler


dreamorosi:
### Use case

As part of the Event Handler implementation (#3251), we need a system to match incoming HTTP requests against registered routes and extract path parameters. 

The Route Matching System provides the **pattern matching engine** that:
1. **Compiles route patterns** from user-friendly syntax (`:param`) to executable regex patterns
2. **Matches incoming requests** against compiled route patterns efficiently
3. **Extracts path parameters** with type safety for dynamic routes
4. **Prioritizes route matching** (static routes before dynamic routes for performance)
5. **Handles edge cases** like trailing slashes, special characters, and conflicting patterns

This system acts as the **matching layer** between route storage (Route Management) and request resolution, ensuring incoming requests are correctly matched to their handlers.

### Solution/User Experience

> [!note]
> The code snippets below are provided as reference only - they are not exhaustive and final implementation might vary.

### Route Matching Engine

```typescript
// Efficient route matching with prioritization
class RouteMatcher {
  private staticRoutes: Map<string, Route[]> = new Map();
  private dynamicRoutes: Route[] = [];
  
  constructor(routes: Route[]) {
    this.organizeRoutes(routes);
  }
  
  // Organize routes for efficient matching
  private organizeRoutes(routes: Route[]): void {
    for (const route of routes) {
      const compiled = RouteCompiler.compile(route.path);
      
      if (compiled.isDynamic) {
        // Dynamic routes need regex matching
        this.dynamicRoutes.push({
          ...route,
          compiledRoute: compiled
        });
      } else {
        // Static routes get O(1) lookup
        const key = `${route.method}:${route.path}`;
        if (!this.staticRoutes.has(key)) {
          this.staticRoutes.set(key, []);
        }
        this.staticRoutes.get(key)!.push(route);
      }
    }
    
    // Sort dynamic routes by specificity (more specific patterns first)
    this.dynamicRoutes.sort(this.compareRouteSpecificity);
  }
  
  // Find matching route for incoming request
  findMatch(method: string, path: string): RouteMatch | null {
    // 1. Try static routes first (fastest)
    const staticKey = `${method.toUpperCase()}:${path}`;
    const staticMatches = this.staticRoutes.get(staticKey);
    
    if (staticMatches && staticMatches.length > 0) {
      return {
        route: staticMatches[0],
        params: {},
        matchType: 'static'
      };
    }
    
[...]

PR #3972, opened 2025-05-23
feat(event-handler): add base router class

dreamorosi:
### Changes

> Please provide a summary of what's being changed

> Please add the issue number below, if no issue is present the PR might get blocked and not be reviewed

**Issue number:** closes #3971

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Disclaimer**: We value your time and bandwidth. As such, any pull requests created on non-triaged issues might not be successful.

PR #4211, opened 2025-07-24
feat(event-handler): add route management system for ApiGw event handler

### LINKED ISSUE #4139: Feature request: Implement Route Management System for Event Handler
### Use case

As part of the Event Handler implementation (#3251), we need a system to store, organize, and manage routes that are registered through the BaseRouter's HTTP method decorators (`get()`, `post()`, etc.).

The Route Management System provides the internal infrastructure to:
1. **Store registered routes** in an organized, efficient manner
2. **Detect route conflicts** and provide helpful warnings to developers
3. **Organize routes** for efficient lookup and processing
4. **Maintain route metadata** needed for resolution and debugging

This system acts as the **storage layer** between route registration (`BaseRouter`) and route matching (separate issue), ensuring routes are properly organized and accessible.

### Solution/User Experience

> [!note]

### Core Route Management Architecture

```typescript
// Internal Route representation
class Route {
  readonly id: string;                    // Unique route identifier
  readonly method: string;                // 'GET', 'POST', etc.
  readonly path: string;                  // Original path: '/users/:id'
  readonly handler: RouteHandler;         // The function to execute
  readonly registeredAt: Date;            // When route was registered
  
  constructor(method: string, path: string, handler: RouteHandler) {
    this.id = `${method}:${path}`;
    this.method = method.toUpperCase();
    this.path = path;
    this.handler = handler;
    this.registeredAt = new Date();
  }
}

// Route storage and organization
class RouteRegistry {
  private routes: Map<string, Route> = new Map();
  private routesByMethod: Map<string, Route[]> = new Map();
  
  // Add a route to the registry
  addRoute(route: Route): void {
    // Conflict detection
    if (this.routes.has(route.id)) {
      this.handleRouteConflict(route);
    }
    
    // Store route
    this.routes.set(route.id, route);
    
    // Organize by method for efficient lookup
    if (!this.routesByMethod.has(route.method)) {
      this.routesByMethod.set(route.method, []);
    }
    this.routesByMethod.get(route.method)!.push(route);
  }
  
  // Retrieve routes for matching
  getRoutesByMethod(method: string): Route[] {
    return this.routesByMethod.get(method.toUpperCase()) || [];
  }
  
  getAllRoutes(): Route[] {
    return Array.from(this.routes.values());
  }
  
  // Development helpers
  getRouteCount(): number {
    return this.routes.size;
  }
  
  getRouteSummary(): RouteSummary {
    // Provide debugging information
  }
}
```

### Integration with BaseRouter

```typescript
// BaseRouter uses RouteRegistry for storage
abstract class BaseRouter {
  protected routeRegistry = new RouteRegistry();
  
  get(path: string, handler: RouteHandler): void {
    const route = new Route('GET', path, handler);
    this.routeRegistry.addRoute(route);
  }
  
  post(path: string, handler: RouteHandler): void {
    const route = new Route('POST', path, handler);
    this.routeRegistry.addRoute(route);
  }
  
  // ... other HTTP methods
  
  // Expose routes for matching system (next issue)
  protected getRegisteredRoutes(): Route[] {
    return this.routeRegistry.getAllRoutes();
  }
  
  protected getRoutesByMethod(method: string): Route[] {
    return this.routeRegistry.getRoutesByMethod(method);
  }
}
```

### Route Compilation System

```typescript
// Convert user-friendly routes to executable patterns
class RouteCompiler {
  private static readonly PARAM_PATTERN = /:(\w+)/g;
  private static readonly SAFE_CHARS = '-._~()\'!*:@,;=+&$';
  private static readonly UNSAFE_CHARS = '%<> \\[\\]{}|^';
  
  static compile(path: string): CompiledRoute {
    // Convert /users/:id/posts/:postId to regex with named groups
    const paramNames: string[] = [];
    
    const regexPattern = path.replace(
      RouteCompiler.PARAM_PATTERN,
      (match, paramName) => {
        paramNames.push(paramName);
        return `(?<${paramName}>[${RouteCompiler.SAFE_CHARS}${RouteCompiler.UNSAFE_CHARS}\\w]+)`;
      }
    );
    
    // Add anchors for exact matching
    const finalPattern = `^${regexPattern}$`;
    
    return {
      originalPath: path,
      regex: new RegExp(finalPattern),
      paramNames,
      isDynamic: paramNames.length > 0
    };
  }
  
  // Validate route patterns for common issues
  static validatePattern(path: string): ValidationResult {
    // Check for invalid characters, malformed parameters, etc.
    const issues: string[] = [];
    
    // Check for malformed parameters
    if (path.includes(':') && !RouteCompiler.PARAM_PATTERN.test(path)) {
      issues.push('Malformed parameter syntax. Use :paramName format.');
    }
    
    // Check for duplicate parameter names
    const params = RouteCompiler.extractParamNames(path);
    const duplicates = params.filter((param, index) => params.indexOf(param) !== index);
    if (duplicates.length > 0) {
      issues.push(`Duplicate parameter names: ${duplicates.join(', ')}`);
    }
    
    return {
      isValid: issues.length === 0,
      issues
    };
  }
  
  private static extractParamNames(path: string): string[] {
    const matches = [...path.matchAll(RouteCompiler.PARAM_PATTERN)];
    return matches.map(match => match[1]);
  }
}

interface CompiledRoute {
  originalPath: string;
  regex: RegExp;
  paramNames: string[];
  isDynamic: boolean;
}
```

### Scope - In Scope:
- **Route class**: Internal representation of registered routes
- **RouteRegistry**: Storage and organization of routes
- **Route conflict detection**: Basic duplicate detection with warnings
- **Route organization**: Efficient storage by HTTP method
- **Development helpers**: Route inspection and debugging tools
- **Integration points**: Clean interface for route matching system
- **Route compilation** (`:param` → regex)

### Scope - Out of Scope (Future Issues):
- Route pattern matching & execution logic - belongs in Route Matching & Resolution issue
- Route-specific configuration (CORS, middleware, etc.) - future features

### Data Flow:
```
BaseRouter.get() → Route → RouteRegistry.addRoute() → Storage
RouteRegistry.getRoutesByMethod() → Route Matching System → Resolution
```

### Alternative solutions

```Markdown
N/A
```

### Acknowledgment

### Future readers

svozza:
## Summary
Adds the route management layer for the API Gateway event handler. The main focus of this PR is to create the registry that will hold all the handlers registered to various routes and integrate it with the `BaseRouter` class added in #3972. Route validation functions were also added.

### Changes

- Added a `Route` class for internally representing routes
- Added the `RouteHandlerRegistry` class, following the pattern established in the `appsync-events` and `appsync-resolver` event handlers.
  - The registry exposes a `register` function where handlers are registered and stored by route id (`HttpMethod:path`) and by HTTP method
  - Handles duplicate added routes in a last one wins fashion
  - Validates routes and logs a warning for invalid paths and skips adding them
  - Provides methods to query the registry for size, routes and routes by HTTP method
- Updated the `BaseRouter` class:
  - `route` method is no longer abstract and registration of routes using the `RouteHandlerRegistry` class is done in one place here
  - Removed unnecessary third argument for HTTP method functions
- Added utilities for working with paths:
  - I have chosen not add these as static methods in a class as suggested in #4139. Instead, I have create a utility module with the new functions. My reasoning is that these are pure functions that do not require the incidental complexity of a class. As the experience with the `EnvService` has shown, should we wish to use these elsewhere in the future, it is easier to use plain functions rather than require consumers to instantiate a class  with multiple methods when all they might want is one function.
  - This module defines a strict regex to define what form parameters in a path (e.g., `/user/:userId`) can take.
    - The first character of a parameter must be a colon
    - A parameter can only contain one colon
    - The first character after the colon must be a letter or an underscore
    - Characters after the first letter must be letters, numbers or underscores
    - Multiple parameters are permitted, e.g., `/:param1/:param2/:param3`
  - The `validatePathPattern` function ensures that parameters conform to the rules described above
  - The `compilePath` function is responsible for converting paths into regexes that will then be used when implementing #4140
- Added tests for all newly added classes and updated the `BaseRouter` tests to reflect the changes made there.

**Issue number:** closes #4139

---

### DIFF
--- a/packages/event-handler/src/rest/Route.ts
+++ b/packages/event-handler/src/rest/Route.ts
@@ -5,14 +5,12 @@ class Route {
   readonly method: string;
   readonly path: Path;
   readonly handler: RouteHandler;
-  readonly registeredAt: Date;
 
   constructor(method: string, path: Path, handler: RouteHandler) {
     this.id = `${method}:${path}`;
     this.method = method.toUpperCase();
     this.path = path;
     this.handler = handler;
-    this.registeredAt = new Date();
   }
 }
 


dreamorosi:
Thank you Stefano, I'll block some time to review this for tomorrow.

svozza:
Sounds good. I've hewn fairly close to your sample implementation, aside from the utility stuff, so hopefully it won't be too much work.

packages/event-handler/src/rest/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+    issues,
+  };
+}
+
+interface CompiledRoute {
+  originalPath: string;
+  regex: RegExp;
+  paramNames: string[];
+  isDynamic: boolean;
+}
```
dreamorosi: Let's extract this in `src/types/rest.ts`

packages/event-handler/src/rest/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+    paramNames,
+    isDynamic: paramNames.length > 0,
+  };
+}
+
+type ValidationResult = {
+  isValid: boolean;
+  issues: string[];
+};
+
```
dreamorosi: Same of the other type, let's extract it under `src/types/rest.ts`

packages/event-handler/src/rest/utils.ts
```diff
@@ -0,0 +1,60 @@
+import type { Path } from '../types/rest.js';
+
+const PARAM_PATTERN = /:([a-zA-Z_]\w*)(?=\/|$)/g;
+const SAFE_CHARS = "-._~()'!*:@,;=+&$";
+const UNSAFE_CHARS = '%<> \\[\\]{}|^';
+
```
dreamorosi: I'd suggest extracting these in the `src/constants.ts` file so it's more consistent with other areas of the codebase

packages/event-handler/tests/unit/rest/RouteHandlerRegistry.test.ts
```diff
@@ -0,0 +1,222 @@
...
+describe('Class: RouteHandlerRegistry', () => {
+  it('should warn when registering a duplicate route', () => {
+    // Prepare
+    const mockLogger = {
+      debug: vi.fn(),
+      warn: vi.fn(),
+      error: vi.fn(),
+    };
+
+    const registry = new RouteHandlerRegistry({ logger: mockLogger });
```
dreamorosi: The global `console` is mocked globally in all tests. We can pass a `console` here and then later in the assertions we can do `expect(console.warn)...`, this way we don't need to define a `mockLogger` in each test.

packages/event-handler/tests/unit/rest/RouteHandlerRegistry.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    'should not register routes with invalid path pattern: $description',
+    ({ path }) => {
+      // Prepare
+      const mockLogger = {
+        debug: vi.fn(),
+        warn: vi.fn(),
+        error: vi.fn(),
+      };
+
+      const registry = new RouteHandlerRegistry({ logger: mockLogger });
```
dreamorosi: Same as the first case, we can use global `console` here.

packages/event-handler/tests/unit/rest/RouteHandlerRegistry.test.ts
```diff
@@ -0,0 +1,222 @@
...
+
+  it('should not register routes with duplicate parameter names', () => {
+    // Prepare
+    const mockLogger = {
+      debug: vi.fn(),
+      warn: vi.fn(),
+      error: vi.fn(),
+    };
+
+    const registry = new RouteHandlerRegistry({ logger: mockLogger });
```
dreamorosi: Same as the first case, we can use global `console` here

packages/event-handler/tests/unit/rest/RouteHandlerRegistry.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    expect(registry.getRouteCount()).toBe(0); // Route should not be registered
+    expect(registry.getAllRoutes()).toHaveLength(0);
+    expect(registry.getRoutesByMethod(HttpVerbs.GET)).toHaveLength(0);
+  });
+
+  describe('getRouteCount', () => {
+    it('returns 0 for empty registry', () => {
+      // Prepare
+      const mockLogger = { debug: vi.fn(), warn: vi.fn(), error: vi.fn() };
+      const registry = new RouteHandlerRegistry({ logger: mockLogger });
```
dreamorosi: We can probably remove the `logger: mockLogger` param here, since we're not making any assertions and the global `console` should be mocked and silenced already.

packages/event-handler/src/rest/RouteHandlerRegistry.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    this.#routes.set(route.id, route);
+
+    if (!this.#routesByMethod.has(route.method)) {
+      this.#routesByMethod.set(route.method, []);
+    }
+    // biome-ignore lint/style/noNonNullAssertion: Map.set operation above ensures Map.get won't return undefined
+    this.#routesByMethod.get(route.method)!.push(route);
+  }
```
dreamorosi: If we change the flow to something like this, we can avoid the non-null assertion:

```ts
const routesByMethod =
this.#routesByMethod.get(route.method) || [];
routesByMethod.push(route);
this.#routesByMethod.set(route.method, routesByMethod);
```

Additionally, if we make `routesByMethod` a set `new Set<Route>()` we can avoid duplicates. If we did that, the `getRoutesByMethod` method would become and tests would still pass:

```ts
public getRoutesByMethod(method: string): Route[] {
  const routes = this.#routesByMethod.get(method.toUpperCase());
  if (!routes) {
    return [];
  }
  return Array.from(routes);
}
```

packages/event-handler/tests/unit/rest/RouteHandlerRegistry.test.ts
```diff
@@ -0,0 +1,222 @@
...
+      const route1 = new Route(HttpVerbs.GET, '/users', handler);
+      const route2 = new Route(HttpVerbs.POST, '/users', handler);
+      const route3 = new Route(HttpVerbs.GET, '/posts', handler);
+
+      // Act
+      registry.register(route1);
+      registry.register(route2);
+      registry.register(route3);
+
+      // Assert
```
dreamorosi, line 187:
```suggestion
      // Assess
```

`s/Assert/Assess/` in the file

packages/event-handler/tests/unit/rest/RouteHandlerRegistry.test.ts
```diff
@@ -0,0 +1,222 @@
...
+      expect(mockLogger.warn).toHaveBeenCalledWith(
+        'Malformed parameter syntax. Use :paramName format.'
+      );
+      expect(registry.getRouteCount()).toBe(0);
+      expect(registry.getAllRoutes()).toHaveLength(0);
+      expect(registry.getRoutesByMethod(HttpVerbs.GET)).toHaveLength(0);
+    }
+  );
+
+  it('should not register routes with duplicate parameter names', () => {
```
dreamorosi, line 65:
```suggestion
  it('doesn\'t register routes with duplicate parameter names', () => {
```

packages/event-handler/tests/unit/rest/RouteHandlerRegistry.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    { path: '/users/:12345id', description: 'parameter beginning with number' },
+    { path: '/users/:', description: 'parameter without name' },
+    { path: '/users/:id-name', description: 'parameter with hyphen' },
+    { path: '/users/:id.name', description: 'parameter with dot' },
+    {
+      path: '/users/:id:name',
+      description: 'consecutive parameters without separator',
+    },
+  ])(
+    'should not register routes with invalid path pattern: $description',
```
dreamorosi, line 44:
```suggestion
    'shouldn\'t register routes with invalid path pattern: $description',
```

packages/event-handler/tests/unit/rest/RouteHandlerRegistry.test.ts
```diff
@@ -0,0 +1,222 @@
+import { describe, expect, it, vi } from 'vitest';
+import { HttpVerbs } from '../../../src/rest/constatnts.js';
+import { Route } from '../../../src/rest/Route.js';
+import { RouteHandlerRegistry } from '../../../src/rest/RouteHandlerRegistry.js';
+import type { Path } from '../../../src/types/rest.js';
+
+describe('Class: RouteHandlerRegistry', () => {
+  it('should warn when registering a duplicate route', () => {
```
dreamorosi, line 8:
```suggestion
  it('warns when registering a duplicate route', () => {
```

packages/event-handler/src/rest/RouteHandlerRegistry.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  constructor(options: RouteRegistryOptions) {
+    this.#logger = options.logger;
+  }
+
+  public register(route: Route): void {
+    const { isValid, issues } = validatePathPattern(route.path);
+    if (!isValid) {
+      for (const issue of issues) {
+        this.#logger.warn(issue);
```
dreamorosi, line 20: I'm not sure about this, should we emit each issue in a single log entry/line or a single one with formatting?

svozza: So are we thinking something like:

```ts
this.#logger.warn(`Route ${route.path} is invalid: \n ${issues.join('\n')}`);
```

packages/event-handler/src/rest/RouteHandlerRegistry.ts
svozza: Ah nice sugggestion for the non-null. We'll have an even nicer solution once `Map.getOrInsert` arrives: https://github.com/tc39/proposal-upsert.

I like the set idea too.

svozza: I'm just thinking, I don't think `Set` will help us avoid duplicates. The`Set` type uses `===` so you don't get the same benefit you do when comparing `Route` objects as you would with primitives like strings:

```ts
const route1 = new Route('GET', '/test', () => {});
const route2 = new Route('GET', '/test', () => {});

const routeSet = new Set([route1, route2])
routeSet.size // => 2;
```

packages/event-handler/src/rest/utils.ts
svozza: done

packages/event-handler/tests/unit/rest/RouteHandlerRegistry.test.ts
svozza: done

packages/event-handler/tests/unit/rest/RouteHandlerRegistry.test.ts
svozza: Done

packages/event-handler/tests/unit/rest/RouteHandlerRegistry.test.ts
svozza: Done

packages/event-handler/tests/unit/rest/RouteHandlerRegistry.test.ts
svozza: Done

packages/event-handler/src/rest/utils.ts

packages/event-handler/src/rest/utils.ts

packages/event-handler/src/rest/RouteHandlerRegistry.ts
dreamorosi: Yea that's an option - but to be honest I am unclear yet about which way we should go.

I'm ok to follow your proposal or even keep it as it is and postpone the decision when we run it on Lambda a few times and see how it feels.

packages/event-handler/src/rest/RouteHandlerRegistry.ts
dreamorosi: You're right - didn't think about them being complex objects.

Do you think it's still worth making it a `Set` to remove the non-null? If not, then we can keep as-is.

dreamorosi (dismissed):
Thanks for the work on the PR and for addressing my previous comments.

I've answered your questions about logs & `Set` usage and I don't feel strongly about either at this stage, so I'm happy for you to make a decision and either resolve the conversations, then merge - or change it + me sending another "Approve" right after that.

packages/event-handler/src/rest/RouteHandlerRegistry.ts
svozza: Yeah, I'm in two minds myself. Let's postpone our decision for now and see where we are when the handler is more fleshed out.

packages/event-handler/src/rest/Route.ts
```diff
@@ -0,0 +1,19 @@
...
+  readonly path: Path;
+  readonly handler: RouteHandler;
+  readonly registeredAt: Date;
+
+  constructor(method: string, path: Path, handler: RouteHandler) {
+    this.id = `${method}:${path}`;
+    this.method = method.toUpperCase();
+    this.path = path;
+    this.handler = handler;
+    this.registeredAt = new Date();
```
leandrodamascena: Just out of curiosity: does knowing when the route was registered matter in any way or does it might just consume bytes of memory?

packages/event-handler/src/rest/RouteHandlerRegistry.ts
svozza: We can still get rid of the null check using arrays like you originally suggested and we don't pay the price of having to iterate the set to create that array in `getRoutesByMethod`.

packages/event-handler/src/rest/Route.ts
svozza: I was following the spec from the original issue but yeah, I'm not sure if we'd use it for anything.

leandrodamascena: I'd remove this field to save some bytes.

dreamorosi: Yea, transparently, it was a GenAI hallucination and an oversight on my side for not catching it.

Good callout @leandrodamascena - let's remove it.

svozza: Done!

dreamorosi (approved):

Issue #4100, opened 2025-07-01
Feature request: Add support for AppSync GraphQL batch resolvers


dreamorosi:
### Use case
In #3999 we added support for AppSync GraphQL single resolver, which allows customers to map a GraphQL action/request to a resolver. We should now add batch resolvers.

### Solution/User Experience
Building on the APIs introduced in #3999, we should add a new method `batchResolver` that accepts a resolver function and a configuration/options object. The config object should work similar to the `resolver` method added in the PR mentioned above and accept `typeName` and `fieldName`. The main difference for this new method is that it passes a list of `AppSyncResolverEvent` to the route handler.

### Alternative solutions
N/A

### Acknowledgment
- [x] This feature request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

arnabrahman:
Please assign me.

arnabrahman:
Hi, I’m having some confusion while resolving non-aggregate batch events.

If we look at the [Python docs example](https://docs.powertools.aws.dev/lambda/python/latest/core/event_handler/appsync/#advanced_batch_resolver_payloadjson_1), we’ll see that `post_id` is inside the `source` key. And if you try the [example](https://docs.powertools.aws.dev/lambda/python/latest/core/event_handler/appsync/#advanced_batch_resolver_individualpy) from the docs and print `post_id`, it’s empty.

The reason is [this](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/event_handler/appsync.py#L243), where `event.arguments` is being resolved. This should instead be `event.source`. In the aggregate [example](https://docs.powertools.aws.dev/lambda/python/latest/core/event_handler/appsync/#advanced_batch_resolverpy) as well, the actual data is parsed from `event.source`.

In my case, for batch events, I always receive the information in `source` while `arguments` is always empty. My batch implementation only works if I resolve `event.source` rather than `event.arguments`. 

Let me know if I am missing any context here, @dreamorosi @leandrodamascena

leandrodamascena:
[...]

PR #4218, opened 2025-07-27
feat(event-handler): add support for AppSync GraphQL batch resolvers

arnabrahman:
## Summary
This PR will add the ability to register batch resolvers for AppSync GraphQL API.
### Changes

- `batchResolver` method is introduced to register handler for batch events   

   ```ts
   
   app.batchResolver<{id: number}>(async (events) => {
     // Process all events in batch
     return events.map(event => ({ id: event.arguments.id, data: 'processed' }));
   }, {
     fieldName: 'getPosts'
   });
   
   export const handler = async (event, context) =>
     app.resolve(event, context);
   ```
   

- Process events individually

   
   If you want to process each event individually instead of receiving all events at once, you can set the
   `aggregate` option to `false`. In this case, the handler will be called once for each event in the batch,
   similar to regular resolvers.
   

   ```ts
   
   app.batchResolver(async (args, { event, context }) => {
     // Process individual request
     return { id: args.id, data: 'processed' };
   }, {
     fieldName: 'getPost',
     aggregate: false
   });
   
   export const handler = async (event, context) =>
     app.resolve(event, context);
   ```
   

- Strict error handling

   If you want stricter error handling when processing events individually, you can set the `raiseOnError` option
   to `true`. In this case, if any event throws an error, the entire batch processing will stop and the error
   will be propagated. Note that `raiseOnError` can only be used when `aggregate` is set to `false`.
   
   ```ts
   
   app.batchResolver(async (args, { event, context }) => {
     // Process individual request
     return { id: args.id, data: 'processed' };
   }, {
     fieldName: 'getPost',
     aggregate: false,
     raiseOnError: true
   });
   
   export const handler = async (event, context) =>
     app.resolve(event, context);
   ```
   
   You can also specify the type of the arguments using generic type parameters for non-aggregated handlers:
   
   ```ts
   
   const app = new AppSyncGraphQLResolver()
   
   app.batchResolver<{ postId: string }>(async (args, { event, context }) => {
     // args is typed as { postId: string }
     return { id: args.postId };
   }, {
     fieldName: 'getPost',
     aggregate: false
   });
   
   export const handler = async (event, context) =>
     app.resolve(event, context);
   ```
   
   As a decorator:
   
   ```ts
   
   class Lambda {
     ⁣@app.batchResolver({ fieldName: 'getPosts' })
     async handleGetPosts(events) {
       // Process batch of events
       return events.map(event => ({ id: event.arguments.id, data: 'processed' }));
     }
   
     ⁣@app.batchResolver({ fieldName: 'getPost', aggregate: false })
     async handleGetPost(args, { event, context }) {
       // Process individual request
       return { id: args.id, data: 'processed' };
     }
   
     ⁣@app.batchResolver({ fieldName: 'getPost', aggregate: false, raiseOnError: true })
     async handleGetPostStrict(args, { event, context }) {
       // Process individual request with strict error handling
       return { id: args.id, data: 'processed' };
     }
   
     async handler(event, context) {
       return app.resolve(event, context, {
         scope: this, // bind decorated methods to the class instance
       });
     }
   }
   
   const lambda = new Lambda();
   ```

- Also, `onBatchQuery` & `onBatchMutation` methods are defined for registering batch `Query` & `Mutation` events. `aggregate` & `raiseOnError` options are also available for these two methods.
     ```ts

   
   app.onBatchQuery<{ id: number }>('getPosts', async (events) => {
     // Process all events in batch
     return events.map(event => ({ id: event.arguments.id, data: 'processed' }));
   });
   
   export const handler = async (event, context) =>
     app.resolve(event, context);
   ```

   ```ts
     
     app.onBatchMutation<{ id: number }>('createPosts', async (events) => {
       // Process all events in batch
       return events.map(event => ({ id: event.arguments.id, status: 'created' }));
     });
     
     export const handler = async (event, context) =>
       app.resolve(event, context);
    ```

**Issue number:** #4100

---


packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
```diff
@@ -98,27 +158,185 @@ class AppSyncGraphQLResolver extends Router {
...
+   *
+   * Logs the provided error message and error object. If the error is an instance of
+   * `InvalidBatchResponseException` or `ResolverNotFoundException`, it is re-thrown.
+   * Otherwise, the error is formatted into a response using `#formatErrorResponse`.
+   *
+   * @param error - The error object to handle.
+   * @param errorMessage - A descriptive message to log alongside the error.
+   * @throws InvalidBatchResponseException | ResolverNotFoundException
+   */
+  #handleError(error: unknown, errorMessage: string) {
```
svozza, line 218: Can we use a better type than `unknown` here?

packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
```diff
@@ -98,27 +158,185 @@ class AppSyncGraphQLResolver extends Router {
...
+      }
+
+      return response;
+    }
+
+    const handler = options.handler as BatchResolverHandlerFn;
+    const results: unknown[] = [];
+
+    if (raiseOnError) {
+      for (const event of events) {
```
svozza, line 310: I think we should give users the option to do these in parallel with `Promise.all`.

packages/event-handler/src/types/appsync-graphql.ts
```diff
@@ -3,6 +3,64 @@ import type { AppSyncResolverEvent, Context } from 'aws-lambda';
 import type { RouteHandlerRegistry } from '../appsync-graphql/RouteHandlerRegistry.js';
 import type { Router } from '../appsync-graphql/Router.js';
 
+// #region BatchResolver fn
```
svozza, line 6: What does this comment mean?

dreamorosi: It's a comment we use in some places to facilitate folding & navigating the code, we have it in a few places throughout the codebase, for example:

Screenshot 2025-07-28 at 09 39 32

svozza: Ah I see!

packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
arnabrahman: When `raiseOnError` is `true`, we stop on the first error. But if we use `Promise.all`,  I believe we won’t be able to stop at the first error, since the promises will run in parallel.


However, it should be possible when `raiseOnError` is `false`. Also, do we want to give the user the option for this or just do parallel processing by default, similar to what has been done in [AppSyncEventResolver](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/event-handler/src/appsync-events/AppSyncEventsResolver.ts#L139)

packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
arnabrahman: The error is coming from a try/catch block, which treats caught errors as unknown by default. Since the error originates from the user’s handler, I can’t safely assume its type, it could be anything. 

That said, if you have a specific suggestion for a better type, I’m open to your recommendation.

packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
svozza: The way `Promise.all` works is that the first error causes all in-flight promises to be rejected so I think that's still OK. But we would need to use `Promise.allSettled` if `raiseOnError` was false. Good point about following what's done in AppSyncEventResolver, I think we should do that.

Small nit about the `raiseOnError` variable name, would `throwOnError` not be more appropriate. The word `raise` is more a Python thing. to my mind.

packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
svozza: Ah yes, of course, you can throw anything not just error objects so we can't make ay assumptions here.

packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
```diff
@@ -98,27 +158,185 @@ class AppSyncGraphQLResolver extends Router {
...
+   */
+  async #callBatchResolver(
+    events: AppSyncResolverEvent<Record<string, unknown>>[],
+    context: Context,
+    options: RouteHandlerOptions<Record<string, unknown>, boolean, boolean>,
+    resolveOptions?: ResolveOptions
+  ): Promise<unknown[]> {
+    const { aggregate, raiseOnError } = options;
+    this.logger.debug(
+      `Aggregate flag aggregate=${aggregate} & Graceful error handling flag raiseOnError=${raiseOnError}`
```
svozza:
```suggestion
      `Aggregate flag aggregate=${aggregate} and graceful error handling flag raiseOnError=${raiseOnError}`
```

packages/event-handler/tests/unit/appsync-graphql/AppSyncGraphQLResolver.test.ts
```diff
@@ -346,4 +472,178 @@ describe('Class: AppSyncGraphQLResolver', () => {
       });
     }
   );
+
+  it('logs a warning and returns early if one of the batch event is not compatible', async () => {
```
svozza:
```suggestion
  it('logs a warning and returns early if one of the batch events is not compatible', async () => {
```

packages/event-handler/tests/unit/appsync-graphql/AppSyncGraphQLResolver.test.ts
```diff
@@ -346,4 +472,178 @@ describe('Class: AppSyncGraphQLResolver', () => {
...
+    // Assess
+    expect(handler).toHaveBeenCalledTimes(1);
+    expect(handler).toHaveBeenCalledWith(events, { event: events, context });
+    expect(result).toEqual([
+      { id: '1', value: 'A' },
+      { id: '2', value: 'B' },
+    ]);
+  });
+
+  it('registers a batch resolver via direct function call and invokes it (aggregate=false) and (raiseOnError=true)', async () => {
```
svozza: Could these combinations of tests with different values of `aggregator` and `raiseOnError` be consolidated by using `it.each`?

packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
dreamorosi: Good callout about `throwOn...`, we use this in several other places in this version of Powertools for AWS:
- [Idempotency](https://docs.powertools.aws.dev/lambda/typescript/latest/features/idempotency/#making-idempotency-key-required)
- [Parameters](https://docs.powertools.aws.dev/lambda/typescript/latest/features/parameters/#partial-transform-failures-with-getmultiple)
- [Batch Processing](https://docs.powertools.aws.dev/lambda/typescript/latest/features/batch/#working-with-full-batch-failures)

As well as probably others, so +1 on changing it to this.

---

Regarding the `Promise.all`, if I remember correctly we didn't use it in the other Event Handler because even though the first rejected promises causes the "overall" promise to also be rejected, it doesn't abort the other in-flight promises, for example:

```ts
const promise1 = new Promise((resolve, reject) => {
  setTimeout(resolve, 2000, 'resolve1');
}).then((a) => {
  console.log('then1');
  return a;
});

const promise2 = new Promise((resolve, reject) => {
  setTimeout(reject, 1000, 'reject2');
}).then((a) => {
  console.log('then2');
  return a;
});

const promise3 = new Promise((resolve, reject) => {
  setTimeout(resolve, 3000, 'resolve3');
}).then((a) => {
  console.log('then3');
  return a;
});

Promise.all([promise1, promise2, promise3])
  .then((values) => {
    console.log('then', values);
  })
  .catch((err) => {
    console.log('catch', err);
  });
```

results in:

```sh
catch reject2
then1
then3
```

svozza: Ah yes, that makes sense!

arnabrahman: renamed `raiseOnError` to `throwOnError`

svozza: I still think we should run the event processing in parallel if `throwOnError` is `false` although not with `Promise.all` but rather `Promise.allSettled`.

arnabrahman: Yes, theoretically, we can use `Promise.allSettled` when `throwOnError` is `false`. I can't think of any obvious issues with it. @dreamorosi, any thoughts?

dreamorosi:
@svozza / @sdangol can you please take a look and leave a review? I believe most of the comments were addressed but unsure.

svozza:
There's one outstanding issue that we'd like your input on: https://github.com/aws-powertools/powertools-lambda-typescript/pull/4218#discussion_r2240400685.

packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
dreamorosi: Is there any scenario in which order does matter here? Does AppSync even guarantee ordering?

If no, then I agree, we could run them in parallel - the only concern I have is we'd be introducing a conditional behavior based on a slightly-unrelated property.

Unless customers are aware of this nuance we discussed above, the mental model would be kind of weird:
- if `throwOnError === false` things run in parallel
- else things run sequentially

Ideally I'd prefer to be more explicit, but I also don't want to add another property just for this.

Maybe we move forward with always sequential, and address this later with a better solution if there's usage/demand.

dreamorosi:
> There's one outstanding issue that we'd like your input on: [#4218 (comment)](https://github.com/aws-powertools/powertools-lambda-typescript/pull/4218#discussion_r2240400685).

Thank you, and sorry - I missed that one.

packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
svozza: `Promise.allSettled` will preserve the order of the array you pass in. As a customer, I would definitely want to be able to at least have the option to run these in parallel. My mental model for Nodejs in general is multiple independent async things should be done concurrently unless there's a very good reason not to. Waiting in the lambda for I/O to resolve before executing the next one is just wasting CPU cycles (and money) when using an event loop. But I agree, this is not a one way door so we can add later.

arnabrahman: Alright then, for now better to keep it simple and always process sequentially.

As for the ordering question, Leandro did mention in the original [issue](https://github.com/aws-powertools/powertools-lambda-typescript/issues/1166#issuecomment-2900275223) that the events sent to AppSync must be in the same order as they were received.

Although `Promise.allSettled` does preserve the order, [reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled#return_value:~:text=in%20the%20order%20of%20the%20promises%20passed%2C%20regardless%20of%20completion%20order)

> The fulfillment value is an array of objects, each describing the outcome of one promise in the iterable, in the order of the promises passed, regardless of completion order.

svozza (approved):

packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
dreamorosi: You're both correct - however I think we need to keep in mind the overall design of the feature and give customers a coherent mental model.

Let's remember that we're specifically working on "Batch Processing", by default this experience gives you a list of items (aka the batch) and you're in complete control over how you want to process them, meaning you get a list as input, and you're expected to return a list as output - how you process it is up to you, including parallelizing it if you want.

Conversely, you can also set `aggregate` to `false`, and in this case we'll iterate through the batch for you. Reading the Python docs, this feature is intended when:
> In rare scenarios, you might want to process each item individually, trading ease of use for increased latency as you handle one batch item at a time.

With this is mind, I think for now we could move forward with sequential processing and then reconsider in the future if needed, since customers will have both options already.

svozza:
Thank you so @arnabrahman for all your hard work on this!

Issue #4134, opened 2025-07-04
Maintenance: Add warning for overwriting dimensions in setDefaultDimensions method


dreamorosi:
## Summary

The Metrics utility currently has inconsistent behavior when overwriting dimensions. The `addDimension()` and `addDimensions()` methods properly emit warnings when overwriting existing dimensions, but the `setDefaultDimensions()` method silently overwrites existing default dimensions without any warning.

This creates an inconsistent user experience where some dimension overwrite scenarios provide helpful warnings while others do not.

## Why is this needed?

1. **Consistency**: All dimension overwrite scenarios should behave consistently by providing warnings
2. **Developer Experience**: Warnings help developers catch potential bugs where they might accidentally overwrite important dimensions  
3. **Alignment with existing patterns**: The codebase already has the infrastructure and patterns for dimension overwrite warnings
4. **Parity with other methods**: `addDimension()` and `addDimensions()` already implement this behavior successfully

## Which area does this relate to?

Metrics

## Solution

Add warning logic to the `setDefaultDimensions()` method to detect when existing default dimensions are being overwritten and emit appropriate warning messages using the existing logger infrastructure.

The implementation should:
- Check if any keys in the new dimensions object already exist in `this.defaultDimensions`
- Emit a warning message similar to the existing ones: `"Dimension \"${key}\" has already been added. The previous value will be overwritten."`
- Use the existing `this.#logger.warn()` pattern for consistency

### Test case that should pass after implementation:

```typescript
it('warns when setDefaultDimensions overwrites existing dimensions', () => {
    // Prepare
    const metrics = new Metrics({
      namespace: DEFAULT_NAMESPACE,
      defaultDimensions: { environment: 'prod' },
    });

    // Act
    metrics.setDefaultDimensions({ region: 'us-east-1' });
    metrics.setDefaultDimensions({
      environment: 'staging', // overwrites default dimension
    });

    // Assess
    expect(console.warn).toHaveBeenCalledOnce();
    expect(console.log).toHaveBeenCalledWith(
      'Dimension "environment" has already been added. The previous value will be overwritten.'
    );
  });
```

## Acknowledgment

- [x] This request meets Powertools for AWS Lambda (TypeScript) Tenets
- [x] Should this be considered in other Powertools for AWS Lambda languages? i.e. Python, Java, and .NET

## Future readers

Please react with 👍 and your use case to help us understand customer demand.

uttam282005:
[...]

PR #4222, opened 2025-07-28
fix(metrics): emit warning when default dimensions are overwritten

uttam282005:
## Summary
Adds warning messages to the Metrics `setDefaultDimensions` method when overwriting existing default dimensions, ensuring consistency and parity with `addDimension`/`addDimensions` methods.

### Changes

This pull request improves the robustness of the `setDefaultDimensions` method in the Metrics utility by:

- Moved setEnvConfig() and setConsole() calls from setOptions() to the constructor to ensure logger is initialized before any method uses it.
- Ensuring it gracefully handles `undefined`, `null`, and invalid values for default dimensions.
- Updated the `setDefaultDimensions()` method in the Metrics utility to emit a warning when a default dimension is being overwritten.
- Logging a warning if an invalid dimension name or value is provided (i.e., non-string or empty).
- Adding new unit tests to verify the above behaviors, including:
  - Ensuring that calling `setDefaultDimensions(undefined)` results in immediate return, with no warning emitted and only default dimensions applied.
  - Verifying that invalid dimension values (undefined, null, empty strings) or names are skipped, a warning is emitted in the console, and only valid dimensions are set.

This PR addresses [#4134](https://github.com/aws-powertools/powertools-lambda-typescript/issues/4134) by improving input validation and error reporting for the `setDefaultDimensions` method in the Metrics module, to prevent invalid dimension names/values from being set.

**Issue number:** #4134

### Tests Added

The following tests were added:

```typescript
it('returns immediately if dimensions is undefined', () => {
  // Prepare
  const metrics = new Metrics({
    singleMetric: true,
    namespace: DEFAULT_NAMESPACE,
  });
  // Act
  metrics.setDefaultDimensions(undefined);
  metrics.addMetric('myMetric', MetricUnit.Count, 1);
  // Assess: No warning, only default dimensions/service
  expect(console.warn).not.toHaveBeenCalled();
  expect(console.log).toHaveEmittedEMFWith(
    expect.objectContaining({
      service: 'hello-world',
    })
  );
});

it.each([
  { value: undefined, name: 'valid-name' },
  { value: null, name: 'valid-name' },
  { value: '', name: 'valid-name' },
  { value: 'valid-value', name: '' },
])(
  'skips invalid default dimension values in setDefaultDimensions ($name)',
  ({ value, name }) => {
    // Arrange
    const metrics = new Metrics({
      singleMetric: true,
      namespace: DEFAULT_NAMESPACE,
    });

    // Act
    metrics.setDefaultDimensions({
      validDimension: 'valid',
      [name as string]: value as string,
    });

    metrics.addMetric('test', MetricUnit.Count, 1);

    // Assert
    expect(console.warn).toHaveBeenCalledWith(
      `The dimension ${name} doesn't meet the requirements and won't be added. Ensure the dimension name and value are non empty strings`
    );

    expect(console.log).toHaveEmittedEMFWith(
      expect.objectContaining({ validDimension: 'valid' })
    );

    expect(console.log).toHaveEmittedEMFWith(
      expect.not.objectContaining({ [name]: value })
    );
  }
);
```

### setDefaultDimensions Update

- Modified to emit a warning if a dimension name or value does not meet requirements (non-empty string).

### DIFF
--- a/packages/metrics/src/Metrics.ts
+++ b/packages/metrics/src/Metrics.ts
@@ -240,8 +240,10 @@ class Metrics extends Utility implements MetricsInterface {
     super();
 
     this.dimensions = {};
-    this.setOptions(options);
+    this.setEnvConfig();
+    this.setConsole();
     this.#logger = options.logger || this.console;
+    this.setOptions(options);
   }
 
   /**
@@ -824,13 +826,39 @@ class Metrics extends Utility implements MetricsInterface {
    * @param dimensions - The dimensions to be added to the default dimensions object
    */
   public setDefaultDimensions(dimensions: Dimensions | undefined): void {
+    if (!dimensions) return;
+
+    const cleanedDimensions: Dimensions = {};
+
+    for (const [key, value] of Object.entries(dimensions)) {
+      if (
+        isStringUndefinedNullEmpty(key) ||
+        isStringUndefinedNullEmpty(value)
+      ) {
+        this.#logger.warn(
+          `The dimension ${key} doesn't meet the requirements and won't be added. Ensure the dimension name and value are non empty strings`
+        );
+        continue;
+      }
+
+      if (Object.hasOwn(this.defaultDimensions, key)) {
+        this.#logger.warn(
+          `Dimension "${key}" has already been added. The previous value will be overwritten.`
+        );
+      }
+
+      cleanedDimensions[key] = value;
+    }
+
     const targetDimensions = {
       ...this.defaultDimensions,
-      ...dimensions,
+      ...cleanedDimensions,
     };
-    if (MAX_DIMENSION_COUNT <= Object.keys(targetDimensions).length) {
+
+    if (Object.keys(targetDimensions).length >= MAX_DIMENSION_COUNT) {
       throw new Error('Max dimension count hit');
     }
+
     this.defaultDimensions = targetDimensions;
   }
 
@@ -1058,8 +1086,6 @@ class Metrics extends Utility implements MetricsInterface {
       functionName,
     } = options;
 
-    this.setEnvConfig();
-    this.setConsole();
     this.setCustomConfigService(customConfigService);
     this.setDisabled();
     this.setNamespace(namespace);
--- a/packages/metrics/src/Metrics.ts
+++ b/packages/metrics/src/Metrics.ts
@@ -828,17 +828,7 @@ class Metrics extends Utility implements MetricsInterface {
    * @param dimensions - The dimensions to be added to the default dimensions object
    */
   public setDefaultDimensions(dimensions: Dimensions | undefined): void {
-    if (isNullOrUndefined(dimensions)) {
-      this.#logger.warn(
-        'No dimensions were supplied to setDefaultDimensions. Skipping update.'
-      );
-      return;
-    }
-
-    if (!isRecord(dimensions)) {
-      this.#logger.warn(
-        'Invalid dimensions type provided to setDefaultDimensions. Expected an object.'
-      );
+    if (isNullOrUndefined(dimensions) || !isRecord(dimensions)) {
       return;
     }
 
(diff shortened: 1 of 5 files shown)


packages/metrics/src/Metrics.ts
```diff
@@ -240,8 +240,10 @@ class Metrics extends Utility implements MetricsInterface {
     super();
 
     this.dimensions = {};
-    this.setOptions(options);
+    this.setEnvConfig();
+    this.setConsole();
     this.#logger = options.logger || this.console;
+    this.setOptions(options);
```
sdangol, line 248: If the `setOptions` is kept there before the logger initialization, it should call the other two methods `    this.setEnvConfig()` and `this.setConsole()`.

packages/metrics/src/Metrics.ts
```diff
@@ -824,13 +826,39 @@ class Metrics extends Utility implements MetricsInterface {
    * @param dimensions - The dimensions to be added to the default dimensions object
    */
   public setDefaultDimensions(dimensions: Dimensions | undefined): void {
+    if (!dimensions) return;
```
sdangol: It might be useful to log a warning here that the dimension was not supplied

packages/metrics/src/Metrics.ts
```diff
@@ -824,13 +826,44 @@ class Metrics extends Utility implements MetricsInterface {
    * @param dimensions - The dimensions to be added to the default dimensions object
    */
   public setDefaultDimensions(dimensions: Dimensions | undefined): void {
+    if (!dimensions) {
```
sdangol: This check could be more specific using `isNullOrUndefined` and `isRecord` utility functions

packages/metrics/src/Metrics.ts
sdangol: Sorry, I get why this was done now. Resolving this comment

packages/metrics/src/Metrics.ts
```diff
@@ -824,13 +828,51 @@ class Metrics extends Utility implements MetricsInterface {
    * @param dimensions - The dimensions to be added to the default dimensions object
    */
   public setDefaultDimensions(dimensions: Dimensions | undefined): void {
+    if (isNullOrUndefined(dimensions)) {
+      this.#logger.warn(
```
sdangol: Hi @uttam282005, really sorry for the back and forth, but I think you initial idea of not logging a warning seems more better. I realized that this gets called every time the class is instantiated and a lot of customers might not be passing the default dimensions. In that case, it wouldn't be appropriate to log a warning because they didn't pass an optional parameter. Can you please change it back to not log a warning. Sorry about that.

Also, you could just do `isNullOrUndefined || !isRecord(dimensions)` for the check in a single `if` block and return

packages/metrics/tests/unit/creatingMetrics.test.ts
```diff
@@ -182,6 +182,9 @@ describe('Creating metrics', () => {
     const metrics = new Metrics({
       singleMetric: false,
       namespace: DEFAULT_NAMESPACE,
+      defaultDimensions: {
```
sdangol: @uttam282005 Can you also remove these `defaultDimensions` that you had to put in the test to prevent the errors. I think this should be ok without them now.

svozza:
Great work @uttam282005! This will be included in today's release.

sdangol:
Hi @uttam282005, we were trying to do a release today but our e2e tests were failing. We realized that there were duplicate logs being generated. This seems to be because of the `setDefaultDimensions` method being called in multiple places without invoking it explicitly. I've created an issue for this: #4227  
Would you like to work on a fix for this?

uttam282005:
Thanks for flagging this @sdangol, and apologies for the oversight. I’d be happy to work on a fix.

svozza:
Absolutely no need to apologise! This is why we have E2E tests because it's very hard to know how all the pieces in a module interact when you're working on a specific piece of the code base.

PR #4226, opened 2025-07-29
Revert "fix(metrics): emit warning when default dimensions are overwritten"

svozza:
Reverts aws-powertools/powertools-lambda-typescript#4222

The change in this PR has caused duplicate log messages to be created that are causing the e2e tests to fail. We will need to handle these duplicate logs appearing before we release the change as it will be confusing for customers.


leandrodamascena (changes requested):
Hey @svozza pls open an issue and keep the acknowledge section in the PR body.

leandrodamascena (approved):

PR #4235, opened 2025-07-30
Fix/service dimension cleanup

uttam282005:
## Summary
This PR restores and improves the warning mechanism for overwritten default dimensions in the `Metrics` utility:

### Changes
* **Restores** the warning logic that was unintentionally removed in a prior revert.
* **Fixes a bug** where calling `captureColdStartMetric()` would emit **duplicate warnings** for the `service` dimension.
* **Refines warning behavior** to only trigger when the overwritten `service` dimension is different from the internal default (i.e., truly user-defined).

---

### Additional Fix

Previously, passing `defaultDimensions` with a `service` key in the constructor triggered **unnecessary warnings**, beacause `service` was set internally as a default dimension via `setService()`:

```ts
new Metrics({
  defaultDimensions: {
    service: 'auth-service', // <-- warning was wrongly emitted 
  },
});
```

Now, this behavior is handled correctly. **No warning is emitted** if the `service` in `defaultDimensions` is the same as the internal default service name.

---

### Final Fix Summary

* Avoids redundant call to `setDefaultDimensions()` inside `captureColdStartMetric()`, which previously caused a second warning.
* Prevents unnecessary or duplicate warnings when instantiating `Metrics` with `service` value in `defaultDimensions`.

**Issue number:** #4134

---

### Note

Instantiating the object like this will still emit a warning as user is trying to overwrite previously passed `serviceName`, even if they are equal. This behaviour is consistent with other methods like `addDimensions` and `addDimension`.
```ts
const metrics = new Metrics({
  serviceName: 'some-service',
  defaultDimensions: {
    service: 'some-service', // ⚠️ Warning: service already added
  },
});
```

### DIFF
--- a/packages/metrics/src/Metrics.ts
+++ b/packages/metrics/src/Metrics.ts
@@ -1,7 +1,9 @@
 import { Console } from 'node:console';
 import {
   isIntegerNumber,
+  isNullOrUndefined,
   isNumber,
+  isRecord,
   isString,
   isStringUndefinedNullEmpty,
   Utility,
@@ -240,8 +242,10 @@ class Metrics extends Utility implements MetricsInterface {
     super();
 
     this.dimensions = {};
-    this.setOptions(options);
+    this.setEnvConfig();
+    this.setConsole();
     this.#logger = options.logger || this.console;
+    this.setOptions(options);
   }
 
   /**
@@ -824,13 +828,41 @@ class Metrics extends Utility implements MetricsInterface {
    * @param dimensions - The dimensions to be added to the default dimensions object
    */
   public setDefaultDimensions(dimensions: Dimensions | undefined): void {
+    if (isNullOrUndefined(dimensions) || !isRecord(dimensions)) {
+      return;
+    }
+
+    const cleanedDimensions: Dimensions = {};
+
+    for (const [key, value] of Object.entries(dimensions)) {
+      if (
+        isStringUndefinedNullEmpty(key) ||
+        isStringUndefinedNullEmpty(value)
+      ) {
+        this.#logger.warn(
+          `The dimension ${key} doesn't meet the requirements and won't be added. Ensure the dimension name and value are non empty strings`
+        );
+        continue;
+      }
+
+      if (Object.hasOwn(this.defaultDimensions, key)) {
+        this.#logger.warn(
+          `Dimension "${key}" has already been added. The previous value will be overwritten.`
+        );
+      }
+
+      cleanedDimensions[key] = value;
+    }
+
     const targetDimensions = {
       ...this.defaultDimensions,
-      ...dimensions,
+      ...cleanedDimensions,
     };
-    if (MAX_DIMENSION_COUNT <= Object.keys(targetDimensions).length) {
+
+    if (Object.keys(targetDimensions).length >= MAX_DIMENSION_COUNT) {
       throw new Error('Max dimension count hit');
     }
+
     this.defaultDimensions = targetDimensions;
   }
 
@@ -1058,8 +1090,6 @@ class Metrics extends Utility implements MetricsInterface {
       functionName,
     } = options;
 
-    this.setEnvConfig();
-    this.setConsole();
     this.setCustomConfigService(customConfigService);
     this.setDisabled();
     this.setNamespace(namespace);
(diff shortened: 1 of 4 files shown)


packages/metrics/tests/unit/dimensions.test.ts
```diff
@@ -552,4 +552,104 @@ describe('Working with dimensions', () => {
...
+    // Prepare
+    const metrics = new Metrics({
+      singleMetric: true,
+      namespace: DEFAULT_NAMESPACE,
+    });
+
+    // Act
+    metrics.addMetric('myMetric', MetricUnit.Count, 1);
+
+    // Assert
```
dreamorosi:
```suggestion
    // Assess
```

dreamorosi: Please make sure to replace all in this file

packages/metrics/src/Metrics.ts
```diff
@@ -824,13 +823,46 @@ class Metrics extends Utility implements MetricsInterface {
...
+        this.#logger.warn(
+          `The dimension ${key} doesn't meet the requirements and won't be added. Ensure the dimension name and value are non empty strings`
+        );
+        continue;
+      }
+
+      if (Object.hasOwn(this.defaultDimensions, key)) {
+        const currentValue = this.defaultDimensions[key];
+        const suppressOverwriteWarning =
+          key === 'service' && currentValue === this.defaultServiceName;
```
sdangol, line 846: @uttam282005 I tried to run the e2e tests against the PR, and [one of the tests](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/layers/tests/e2e/layerPublisher.test.ts#L141) is failing. It seems like there's still one warning that gets logged if we use `captureColdStartMetric`.
I'm not sure, but I think it has to do with `this.defaultServiceName`. We're doing this in setService to get the service, do we need something similar?
```typescript
service ||
this.getCustomConfigService()?.getServiceName() ||
this.#envConfig.serviceName ||
this.defaultServiceName;
```

uttam282005: I believe the warning occurs because the `service` dimension is initially set via `setService()`, this value can come from a user-provided `serviceName`, an environment variable, or a custom config service. Later, when `setDefaultDimensions()` is called (and includes a `service` key), we get a warning indicating the `service` dimension is being overwritten.

In many cases, this warning is correct, it's a user attempting to override a previously set user-defined `service` value, which we want to surface. However, the problem arises when `setDefaultDimensions()` is called internally (e.g., within `captureColdStartMetric()` or during initialization like `singleMetric()` ). In these cases, the overwrite is intentional and should not trigger a warning.

To address this, the cleanest solution I can think of is introducing a private boolean flag that marks such internal calls. This flag would let us suppress the warning when `setDefaultDimensions()` is invoked by internal logic rather than user input, preserving the intent of the warning without polluting logs in expected flows.

uttam282005: The root cause is the call to single metric which is emitting the same behaviour that I have written in the note section of the PR.

packages/metrics/src/Metrics.ts
```diff
@@ -1058,8 +1090,6 @@ class Metrics extends Utility implements MetricsInterface {
       functionName,
     } = options;
 
-    this.setEnvConfig();
```
dreamorosi, line 1061: Why were these two removed and extracted in the constructor?

uttam282005: setEnvConfig() and setConsole() calls from setOptions() were moved to the constructor to ensure logger is initialized before any method (eg. setDefaultDimensions) uses it.

uttam282005:
@dreamorosi @sdangol I’d like to get your thoughts on how best to solve this.

sdangol:
@uttam282005 I think having the flag to maintain the internal calls would probably add some complexity to this. Let me think this through for some more time, discuss with the team and get back to you

uttam282005:
```ts
private static supressWarnings = false;

private static supressSetDefaultDimensionsWarnings<T>(fn: () => T): T {
  Metrics.supressWarnings = true;
  try {
    return fn();
  } finally {
    Metrics.supressWarnings = false;
  }
}
```

How is this approach?

uttam282005:
I will close this PR and create a new one.

PR #4251, opened 2025-08-02
refactor(logger): replace EnvironmentVariablesService class with helper functions

### LINKED ISSUE #4116: Maintenance: replace `EnvironmentVariablesService` class with helper functions in Logger
### Summary

In https://github.com/aws-powertools/powertools-lambda-typescript/pull/3945 we added new zero-dependency functional utilities to read and parse environment variables. These new functions supersede the old class based environment service approach. We should should refactor any use of the class based system to the new function based approach.

### Why is this needed?

The rationale for the introduction of these utilities as per https://github.com/aws-powertools/powertools-lambda-typescript/pull/3945 is as follows:

> [M]ost of the advantage to customers will come in the form of smaller utilities over time, since we'll be able to use these helpers across other Powertools for AWS utilities and replace the existing EnvironmentVariablesService class-based model which is extremely verbose and requires props drilling if you want to access env variables deep into an inheritance stack.

This change will help us to eventually remove the `EnvironmentVariablesService` entirely from the code base.

Logger

### Solution

The logger currently extends the `EnvironmentVariablesService` class with logger specific environment variable methods. The main usage of this class is in the `Logger.ts` file in the [setOptions](https://github.com/aws-powertools/powertools-lambda-typescript/blob/263db2d74e558adb9b088174a5500de6c29488d0/packages/logger/src/Logger.ts#L1284) method. This method is called by the constructor on instantiation and ensures that the environment variables are only read once.

```ts
import { EnvironmentVariablesService } from '@aws-lambda-powertools/commons';

// ...

public constructor(options: ConstructorOptions = {}) {
    super();
    const { customConfigService, ...rest } = options;
    this.setCustomConfigService(customConfigService);
    this.setOptions(rest);
    // ...
  }

private setOptions(
    options: Omit<ConstructorOptions, 'customConfigService'>
  ): this {
    const {
      logLevel,
      serviceName,
      sampleRateValue,
      logFormatter,
      persistentKeys,
      persistentLogAttributes, // deprecated in favor of persistentKeys
      environment,
      jsonReplacerFn,
      logRecordOrder,
      logBufferOptions,
      correlationIdSearchFn,
    } = options;
    
   // ...

    // configurations that affect Logger behavior
    const AlcLogLevel = this.getEnvVarsService().getAwsLogLevel();
    if (this.isValidLogLevel(AlcLogLevel)) {
      this.#alcLogLevel = AlcLogLevel;
    }
    
    this.setLogEvent();
    this.setInitialLogLevel(logLevel);
    this.setInitialSampleRate(sampleRateValue);

    // configurations that affect how logs are printed
    this.setLogFormatter(logFormatter, logRecordOrder);
    this.setConsole();
    this.setLogIndentation();
    this.#jsonReplacerFn = jsonReplacerFn;
    this.#setLogBuffering(logBufferOptions);
    this.#correlationIdSearchFn = correlationIdSearchFn;

    return this;
  }
```

Note that the multiple `set*` methods called in `setOptions` also read from the environment such as [setLogEvent](https://github.com/aws-powertools/powertools-lambda-typescript/blob/263db2d74e558adb9b088174a5500de6c29488d0/packages/logger/src/Logger.ts#L1243):

```ts
private setLogEvent(): void {
    if (this.getEnvVarsService().getLogEvent()) {
      this.logEvent = true;
    }
  }
```

We could refactor this call to the `this.getEnvVarsService()` method like so:

```ts
import { getBooleanFromEnv } from '@aws-lambda-powertools/commons/utils/env';

private setLogEvent(): void {
    if (getBooleanFromEnv('POWERTOOLS_LOGGER_LOG_EVENT')) {
      this.logEvent = true;
    }
  }
```

Other calls to `this.getEnvVarsService()` could be refactored similarly using the new functional utilities.

### Acknowledgment

### Future readers

jaimellamasi:
### Changes

This PR removes the legacy class-based EnvironmentVariablesService methods from the Logger package.

**Issue number: closes #4116** 

---


packages/logger/tests/unit/configFromEnv.test.ts
```diff
@@ -169,6 +182,21 @@ describe('Class: EnvironmentVariablesService', () => {
     expect(value).toEqual(0.01);
   });
 
+  it.each([undefined, ''])(
```
dreamorosi: Since we removed all usage of `EnvironmentVariablesService` from the `Logger` class, we should also remove the files for the `EnvironmentVariablesService` itself - this should also make this test redundant.

jaimellamasi: in order to fully remove `EnvironmentVariablesService` class from `Logger`, I think we need to remove it from `PowertoolsLogFormatter` constructor and so also refactor `LogFormatter`. I was not sure if this was expected on this PR, but sure I can do it just let me know

dreamorosi: Hi sorry if the issue wasn't clear-  yes, please, let's do that too if you don't mind!

jaimellamasi:
@dreamorosi I just pushed the changes to remove `EnvironmentVariablesService` from `Logger`. You will notice some assertion changed on the unit tests related to `stack` because of the behaviour of `isDevMode()` compared to `this. envVarsService()?.isDevMode()`. I am not 100% sure changes are good. I wait your feedback, thanks!

dreamorosi:
> @dreamorosi I just pushed the changes to remove `EnvironmentVariablesService` from `Logger`. You will notice some assertion changed on the unit tests related to `stack` because of the behaviour of `isDevMode()` compared to `this. envVarsService()?.isDevMode()`. I am not 100% sure changes are good. I wait your feedback, thanks!

Sorry, GitHub was having issues and wasn't able to engage with the PR.

I have reviewed the PR and also ran the end to end tests [which are passing](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/16755735734). I am unsure why the previous tests were passing but the current assertions look correct to me.

PR #4261, opened 2025-08-04
test(event-handler): coverage 100% for AppSync GraphQL

### LINKED ISSUE #4260: Maintenance: test coverage misreported in AppSync Event Handler
### Summary

In #4100 we added new features to the AppSync GraphQL event handler. The coverage for the new code and package appeared as 100%, however when running the unit tests locally (which uses a different command) the coverage drops.

We should consolidate the commands so they run in the same way regardless of the scope, and potentially fix the coverage issues if still there.

### Why is this needed?

So that we can maintain high coverage and make sure contributors don't encounter friction when running the local dev automation.

Event Handler

### Solution

_No response_

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR refactors slightly the AppSync GraphQL event handler so that the `aggregate` and `throwOnError` flags have default values at the registry level, rather than when registering the handler - this helps remove code branches and increase test coverage.

Additionally, the PR adds missing unit test cases for decorator-based usage of batch query and mutation resolvers.

Finally, the PR consolidates the Vitest config according to the latest guidance.

**Issue number:** closes #4260

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,7 +36,7 @@
         "markdownlint-cli2": "^0.18.1",
         "middy5": "npm:@middy/core@^5.4.3",
         "middy6": "npm:@middy/core@^6.0.0",
-        "typedoc": "^0.28.8",
+        "typedoc": "^0.28.9",
         "typedoc-plugin-missing-exports": "^4.0.0",
         "typescript": "^5.8.3",
         "vitest": "^3.0.9"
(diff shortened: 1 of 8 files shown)


packages/event-handler/src/appsync-graphql/RouteHandlerRegistry.ts
```diff
@@ -40,7 +40,13 @@ class RouteHandlerRegistry {
   public register(
     options: RouteHandlerOptions<Record<string, unknown>, boolean, boolean>
   ): void {
-    const { fieldName, handler, typeName, throwOnError, aggregate } = options;
+    const {
+      fieldName,
+      handler,
+      typeName,
+      throwOnError = false,
```
svozza, line 47: The behaviour downstream will be subtly different now, default parameters will only only occur if the value is `undefined` whereas the the nullish coalescing operator used later would have caught both`null` and `undefined`. Don't think it's a issue but just wanted mention it.

svozza (approved):

packages/event-handler/src/appsync-graphql/RouteHandlerRegistry.ts
dreamorosi: I see and you are right - customers should not pass `null` when registering the handler and TypeScript would complain since the parameters are set to `boolean | undefined`, however it might happen if you're using JavaScript.

Do you think I should make a commit to handle this? I could replace `throwOnError = false` with:

```ts
const {
  // other props
  throwOnError,
  // other props
} = props;

let actualThrowOnError = isNullOrUndefined(throwOnError) ? false : true;
```

what do you think?

svozza: This issue has actually burned me in pure JS projects before but I would be quite surprised if people were passing `null` in this options object in a JS project, it feels very unnatural to me. I think we're OK leaving as is.

PR #4287, opened 2025-08-07
improv(metrics): emit warning when default dimensions are overwritten

### LINKED ISSUE #4134: Maintenance: Add warning for overwriting dimensions in setDefaultDimensions method

uttam282005:
### Changes

This PR restores and improves the warning mechanism for overwritten default dimensions in the Metrics utility:

### Key Updates:

* Introduced a static utility method `suppressSetDefaultDimensionsWarnings()` to temporarily suppress warnings during internal `setDefaultDimensions()` calls.
* Fixes a bug where calling captureColdStartMetric() would emit duplicate warnings for the service dimension.
* Refines warning behavior to only trigger when the overwritten service dimension is different from the internal default (i.e., truly user-defined).

### Affected Code:

* `metrics.ts` (suppression logic added)
* `dimension.test.ts` (added tests covering the changes)
* `captureColdStartMetric` updated to wrap internal `setDefaultDimensions` call
* Constructor logic updated to avoid redundant warnings when values match

**Issue number:** closes #4134

---


packages/metrics/tests/unit/customTimestamp.test.ts
dreamorosi, line 1: This seems unrelated to the PR, was the change intentional?

uttam282005: This was actually part of my previous commit where I have to update metric initialisations to avoid warnings related to default service name.  It's there because I cherry picked that commit and made changes on top of it.
I can change it back if you want but it doesn't really have any functional significance.

dreamorosi: Understand, thanks for explaining - please revert all the unrelated changes so we can keep the diff clean and focused on what's actually changing in this PR.

uttam282005:
sure.

dreamorosi:
@svozza I made a few commits to remove the workaround used to avoid the double warning, and refactored some of the internal logic to remove the root cause of the double warning.

The end to end tests [are passing](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/16991573186), if you can on Monday please take a look, since I'm the last committer I can't self-approve it.

PR #4297, opened 2025-08-08
feat(event-handler): implement route matching & resolution system for rest handler

### LINKED ISSUE #4140: Feature request: Implement Route Matching & Resolution System for Event Handler

svozza:
## Summary
Implements the route matching and resolution system by following the pattern of the `appsync-event` handler and centralising the resolution logic in the `RouteHandlerRegistry`. This different from the proposed solution in #4140, which envisioned multiple classes involved in the routing and matching logic. This design is simpler and has less moving parts.

### Changes

- `RouteHandlerRegistry`
  - Changes the internal representation of the registry to contain registries for static and dynamic routes. The `register` function is responsible for deciding which goes where.
  - Added a method to determine the specificity of dynamic routes: routes with fewer parameters and more path segments are considered more specific.
  - When routes are added to the dynamic routes array, the addition is done in a way to ensure the array stays sorted in ascending order of specificity, i.e., the most specific route will be the first element etc. Doing so removes the need for the `refresh` method proposed that would have been called each time a route was added. This is also done in the `register` method.
  - The `resolve` method takes a HTTP method and a path
    - first we check the static route registry to find the handler
    - if that fails we check the dynamic registry, if we find a matching handler we extract any parameters from the path, parse and validate them, and return the handler, the parsed parameters and the raw parameters.
    - if that fails we return `null`.
- We have added two new functions to the `utils` module, although I am leaning towards moving them to be private methods in the `RouteHandlerRegistry` class as I don't think they're going to be used anywhere else:
  - `processParams`, which URL decodes any parameters in the path
  - `validateParams`, which throws a validation error if the parameters have invalid values
- Added unit tests with particular focus on ensuring the correctness of the specificity sorting. Note: the diff of the test file is so larger that GitHub automatically hides it.

### Note

There was one part of the initial implementation that I wasn't able to incorporate. There was a recursive type that was defined but it was never used anywhere so I wasn't sure what to do with it:

```ts
type ExtractParams<T extends string> = 
  T extends `${infer Start}:${infer Param}/${infer Rest}`
    ? { [K in Param]: string } & ExtractParams<`${Start}${Rest}`>
    : T extends `${infer Start}:${infer Param}`
    ? { [K in Param]: string }
    : {};
```

**Issue number:** closes #4140

---

### DIFF
--- a/packages/event-handler/src/rest/RouteHandlerRegistry.ts
+++ b/packages/event-handler/src/rest/RouteHandlerRegistry.ts
@@ -5,15 +5,11 @@ import type {
   Path,
   RouteHandlerOptions,
   RouteRegistryOptions,
+  ValidationResult,
 } from '../types/rest.js';
 import { ParameterValidationError } from './errors.js';
 import type { Route } from './Route.js';
-import {
-  compilePath,
-  processParams,
-  validateParams,
-  validatePathPattern,
-} from './utils.js';
+import { compilePath, validatePathPattern } from './utils.js';
 
 class RouteHandlerRegistry {
   readonly #staticRoutes: Map<string, Route> = new Map();
@@ -48,6 +44,39 @@ class RouteHandlerRegistry {
     const bSegments = b.path.split('/').length;
 
     return bSegments - aSegments;
+  }
+  /**
+   * Processes route parameters by URL-decoding their values.
+   * @param params - Raw parameter values extracted from the route path
+   * @returns Processed parameters with URL-decoded values
+   */
+  #processParams(params: Record<string, string>): Record<string, string> {
+    const processed: Record<string, string> = {};
+
+    for (const [key, value] of Object.entries(params)) {
+      processed[key] = decodeURIComponent(value);
+    }
+
+    return processed;
+  }
+  /**
+   * Validates route parameters to ensure they are not empty or whitespace-only.
+   * @param params - Parameters to validate
+   * @returns Validation result with success status and any issues found
+   */
+  #validateParams(params: Record<string, string>): ValidationResult {
+    const issues: string[] = [];
+
+    for (const [key, value] of Object.entries(params)) {
+      if (!value || value.trim() === '') {
+        issues.push(`Parameter '${key}' cannot be empty`);
+      }
+    }
+
+    return {
+      isValid: issues.length === 0,
+      issues,
+    };
   }
   /**
    * Registers a route in the registry after validating its path pattern.
@@ -99,7 +128,6 @@ class RouteHandlerRegistry {
       this.#staticRoutes.set(route.id, route);
     }
   }
-
   /**
    * Resolves a route handler for the given HTTP method and path.
    *
@@ -142,9 +170,9 @@ class RouteHandlerRegistry {
       if (match?.groups) {
         const params = match.groups;
 
-        const processedParams = processParams(params);
+        const processedParams = this.#processParams(params);
 
-        const validation = validateParams(processedParams);
+        const validation = this.#validateParams(processedParams);
 
         if (!validation.isValid) {
           throw new ParameterValidationError(validation.issues);
--- a/packages/event-handler/src/rest/utils.ts
+++ b/packages/event-handler/src/rest/utils.ts
@@ -43,32 +43,3 @@ export function validatePathPattern(path: Path): ValidationResult {
     issues,
   };
 }
-
-export function processParams(
-  params: Record<string, string>
-): Record<string, string> {
-  const processed: Record<string, string> = {};
-
-  for (const [key, value] of Object.entries(params)) {
-    processed[key] = decodeURIComponent(value);
-  }
-
-  return processed;
-}
-
-export function validateParams(
-  params: Record<string, string>
-): ValidationResult {
-  const issues: string[] = [];
-
-  for (const [key, value] of Object.entries(params)) {
-    if (!value || value.trim() === '') {
-      issues.push(`Parameter '${key}' cannot be empty`);
-    }
-  }
-
-  return {
-    isValid: issues.length === 0,
-    issues,
-  };
-}


packages/event-handler/src/rest/RouteHandlerRegistry.ts
```diff
@@ -1,10 +1,24 @@
...
+  validateParams,
+  validatePathPattern,
+} from './utils.js';
 
 class RouteHandlerRegistry {
-  readonly #routes: Map<string, Route> = new Map();
+  readonly #staticRoutes: Map<string, Route> = new Map();
+  readonly #dynamicRoutesIndexMap: Map<string, number> = new Map();
+  readonly #dynamicRoutes: DynamicRoute[] = [];
   readonly #routesByMethod: Map<string, Route[]> = new Map();
```
svozza: I haven't removed it here but I don't think we need this `#routesByMethod` map anymore.

packages/event-handler/src/rest/RouteHandlerRegistry.ts
```diff
@@ -22,29 +86,118 @@ class RouteHandlerRegistry {
...
+    }
+
+    return null;
+  }
+
+  /**
+   * Returns the total number of registered routes.
+   * @returns Total count of registered routes
+   */
   public getRouteCount(): number {
```
svozza: I think we can probably get rid all three of these methods, I don't think we need them.

packages/event-handler/src/rest/utils.ts
```diff
@@ -43,3 +44,38 @@ export function validatePathPattern(path: Path): ValidationResult {
...
   };
 }
+
+type ProcessedParams = {
+  raw: Record<string, string>;
+  processed: Record<string, any>;
+  route: Route;
+};
+
+export function processParams(
```
svozza: As mentioned in the body of the PR, I think both these utilities can probably live as private methods in `RouteHandlerRegistry`. Thoughts?

packages/event-handler/src/rest/RouteHandlerRegistry.ts
```diff
@@ -13,6 +27,56 @@ class RouteHandlerRegistry {
...
+   * Adds a dynamic route to the registry, ensuring that the dynamic
+   * routes array says sorted in ascending order of specificity.
+   * @param route - The dynamic route to add
+   * @returns The index in the specific routes array where the route was stored
+   */
+  #addDynamicRoute(route: DynamicRoute): number {
+    for (let i = 0; i < this.#dynamicRoutes.length; i++) {
+      const existingRoute = this.#dynamicRoutes[i];
+      if (this.#compareRouteSpecificity(route, existingRoute) < 0) {
+        this.#dynamicRoutes.splice(i, 0, route);
```
svozza: Note, the time complexity of `splice` is basically O(n) inside this loop when the item to be inserted is towards the end of the list. It is very unlikely that users will ever register enough routes for this to become a problem.

svozza: Ignore this, I realised we can just use a simple binary search and get O(n) performance.

packages/event-handler/src/rest/utils.ts
```diff
@@ -43,3 +43,32 @@ export function validatePathPattern(path: Path): ValidationResult {
     issues,
   };
 }
+
+export function processParams(
+  params: Record<string, string>
+): Record<string, any> {
+  const processed: Record<string, any> = {};
```
svozza: I took this implementation direct from #4140 but I think this should be `Record<string, string>`. Am I missing something?

packages/event-handler/src/rest/RouteHandlerRegistry.ts
svozza: Third pass at this. Lol. One of the reasons I initially went down this path was because there was a `refresh` method in #4140 that would be called every time a route would be added which sorted the list. That meant that every time you added a route it would taken O(nlogn) time.

I've realised that rather than doing the sort eagerly, we can do it lazily instead by setting a flag to indicate whether the list should be sorted that the `resolve` method checks. As `resolve` will only ever be called after all the routes have been registered, the call to sort will only ever be done once. 

This means we can just use `push` to add routes to the dynamic routes array, which only take O(1) time instead of using `splice` which as mentioned is an O(n) operation. I've tested this and it improves the performance of registering routes by 50%, with no impact on the performance of `resolve`. I used this benchmark:

```ts
import { bench } from 'vitest';
import { HttpVerbs } from '../../src/rest/constants.js';
import { Route } from '../../src/rest/Route.js';
import { RouteHandlerRegistry } from '../../src/rest/RouteHandlerRegistry.js';
import type { Path, HttpMethod } from '../../src/types/rest.js';

// Setup shared data
const handler = () => 'test';
const methods = [HttpVerbs.GET, HttpVerbs.POST, HttpVerbs.PUT, HttpVerbs.DELETE];
const pathTemplates = [
  '/:id',
  '/users/:id',
  '/api/:version/users/:id',
  '/:category/:subcategory/:id',
  '/api/:version/:service/:resource/:id',
  '/complex/:a/:b/:c/:d/:e',
  '/simple/:param',
  '/:x/nested/:y/deep/:z',
  '/service/:service/action/:action/item/:item',
  '/:root/:branch/:leaf/:detail/:extra/:more'
];

// Pre-generate routes for registration benchmark
const preGeneratedRoutes: Route[] = [];
for (let i = 0; i < 300; i++) {
  const method = methods[Math.floor(Math.random() * methods.length)];
  const template = pathTemplates[Math.floor(Math.random() * pathTemplates.length)];
  const path = template.replace(/:([a-z]+)/g, `:$1${i}`);
  preGeneratedRoutes.push(new Route(method, path as Path, handler));
}

// Setup registry with routes for resolution benchmark
const registry = new RouteHandlerRegistry({ logger: console });
const testPaths: string[] = [];
for (let i = 0; i < 300; i++) {
  const method = methods[Math.floor(Math.random() * methods.length)];
  const template = pathTemplates[Math.floor(Math.random() * pathTemplates.length)];
  const path = template.replace(/:([a-z]+)/g, `:$1${i}`);
  registry.register(new Route(method, path as Path, handler));
  
  // Generate corresponding test path for resolution
  const testPath = template.replace(/:([a-z]+)/g, 'value$1');
  testPaths.push(`${method}:${testPath}`);
}

bench('route registration with 300 pre-generated routes', () => {
  const newRegistry = new RouteHandlerRegistry({ logger: console });
  for (const route of preGeneratedRoutes) {
    newRegistry.register(route);
  }
});

bench('route resolution with 300 registered routes', () => {
  for (const pathWithMethod of testPaths) {
    const [method, path] = pathWithMethod.split(':');
    registry.resolve(method as HttpMethod, path as Path);
  }
});
```

Binary search version:
Screenshot 2025-08-10 at 14 13 53

Sort version:
Screenshot 2025-08-10 at 14 13 08

This is a good example where time complexity analysis needs nuance. If you only ever add one route then the performance of the new version `register` will be O(nlogn) vs O(n) for the old one. But the more routes you add, the more the cost of the one off `sort` operation gets amortized and it effectively becomes O(1).

packages/event-handler/src/rest/RouteHandlerRegistry.ts
```diff
@@ -1,19 +1,66 @@
...
+   * Registers a route in the registry after validating its path pattern.
+   *
+   * The function decides whether to store the route in the static registry
+   * (for exact paths like `/users`) or dynamic registry (for parameterized
+   * paths like `/users/:id`) based on the compiled path analysis.
+   *
+   * @param route - The route to register
+   */
   public register(route: Route): void {
+    this.#shouldSort = true;
```
svozza, line 91: This handles the edge case where for some reason someone registers a route after `resolve` has been called.

packages/event-handler/src/rest/RouteHandlerRegistry.ts
dreamorosi: > I've realised that rather than doing the sort eagerly, we can do it lazily instead by setting a flag to indicate whether the list should be sorted that the resolve method checks. As resolve will only ever be called after all the routes have been registered, the call to sort will only ever be done once.

Smart - well done with this analysis.

packages/event-handler/src/rest/RouteHandlerRegistry.ts
dreamorosi: Good to have it, but I agree that this is not a pattern we should encourage

packages/event-handler/src/rest/RouteHandlerRegistry.ts
dreamorosi: Feel free to remove it, it might've been a remnant from a previous pass.

packages/event-handler/src/rest/RouteHandlerRegistry.ts
dreamorosi: Feel free to remove it, can't think of any use for this - and if we need it we can always add it back.

packages/event-handler/src/rest/utils.ts
dreamorosi: No, you're correct I think.

packages/event-handler/src/rest/RouteHandlerRegistry.ts
svozza: Done

packages/event-handler/src/rest/utils.ts

dreamorosi (approved):
Amazing work here, especially with the deep dive in time complexity and optimization.

The code looks good even though I haven't tested it in a function, I think for now we can merge it and progress - either way we'll have a team bug bash when we're near the release.

The only thing potentially missing, which I am ok to tackle in a separate issue is type safety.

The linked issue had an ugly recursive type (which I could swear I saw a question from you about but can't find anymore) that basically aimed at giving the experience described in the [RFC](https://github.com/aws-powertools/powertools-lambda-typescript/discussions/3500) in the section "Dynamic routes" which would allow you to do this in the route handler:

Image

As I said, I'm ok to work on this separately and in the near future - I have a local PoC with this working (in terms of types) that I made over a year ago when I was working on the RFC, so we can potentially take some inspiration there.

svozza:
> The only thing potentially missing, which I am ok to tackle in a separate issue is type safety.

> The linked issue had an ugly recursive type (which I could swear I saw a question from you about but can't find anymore) that basically aimed at giving the experience described in the https://github.com/aws-powertools/powertools-lambda-typescript/discussions/3500 in the section "Dynamic routes" which would allow you to do this in the route handler:

Yeah, I mentioned that recursive type in the PR body at the top, I wasn't sure what to do with it. That autocomplete use case is so nice, defo something we want! But agreed we can tackle it in a separate PR, this one is big enough as it is.

packages/event-handler/src/rest/utils.ts
svozza: Before I merge, do you have any strong feelings on this @dreamorosi? If not I will move it to be a private method.

dreamorosi: I don't at this stage - but my rule of thumb recommendation would be to start everything as `#private`, move to `protected` if we think customers might want to extend it (or they ask us), then leave `public` only things we are 100% intending them to use.

svozza: Yeah, that was thinking too.

dreamorosi (approved):

Issue #4141, opened 2025-07-06
Feature request: Implement Error Handling System for Event Handler


dreamorosi:
### Use case

As part of the Event Handler implementation (#3251), we need a comprehensive error handling system that provides:

1. **Centralized exception management** for consistent error responses across all routes
2. **Pre-built HTTP error classes** for common scenarios (400, 401, 404, 500, etc.)
3. **Custom error handler registration** allowing developers to handle specific error types
4. **Automatic error response formatting** that converts exceptions to proper HTTP responses
5. **Development-friendly error reporting** with detailed information in debug mode
6. **Integration with route resolution** to catch and handle errors during request processing

This system ensures that errors are handled consistently, securely, and provide good developer experience while maintaining production-ready error responses.

### Solution/User Experience

> [!note]
> The code snippets below are provided as reference only - they are not exhaustive and final implementation might vary.

```typescript
// Base error class for all HTTP errors
abstract class ServiceError extends Error {
  abstract readonly statusCode: number;
  abstract readonly errorType: string;
  public readonly details?: Record<string, unknown>;
  
  constructor(
    message: string,
    options?: ErrorOptions,
    details?: Record<string, unknown>
  ) {
    super(message);
    this.name = 'ServiceError';
  }
  
  toJSON(): ErrorResponse {
    return {
      statusCode: this.statusCode,
      error: this.errorType,
      message: this.message,
      ...(this.details && { details: this.details })
    };
  }
}
```

Other error classes to be implemented, which extend `ServiceError` above:
- BadRequestError
- UnauthorizedError
- ForbiddenError
- NotFoundError
- MethodNotAllowedError
- InternalServerError
- ServiceUnavailableError

When implementing, also double check in the Powertools for AWS Lambda (Python) repo if more error classes are present.

### Exception Handler Manager

```typescript
// Centralized exception handling
class ExceptionHandlerManager {
  private handlers: Map<ErrorConstructor, ErrorHandler> = new Map();
  private defaultHandler?: ErrorHandler;
  
  // Register custom error handler for specific error types
  register<T extends Error>(
    errorType: ErrorConstructor<T> | ErrorConstructor<T>[],
    handler: ErrorHandler<T>
  ): void {
    const errorTypes = Array.isArray(errorType) ? errorType : [errorType];
    
    for (const type of errorTypes) {
      this.handlers.set(type, handler as ErrorHandler);
    }
  }
  
  // Set default handler for unhandled errors
[...]

PR #4307, opened 2025-08-11
feat(event-handler): add event handler registry

svozza:
## Summary
This PR creates the error handler registry class. It follows the same pattern as the route registry by exposing `register` and  `resolve` methods. A very rudimentary integration with the `BaseRouter` class has also been done, the next PR will fully implement that integration.

### Changes

- Created the `ErrorHandlerRegistry` class
- **Added `errorHandler()` method to BaseRouter**: Allows registration of error handlers for specific error types using the `ErrorHandlerRegistry`
- **Integrated ErrorHandlerRegistry**: `BaseRouter` now instantiates and manages an `ErrorHandlerRegistry`
- **Added unit tests**: Created tests for error handler registry and the base router integration

**Issue number:** #4141 

---


packages/event-handler/src/rest/BaseRouter.ts
```diff
@@ -32,7 +36,7 @@ abstract class BaseRouter {
    */
   protected readonly isDev: boolean = false;
 
-  public constructor(options?: RouterOptions) {
+  protected constructor(options?: RouterOptions) {
```
dreamorosi, line 39: Shouldn't the constructor be `public` so the class can be instantiated? I've never seen it declared as `protected` because essentially, even when extending you can't overload it but should instead use `super()` in the child class.

packages/event-handler/src/rest/BaseRouter.ts
```diff
@@ -44,9 +48,19 @@ abstract class BaseRouter {
       warn: console.warn,
     };
     this.routeRegistry = new RouteHandlerRegistry({ logger: this.logger });
+    this.errorHandlerRegistry = new ErrorHandlerRegistry({
+      logger: this.logger,
+    });
     this.isDev = isDevMode();
   }
 
+  errorHandler<T extends Error>(
```
dreamorosi: Should this be `public`?

packages/event-handler/src/rest/ErrorHandlerRegistry.ts
```diff
@@ -0,0 +1,75 @@
...
+  ErrorHandlerRegistryOptions,
+  RouteRegistryOptions,
+} from '../types/rest.js';
+
+export class ErrorHandlerRegistry {
+  readonly #handlers: Map<ErrorConstructor, ErrorHandler> = new Map();
+
+  readonly #logger: Pick<GenericLogger, 'debug' | 'warn' | 'error'>;
+
+  constructor(options: ErrorHandlerRegistryOptions) {
```
dreamorosi: should this be `public`? I think let's try to be explicit as much as possible - anyway these are type definitions only, so there's no perf impact in production.

packages/event-handler/src/rest/BaseRouter.ts
svozza: Yeah, this was a suggestion from my IDE that I just accepted without thinking too much. I will revert. 

Edit: this is an abstract class so it can't be instantiated, which I think is why this is OK.

packages/event-handler/src/rest/ErrorHandlerRegistry.ts
svozza: Yeah, we are explicitly using `public` everywhere else so we should be consistent.

packages/event-handler/src/rest/BaseRouter.ts
svozza: *fixed

packages/event-handler/src/rest/BaseRouter.ts
svozza: This is the rule from my IDE: https://www.jetbrains.com/help/inspectopedia/TypeScriptAbstractClassConstructorCanBeMadeProtected.html

dreamorosi: I see, I didn't know about this specific nuance in abstract classes

svozza: Neither did I!

dreamorosi:
For some reason Sonar is failing but there were findings (https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&sinceLeakPeriod=true&pullRequest=4307&id=aws-powertools_powertools-lambda-typescript) usually making another commit fixes it, I think we can address these items and it'll work.

svozza:
> For some reason Sonar is failing but there were findings (https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&sinceLeakPeriod=true&pullRequest=4307&id=aws-powertools_powertools-lambda-typescript) usually making another commit fixes it, I think we can address these items and it'll work.

Yeah, that fixed it!

dreamorosi (approved):

PR #4312, opened 2025-08-12
fix(tracer): pass args of decorated method as they are

### LINKED ISSUE #4306: Bug: Warning callback-based function handlers caused by @tracer.captureLambdaHandler()
### Expected Behavior

I would expect to be able to add `@tracer.captureLambdaHandler()` to my async lambda handler and not recieve the following warning

>  AWS Lambda plans to remove support for callback-based function handlers starting with Node.js 24. You will need to update this function to use an async handler to use Node.js 24 or later. For more information and to provide feedback on this change, see https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/issues/137. To disable this warning, set the AWS_LAMBDA_NODEJS_DISABLE_CALLBACK_WARNING environment variable.

### Current Behavior

We recently started getting:

>  AWS Lambda plans to remove support for callback-based function handlers starting with Node.js 24. You will need to update this function to use an async handler to use Node.js 24 or later. For more information and to provide feedback on this change, see https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/issues/137. To disable this warning, set the AWS_LAMBDA_NODEJS_DISABLE_CALLBACK_WARNING environment variable.

We are not using callback based lambda handler definitions and after removing the `@tracer.captureLambdaHandler()` from our handler function the warning goes away. 

### Code snippet

```
  @tracer.captureLambdaHandler()
  @logger.injectLambdaContext()
  public async handler(event: any, context: Context): Promise<any> {
    await this.preRun(event, context);
    return await this.main(event, context);
  }



export const handler = new Handler();
export const main = handler.handler.bind(handler);
```

### Steps to Reproduce

1. Create a simple async handler
2. Deploy and run and see no warning in logs
3. Add `@tracer.captureLambdaHandler()` to handler() definition
4. deploy and run and see the warning

### Possible Solution

```
public captureLambdaHandler(
  options?: CaptureLambdaHandlerOptions
): HandlerMethodDecorator {
  return (_target, _propertyKey, descriptor) => {
    const originalMethod = descriptor.value!;
    const tracerRef = this;
    
    // Check if the original method expects a callback parameter
    const originalMethodLength = originalMethod.length;
    const expectsCallback = originalMethodLength >= 3;
    
    // Create the wrapper function with the appropriate signature
    if (expectsCallback) {
      // Original method expects callback - create wrapper with callback parameter
      descriptor.value = function (this: Handler, event, context, callback) {
        // ... existing logic with callback parameter
      } as SyncHandler<Handler> | AsyncHandler<Handler>;
    } else {
      // Original method doesn't expect callback - create wrapper WITHOUT callback parameter
      descriptor.value = function (this: Handler, event, context) {
        // ... existing logic without callback parameter
      } as SyncHandler<Handler> | AsyncHandler<Handler>;
    }

    return descriptor;
  };
}
```
ref https://github.com/aws-powertools/powertools-lambda-typescript/blob/081a514b6e1cb81513080352e5dbcadc89ac0574/packages/tracer/src/Tracer.ts#L422-L473

latest

22.x

### Packaging format used

npm

### Execution logs

```Shell
INIT_START Runtime Version: nodejs:20.v70	Runtime Version ARN: arn:aws:lambda:us-east-1::runtime:c11827d1307a63ca6bbb996dbffba66628e1328ed4714070f4b891884a323e0b
2025-08-11T15:27:00.879Z	undefined	WARN	AWS Lambda plans to remove support for callback-based function handlers starting with Node.js 24. You will need to update this function to use an async handler to use Node.js 24 or later. For more information and to provide feedback on this change, see https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/issues/137. To disable this warning, set the AWS_LAMBDA_NODEJS_DISABLE_CALLBACK_WARNING environment variable.
```

dreamorosi:
### Changes

This PR updates the internal implementation of the `captureLambdaHandler` class method decorator in Tracer so that it forwards the `args` of the decorated handler method as they appear at runtime rather than destructuring them and forcing the third parameter (`callback`) to be `undefined`.

Its presence, even when not defined, caused the runtime to interpret the handler as using callback style, which in turn caused deprecation warnings to show up.

**Issue number:** closes #4306

---

### DIFF
--- a/packages/tracer/src/Tracer.ts
+++ b/packages/tracer/src/Tracer.ts
@@ -428,9 +428,12 @@ class Tracer extends Utility implements TracerInterface {
       const tracerRef = this;
       // Use a function() {} instead of an () => {} arrow function so that we can
       // access `myClass` as `this` in a decorated `myClass.myMethod()`.
-      descriptor.value = function (this: Handler, event, context, callback) {
+      descriptor.value = function (
+        this: Handler,
+        ...args: Parameters<Handler>
+      ) {
         if (!tracerRef.isTracingEnabled()) {
-          return originalMethod.apply(this, [event, context, callback]);
+          return originalMethod.apply(this, args);
         }
 
         return tracerRef.provider.captureAsyncFunc(
@@ -440,11 +443,7 @@ class Tracer extends Utility implements TracerInterface {
             tracerRef.addServiceNameAnnotation();
             let result: unknown;
             try {
-              result = await originalMethod.apply(this, [
-                event,
-                context,
-                callback,
-              ]);
+              result = await originalMethod.apply(this, args);
               if (options?.captureResponse ?? true) {
                 tracerRef.addResponseAsMetadata(result, process.env._HANDLER);
               }


dreamorosi:
@sdangol can you please review the implementation of all the other decorators in the codebase and see if they need the same change? If needed, you can use the [triaging runbook](https://github.com/dreamorosi/powertools-triaging-template) and deploy a functions to verify whether decorating causes a warning:

```ts
import type { Callback } from 'aws-lambda';

class Handler {
  // insert decorator here - should not cause a warning
  handler(event: unknown, context: unknown) {
    return {
      statusCode: 200,
      body: JSON.stringify({ message: 'Hello from Lambda!' }),
    };
  }
}

const lambda = new Handler();

```

If you find any, can you open a group issue to apply the same fix and link it back in #4312?

PR #4316, opened 2025-08-13
feat(event-handler): add error handling functionality to BaseRouter

svozza:
## Summary
This PR adds error handling logic to the `BaaseRouter` class

### Changes

`BaseRouter`
- Added specific methods to register `NotFound` and `MethodNotAllowed` handlers
- Added error handler methods, these take an error and check the registry to see if there is a registered handler. If not it has a fallback to a default error handler that will return a 500 error
- Unit tests that test the error registry logic is handled correctly

**Issue number:** #4141 

Final part of the error handling piece is to add support for decorators. I have left that till last because I wanted the logic to be easily reviewable before we added the method overloads and bolierplate to support them.

---

### DIFF
--- a/.github/workflows/layers_partitions.yml
+++ b/.github/workflows/layers_partitions.yml
@@ -182,9 +182,11 @@ jobs:
           REMOTE_DESCRIPTION=$(jq -r '.Description' $layer_output)
           LOCAL_DESCRIPTION=$(jq -r '.Description' AWSLambdaPowertoolsTypeScriptV2.json)
           test "$REMOTE_DESCRIPTION" == "$LOCAL_DESCRIPTION" && echo "Version number OK: ${LOCAL_DESCRIPTION}" || exit 1
-          REMOTE_LAYER_VERSION=$(jq -r '.LayerVersionArn' $layer_output | sed 's/.*://')
-          LOCAL_LAYER_VERSION=$(jq -r '.LayerVersionArn' AWSLambdaPowertoolsTypeScriptV2.json | sed 's/.*://')
-          test "$REMOTE_LAYER_VERSION" == "$LOCAL_LAYER_VERSION" && echo "Layer Version number OK: ${LOCAL_LAYER_VERSION}" || exit 1
+          if [ "${{ inputs.environment }}" == "Prod" ]; then
+            REMOTE_LAYER_VERSION=$(jq -r '.LayerVersionArn' $layer_output | sed 's/.*://')
+            LOCAL_LAYER_VERSION=$(jq -r '.LayerVersionArn' AWSLambdaPowertoolsTypeScriptV2.json | sed 's/.*://')
+            test "$REMOTE_LAYER_VERSION" == "$LOCAL_LAYER_VERSION" && echo "Layer Version number OK: ${LOCAL_LAYER_VERSION}" || exit 1
+          fi
           jq -s -r '["Layer Arn", "Runtimes", "Version", "Description", "SHA256"], ([.[0], .[1]] | .[] | [.LayerArn, (.CompatibleRuntimes | join("/")), .Version, .Description, .Content.CodeSha256]) |@tsv' AWSLambdaPowertoolsTypeScriptV2.json $layer_output | column -t -s $'\t'
 
       - name: Store Metadata - ${{ matrix.region }}


packages/event-handler/src/rest/BaseRouter.ts
```diff
@@ -76,6 +105,68 @@ abstract class BaseRouter {
...
+  /**
+   * Handles errors by finding a registered error handler or falling
+   * back to a default handler.
+   *
+   * @param error - The error to handle
+   * @returns A Response object with appropriate status code and error details
+   */
+  protected handleError(error: Error): Response {
+    const handler = this.errorHandlerRegistry.resolve(error);
+    if (handler != null) {
```
dreamorosi: Why not use strict checks here? I.e. `!==`

packages/event-handler/src/rest/BaseRouter.ts
```diff
@@ -76,6 +105,68 @@ abstract class BaseRouter {
...
+      try {
+        const body = handler(error);
```
dreamorosi: Should we `await` this?

packages/event-handler/src/rest/BaseRouter.ts
```diff
@@ -76,6 +105,68 @@ abstract class BaseRouter {
...
+
+  /**
+   * Default error handler that returns a 500 Internal Server Error response.
+   * In development mode, includes stack trace and error details.
+   *
+   * @param error - The error to handle
+   * @returns A Response object with 500 status and error details
+   */
+  #defaultErrorHandler(error: Error): Response {
+    const isProduction =
```
dreamorosi: I don't think we should add this, we can just assume that if it's not dev mode, then it's prod. I don't know how many people actually set this

packages/event-handler/src/rest/errors.ts
```diff
@@ -19,12 +19,12 @@ export class ParameterValidationError extends RouteMatchingError {
...
 }
 
-abstract class ServiceError extends Error {
+export abstract class ServiceError extends Error {
   abstract readonly statusCode: HttpStatusCode;
   abstract readonly errorType: string;
   public readonly details?: Record<string, unknown>;
 
-  constructor(
+  protected constructor(
```
dreamorosi, line 27: This might need to be `public`

dreamorosi (commented):
Thanks for this PR - left a few comments

packages/event-handler/src/rest/BaseRouter.ts
svozza: It's quite common to set this environment variable but yeah, we can remove it. I was just following what was in the issue.

packages/event-handler/src/rest/errors.ts
svozza: This is like that other one, `SeriveError` is an abstract class so it can't be instantiated.

packages/event-handler/src/rest/BaseRouter.ts
svozza: Yeah, I wasn't sure whether to make the error handlers async. I guess there's definitely cases where user might want to, e.g., do something that involves a network call in their error handlers so we probably should.

packages/event-handler/src/rest/BaseRouter.ts
svozza: I'm just being overly defensive here as this will catch  the case of `undefined` too. Will change it.

packages/event-handler/src/rest/BaseRouter.ts
dreamorosi: I usually make everything async, so even if they set a sync handler it becomes a no-op with minimal overhead.

The opposite would cause a serialization error due to it being a promise

packages/event-handler/src/rest/BaseRouter.ts
dreamorosi: Sorry, I will pay more attention to issues in the future. I've noticed a few of these instances where I missed details.

packages/event-handler/src/rest/BaseRouter.ts
svozza: Done

packages/event-handler/src/rest/BaseRouter.ts
svozza: Don't worry, impossible to make such detailed issues without occasionally missing something. I much prefer not checking this variable because it makes it far harder for the handler to accidentally send unwanted details back in a response.

packages/event-handler/src/rest/BaseRouter.ts
svozza: Yeah, makes sense!

PR #4317, opened 2025-08-13
feat(event-handler): add support for error handling in AppSync GraphQL

### LINKED ISSUE #4130: Feature request: support error handling in AppSync GraphQL
### Use case

The AppSync GraphQL resolver in Event Handler should allow customers to specify handlers for specific error classes which are called whenever a resolver throws that error.

### Solution/User Experience

The new feature should add a new method to the `AppSyncGraphQLResolver` class called `exceptionHandler`.

This method accepts an error class as first parameter (i.e. `AssertionError`) and a function handler as second:

```ts

const app = new AppSyncGraphQLResolver({ logger });

app.exceptionHandler(AssertionError, async (error) => {
  return {
    message: 'Error foo occurred'
  }
});

app.onQuery('getTodo', async () => {
  throw new AssertionError('foo')

  app.resolve(event, context);
```

Whenever an error in any of the other route handler is thrown, we should intercept it (around [here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/fc84347fd5fa01fbeb4b4d528da7d9e1e79edb69/packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts#L117)) and check if there's any exception handler registered (we'll need a registry) that matches the error class being thrown.

If yes, we should call the exception handler and pass the error instance as first argument, and return its response as-is; if not, we should carry on the processing as it's implemented today. 

The `exceptionHandler()` method should also be available as TypeScript class method decorator, similar to how `onQuery()` can.

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

arnabrahman:
## Summary
This PR adds support for error handling in AppSync GraphQL described in the issue
### Changes

- `exceptionHandler`  method is introduced to register an exception and its handler.
- `ExceptionHandlerRegistry` is introduced to keep track of registered exception handlers
- `exceptionHandler` is also exposed as a decorator method
- Unit tests are written for 100% coverage
- Added documentation for exception handling with examples
- Minor housekeeping(ex: dockerfile nodejs version update due to error & `endregion` added)

Lambda handler:
```typescript
import { AssertionError } from 'assert';

const logger = new Logger({
  serviceName: 'MyService',
});
const app = new AppSyncGraphQLResolver({ logger });

app.exceptionHandler(AssertionError, async (error) => {
  return {
    error: {
      message: error.message,
      type: error.name,
    },
  };
});

app.onQuery('createSomething', async () => {
  throw new AssertionError({
    message: 'This is an assertion Error',
  });
});

  app.resolve(event, context);
```
APPSYNC JS resolver

```javascript
export function request(ctx) {
  return {
    operation: 'Invoke',
    payload: ctx,
  };
}

export function response(ctx) {
  if (ctx.result.error) {
    return util.error(ctx.result.error.message, ctx.result.error.type);
  }
  return ctx.result;
}
```
Response mapping template
```velocity
#if (!$util.isNull($ctx.result.error))
  $util.error($ctx.result.error.message, $ctx.result.error.type)
#end

$utils.toJson($ctx.result)
```

Response:
```json
{
  "data": {
    "createSomething": null
  },
  "errors": [
    {
      "path": [
        "createSomething"
      ],
      "data": null,
      "errorType": "AssertionError",
      "errorInfo": null,
      "locations": [
        {
          "line": 2,
          "column": 3,
          "sourceName": null
        }
      ],
      "message": "This is an assertion Error"
    }
  ]
}
```

**Issue number:** closes #4130 

---


arnabrahman:
I have one observation: while looking into the Python [documentation](https://docs.powertools.aws.dev/lambda/python/latest/core/event_handler/appsync/#exception-handling), I realized they also accept list of errors for exception handling. Should we support that as well?

svozza:
> I have one observation: while looking into the Python [documentation](https://docs.powertools.aws.dev/lambda/python/latest/core/event_handler/appsync/#exception-handling), I realized they also accept list of errors for exception handling. Should we support that as well?

Yes, that's what I'm doing in the API Gateway event handler so we should be consistent: https://github.com/aws-powertools/powertools-lambda-typescript/blob/92c7f360ba255a8289cab7e870b7275f669bf9b6/packages/event-handler/src/rest/BaseRouter.ts#L57

docs/Dockerfile
```diff
@@ -2,7 +2,7 @@
 FROM squidfunk/mkdocs-material@sha256:bb7b015690d9fb5ef0dbc98ca3520f153aa43129fb96aec5ca54c9154dc3b729
 
 # Install Node.js
-RUN apk add --no-cache nodejs=22.13.1-r0 npm
```
svozza: Is this intentional?

docs/features/event-handler/appsync-graphql.md
```diff
@@ -148,6 +148,30 @@ You can access the original Lambda event or context for additional information.
 
     1. The `event` parameter contains the original AppSync event and has type `AppSyncResolverEvent` from the `@types/aws-lambda`.
 
+### Exception Handling
+
+You can use the **`exceptionHandler`** method with any exception. This allows you to handle common errors outside your resolver and return a custom response.
+
+You can use a VTL response mapping template to detect these custom responses and forward them to the client gracefully.
```
svozza: The recommended way to do response mapping is JS now, not VTL, so we should include an example of how to do it that way.

packages/event-handler/src/appsync-graphql/ExceptionHandlerRegistry.ts
```diff
@@ -0,0 +1,80 @@
...
+    this.#logger.debug(`Looking for exception handler for error: ${errorName}`);
+
+    const handlerOptions = this.handlers.get(errorName);
+    if (handlerOptions) {
+      this.#logger.debug(`Found exact match for error class: ${errorName}`);
+      return handlerOptions.handler;
+    }
+
+    this.#logger.debug(`No exception handler found for error: ${errorName}`);
+    return undefined;
```
svozza: We return `null` in the API Gateway handler so let's do the same here.

packages/event-handler/tests/unit/appsync-graphql/AppSyncGraphQLResolver.test.ts
```diff
@@ -706,4 +727,348 @@ describe('Class: AppSyncGraphQLResolver', () => {
...
+    const validationError = await handler(
+      onGraphqlEventFactory('getUser', 'Query', {}),
+      context
+    );
+    const notFoundError = await handler(
+      onGraphqlEventFactory('getUser', 'Query', { id: '0', name: 'John Doe' }),
+      context
+    );
+
+    // Assess
```
svozza, line 1062: This test should have have an assertion somewhere to ensure that the `this` context is being preserved.

packages/event-handler/src/appsync-graphql/ExceptionHandlerRegistry.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  /**
+   * Checks if there are any registered exception handlers.
+   */
+  public hasHandlers(): boolean {
```
svozza: Do we need this function? I think it's sufficient for the client to call `resolve` and check whether it returns `null` or not.

docs/Dockerfile
arnabrahman: Yes, as other version is failing in my local

packages/event-handler/src/appsync-graphql/ExceptionHandlerRegistry.ts
arnabrahman: Yeah, that can also be done too

docs/Dockerfile
dreamorosi: I remember you modified this in a previous PR and had to revert it back - I am unsure what's happening here, but I'd say let's not include it as part of this PR. 

If you need this to get the docs running locally for now keep the change in your local branch if possible.

arnabrahman: Apologies for the trouble caused. I wasn’t aware of this. Reverted.

leandrodamascena:
Hi @arnabrahman, thanks for responding to all the comments and making the necessary changes! We'll be back with this PR and review next week.

docs/features/event-handler/appsync-graphql.md
```diff
@@ -148,6 +148,38 @@ You can access the original Lambda event or context for additional information.
 
     1. The `event` parameter contains the original AppSync event and has type `AppSyncResolverEvent` from the `@types/aws-lambda`.
 
+### Exception Handling
+
+You can use the **`exceptionHandler`** method to handle any exception. This allows you to handle common errors outside your resolver and return a custom response.
```
dreamorosi:
```suggestion
You can use the `exceptionHandler` method to handle any exception. This allows you to handle common errors outside your resolver and return a custom response.
```

docs/features/event-handler/appsync-graphql.md
```diff
@@ -148,6 +148,38 @@ You can access the original Lambda event or context for additional information.
...
+
+The **`exceptionHandler`** method also supports passing an array of exceptions that you wish to handle with a single handler.
```
dreamorosi:
```suggestion
The `exceptionHandler` method also supports passing an array of exceptions that you wish to handle with a single handler.
```

examples/snippets/event-handler/appsync-graphql/exceptionHandling.ts
```diff
@@ -0,0 +1,27 @@
+import { AppSyncGraphQLResolver } from '@aws-lambda-powertools/event-handler/appsync-graphql';
+import { Logger } from '@aws-lambda-powertools/logger';
+import { AssertionError } from 'assert';
```
dreamorosi:
```suggestion
import { AssertionError } from 'node:assert';
```

examples/snippets/event-handler/appsync-graphql/exceptionHandling.ts
```diff
@@ -0,0 +1,27 @@
...
+    error: {
+      message: error.message,
+      type: error.name,
+    },
+  };
+});
+
+app.onQuery('createSomething', async () => {
+  throw new AssertionError({
+    message: 'This is an assertion Error',
```
dreamorosi:
```suggestion
    message: 'This is an assertion error',
```

packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
```diff
@@ -189,17 +191,20 @@ class AppSyncGraphQLResolver extends Router {
...
     fn: () => Promise<unknown>,
-    event: AppSyncResolverEvent<Record<string, unknown>>
+    event: AppSyncResolverEvent<Record<string, unknown>>,
+    options?: ResolveOptions
   ): Promise<unknown> {
     try {
       return await fn();
     } catch (error) {
-      return this.#handleError(
+      return await this.#handleError(
```
dreamorosi:
```suggestion
      return this.#handleError(
```

dreamorosi: We don't need to `await` it, the Lambda runtime will already `await` the function's response at the end before returning to AppSync.

packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
```diff
@@ -209,16 +214,39 @@ class AppSyncGraphQLResolver extends Router {
...
+          this.logger.debug(
+            `Calling exception handler for error: ${error.name}`
+          );
+          return await exceptionHandler.apply(options?.scope ?? this, [error]);
+        } catch (handlerError) {
+          this.logger.error(
+            `Exception handler for ${error.name} threw an error`,
+            handlerError
+          );
+        }
```
dreamorosi, line 246: Should we throw a generic default error here? If we enter this `catch` block we're returning `undefined` and I'm not sure if this will make the error opaque to the caller of the operation.

packages/event-handler/src/types/appsync-graphql.ts
dreamorosi, line 1: No need to change it if you don't want to, but the `// #endregion` comments are not needed, VSCode should be able to infer it based on the next `// #region ...` starting.

packages/event-handler/src/types/appsync-graphql.ts
```diff
@@ -178,6 +183,51 @@ type GraphQlBatchRouteOptions<
...
+type ExceptionSyncHandlerFn<T extends Error> = (error: T) => unknown;
+
+type ExceptionHandlerFn<T extends Error> = (error: T) => Promise<unknown>;
+
+type ExceptionHandler<T extends Error = Error> =
+  | ExceptionSyncHandlerFn<T>
+  | ExceptionHandlerFn<T>;
+
+// biome-ignore lint/suspicious/noExplicitAny: this is a generic type that is intentionally open
+type ErrorClass<T extends Error> = new (...args: any[]) => T;
```
dreamorosi, line 200: is there any way that we can convert that `any[]` to `unknown[]`?

packages/event-handler/src/types/appsync-graphql.ts
```diff
@@ -178,6 +183,51 @@ type GraphQlBatchRouteOptions<
...
+   */
+  error: ErrorClass<T> | ErrorClass<T>[];
+  /**
+   * The handler function to be called when the error is caught
+   */
+  handler: ExceptionHandler<T>;
+};
+
+/**
+ * Options for the {@link ExceptionHandlerRegistry | `ExceptionHandlerRegistry`} class
```
dreamorosi, line 219: I think we need to add `import type { ExceptionHandlerRegistry } ...` at the top for this to work.

packages/event-handler/tests/unit/appsync-graphql/AppSyncGraphQLResolver.test.ts
```diff
@@ -1,13 +1,34 @@
 import context from '@aws-lambda-powertools/testing-utils/context';
+import { AssertionError } from 'assert';
```
dreamorosi:
```suggestion
import { AssertionError } from 'node:assert';
```

packages/event-handler/tests/unit/appsync-graphql/AppSyncGraphQLResolver.test.ts
```diff
@@ -706,4 +727,615 @@ describe('Class: AppSyncGraphQLResolver', () => {
...
+    {
+      errorClass: URIError,
+      message: 'URI failed',
+    },
+    {
+      errorClass: AggregateError,
+      message: 'Aggregation failed',
+    },
+  ])(
+    'should invoke exception handler for %s',
```
dreamorosi: Can we change the test headers to use imperative voice, i.e. "[it] invokes ..."

packages/event-handler/tests/unit/appsync-graphql/AppSyncGraphQLResolver.test.ts
```diff
@@ -706,4 +727,615 @@ describe('Class: AppSyncGraphQLResolver', () => {
...
+      errorClass,
+      message,
+    }: {
+      errorClass: ErrorClass<Error>;
+      message: string;
+    }) => {
+      // Prepare
+      const app = new AppSyncGraphQLResolver();
+
+      app.exceptionHandler(errorClass, async (err) => {
```
dreamorosi:
```suggestion
      app.exceptionHandler(errorClass, async (err) => ({
          message,
          errorName: err.constructor.name,
      });
```

packages/event-handler/src/appsync-graphql/ExceptionHandlerRegistry.ts
```diff
@@ -0,0 +1,93 @@
...
+   * @param options - The options containing the error class(es) and their associated handler.
+   * @param options.error - A single error class or an array of error classes to handle.
+   * @param options.handler - The exception handler function that will be invoked when the error occurs.
+   */
+  public register(options: ExceptionHandlerOptions<Error>): void {
+    const { error, handler } = options;
+    const errors = Array.isArray(error) ? error : [error];
+
+    for (const err of errors) {
+      this.registerErrorHandler(err, handler as ExceptionHandler);
```
dreamorosi: Is this `as ExceptionHandler` necessary? Since we're destructuring `ExceptionHandlerOptions`, don't we get the correct type from there?

packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
arnabrahman: Your concern is valid about error opacity, but in this case the method doesn't actually return `undefined`. 

When the exception handler fails and we enter the catch block, we log the handler error but then continue execution - the method falls through to `return this.#formatErrorResponse(error);` So callers always receive a formatted error response.  The exception handler failure gets logged for debugging, but we ensure consistent error handling by falling back to the standard format.

Let me know if you think there's a better approach here!

packages/event-handler/src/types/appsync-graphql.ts
arnabrahman: Yes, you’re right, this is not needed for region detection. However, with this, you get the ability to `expand/collapse` regions (at least that’s how it works in my VS Code), which is quite handy. 

So, I’m in favor of keeping the `endregion` comments.

screenshot

packages/event-handler/src/appsync-graphql/ExceptionHandlerRegistry.ts
arnabrahman: You are right

packages/event-handler/src/types/appsync-graphql.ts
arnabrahman: Initially, when implementing this, I considered using `unknown[]` instead of `any[]`, but it creates TypeScript compatibility issues with error constructors that have specific parameter types. The implementation is also kinda  [similar](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/event-handler/src/types/rest.ts#L31) to what we have for rest-api.

The problem is that `unknown[]` makes TypeScript more strict about constructor parameters. For [example](https://github.com/aws-powertools/powertools-lambda-typescript/pull/4317/files#diff-2a1a7e8e6950a9164d4efee63a534f39fe440028d5ba3328f8e36c38d928be5eR803-R809), `ValidationError` expects a string as its first parameter, but with `unknown[]`, TypeScript can't verify that the constructor signature is compatible:

`Type 'new (message: string, options?: ErrorOptions | undefined) => ValidationError' is not assignable to type 'new (...args: unknown[]) => ValidationError'.`

packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
dreamorosi: You're right - my bad, the lines in the diff were collapsed and I didn't see the line below. I agree!

packages/event-handler/src/types/appsync-graphql.ts
dreamorosi: Didn't know that - agreed!

svozza:
Fantastic work @arnabrahman. Another fine addition!

PR #4319, opened 2025-08-13
improv(logger): resolve issue when no logger region is set

### LINKED ISSUE #4318: Maintenance: AWS_REGION environment variable required after upgrading to @aws-lambda-powertools/logger v2.25.0
### Summary

After upgrading `@aws-lambda-powertools/logger` from version 2.24.1 to 2.25.0, my local tests fail with the following error:

```Environment variable AWS_REGION is required```

Previously, I could run tests locally without setting AWS_REGION, since all AWS calls are mocked and my tests only verify log output. Requiring AWS_REGION for local/unit tests feels unnecessary and makes testing more cumbersome.

Is this an intentional change? If so, could there be a way to relax this requirement for local/test environments?

Thanks for your help!

### Why is this needed?

This will make local & unit testing easier and fixes an issue that was created between version `2.24.1` and `2.25.0`. It enhances backwards compatibility.

_No response_

### Solution

I believe the fix could be to add a `defaultValue: ''` for AWS_REGION when reading the environment variable, similar to how other environment variables are handled.

### Acknowledgment

### Future readers

### PR DESCRIPTION
**Pull request by MaxOpperman**
### Changes

- Add an empty default logger region for local environments.

_Edit:_ I tested this locally with my project that raised issue https://github.com/aws-powertools/powertools-lambda-typescript/issues/4318 by building the powertools-lambda package and adding it to my `node_modules`. This fix resolved the issue.

> Please add the issue number below, if no issue is present the PR might get blocked and not be reviewed.

**Issue number:** closes #4318 

---

### DIFF
--- a/packages/logger/tests/unit/setPowertoolsLogData.test.ts
+++ b/packages/logger/tests/unit/setPowertoolsLogData.test.ts
@@ -1,38 +1,44 @@
 import { Logger } from '../../src/Logger.js';
-import { describe, it, expect, vi } from 'vitest';
+import { describe, it, expect, vi, afterEach } from 'vitest';
 
-describe('Logger PowertoolsLogData AWS_REGION', () => {
+describe('Logger PowertoolsLogData environment variables', () => {
+  afterEach(() => {
+    vi.unstubAllEnvs();
+  });
+
   it.each([
     [undefined, ''],
     ['us-west-2', 'us-west-2'],
-  ])('should set awsRegion to %s when AWS_REGION env var is %s', (envValue, expected) => {
+  ])('sets awsRegion to %s when AWS_REGION env var is %s', (envValue, expected) => {
+    // Prepare
     vi.stubEnv('AWS_REGION', envValue);
+    // Act
     const logger = new Logger();
+    // Assess
     expect(logger['powertoolsLogData'].awsRegion).toBe(expected);
-    vi.unstubAllEnvs();
   });
-});
 
-describe('Logger PowertoolsLogData ENVIRONMENT', () => {
   it.each([
     [undefined, ''],
     ['prd', 'prd'],
-  ])('should set environment to %s when ENVIRONMENT env var is %s', (envValue, expected) => {
+  ])('sets environment to %s when ENVIRONMENT env var is %s', (envValue, expected) => {
+    // Prepare
     vi.stubEnv('ENVIRONMENT', envValue);
+    // Act
     const logger = new Logger();
+    // Assess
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/logger/src/Logger.ts [hunk 1]
```diff
@@ -1362,6 +1362,7 @@ class Logger extends Utility implements LoggerInterface {
     this.addToPowertoolsLogData({
       awsRegion: getStringFromEnv({
         key: 'AWS_REGION',
+        defaultValue: '',
```
svozza, line 1365: Can we add a unit test for this? You can use the in built-in Vitest [env var functions](https://vitest.dev/api/vi.html#vi-stubenv) to ensure the environment variable isn't set in your test:

```ts
// Prepare
vi.stubEnv('AWS_REGION', undefined)

// rest of test...
```

packages/logger/tests/unit/setPowertoolsLogData.test.ts
```diff
@@ -0,0 +1,38 @@
...
+
+describe('Logger PowertoolsLogData AWS_REGION', () => {
+  it.each([
+    [undefined, ''],
+    ['us-west-2', 'us-west-2'],
+  ])('should set awsRegion to %s when AWS_REGION env var is %s', (envValue, expected) => {
+    vi.stubEnv('AWS_REGION', envValue);
+    const logger = new Logger();
+    expect(logger['powertoolsLogData'].awsRegion).toBe(expected);
+    vi.unstubAllEnvs();
```
svozza: We can move these to a `beforeEach` block so we don't have to repeatedly call `vi.unstubAllEnvs();`

dreamorosi: also please look at the patterns in the other test files:
- use a single `describe` block (if any at all)
- test case descriptions should read in active voice `sets awsRegion to %s when AWS_REGION env var is %s`
- sections/blocks within each test case should be delimited with comments `// Prepare`, `// Act`, `// Assess`

svozza:
Thank you so much for such a quick turnaround on identifying and fixing this issue @MaxOpperman!

dreamorosi (approved):

ncino-esselman:
I was just opening an issue! When will this be packaged?

dreamorosi:
> I was just opening an issue! When will this be packaged?
> 
> 

Most likely in the next 12-18hrs, unless issues arise.

svozza:
Just a note that this fix has now been released as v2.25.1 on NPM. Once again, a big thank you to @MaxOpperman!

PR #4329, opened 2025-08-14
fix(event-handler): pass event, context and request objects into handler

### LINKED ISSUE #4340: Feature request: pass event, context and request objects into handler
### Use case

When customers create event handlers, they should have access to the raw Lambda event and the Lambda context. We should also expose a fetch Request object for customers who wish to use familiar web standards in their handlers. This is need for both route handlers and error handlers.

### Solution/User Experience

Handlers will take an options object with theese objects. The types will look something like this:

```ts
type RequestOptions = {
  request: Request;
  event: APIGatewayProxyEvent;
  context: Context;
};

type ErrorResolveOptions = RequestOptions & ResolveOptions;

type ErrorHandler<T extends Error = Error> = (
  error: T,
  options?: RequestOptions
) => Promise<ErrorResponse>;

type RouteHandler<
  TParams = Record<string, unknown>,
  TReturn = Response | JSONObject,
> = (args: TParams, options?: RequestOptions) => Promise<TReturn>;
```

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

svozza:
## Summary
This begins the work for for repsonse handling piece of the REST API event handler. As with other event handlers we need to pass the raw event and the lambda context as option into the handler. We will also pass a `Request` object in, should users prefer to use this web standard when defining their handlers.

### Changes

- Updated the type definition of `RouteHandler` and `ErrorHandler` to support passing the new parameters in.
- Tightened the return type of `RouteHandler` from `any` to `Response | JSONObject`
- Updated the unit tests to reflect these new changes
- Added factory function to unit tests for creating events of type `APIGatewayProxyEvent`
- Removed `trace` and `connect` HTTP methods [see note]

### Note
When attempting to create a `Request` object for the `trace` and `connect` tests, I received an error from the constructor saying they were not supported. On investigation I discovered that these methods are forbidden by the [Fetch specification](https://fetch.spec.whatwg.org/#methods) due to security concerns. I do not think we should support these methods in the event handler so have removed them.

**Issue number:** closes #4340

---

### DIFF
--- a/packages/event-handler/src/rest/BaseRouter.ts
+++ b/packages/event-handler/src/rest/BaseRouter.ts
@@ -8,6 +8,7 @@ import type { ResolveOptions } from '../types/index.js';
 import type {
   ErrorConstructor,
   ErrorHandler,
+  ErrorResolveOptions,
   HttpMethod,
   Path,
   RouteHandler,
@@ -157,12 +158,19 @@ abstract class BaseRouter {
    */
   protected async handleError(
     error: Error,
-    options?: ResolveOptions
+    options: ErrorResolveOptions
   ): Promise<Response> {
     const handler = this.errorHandlerRegistry.resolve(error);
     if (handler !== null) {
       try {
-        const body = await handler.apply(options?.scope ?? this, [error]);
+        const body = await handler.apply(options.scope ?? this, [
+          error,
+          {
+            request: options.request,
+            event: options.event,
+            context: options.context,
+          },
+        ]);
         return new Response(JSON.stringify(body), {
           status: body.statusCode,
           headers: { 'Content-Type': 'application/json' },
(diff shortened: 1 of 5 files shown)


packages/event-handler/src/rest/BaseRouter.ts
```diff
@@ -273,24 +281,6 @@ abstract class BaseRouter {
   ): MethodDecorator | undefined {
     return this.#handleHttpMethod(HttpVerbs.OPTIONS, path, handler);
   }
-
-  public connect(path: Path, handler: RouteHandler): void;
```
dreamorosi, line 277: Saw the note about why we removed these.

I understand why they're forbidden in the `Request` object, but I think (might be wrong) the security implications apply only when it's an outgoing request, which is why the object doesn't allow them.

In this case we're representing an incoming request which customers might still want to handle in their API, so perhaps in the future we should reconsider if there's demand - even if it complicates the typing given we can't have `Request` but only the original raw event when the request has one of these two methods.

Also, even if we removed the two named methods, can customers still create a handler using the [`route()` method](https://github.com/aws-powertools/powertools-lambda-typescript/blob/a6df09c952a85ce495888f1ebaa576ce6d00d105/packages/event-handler/src/rest/BaseRouter.ts#L142-L149)?

svozza: > Also, even if we removed the two named methods, can customers still create a handler using the [route() method](https://github.com/aws-powertools/powertools-lambda-typescript/blob/a6df09c952a85ce495888f1ebaa576ce6d00d105/packages/event-handler/src/rest/BaseRouter.ts#L142-L149)?

~They could now but that's only because I forgot to remove the verbs from the `HttpMethod` type.~ Although I guess, ultimately this is JS so I'm sure someone could hack around the type system.

Edit: turns out I did remove them! So if yoy try to create a route with those verbs your code won't compile but see my original comment.

I agree though, if there is demand we could add them. Maybe we could extend the `Request` class and special case the two verbs.

dreamorosi (approved):
Great stuff!

PR #4343, opened 2025-08-14
add function to convert Lambda proxy event to web response object

### LINKED ISSUE #4342: Feature request: Add function to convert Lambda event to web Request object
### Use case

The event handler will eventually support multiple event type (ALB etc). types, as such we need a common format to represent these requests. We should use the standard Fetch `Request` object for this purpose.

### Solution/User Experience

We will have a utility function to convert an `APIGatewayProxyEvent` to a `Request`. We should also add a type guard to test whether the event to be converted is actually is of type `APIGatewayProxyEvent`. 

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

svozza:
## Summary
The event handler will eventually support multiple lambda event type (ALB etc). types, as such we need a common format to represent these requests. This PR adds a utility function to convert the event to the web Request types

### Changes
- Adds the `proxyEventToWebRequest` function to handle to converion.
- Adds a type guard function to ensure that the event to be converted is the correct type.
- Unit tests for both functions. 

**Issue number:** closes #4342

---


packages/event-handler/src/rest/converters.ts
```diff
@@ -0,0 +1,48 @@
+import type { APIGatewayProxyEvent } from 'aws-lambda';
+
+function createBody(body: string | null, isBase64Encoded: boolean) {
```
dreamorosi: We generally try to use arrow functions across the project, except for cases where we need the scope.

svozza: Fair enough, I prefer to use this syntax because function definitions get hoisted so you don't have to worry about the order you declare your functions in a module. Changed.

packages/event-handler/src/rest/converters.ts
```diff
@@ -0,0 +1,48 @@
...
+const createBody = (body: string | null, isBase64Encoded: boolean) => {
+  if (body === null) return null;
+
+  if (!isBase64Encoded) {
+    return body;
+  }
+  return Buffer.from(body, 'base64').toString('utf8');
+};
+
+export function proxyEventToWebRequest(event: APIGatewayProxyEvent) {
```
dreamorosi:
```suggestion
export const proxyEventToWebRequest = (event: APIGatewayProxyEvent) => {
```

PR #4355, opened 2025-08-18
improv(ci): Automate the layer version tracking during the Make Release workflow

### LINKED ISSUE #4339: Maintenance: Automate the layer version tracking during the `Make Release` workflow
### Summary

Currently, we have to manually track the Layer version number and pass the correct version when running the `Make Release` workflow. This could be automated to improve the workflow. 

### Why is this needed?

This would reduce manual interventions and manual tracking of the layer version and make the process more efficienct

Automation

### Solution

We could simply have a file which contains the layer number which gets incremented when updating the docs arn

### Acknowledgment

### Future readers

sdangol:
## Summary

This PR automated the manual process of checking the current layer version, increment the version and input it to the release workflow. Instead, it makes use of an SSM Parameter to automate the process.

### Changes

This PR updates the layer arn update script to retrieve the current layer arn from SSM Parameter and increment it by 1 to use it as a new layer version instead of getting it passed from the workflow input. Manually doing the increment and passing it to the workflow can be error prone, so it automates that step.

**Issue number:** closes #4339 

---

### DIFF
--- a/.github/scripts/update_layer_arn.sh
+++ b/.github/scripts/update_layer_arn.sh
@@ -8,10 +8,13 @@
 # see .github/workflows/publish_layer.yml
 
 
-# Get the current layer version from SSM Parameter
-current_layer_arn=$(aws ssm get-parameter --name /aws/service/powertools/typescript/generic/all/latest --query Parameter.Value --output text --region us-east-1)
-current_layer_version=$(echo $current_layer_arn | sed 's/.*://')
-new_version=$((current_layer_version + 1))
+# Get the new layer arn from the first command-line argument
+new_layer_arn=$1
+if [ -z "$new_layer_arn" ]; then
+    echo "Usage: $0 <new_layer_arn>"
+    exit 1
+fi
+new_version=$(echo $new_layer_arn | sed 's/.*://')
 
 # Find all files with specified extensions in ./docs and ./examples directories
 # -type f: only find files (not directories)
--- a/.github/scripts/update_layer_arn.sh
+++ b/.github/scripts/update_layer_arn.sh
@@ -9,12 +9,9 @@
 
 
 # Get the current layer version from SSM Parameter
-if [ -z "$1" ]; then
-    echo "Usage: $0 <new_version>"
-    exit 1
-fi
-current_layer_arn=$(aws ssm get-parameter --name /aws/service/powertools/typescript/generic/all/$1 --query Parameter.Value --output text --region us-east-1)
+current_layer_arn=$(aws ssm get-parameter --name /aws/service/powertools/typescript/generic/all/latest --query Parameter.Value --output text --region us-east-1)
 current_layer_version=$(echo $current_layer_arn | sed 's/.*://')
+new_version=$((current_layer_version + 1))
 
 # Find all files with specified extensions in ./docs and ./examples directories
 # -type f: only find files (not directories)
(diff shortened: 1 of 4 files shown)


.github/workflows/publish_layer.yml
```diff
@@ -97,15 +90,15 @@ jobs:
     permissions:
       contents: write
       pull-requests: write
-      id-token: none
+      id-token: write
```
dreamorosi: How come did this change?

sdangol: I thought this was needed in order to get the AWS credentials to get the parameter. But, now I see that there's another separate action for that. Let me change it

.github/workflows/publish_layer.yml
```diff
@@ -103,9 +96,15 @@ jobs:
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8  # v5.0.0
         with:
           ref: ${{ github.sha }}
+      - id: creds
```
sdangol: @sthulb Do we need this to read the the public SSM parameter?

sthulb: Yes, we need credentials of some kind to pull info if we plan to read SSM params

.github/scripts/update_layer_arn.sh
```diff
@@ -8,8 +8,10 @@
 # see .github/workflows/publish_layer.yml
 
 
-# Get the new version number from the first command-line argument
-new_version=$1
+# Get the current layer version from SSM Parameter
+current_layer_arn=$(aws ssm get-parameter --name /aws/service/powertools/typescript/generic/all/latest --query Parameter.Value --output text --region us-east-1)
```
sthulb: This might not be the safest way to get the current version – there's no guarantee that /latest contains the latest version. You're better off finding the current server version number and querying the parameter directly: `/aws/service/powertools/typescript/generic/all/2.14.0`

svozza: I think it's an OK assumption that the `latest` will contain the latest version. The `Make Release` GitHub job is ran long before the job that updates the SSM parameters so it in terms of the whole workflow it will be pointing to the version we care about. 

Also, if `latest` is not pointing to the right version then I think we have bigger problems that this particular job.

sthulb: potentially, but I wouldn't chance it. Always read the latest version tag

sdangol: If we use the layer with the latest version tag that was just published, I guess we wouldn't even need to do the increment, right?

svozza: Actually that's a good point

sdangol: I've updated it now to use the latest version tag

dreamorosi: Remember that at this point of the workflow we haven't updated the SSM params, that happens later & manually, so we still need to increment.

The only sure-fire ways to know the version would be:
- use the CDK output file from any of the regions in a previous step, which contains the ARN of the layer just published
- list all the layers, find the latest just published, use that

With that said, this is a relatively low-risk action. We're gonna use this version number **only** to update the docs, which happens in a PR opened by this automation and will be reviewed.

As Stefano said, if the ARN is wrong there we can easily notice it before merging, but also it means something in the release went wrong.

sdangol: Yeah, I got confused with the published layer and the layer arn in the SSM parameter :) Will revert back the increment changes

sdangol:
@dreamorosi As dicussed during our sync call, I've reused the layer version that was output from the deployment job and used it to update the docs

sthulb (approved):

PR #4364, opened 2025-08-21
improv(ci): removed the secret inheritance and replaced with the needed secrets

### LINKED ISSUE #4363: Maintenance: Pass only the required secret to workflow
### Summary

Provide reusable workflows with access only to the specific secrets they require and use individual secret references instead of inheriting all secrets. This was alerted by a [SonarQube finding](https://sonarcloud.io/project/security_hotspots?sinceLeakPeriod=true&id=aws-powertools_powertools-lambda-typescript&tab=how_to_fix)

In file [.github/workflows/rebuild_latest_docs.yml](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/.github/workflows/rebuild_latest_docs.yml#L33)
```yml
jobs:
  release-docs:
    permissions:
      id-token: write  # trade JWT token for AWS credentials in AWS Docs account
      contents: read  # read from this repo to publish docs
    secrets : inherit <--- Use only required secrets instead of inherit
```

Similarly, secrets are being inherited in these files
- [make-release.yml](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/.github/workflows/make-release.yml#L100)
- [on_doc_merge.yml](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/.github/workflows/on_doc_merge.yml#L19)

### Why is this needed?

GitHub Actions workflows should follow the principle of least privilege by providing reusable workflows with access only to the specific secrets they require. When secrets: inherit is used to call a reusable workflow, all repository secrets become available to the reusable workflow, creating unnecessary security risks.

This practice increases the attack surface and potential for secret exposure, as any vulnerability in the reusable workflow could compromise all secrets rather than just the ones actually needed. Additionally, it violates the principle of least privilege by granting broader access than necessary.

Automation

### Solution

```yml
jobs:
  release-docs:
    permissions:
      id-token: write  # trade JWT token for AWS credentials in AWS Docs account
      contents: read  # read from this repo to publish docs
    secrets:
      AWS_DOCS_ROLE_ARN: ${{ secrets.AWS_DOCS_ROLE_ARN }}
      AWS_DOCS_BUCKET: ${{ secrets.AWS_DOCS_BUCKET }}
```

### Acknowledgment

### Future readers

sdangol:
## Summary

This PR replaces the places where a secret was inherited to a reusable workflow and instead passes the secrets that are only required which was flagged by SonarQube.

### Changes

Changed the workflows which call a reusable workflow to not inherit all the secrets and passed the secrets which are only required.

**Issue number:** closes #4363 

---

### DIFF
--- a/.github/workflows/make-release.yml
+++ b/.github/workflows/make-release.yml
@@ -94,7 +94,7 @@ jobs:
     secrets:
       AWS_LAYERS_BETA_ROLE_ARN: ${{ secrets.AWS_LAYERS_BETA_ROLE_ARN }}
       AWS_LAYERS_PROD_ROLE_ARN: ${{ secrets.AWS_LAYERS_PROD_ROLE_ARN }}
-      TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      TOKEN_GITHUB: ${{ secrets.GITHUB_TOKEN }}
     permissions:
       id-token: write
       contents: write
--- a/.github/workflows/publish_layer.yml
+++ b/.github/workflows/publish_layer.yml
@@ -18,7 +18,7 @@ on:
         description: "Role ARN for deploying the Layer to Beta"
       AWS_LAYERS_PROD_ROLE_ARN:
         description: "Role ARN for deploying the Layer to Prod"
-      TOKEN:
+      TOKEN_GITHUB:
         description: "GitHub Token to interact with GitHub"
     inputs:
       latest_published_version:
@@ -124,4 +124,4 @@ jobs:
         with: 
           temp_branch_prefix: 'ci-layer-docs'
           pull_request_title: 'chore(ci): update layer ARN on documentation'
-          github_token: ${{ secrets.TOKEN }}
+          github_token: ${{ secrets.TOKEN_GITHUB }}


.github/workflows/make-release.yml
```diff
@@ -97,7 +97,10 @@ jobs:
   # publish_layer -> reusable_deploy_layer_stack -> reusable_update_layer_arn_docs
   publish_layer:
     needs: publish-npm
-    secrets: inherit
+    secrets:
+      AWS_LAYERS_BETA_ROLE_ARN: ${{ secrets.AWS_LAYERS_BETA_ROLE_ARN }}
+      AWS_LAYERS_PROD_ROLE_ARN: ${{ secrets.AWS_LAYERS_PROD_ROLE_ARN }}
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
dreamorosi: This is causing the CI to fail https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/17130289519

sdangol: I've removed the `GITHUB_TOKEN` from there since it should be available by default to the workflows

sdangol: An [example](https://docs.github.com/en/enterprise-cloud@latest/actions/how-tos/reuse-automations/reuse-workflows#example-caller-workflow) was passing the `GITHUB_TOKEN` secret, so probably needed? @sthulb Could you review this?

sthulb (approved):

dreamorosi:
I checked out the branch and noticed that some of the secrets are getting flagged by linting

image

Can we check if it's fine to leave as is or there's a better way?

sdangol:
@dreamorosi I think it's because we need to pass in the correct environment for this. Do you know which environment these secrets are scoped to?

dreamorosi:
Should be `Docs`

sdangol:
@dreamorosi I tried adding the environment there, but the linter then complains that the property is not allowed. It seems like when using reusable workflows, we don't define the environments in the calling workflow.

Maybe the warning is because of this [issue with the extension](https://github.com/github/vscode-github-actions/issues/222) and is a false positive?

dreamorosi (approved):

PR #4370, opened 2025-08-22
improv(ci): Moved the secrets expanded in the run block to env variables

### LINKED ISSUE #4369: Maintenance: Avoid Expaning Secrets in a run block in a workflow
### Summary

As per [some SonarQube findings](https://sonarcloud.io/project/security_hotspots?sinceLeakPeriod=true&id=aws-powertools_powertools-lambda-typescript&tab=how_to_fix), we should pass secrets to steps through environment variables instead of command-line arguments.

The following files are flagged with the finding:
- .github/workflows/bootstrap_region.yml
- .github/workflows/dispatch_analytics.yml
- .github/workflows/layer_balance.yml
- .github/workflows/publish-package.yml

### Why is this needed?

GitHub Actions workflows should handle secrets securely by passing them through environment variables rather than directly expanding them in run commands. When secrets are expanded directly in run commands using GitHub’s expression syntax `(${{ secrets.SECRET_NAME }})`, they may be exposed in process lists, shell history, or files.

This practice increases the risk of accidental secret exposure, as command-line arguments and process information can be visible to other processes running on the same system.

Automation

### Solution

```yaml
name: Example

on:
  pull_request:

jobs:
  main:
    steps:
      - name: example
        env:
          SECRET: ${{ secrets.EXAMPLE_SECRET }}
        run: |
          example-command "$SECRET"
```

### Acknowledgment

### Future readers

sdangol:
## Summary

This PR moves the secrets that were expanded in the run block of the workflows as an env variable of each step. This is done to fix the [security finding from SonarQube](https://sonarcloud.io/project/security_hotspots?sinceLeakPeriod=true&id=aws-powertools_powertools-lambda-typescript&tab=how_to_fix)

### Changes

Created an env variable for the secret expanded in the run block and used that env variable instead.

**Issue number:** closes #4369 

---


.github/workflows/dispatch_analytics.yml
dreamorosi, line 1: We can remove this workflow, we don't use it anymore

.github/workflows/layer_balance.yml
```diff
@@ -62,8 +62,12 @@ jobs:
       - id: run-balance-new-region
         name: Run Balance
         if: ${{ inputs.start_at == '' }}
-        run: balance -read-region us-east-1 -write-region ${{ inputs.region }} -write-role ${{ secrets.BALANCE_ROLE_ARN }} -layer-name AWSLambdaPowertoolsTypeScriptV2 -dry-run=false
+        env:
```
dreamorosi: Should we set this at the job level, rather than at the step level?

.github/workflows/dispatch_analytics.yml
sdangol: Removed it now

.github/workflows/layer_balance.yml
sdangol: Yeah makes sense, moved it to the job level

dreamorosi (approved):
Thank you!

PR #4374, opened 2025-08-22
chore(commons): concatenate PT UA to AWS_SDK_UA_APP_ID when one is set

### LINKED ISSUE #4373: Maintenance: concatenate Powertools for AWS UA to existing `AWS_SDK_UA_APP_ID`
### Summary

Customer can set a `AWS_SDK_UA_APP_ID` environment variable in their AWS Lambda config. 

When this happens, the customer’s value should preserved and our UA string should be appended.

For example, if customer has set the variable in their Lambda config, we’d see `APP/CUSTOMER_STRING_FROM_LAMBDA_CONFIG/PT/NO-OP/Version`. 

### Why is this needed?

So that we can continue appending our user agent to AWS SDK client requests even when customers set their own app id.

_No response_

### Solution

_No response_

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR updates the logic for the user agent string so that when the `AWS_SDK_UA_APP_ID` environment variable is already set, it concatenates the Powertools for AWS UA string to the existing one.

**Issue number:** closes #4373

---

### DIFF
--- a/packages/commons/tests/unit/awsSdkUtils.test.ts
+++ b/packages/commons/tests/unit/awsSdkUtils.test.ts
@@ -1,4 +1,4 @@
-import { beforeAll, describe, expect, it, vi } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
 import { customUserAgentMiddleware } from '../../src/awsSdkUtils.js';
 import {
   addUserAgentMiddleware,
@@ -8,14 +8,11 @@ import {
 
 vi.hoisted(() => {
   process.env.AWS_EXECUTION_ENV = '';
+  process.env.AWS_SDK_UA_APP_ID = 'test';
 });
 
 describe('Helpers: awsSdk', () => {
   describe('Function: userAgentMiddleware', () => {
-    beforeAll(() => {
-      vi.spyOn(console, 'warn').mockImplementation(() => ({}));
-    });
-
     it('handles gracefully failures in adding a middleware and only log a warning', () => {
       // Prepare
       const client = {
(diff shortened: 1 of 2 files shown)


packages/commons/tests/unit/awsSdkUtils.test.ts
```diff
@@ -88,6 +82,11 @@ describe('Helpers: awsSdk', () => {
     });
   });
 
+  it('concatenates the PT AWS_SDK_UA_APP_ID when one is already set', () => {
+    // Assess
+    expect(process.env.AWS_SDK_UA_APP_ID).toEqual('test/PT/TEST/2.25.1');
```
sdangol: This might fail when the version changes.
```suggestion
    expect(process.env.AWS_SDK_UA_APP_ID).toEqual(`test/PT/TEST/${version}`);
```

sdangol (changes requested):

packages/commons/tests/unit/awsSdkUtils.test.ts
```diff
@@ -8,14 +8,11 @@ import {
 
 vi.hoisted(() => {
   process.env.AWS_EXECUTION_ENV = '';
+  process.env.AWS_SDK_UA_APP_ID = 'test';
```
sdangol, line 11: Should we use `vi.stubEnv('AWS_SDK_UA_APP_ID','test');`

dreamorosi: Not here, we're putting them directly at the top of the file via `vi.hoisted` because we want to test the logic being run when you import the package.

sdangol (approved):

PR #4408, opened 2025-09-03
feat(parser): integrate parser with Batch Processing

### LINKED ISSUE #4394: Feature request: Parser integration for Batch Processing
### Use case

As a customer, when working with Batch Processing, I want to have the utility parse and validate payloads before they're passed to my record handler[^1], so that I can focus on my business logic.

Today, assuming I'm working with SQS - but this applies to all the sources supported by the Batch Processing utility - I need to manually parse/validate payloads against my schemas manually in the body of the record handler like this:

```ts
import {
  BatchProcessor,
  EventType,
  processPartialResponse,
} from '@aws-lambda-powertools/batch';
import { parse } from '@aws-lambda-powertools/parser';
import {
  SqsRecordSchema,
  SqsSchema,
} from '@aws-lambda-powertools/parser/schemas/sqs';
import { z } from 'zod';
import type { SQSHandler, SQSRecord } from 'aws-lambda';

const customSchema = z.object({
  name: z.string(),
  age: z.number(),
});

const SqsExtendedSchema = SqsSchema.extend({
  Records: z.array(
    SqsRecordSchema.extend({
      body: JSONStringified(customSchema),
    })
  ),
});

const processor = new BatchProcessor(EventType.SQS); 

const recordHandler = async (record: SQSRecord) => { 
  const payload = parse(record, undefined, SqsExtendedSchema);
  const { body: { name, age } } = payload; // this is safe to use because it's parsed
};

export const handler: SQSHandler = async (event, context) =>
  processPartialResponse(event, recordHandler, processor, { 
    context,
  });
```

We could simplify this experience quite a bit by making some assumptions (see below).

[^1]: A record handler is the function designated to be called by the Batch Processing utility for each item

### Solution/User Experience

Because we always know the `EventType` at the time of instantiation of the `BatchProcessor` we can allow customers to simply pass us a schema for the record/item like so:

```ts
import {
  BatchProcessor,
  EventType,
  processPartialResponse,
} from '@aws-lambda-powertools/batch';
import { z } from 'zod';
import type { SQSHandler, SQSRecord } from 'aws-lambda';

const customSchema = z.object({
  name: z.string(),
  age: z.number(),
});

const processor = new BatchProcessor(EventType.SQS, customSchema);

const recordHandler = async (
  { body: { name, age } }: SQSRecord & { body: z.infer<typeof customSchema> }
) => { 
   // this is safe to use because it's parsed
 };

 export const handler: SQSHandler = async (event, context) =>
   processPartialResponse(event, recordHandler, processor, { 
     context,
   });
```

Under the hood, we can then:
- dynamically require the `parse` function, `JSONStringified` helper, and `SqsSchema` schema
- extend the `SqsSchema` with the customer-provided schema
- call `parse(<data>, undefined, <customer-provided-schema>)`
- mark the item as failed if parsing fails, pass it to the record handler if parsing is successful

**Open questions:**

Note that this will require us to [use dynamic imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import) - this should be fine since the record handler is meant to be async, so that code path is already async. Alternatively, we'll need to make the customer pass us the `parser` function and the schema already extended, which basically saves one line.

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

sdangol:
## Summary

This PR integrates the Parser functionality with the Batch Processor so that customers can parse and validate payloads before they're passed to the record handler. It supports both the extended schemas for the SQSRecord, KinesisRecord, and DynamoDBRecord as well as the inner payload schema.

### Changes

- Added parser as dev dependency
- Added a config to the constructor of `BasePartialBatchProcessor` to set the schema property
- Added a `parseRecord` method to the `BatchProcessor` to do the parsing by dynamically importing the parse function and the appropriate schema

**Issue number:** closes #4394 

---


package-lock.json
dreamorosi, line 1: Should Parser be added as dev dependency here instead?

packages/batch/src/BatchProcessor.ts
```diff
@@ -121,6 +138,44 @@ class BatchProcessor extends BasePartialBatchProcessor {
...
+          return parse(record, undefined, schema) as SQSRecord;
+        } catch (error) {
+          const { JSONStringified } = await import(
+            '@aws-lambda-powertools/parser/helpers'
+          );
+          const { SqsRecordSchema } = await import(
+            '@aws-lambda-powertools/parser/schemas/sqs'
+          );
+          const extendedSchema = SqsRecordSchema.extend({
+            // biome-ignore lint/suspicious/noExplicitAny: at least for now, we need to broaden the type because the JSONstringified helper method is not typed with StandardSchemaV1 but with ZodSchema
```
dreamorosi: In practice, I don't think we can support this use case at runtime anyway.

If someone passes us a complete StandardSchema schema from another library, we'll be able to use it as-is (`try` block).

For us to be able to extend the built-in `SqsRecordSchema` however we must receive a Zod schema or both the `extend` and `JSONStringified` will fail at runtime - so the type error is correct here.

With this in mind, we might want to use the `vendor` key on the schema (see [spec](https://github.com/standard-schema/standard-schema?tab=readme-ov-file#the-interface)) to check if it's a Zod schema and throw either 1/ throw an error or 2/ log a warning and ignore parsing if the schema is from another library we can't extend with.

If instead we find this conditional behavior confusing, we'll have to restrict the type of `schema` to Zod schemas only.

dreamorosi:
Can you also please address all the Sonar findings?

packages/batch/src/BatchProcessor.ts
svozza: I'm in favour of throwing here, if we just log a warning them we run the risk of passing on invalid events.

dreamorosi: Both throwing an error and continue processing an invalid event would result in the item being marked as failed by the Batch Processor and sent back to the source.

Most likely we should also log a warning, otherwise all operators would see is the item being reprocessed and eventually be sent to a DLQ or lost.

packages/batch/src/BatchProcessor.ts
```diff
@@ -121,6 +138,44 @@ class BatchProcessor extends BasePartialBatchProcessor {
...
+  /**
+   * Parse the record according to the schema passed.
+   *
+   * If the schema is not provided, it returns the record as is.
+   *
+   * @param record The record to be parsed
+   * @param eventType The type of event to process
+   * @param schema The StandardSchema to be used for parsing
+   */
+  public async parseRecord(
```
svozza: This method strikes me as odd, it's called `parseRecord` but it has a code path where the record is not parsed. I think logically this function should only do parsing and the conditional on whether to parse or not moves to L115. Something like this:

```ts
const recordToProcess = this.schema == null ? record : (await this.parseRecord(
        record,
        this.eventType,
        this.schema
      ));
```

packages/batch/src/BatchProcessor.ts
svozza: Yeah, that's a good point, it will be much easier to diagnose if we emit a log statement.

package-lock.json
sdangol: I saw the logger was using the jmespath package as a peer dependency, so I went with the same pattern https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/logger/package.json#L78
If we do dynamic imports, does it not need to be a peer dependency?

packages/batch/src/BatchProcessor.ts
svozza: Also, do we expect users to use this function as a standalone (in which case, it should probably be marked as `static`) or can it be a private method?

packages/batch/src/BatchProcessor.ts
sdangol: In order to cast the schema to Zod schema, do we need to dynamically import Zod as well?

packages/batch/src/BatchProcessor.ts
sdangol: The customers aren't expected to use this directly. I'll mark it as private

packages/batch/src/BatchProcessor.ts
dreamorosi: I'm ok to not use the correct type and typecast to `any`, as long as we check the `vendor` field at runtime and make sure to stop when the schema is not from Zod.

packages/batch/src/BatchProcessor.ts
```diff
@@ -121,6 +137,54 @@ class BatchProcessor extends BasePartialBatchProcessor {
...
+   *
+   * @param record The record to be parsed
+   * @param eventType The type of event to process
+   * @param schema The StandardSchema to be used for parsing
+   */
+  private async parseRecord(
+    record: EventSourceDataClassTypes,
+    eventType: keyof typeof EventType,
+    schema: StandardSchemaV1
+  ): Promise<SQSRecord | KinesisStreamRecord | DynamoDBRecord> {
```
svozza: The return type of this function is `Promise<SQSRecord | KinesisStreamRecord | DynamoDBRecord>` but it only seems to handle SQS payloads, can we remove th other return types and maybe rename the function `parseSqsRecord` or are we planning to add other record types here in a follow up PR? 

Also, I think we can remove some indentation here by inverting the event type check and returning early:

```ts
async function parseRecord(
    record: EventSourceDataClassTypes,
    eventType: keyof typeof EventType,
    schema: StandardSchemaV1
): Promise<SQSRecord | KinesisStreamRecord | DynamoDBRecord> {
    if (eventType !== EventType.SQS) {
        console.warn(
            `The event type provided is not supported. Supported events: ${Object.values(EventType).join(',')}`
        );
        throw new Error('Unsupported event type');    
    }
    
    const {parse} = await import('@aws-lambda-powertools/parser');
    const extendedSchemaParsing = parse(record, undefined, schema, true);
    if (extendedSchemaParsing.success)
        return extendedSchemaParsing.data as SQSRecord;
    // ...
}
```

dreamorosi: This PR should probably have been a draft.

We agreed offline that we were going to do a PoC with SQS only and then add the others, so I think @sdangol will add them soon.

sdangol: Yeah, I'll mark it as draft and add the other ones.

sdangol:
@dreamorosi I'm still a bit confused about extending the `KinesisDataStreamRecord`.
```typescript
if (eventType === EventType.KinesisDataStreams) {
      const extendedSchemaParsing = parse(record, undefined, schema, true);
      if (extendedSchemaParsing.success)
        return extendedSchemaParsing.data as KinesisStreamRecord;
      if (schema['~standard'].vendor === SchemaType.Zod) {
        const { JSONStringified } = await import(
          '@aws-lambda-powertools/parser/helpers'
        );
        const { KinesisDataStreamRecord } = await import(
          '@aws-lambda-powertools/parser/schemas/kinesis'
        );
        const extendedSchema = KinesisDataStreamRecord.extend({
          // biome-ignore lint/suspicious/noExplicitAny: The vendor field in the schema is verified that the schema is a Zod schema
          data: // Decompress and decode the data to match against schema
        });
        return parse(record, undefined, extendedSchema);
      }
      console.warn(
        'The schema provided is not supported. Only Zod schemas are supported for extension.'
      );
      throw new Error('Unsupported schema type');
    }
```

To extend it, should I create another helper which does the decompressions and decoding?
Or should I just use `Envelopes` for the whole thing. But, using `Envelopes` would just return the internal payload and our record handler expects the whole Record. We could maybe update the Record with the parsed payload and return it.
What would you suggest?

dreamorosi:
I've been thinking about your question for a while and there's no straightforward way to do it with our current feature set.

We can't use envelopes because of what you mentioned, and doing the parsing in two parts while possible, is suboptimal for two reasons:
- we'd be iterating through the entire batch in order to parse each record and then we'd have to recompose the object
- in case of parsing errors the path of the failed field(s) would be messed up since we'd no longer parse a schema (i.e. there's an error at the 2nd item in the batch, the error path should be something like `Records.1.kinesis.data.foo`)

With this in mind, I think we'll have to extract the `transform` logic that we have [here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/8b25ce7ff2c74a1db111c82e0f42970151c4cbc2/packages/parser/src/schemas/kinesis.ts#L14-L32) into its own helper (with a dedicated PR) and then use the helper in the Batch Processing utility when extending.

dreamorosi:
@sdangol can you look at the SonarCloud issue?

sdangol:
@dreamorosi @svozza This is now ready to be reviewed.

dreamorosi:
The SonarCloud message still shows an issue

packages/batch/src/BasePartialBatchProcessor.ts
```diff
@@ -42,12 +44,20 @@ abstract class BasePartialBatchProcessor extends BasePartialProcessor {
    */
   public eventType: keyof typeof EventType;
 
+  /**
+   * The schema of the body of the event record for parsing
+   */
+  public schema?: StandardSchemaV1;
```
svozza: Could this be `protected` rather than `public`?

dreamorosi: or even fully private like `#schema` if not used by children.

sdangol: It is being used by the child class `BatchProcessor`. So, I'll mark it as protected.

packages/batch/src/BatchProcessor.ts
```diff
@@ -121,6 +138,108 @@ class BatchProcessor extends BasePartialBatchProcessor {
...
+            OldImage: DynamoDBMarshalled<StreamRecord['OldImage']>(
+              schema as any
+            ).optional(),
+            NewImage: DynamoDBMarshalled<StreamRecord['NewImage']>(
+              schema as any
+            ).optional(),
+          }),
+        });
+      }
+      default:
```
svozza: Let's wrap the default case here in braces for block scope like we do the other cases

svozza (approved):

packages/batch/src/BatchProcessor.ts
dreamorosi: I'm testing the package on a function and it works, but I think we forgot to emit some kind of warning/log/error when the parsing fails.

Is the Python implementation doing anything in this regard?

sdangol: I think this is the [python implementation](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/utilities/batch/base.py#L377-L386) for handling parsing failure. Should we do something similar?

sdangol: I was trying to test it for Kinesis using https://github.com/dreamorosi/kinesis-batch and it was failing, trying to debug.

packages/batch/src/BatchProcessor.ts
```diff
@@ -121,6 +138,109 @@ class BatchProcessor extends BasePartialBatchProcessor {
...
+    schema: StandardSchemaV1
+  ) {
+    switch (eventType) {
+      case EventType.SQS: {
+        const [{ JSONStringified }, { SqsRecordSchema }] = await Promise.all([
+          import('@aws-lambda-powertools/parser/helpers'),
+          import('@aws-lambda-powertools/parser/schemas/sqs'),
+        ]);
+        return SqsRecordSchema.extend({
+          body: JSONStringified(schema as any),
```
sdangol: @dreamorosi The body could just be a string only as well, right? If we try to extend using `JSONStringified`, it currently fails for when we pass a string.

sdangol: Should we not need to use the `JSONStringified` helper here, or any other helpers in other events? and assume customers will use the necessary helpers if they need to?

packages/batch/src/BatchProcessor.ts
dreamorosi: Message me, I got it working fine

dreamorosi: > Should we do something similar?

Yes, but in order to be able to do that we'll need to add a logger feature which we currently don't have.

Let's put a pin on this and I'll open a separate issue.

packages/batch/tests/unit/BatchProcessor.test.ts
```diff
@@ -286,4 +293,1018 @@ describe('Class: AsyncBatchProcessor', () => {
...
+
+      // Assess
+      await expect(processor.process()).rejects.toThrowError(
+        FullBatchFailureError
+      );
+    });
+  });
+
+  describe('Batch processing with Parser Integration: Passing Internal Kinesis Record Schema without transformers', () => {
+    it.skip('completes the processing with failures if some of the payload does not match the passed schema', async () => {
```
sdangol: @dreamorosi There was this usecase which seems problematic when trying to parse as we had discussed offline. If the customer passes a string schema without transformation and expects the parser to do the base64 decoding, it wouldn't happen because the normal schema would also pass in this case.

sdangol: Expected to get an encoded string as per our slack conversation

packages/batch/src/constants.ts
```diff
@@ -17,6 +17,13 @@ const EventType = {
   DynamoDBStreams: 'DynamoDBStreams',
 } as const;
 
+/**
+ * Enum of supported schema types for the utility
+ */
+const SchemaType = {
```
dreamorosi:
```suggestion
const SchemaVendor = {
```

packages/batch/src/BatchProcessor.ts
dreamorosi, line 1: For new methods, I'd prefer creating signatures with objects rather than positional args since it makes it easier to read/understand at the call site.

For example, with the current signatures calling `parseRecord` looks like: `await this.parseRecord(record, this.eventType, this.schema);` which is not too bad because the args being passed are named correctly.

Calling `createExtendedSchema` instead looks much more opaque: `await this.createExtendedSchema(eventType, schema, false)` making it unclear what `false` stands for.

If the signature was:
```ts
private async createExtendedSchema(options: {
  eventType: keyof typeof EventType,
  schema: StandardSchemaV1,
  useTransformers: boolean
})
```

calling it would look like: await this.createExtendedSchema({ eventType, schema, useTransformers : false })` which is much clearer.

Also, unless the methods are `public` or `protected`, let's use `#methodName` instead of `private methodName` to make them truly private at runtime.

packages/batch/src/BatchProcessor.ts
```diff
@@ -121,6 +138,156 @@ class BatchProcessor extends BasePartialBatchProcessor {
...
+        undefined,
+        extendedSchemaWithoutTransformers
+      ) as EventSourceDataClassTypes;
+    } catch {
+      const extendedSchemaWithTransformers = await this.createExtendedSchema(
+        eventType,
+        schema,
+        true
+      );
+      return parse(
```
dreamorosi: Using `safeParse: true` in this call will make it easier to add debug logs later since right now a failed parsing exits this scope.

Similarly, for the `parse` call above we could achieve the same without generating a stack trace, which can be more performant.

packages/batch/src/BatchProcessor.ts
```diff
@@ -121,6 +138,156 @@ class BatchProcessor extends BasePartialBatchProcessor {
...
+
+  /**
+   * Parse the record according to the schema passed.
+   *
+   * If the passed schema is already an extended schema,
+   * it directly uses the schema to parse the record
+   *
+   * Only Zod Schemas are supported for automatic schema extension
+   *
+   * @param record The record to be parsed
```
dreamorosi:
```suggestion
   * @param record - The record to be parsed
```

dreamorosi: Same for the other `@param` added in this diff

packages/batch/src/BatchProcessor.ts
```diff
@@ -121,6 +138,156 @@ class BatchProcessor extends BasePartialBatchProcessor {
...
+    }
+  }
+
+  /**
+   * Parse the record according to the schema passed.
+   *
+   * If the passed schema is already an extended schema,
+   * it directly uses the schema to parse the record
+   *
+   * Only Zod Schemas are supported for automatic schema extension
```
dreamorosi: Let's use punctuation correctly throughout the docstring.

packages/batch/src/types.ts
```diff
@@ -89,6 +91,10 @@ type PartialItemFailures = { itemIdentifier: string };
  */
 type PartialItemFailureResponse = { batchItemFailures: PartialItemFailures[] };
 
+type BasePartialBatchProcessorConfig = {
+  schema: StandardSchemaV1;
```
dreamorosi, line 106: We need to document `schema` so that customer know what it does and all its nuances.

dreamorosi (changes requested):
Only a couple of last styling and documentation details and then we're good to merge!

packages/batch/src/BatchProcessor.ts
```diff
@@ -121,6 +138,156 @@ class BatchProcessor extends BasePartialBatchProcessor {
...
+    schema: StandardSchemaV1
+  ): Promise<EventSourceDataClassTypes> {
+    const { parse } = await import('@aws-lambda-powertools/parser');
+    // Try parsing with the original schema first
+    const extendedSchemaParsing = parse(record, undefined, schema, true);
+    if (extendedSchemaParsing.success) {
+      return extendedSchemaParsing.data as EventSourceDataClassTypes;
+    }
+    // Only proceed with schema extension if it's a Zod schema
+    if (schema['~standard'].vendor !== SchemaType.Zod) {
```
svozza: Should we not do this check as the very first thing in this function? We're wasting work doing the parsing if all we're going to do is throw based on information we had before we began parsing.

sdangol: The first parsing that's being done, doesn't necessarily need it to be a Zod schema. We only need the Zod schema when we're expecting to use the transformers.

svozza: Ah I see.

PR #4410, opened 2025-09-03
improv(tests): fixed the logger test breaking with Nodejs v22.19.0

### LINKED ISSUE #4409: Maintenance: Logger unit test fails with Nodejs v22.19.0 in the pipeline
### Summary

One of the unit tests in the Logger utility is failing
https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/17432536582/job/49494081506

There seems to be an extra parameter `diff` that was passed in error. This is only seen in v22.19.0. Previous versions of Nodejs didn't pass that parameter.

### Why is this needed?

The test is currently failing and blocking the pipeline

Tests

### Solution

We can either make the test less strict by using `toMatchObject` instead of `toEqual`.
or, we could just include the `diff` parameter to be expected.

### Acknowledgment

### Future readers

sdangol:
## Summary

This PR fixes the breaking tests in Nodejs v22.19.0 by replacing the deep checking of error properties with subset checking of the error properties.

### Changes

- Added the `diff` property in the expected object if running on Nodejs version >=22.19.0

**Issue number:** closes #4409 

---

### DIFF
--- a/packages/logger/tests/unit/formatters.test.ts
+++ b/packages/logger/tests/unit/formatters.test.ts
@@ -371,7 +371,7 @@ describe('Formatters', () => {
         operator: 'strictEqual',
         code: 'ERR_ASSERTION',
         generatedMessage: true,
-        diff: 'simple',
+        diff: process.versions.node >= '22.19.0' ? 'simple' : undefined,
       },
     },
     {


packages/logger/tests/unit/formatters.test.ts
```diff
@@ -450,7 +450,7 @@ describe('Formatters', () => {
       const formattedError = formatter.formatError(error);
 
       // Assess
-      expect(formattedError).toEqual({
+      expect(formattedError).toMatchObject({
```
dreamorosi: Can we just add the missing property in the `expectedFields` of the error that is causing issues rather than relaxing the assertion for all of them?

While I agree that this test is brittle and subject to changes of this kind, we have it as first canary to ensure that error objects are parsed consistently and as expected.

If we start matching subsets of these objects we might miss some regression or breaking behavior in fields that are excluded from the assertion, but that would break customers.

svozza: I think we'd need to do a check for the Node version when creating `expectedFields` because just adding the expected field would break the Node 20 tests.

packages/logger/tests/unit/formatters.test.ts
```diff
@@ -371,6 +371,7 @@ describe('Formatters', () => {
         operator: 'strictEqual',
         code: 'ERR_ASSERTION',
         generatedMessage: true,
+        diff: process.versions.node >= '22.19.0' ? 'simple' : undefined,
```
dreamorosi, line 374:
```suggestion
        diff: expect.anything(),
```

dreamorosi (approved):

PR #4428, opened 2025-09-05
feat(event-handler): add middleare registration and composition to rest handler

### LINKED ISSUE #4427: Feature request: Implement middleware composition in event handler
### Use case

Customers should be able to register multiple middlewares and have them applied in the order they were added. As all middleware share the same type signature this is ultimately just function composition. Using the familiar pattern of using a callback, typically named `next` to proceeed to the next middleware, users should also have the option to shortcircuit the composition and return early from the middleware chain.

### Register middleware

Users can register middleware with the `use` method:

```ts
app.use(async (params, options, next) => {
    console.log('middleware1-start');
    await next();
    console.log('middleware1-end');
});

app.use(async (params, options, next) => {
    console.log('middleware2-start');
    await next();
    console.log('middleware2-end');
});

app.get('/test', async () => {
    console.log('handler');
    return { success: true };
});

const res = await app.resolve(createTestEvent('/test', 'GET'), context);

/** prints
 'middleware1-start',
 'middleware2-start',
 'handler',
 'middleware2-end',
 'middleware1-end',
 **/   
```

### Compose middleware

We expose a function called `composeMiddleware`, which allows users to create re-usuable middleware stacks and we use this function in BaseRouter itself to chain the middleware.

```ts
import {composeMiddleware} from 'from '@aws-lambda-powertools/event-handler/rest/utils.jsl';'

const addCustomHeaders: Middleware = async (params, { request, event }, next) => {
  request.headers.set('x-request-id', event.requestContext.requestId);
  request.headers.set('x-source-ip', event.requestContext.identity.sourceIp);
  
  return next();
};

const logRequest: Middleware = async (params, { request, event }, next) => {
  console.log(`Processing ${event.httpMethod} ${event.path} from ${event.requestContext.identity.sourceIp}`);
  
  return next();
};

const customMiddleware = composeMiddleware([addCustomHeaders, logRequest]);

app.use(customMiddleware)
```

### Shortcircuit

```ts
app.use(async (params, options, next) => {
    await next();
});

app.use(async (params, options, next) => {
    return new Response('Short-circuited', { status: 401 });
});

app.get('/test', async () => {
    return { success: true };
});

// Act
const result = await app.resolve(
    createTestEvent('/test', 'GET'),
    context
);

// result is 401 response not 200 with { success: true } body
```

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

svozza:
## Summary

This PR adds the functionality to register middleware and to compose mutliple middlewares to create reusable middleware stacks.

**Note**: currently there is no way to alter the response inside a middleware function, this will be done in the next PR.

#### 1. **Middleware Composition Logic** (`src/rest/utils.ts`)
- Implemented `composeMiddleware()` function for chaining middleware execution
- Added protection against multiple `next()` calls
- Supports middleware short-circuiting by returning responses

#### 2. **Router Integration** (`src/rest/BaseRouter.ts`)
- Added `use()` method for middleware registration
- Updated `resolve()` method to execute middleware chain before route handlers
- Proper scope binding for class-based handlers

#### 3. **Added unit tests** (`tests/unit/rest/BaseRouter.test.ts`)
- **Execution Order**: Middleware runs before handlers in registration order
- **Short-circuiting**: Middleware can return responses to bypass remaining chain
- **Parameter Passing**: Route params and request options passed correctly
- **Error Handling**: Errors in middleware before/after `next()` and in handlers
- **Edge Cases**: Multiple `next()` calls, middleware not calling `next()`
- **Response Types**: Support for both Response objects and JSON objects
- **Class Decorators**: Integration with class-based handlers preserving `this` scope

**Issue number:** closes #4427 

---


packages/event-handler/src/rest/BaseRouter.ts
```diff
@@ -185,14 +195,24 @@ abstract class BaseRouter {
...
+
+      const result = await middleware(
         route.params,
         {
           event,
           context,
           request,
         },
-      ]);
+        () => handler(route.params, { event, context, request })
```
svozza, line 235: Note here that we are just treating the handler like another middleware that doesn't call next

dreamorosi: neat

packages/event-handler/src/rest/BaseRouter.ts
```diff
@@ -185,14 +195,24 @@ abstract class BaseRouter {
...
+      );
+
+      if (result === undefined) throw new InternalServerError();
```
dreamorosi, line 242: How come did we add this and should this be a 500 and not a 200?

packages/event-handler/src/rest/utils.ts
```diff
@@ -105,3 +108,39 @@ export const isAPIGatewayProxyResult = (
       typeof result.isBase64Encoded === 'boolean')
   );
 };
+
+export const composeMiddleware = (middlewares: Middleware[]): Middleware => {
```
dreamorosi, line 146: This function could use a bit of documentation, it's not immediately clear what it does, but I am ok for you to add it in one of the next PRs.

dreamorosi (dismissed):
Only left a couple notes, great work!

packages/event-handler/src/rest/BaseRouter.ts
svozza: This is an artifact of treating the handler like middleware. The compiler can't figure out that we will always return a result because the type of `NextFunction` is  `() => Promise<HandlerResponse | void>`. We should never get `undefined` here because the handler should always respond with `type HandlerResponse = Response | JSONObject` so I added this to appease the compiler.

packages/event-handler/src/rest/utils.ts
svozza: Oh yeah sorry meant to add that. Will do it now.

packages/event-handler/src/rest/BaseRouter.ts
dreamorosi: Makes sense, let's add a comment above it to document this.

svozza: Yep, good shout.

dreamorosi (approved):
Great stuff!

After we're done with the feature, let's see if we can bring down the code duplication in the unit test files a bit.

PR #4437, opened 2025-09-05
feat(event-handler): add route specific middleware registration and execution

### LINKED ISSUE #4430: Feature request: Add Route Specific Middleware for REST Event Hndler
### Use case

Users should be able to add middleware to specific routes that are ran after any global middlewares have ran.

### Register middleware

Middleware for routes is added as an argument to the HTTP verb methods exposed by the router:

```ts
const middleware1 = async (params, options, next) => {
    console.log('middleware1-start');
    await next();
    console.log('middleware1-end');
};

const middleware2 = async (params, options, next) => {
    console.log('middleware2-start');
    await next();
    console.log('middleware2-end');
};

const middleware3 = async (params, options, next) => {
    console.log('middleware3-start');
    await next();
    console.log('middleware3-end');
};

app.use(middleware1);

app.get('/global', [], async () => {
    console.log('handler');
    return { success: true };
});

app.get('/route-specific', [middleware2, middleware3], async () => {
    console.log('handler');
    return { success: true };
});

await app.resolve(createTestEvent('/global', 'GET'), context);

/** prints
 'middleware1-start',
 'handler',
 'middleware1-end',
 **/

await app.resolve(createTestEvent('/route-specific', 'GET'), context);

/** prints
 'middleware1-start',
 'middleware2-start',
 'middleware3-start',
 'handler',
 'middleware3-end',
 'middleware2-end',
 'middleware1-end',
 **/
```

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

svozza:
## Summary
This PR implements route-specific middleware functionality for the REST event handler, allowing middleware to be applied to individual routes in addition to global middleware. E.g.,

```ts
// functional
const middleware1 = async (params, options, next) => {
    console.log('middleware1-start');
    await next();
    console.log('middleware1-end');
};

app.get('/test', [middleware1], (params, {event, context, request}) => {
  return { result: 'get-decorator-middleware' };
});

// decorators
@app.get('/test', [middleware1])
public async getTest() {
    return { result: 'get-decorator-middleware' };
}
```

#### Route Class (`packages/event-handler/src/rest/Route.ts`)
- Added `middleware` property to Route class constructor with default empty array

#### BaseRouter (`packages/event-handler/src/rest/BaseRouter.ts`)
- Updated all HTTP method signatures to support optional middleware parameter: `get(path, middleware?, handler)`
- Added overloads for decorator syntax with middleware.
- Modified `resolve()` method to compose global and route-specific middleware
- Route-specific middleware executes after global middleware in execution chain

#### RouteHandlerRegistry (`packages/event-handler/src/rest/RouteHandlerRegistry.ts`)
- Updated `resolve()` method to include middleware in returned `RouteHandlerOptions`

### Tests
- Created factory functions for creating middlewares to reduce code duplication
- Tests for functional middleware
- Decorator tests

**Issue number:** closes #4430 

---


packages/event-handler/src/rest/Route.ts
```diff
@@ -1,16 +1,28 @@
...
   readonly handler: RouteHandler;
+  readonly middleware: Middleware[];
 
-  constructor(method: HttpMethod, path: Path, handler: RouteHandler) {
+  constructor(
+    method: HttpMethod,
+    path: Path,
+    handler: RouteHandler,
+    middleware: Middleware[] = []
+  ) {
```
dreamorosi, line 20: Should we make this an object? It's starting to become unwieldy.

packages/event-handler/tests/unit/rest/BaseRouter.test.ts
```diff
@@ -37,6 +40,50 @@ const createTestEvent = (
...
+  executionOrder: string[],
+  response: any
+): Middleware => {
+  return async (_params, _options, _next) => {
+    executionOrder.push(name);
+    return response;
+  };
+};
+
+const createNoNextMiddleware = (
```
dreamorosi: I'd consider moving these to a dedicated file under `tests/unit/rest/helpers.ts`

packages/event-handler/tests/unit/rest/BaseRouter.test.ts
```diff
@@ -179,23 +226,14 @@ describe('Class: BaseRouter', () => {
     expect(logger.debug).toHaveBeenCalledWith('test debug');
   });
 
-  describe('middleware', () => {
+  describe('middleware - global', () => {
```
dreamorosi, line 170: I think if possible, I'd remove the usage of nested `describe` blocks in favor of `// #region middleware` & `// endregion`  [comments](https://stackoverflow.com/a/76568099), so we avoid the nesting but we still get the folding. Same for the other route specific middleware group.

This will also remove the SonarCloud issue.

packages/event-handler/tests/unit/rest/BaseRouter.test.ts
```diff
@@ -583,6 +1021,133 @@ describe('Class: BaseRouter', () => {
...
+          public async optionsTest() {
+            return { result: 'options-decorator-middleware' };
+          }
+
+          public async handler(event: unknown, context: Context) {
+            return app.resolve(event, context);
+          }
+        }
+
+        const lambda = new Lambda();
```
dreamorosi, line 1068: Should we change one of these tests to use the `const handler = lambda.handler.bind(lambda)` and add the `{ scope: this }` in the resolve call, so we test the scope preservation?

As far as I can see we're not doing it in this test, and I see we've modified decorator code - but if I missed it or it's covered elsewhere you can ignore, although I try to write it like that always so we spot regressions in multiple places.

dreamorosi (commented):
Left a few comments / questions - the implementation is quite involved but looks ok.

Fell free to request another review when ready.

packages/event-handler/tests/unit/rest/BaseRouter.test.ts
svozza: There are tests for scope preservation eleswhere in the suite, we don't preserve `this` in middleware so I didn't include it but it does no harm to test it here anyway.

packages/event-handler/tests/unit/rest/BaseRouter.test.ts
svozza: As mentioned offline, I will break this test suite into multiple files in a separate PR as it's getting too big.

packages/event-handler/tests/unit/rest/BaseRouter.test.ts
svozza: Agrred

packages/event-handler/src/rest/Route.ts
svozza: You mean don't have it as a class and just use a `type`? Definitely happy to do so.

svozza: I guess the only thing here is that we create the id in the constructor. If we allow the person constructing the object to create the ID themselves, they could get it wrong. We rely on that specific format in some places in the code.

dreamorosi: What I meant was changing the signature to something like:

```ts
type RouteOptions = {
    method: HttpMethod,
    path: Path,
    handler: RouteHandler,
    middleware: Middleware[] = []
};

constructor(options: RouteOptions) {
```

mainly because it's getting crowded, but it's only a style preference since they're all mandatory params.

svozza: Ah I see what you mean. I will do that in a separate PR but I agree, a new type would be nicer here.

PR #4439, opened 2025-09-08
feat(event-handler): implement mechanism to manipulate response in middleware

### LINKED ISSUE #4393: Feature request: Implement Middleware for REST Event Handler
### Use case

The REST API event handler should provide a way to apply middleware to requests both globally and at route level. The order in which middleware are passed to the event handler are the order in which they will be applied. Users will also be able to return early from from middleware and any errors thrown will be caught and handled in the same manner as errors from handlers. 

## Global middleware
Middleware that should apply to all routes would be register as so:

```ts
app.use(async ({req, res, context, event}}, next) => {
  // this will run before the handler executes
  console.log(`Http method is: ${req.method}`);
  // this will ensure the next middleware in the chain is executed
  await next();
  // this will execute after the handler executes
  res.headers.set('X-post-req-header', 'some-value')
});
```

## Route specific middleware
Route specific middleware can be specified when registering a router:

```ts
const middleware1(async ({req, res, context, event}}, next) => {
  // ....
});

const middleware2(async ({req, res, context, event}}, next) => {
  // ....
});

app.get('/standard', [middleware1, middleware2], () => {
  return new Response(JSON.stringify({message: 'OK'}), {
    status: 200,
    headers: {'Content-Type': 'application/json'}
  });
});
```

## Early returns

Users can also return early from middleware. This will prevent the execution of any other middleware further down the chain:

```ts
app.use(async ({req, res, context, event}}, next) => {
  if(someCondition) {
    res.header.set('X-return-early', 'some-value');
    return res; // you may also return a JSON object or an APIGatewayProxyResult
  }
  // this will ensure the next middleware in the chain is executed
  await next();
});
```

## Errors

Errors thrown in middleware will stop processing and result in the thrown error being handled in the same manner as errors from handlers.

```ts
app.use(async ({req, res, context, event}}, next) => {
  if(someCondition) {
    throw new BadRequestError();
  }
  await next();
});
```

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

svozza:
## Summary
This commit introduces a mechanism for middleware to manipulate HTTP responses sent by the handler. This is required for middleware such as CORS.

### Changes
**BaseRouter.ts**
- `HandlerOptions` now exposes a `res` field. This is a Web API `Response` object that represents the response that will be sent. This allows middleware that runs after the handler has been executed to manipulate the response before it is finally sent
- The middleware that runs the handler now does not return the result but mutates the `res` field and adds the response there
- Renamed `RequestOptions` to `RequestContext` because it now deals with responses as well as requests

**Converters.ts**
- Created new function to convert `HandlerResponse` for a Web API `Response` object.

**Tests:**
- Added middleware header precedence testing ('last write wins' behavior)
- Added tests for new `handlerResultToResponse` fucntion
- Implemented direct unit tests for `composeMiddleware`

**Issue number:** closes #4393

---


packages/event-handler/src/rest/BaseRouter.ts
```diff
@@ -223,34 +234,38 @@ abstract class BaseRouter {
...
-      // In practice this we never happen because the final 'middleware' is
-      // the handler function that allways returns HandlerResponse. However, the
-      // type signature of of NextFunction includes undefined so we need this for
-      // the TS compiler
-      if (result === undefined) throw new InternalServerError();
+      // middleware result takes precedence to allow short-circuiting
+      const result = middlewareResult ?? handlerOptions.res;
 
-      return await handlerResultToProxyResult(result);
+      return handlerResultToProxyResult(result);
```
svozza, line 262: I wanted to move this to its own middleware that would always run at the end but we can't have the early return/short-circuit behaviour of the middleware if we do that.

packages/event-handler/src/types/rest.ts
```diff
@@ -14,17 +14,18 @@ type ErrorResponse = {
   message: string;
 };
 
-type RequestOptions = {
+type HandlerOptions = {
```
svozza: Should we call this something like `HandlerContext` or `RequestContext`? `Options` doesn't feel right for a mutable container for state. I initially chose this`HandlerOptions` name because I didn't want users to confuse it with the built-in lambda context.

dreamorosi: If we see this as being manipulated across different parts of the request lifecycle, for example in a middleware, then  I'd call it `RequestContext`. If not, `HandlerContext` could work.

For me options suggests a configuration rather than arbitrary state.

svozza: Yeah, I think I'll go with `RequestContext` as you say, options definitely feels more static and config-like.

svozza: This has been renamed

dreamorosi (approved):

PR #4458, opened 2025-09-10
chore(event-handler): expose rest handler functionality

### LINKED ISSUE #4456: Maintenance: Expose REST Event Handler Under `experimental-rest` Path
### Summary

To allow users to use the beta version of the REST event handler we need to expose the API:

```ts
import {Router} from '@aws-lambda-powertools/event-handler/experimental-rest';
```

This will require adding an `index.ts` file to export the `Router` class, converters, utils and constants.

We also need to update the `exports` and `typesVersions` sections of `package.json`.

### Why is this needed?

Users will not be able to import the package otherwise.

_No response_

### Solution

_No response_

### Acknowledgment

### Future readers

svozza:
## Summary
This PR exposes the REST API event handler under the namespace `experimental-rest`:

```ts
import {Router} from '@aws-lambda-powertools/event-handler/experimental-rest';
```

### Changes

- Created an `index.ts` file so classes and functions can be imported from the the top level namespace
- Updated the `exports` and `typesVersions` sections of `package.json`
- Where applicable, updated the tests to use the the `index.ts` file to import classes and utilities.
- Renamed conversion functions `*ToResponse` to `*ToWebResponse`.

**Issue number:** closes #4456

---

### DIFF
--- a/packages/event-handler/src/rest/index.ts
+++ b/packages/event-handler/src/rest/index.ts
@@ -1,0 +1,5 @@
+export { HttpErrorCodes, HttpVerbs } from './constants.js';
+export * from './converters.js';
+export * from './errors.js';
+export { Router } from './Router.js';
+export * from './utils.js';
--- a/packages/event-handler/src/rest/index.ts
+++ b/packages/event-handler/src/rest/index.ts
@@ -1,5 +1,28 @@
 export { HttpErrorCodes, HttpVerbs } from './constants.js';
-export * from './converters.js';
-export * from './errors.js';
+export {
+  handlerResultToProxyResult,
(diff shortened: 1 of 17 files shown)


packages/event-handler/src/rest/index.ts
dreamorosi, line 1: Even if more verbose, I'd rather we export only and exactly the things we want customers to be able to use. Even if today it's all of the ones in these files, I'd rather be explicit so we avoid future issues with leaking APIs that were not intended for usage.

svozza: Ah yes, good shout.

dreamorosi (approved):

PR #4460, opened 2025-09-10
chore(event-handler): rename variables to reflect that options object is now a RequestContext

### LINKED ISSUE #4459: Maintenance: Reflect Change From HandlerOptions to RequestContext in REST API Resolvers
### Summary

In #4439 we renamed `HandlerOptions` to `RequestContext` to reflect that it is more of a container for state now rather than configuration. We should update all the tests that use the name `options` or similar when referring to `reqCtx` to make this change clear.

### Why is this needed?

It is confusing for us to refer to a context object as an options object when the latter implies it only contains configuration.

Event Handler

### Solution

Rename any parameters in middleware or route handlers that are of type `RequestContext` from `options` to `reqCtx`:

```ts
// old
const middleware: Middleware = async (params, options, next) => {
  options.res.headers.set('X-Before', 'Before');
  await next();
  options.res.headers.set('X-After', 'After');
}

// new
const middleware: Middleware = async (params, reqCtx, next) => {
  reqCtx.res.headers.set('X-Before', 'Before');
  await next();
  reqCtx.res.headers.set('X-After', 'After');
}
```

### Acknowledgment

### Future readers

svozza:
## Summary
This PR reflects the change of type from `HandlerOptions` to `RequestContext` made in #4439 by updating class variables, JSDocs and the tests.

### Changes

- Update all unit tests that define handlers or middleware to use `reqCtx` instead of `options` as the name for any parameter of type `RequestContext`
- Updated the type definitions `reqCtx` instead of `options` as the name for any parameter of type `RequestContext`
- Updated the JSDocs and examples to reflect this change

**Issue number:** closes #4459

---

### DIFF
--- a/packages/event-handler/src/rest/Router.ts
+++ b/packages/event-handler/src/rest/Router.ts
@@ -79,7 +79,7 @@ class Router {
    * Registers a custom error handler for specific error types.
    *
    * @param errorType - The error constructor(s) to handle
-   * @param handler - The error handler function that returns an ErrorResponse
+   * @param handler - The error handler function that returns a Promise<ErrorResponse>
    */
   public errorHandler<T extends Error>(
     errorType: ErrorConstructor<T> | ErrorConstructor<T>[],
@@ -106,7 +106,7 @@ class Router {
   /**
    * Registers a custom handler for 404 Not Found errors.
    *
-   * @param handler - The error handler function for NotFoundError
+   * @param handler - The error handler function that returns a Promise<ErrorResponse>
    */
   public notFound(handler: ErrorHandler<NotFoundError>): void;
   public notFound(): MethodDecorator;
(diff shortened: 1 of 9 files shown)


sdangol (approved):

packages/event-handler/src/rest/Router.ts
```diff
@@ -79,7 +79,7 @@ class Router {
    * Registers a custom error handler for specific error types.
    *
    * @param errorType - The error constructor(s) to handle
-   * @param handler - The error handler function that returns an ErrorResponse
+   * @param handler - The error handler function that returns a Promise<ErrorResponse>
```
dreamorosi: As a minor nit, I'd avoid explicitly mentioning type names in this area of the jsdocs, primarily because there's a high chance they'll go out of sync if we refactor.

In this case I'd just say `The error handler function that returns a response` or `The error handler that returns an response with error` or similar.

svozza: Yeah, makes sense.

packages/event-handler/src/rest/Router.ts
```diff
@@ -79,7 +79,8 @@ class Router {
    * Registers a custom error handler for specific error types.
    *
    * @param errorType - The error constructor(s) to handle
-   * @param handler - The error handler function that returns an ErrorResponse
+   * @param handler - The error handler that returns an
+   * error response>
```
sdangol: missed a formatting issue here

svozza: Fixed

PR #4477, opened 2025-09-12
feat(event-handler): add CORS middleware support

### LINKED ISSUE #4446: Feature request: Implement CORS middleware for REST API
### Use case

We should have a middleware that adds [CORS headers](https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-cors.html) to responses. The middelware should have sane defaults than can be overridden when initilising the middleware.

### Solution/User Experience

The middleware will be applied like any other:

```ts
import {cors} from from '@aws-lambda-powertools/event-handler/rest/middleware';

// use defaults
app.use(cors());

// custom configuration
app.use(cors({
  allowHeaders: ['X-My-Header'],
  allowMethods: ['GET', 'POST', 'OPTIONS'],
  credentials: true,
  exposeHeaders: ['Content-Length'],
  maxAge: 300,
  origin: 'http://mysite.com',
}));
```

As per the Python implementation the defaults are:

```ts
{
  origin: '*',
  allowHeaders: ['Authorization', 'Content-Type', 'X-Amz-Date', 'X-Api-Key', 'X-Amz-Security-Token'],
  allowMethods: ['DELETE', 'GET', 'HEAD', 'PATCH', 'POST', 'PUT'  ],
  exposeHeaders: [],
  credentials: false,
}
```

Pre-flight (OPTIONS) calls are typically handled at the API Gateway or Lambda Function URL so no Lambda integration is necessary. However, ALB expects you to handle pre-flight requests so our implementation should still check for the `OPTIONS` verb and short-circuit so that we can future proof. This involves returning a Web `Response` with the necessary headers and a HTTP status code of 204.

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

dcabib:
## Description

This PR implements CORS middleware support for the Event Handler as requested in issue #4446.

## Implementation

- CORS middleware with configurable options
- OPTIONS preflight handling with 204 status
- string and array based origin validation

## Configuration Options

The middleware supports all standard CORS options:
- `origin`: string | string[]
- `methods`: string[]
- `allowedHeaders`: string[]
- `exposedHeaders`: string[]
- `credentials`: boolean
- `maxAge`: number

## Usage

```typescript
// Global CORS
router.use(cors({ origin: '*' }));

// Route-specific CORS  
router.get('/api/data', [cors({ origin: 'https://myapp.com' })], handler);
```

## Acknowledgment

Closes #4446

### DIFF
--- a/packages/event-handler/tests/unit/rest/middleware/cors.test.ts
+++ b/packages/event-handler/tests/unit/rest/middleware/cors.test.ts
@@ -32,11 +32,14 @@ describe('CORS Middleware', () => {
   });
 
   it('uses default configuration when no options are provided', async () => {
+    // Prepare
     const corsHeaders: { [key: string]: string } = {};
     app.get('/test', [createHeaderCheckMiddleware(corsHeaders)], async () => ({ success: true }));
 
+    // Act
     const result = await app.resolve(getRequestEvent, context);
 
+    // Assess
     expect(result.headers?.['access-control-allow-origin']).toEqual(DEFAULT_CORS_OPTIONS.origin);
     expect(result.multiValueHeaders?.['access-control-allow-methods']).toEqual(DEFAULT_CORS_OPTIONS.allowMethods);
     expect(result.multiValueHeaders?.['access-control-allow-headers']).toEqual(DEFAULT_CORS_OPTIONS.allowHeaders);
@@ -45,12 +48,15 @@ describe('CORS Middleware', () => {
   });
 
   it('merges user options with defaults', async () => {
+    // Prepare
     const corsHeaders: { [key: string]: string } = {};
-    const customApp = new Router();
-    customApp.get('/test', [cors(customCorsOptions), createHeaderCheckMiddleware(corsHeaders)], async () => ({ success: true }));
+    const app = new Router();
+    app.get('/test', [cors(customCorsOptions), createHeaderCheckMiddleware(corsHeaders)], async () => ({ success: true }));
 
-    const result = await customApp.resolve(getRequestEvent, context);
+    // Act
+    const result = await app.resolve(getRequestEvent, context);
 
+    // Assess
     expect(result.headers?.['access-control-allow-origin']).toEqual('https://example.com');
     expect(result.multiValueHeaders?.['access-control-allow-methods']).toEqual(['GET', 'POST']);
     expect(result.multiValueHeaders?.['access-control-allow-headers']).toEqual(['Authorization', 'Content-Type']);
@@ -69,28 +75,37 @@ describe('CORS Middleware', () => {
     ['matching', 'https://app.com', 'https://app.com'],
     ['non-matching', 'https://non-matching.com', '']
   ])('handles array origin with %s request', async (_, origin, expected) => {
-    const customApp = new Router();
-    customApp.get('/test', [cors({ origin: ['https://app.com', 'https://admin.app.com'] })], async () => ({ success: true }));
+    // Prepare
+    const app = new Router();
+    app.get('/test', [cors({ origin: ['https://app.com', 'https://admin.app.com'] })], async () => ({ success: true }));
 
-    const result = await customApp.resolve(createTestEvent('/test', 'GET', { 'Origin': origin }), context);
+    // Act
+    const result = await app.resolve(createTestEvent('/test', 'GET', { 'Origin': origin }), context);
 
+    // Assess
     expect(result.headers?.['access-control-allow-origin']).toEqual(expected);
   });
 
   it('handles OPTIONS preflight requests', async () => {
+    // Prepare
     app.options('/test', async () => ({ foo: 'bar' }));
 
+    // Act
     const result = await app.resolve(createTestEvent('/test', 'OPTIONS', { 'Access-Control-Request-Method': 'GET' }), context);
 
+    // Assess
     expect(result.statusCode).toBe(204);
   });
 
   it('calls the next middleware if the Access-Control-Request-Method is not present', async () => {
+    // Prepare
     const corsHeaders: { [key: string]: string } = {};
     app.options('/test', [createHeaderCheckMiddleware(corsHeaders)], async () => ({ success: true }));
 
+    // Act
     await app.resolve(optionsRequestEvent, context);
 
+    // Assess
     expect(corsHeaders).toMatchObject(expectedDefaultHeaders);
   });
 });


examples/snippets/event-handler/rest/cors_dynamic_origin.ts
```diff
@@ -0,0 +1,40 @@
...
+    console.log('CORS request from:', origin);
+    
+    // Return boolean: true allows the origin, false denies it
+    return origin ? allowedOrigins.includes(origin) : false;
+  },
+  credentials: true,
+  allowHeaders: ['Content-Type', 'Authorization'],
+}));
+
+// Route-specific CORS for public API
```
svozza: This will get over-written. Please study the the order of exectuion of middleware in the docs:

https://docs.powertools.aws.dev/lambda/typescript/latest/features/event-handler/rest/#middleware

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
...
+   * The Access-Control-Max-Age header value in seconds.
+   * Only applicable for preflight requests.
+   */
+  maxAge?: number;
+}
+
+/**
+ * Resolved CORS configuration with all defaults applied
+ */
+interface ResolvedCorsConfig {
```
svozza: Any particular reason to favour an `interface` over a `type` here?

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
...
+  
+  if (Array.isArray(originConfig)) {
+    return origin && originConfig.includes(origin) ? origin : '';
+  }
+  
+  if (typeof originConfig === 'string') {
+    return originConfig;
+  }
+  
+  return DEFAULT_CORS_OPTIONS.origin as string;
```
svozza: Why do we need this cast?

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
...
+
+/**
+ * Handles preflight OPTIONS requests
+ */
+function handlePreflight(config: ResolvedCorsConfig, reqCtx: RequestContext): Response {
+  const { request } = reqCtx;
+  const requestOrigin = request.headers.get('Origin');
+  const resolvedOrigin = resolveOrigin(config.origin, requestOrigin, reqCtx);
+
+  const headers = new Headers();
```
svozza: We should be mutating the headers object on `res` here because we have no idea if any other middleware has ran before us.

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
...
+export const cors = (options: CorsOptions = {}): Middleware => {
+  const config = resolveConfiguration(options);
+  
+  return async (_params: Record<string, string>, reqCtx: RequestContext, next: () => Promise<HandlerResponse | void>) => {
+    const { request } = reqCtx;
+    const method = request.method.toUpperCase();
+    
+    // Handle preflight OPTIONS request
+    if (method === HttpVerbs.OPTIONS) {
+      return handlePreflight(config, reqCtx);
```
svozza: If the `Access-Control-Request-Method` header is not present, we need to call `next` as the `OPTIONS` request is not a preflight request.

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
...
+    }
+    
+    // Continue to next middleware/handler first
+    await next();
```
svozza, line 96: CORS is one of the first things that needs to be done, this call to `next` needs to be called after the headers are set.

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
...
+    
+    // Add CORS headers to the response after handler
+    addCorsHeaders(config, reqCtx);
```
svozza: I don't think we need this extra layer of indirection by pulling this into a function, let's inline it.

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
...
+    status: HttpErrorCodes.NO_CONTENT, // 204
+    headers,
+  });
+}
+
+/**
+ * Adds CORS headers to regular requests
+ */
+function addCorsHeaders(config: ResolvedCorsConfig, reqCtx: RequestContext): void {
+  const { request, res } = reqCtx;
```
svozza: Please see the note about desctructuring the `res` object: https://docs.powertools.aws.dev/lambda/typescript/latest/features/event-handler/rest/#avoiding-destructuring-pitfalls

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
...
+  const requestOrigin = request.headers.get('Origin');
+  const resolvedOrigin = resolveOrigin(config.origin, requestOrigin, reqCtx);
+
+  if (resolvedOrigin) {
+    res.headers.set('Access-Control-Allow-Origin', resolvedOrigin);
+  }
+  
+  if (config.exposeHeaders.length > 0) {
```
svozza: We can loop through `exposeHeaders` and use the `Headers.append` method.

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
...
+    headers,
+  });
+}
+
+/**
+ * Adds CORS headers to regular requests
+ */
+function addCorsHeaders(config: ResolvedCorsConfig, reqCtx: RequestContext): void {
+  const { request, res } = reqCtx;
+  const requestOrigin = request.headers.get('Origin');
```
svozza: We need to set the Vary header, see more info here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Allow-Origin#cors_and_caching

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
...
+  const resolvedOrigin = resolveOrigin(config.origin, requestOrigin, reqCtx);
+
+  const headers = new Headers();
+  
+  if (resolvedOrigin) {
+    headers.set('Access-Control-Allow-Origin', resolvedOrigin);
+  }
+  
+  if (config.allowMethods.length > 0) {
+    headers.set('Access-Control-Allow-Methods', config.allowMethods.join(', '));
```
svozza: We can loop through exposeHeaders and use the `headers.append` method.

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
...
+  }
+  
+  if (config.allowHeaders.length > 0) {
```
svozza: We can loop through `exposeHeaders` and use the `headers.append` method.

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
...
+ *   }
+ * }));
+ * ```
+ */
+export const cors = (options: CorsOptions = {}): Middleware => {
+  const config = resolveConfiguration(options);
+  
+  return async (_params: Record<string, string>, reqCtx: RequestContext, next: () => Promise<HandlerResponse | void>) => {
+    const { request } = reqCtx;
+    const method = request.method.toUpperCase();
```
svozza: We can remove `toUpperCase` because method is always capitalised on a `Request` object.

packages/event-handler/tests/unit/rest/Router/cors-integration.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    const app = new Router();
+    
+    // Add CORS middleware using working pattern
+    app.use(cors());
+    
+    app.get('/test', async () => ({ success: true }));
+
+    const result = await app.resolve(createTestEvent('/test', 'GET'), mockContext);
+    
+    // Use exact pattern from working middleware tests
```
svozza: Let's remove this comment

packages/event-handler/tests/unit/rest/Router/cors-integration.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  logStreamName: '2021/01/01/[$LATEST]test-stream',
+  getRemainingTimeInMillis: () => 30000,
+  done: () => {},
+  fail: () => {},
+  succeed: () => {},
+};
+
+describe('Router CORS Integration', () => {
+  it('should apply CORS headers with default configuration', async () => {
+    const app = new Router();
```
svozza: Please follow the `// Prepare`, `// Act`, `// Assess` comment pattern that we use in all the other tests.

packages/event-handler/tests/unit/rest/Router/cors-integration.test.ts
```diff
@@ -0,0 +1,83 @@
...
+      body: JSON.stringify({ success: true }),
+      headers: {
+        'content-type': 'application/json',
+        'access-control-allow-origin': '*',
+      },
+      isBase64Encoded: false,
+    });
+  });
+
+  it('should handle OPTIONS preflight without route registration', async () => {
```
svozza: This behaviour is already tested elsewhere.

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
```diff
@@ -0,0 +1,402 @@
...
+  logGroupName: '/aws/lambda/test-function',
+  logStreamName: '2021/01/01/[$LATEST]test-stream',
+  getRemainingTimeInMillis: () => 30000,
+  done: () => {},
+  fail: () => {},
+  succeed: () => {},
+};
+
+// Test utilities
+function createMockAPIGatewayEvent(
```
svozza: There is already a uitility function that does this in the `packages/event-handler/tests/unit/rest/helpers.ts` file.

packages/event-handler/tests/unit/rest/Router/cors-integration.test.ts
```diff
@@ -0,0 +1,83 @@
...
+ * @group unit/event-handler/rest/router/cors-integration
+ */
+
+import { describe, expect, it } from 'vitest';
+import { Router } from '../../../../src/rest/Router.js';
+import { cors } from '../../../../src/rest/middleware/cors.js';
+import { createTestEvent } from '../helpers.js';
+
+// Mock context - simplified
+const mockContext = {
```
svozza: Please use the mock context library:

```ts
```

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
```diff
@@ -0,0 +1,402 @@
...
+ */
+
+import type { APIGatewayProxyEvent, Context } from 'aws-lambda';
+import { describe, expect, it, vi } from 'vitest';
+import type { RequestContext } from '../../../../src/types/rest.js';
+import { cors } from '../../../../src/rest/middleware/cors.js';
+import type { CorsOptions } from '../../../../src/rest/middleware/cors.js';
+
+// Mock context
+const mockContext: Context = {
```
svozza: As above:

As above, use the mock:

```ts
```

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
```diff
@@ -0,0 +1,402 @@
...
+  return {
+    request,
+    res: response,
+    event: createMockAPIGatewayEvent(method, origin ? { Origin: origin } : {}),
+    context: mockContext,
+  };
+}
+
+describe('CORS Configuration', () => {
+  it('should use default configuration when no options provided', async () => {
```
svozza: Please declarative test names such as `it('uses default configuration when no options provided'` rather than the 'should' construction.

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
```diff
@@ -0,0 +1,402 @@
...
+    const reqCtx = createMockRequestContext('OPTIONS', 'https://example.com');
+    const next = vi.fn().mockResolvedValue(undefined);
+
+    const result = (await middleware({}, reqCtx, next)) as Response;
+
+    expect(result.headers.get('Access-Control-Allow-Origin')).toBe('https://example.com');
+    expect(next).not.toHaveBeenCalled();
+  });
+
+  it('should include Access-Control-Allow-Methods header in preflight', async () => {
```
svozza: We can check all these headers in one test

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
```diff
@@ -0,0 +1,402 @@
...
+    const reqCtx = createMockRequestContext('GET');
+    const next = vi.fn().mockResolvedValue(undefined);
+
+    await middleware({}, reqCtx, next);
+
+    expect(reqCtx.res.headers.get('Access-Control-Allow-Origin')).toBe('*');
+    expect(next).toHaveBeenCalledTimes(1);
+  });
+
+  it('should call next() for non-OPTIONS requests', async () => {
```
svozza: This is already tested multiple times, we can remove

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
```diff
@@ -0,0 +1,402 @@
...
+    event: createMockAPIGatewayEvent(method, origin ? { Origin: origin } : {}),
+    context: mockContext,
+  };
+}
+
+describe('CORS Configuration', () => {
+  it('should use default configuration when no options provided', async () => {
+    const middleware = cors();
+    const reqCtx = createMockRequestContext('GET');
+    const next = vi.fn().mockResolvedValue(undefined);
```
svozza: Let's just define this once and then have a `befeEach` block that resets the mocks.

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
```diff
@@ -0,0 +1,402 @@
...
+
+    await middleware({}, reqCtx, next);
+
+    expect(reqCtx.res.headers.get('Access-Control-Expose-Headers')).toBe('X-Total-Count');
+    expect(next).toHaveBeenCalledTimes(1);
+  });
+});
+
+describe('Python Implementation Parity', () => {
```
svozza: We don't need this, I think, it's alredy tested in the default options test.

packages/event-handler/tests/unit/rest/Router/cors-integration.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  done: () => {},
+  fail: () => {},
+  succeed: () => {},
+};
+
+describe('Router CORS Integration', () => {
+  it('should apply CORS headers with default configuration', async () => {
+    const app = new Router();
+    
+    // Add CORS middleware using working pattern
```
svozza: Let's remove this comment

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
+import type { Middleware, RequestContext, HandlerResponse } from '../../types/rest.js';
```
svozza: Let's put the CORS code in it's own directory: `packages/event-handler/src/rest/middleware/cors/index.ts

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
...
+import { HttpErrorCodes, HttpVerbs } from '../constants.js';
+
+/**
+ * Configuration options for CORS middleware
+ */
+export interface CorsOptions {
```
svozza: Any reason we're using an interface here rather than a type?

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
...
+  allowMethods: ['DELETE', 'GET', 'HEAD', 'PATCH', 'POST', 'PUT'],
+  allowHeaders: ['Authorization', 'Content-Type', 'X-Amz-Date', 'X-Api-Key', 'X-Amz-Security-Token'],
+  exposeHeaders: [],
+  credentials: false,
+};
+
+/**
+ * Resolves and validates the CORS configuration
+ */
+function resolveConfiguration(userOptions: CorsOptions): ResolvedCorsConfig {
```
svozza: Do we actually need this function. Can we not use the spread operator or similar in the factory function. Something like:

```ts
const resolveOptions = {
  ...DEFAULT_CORS_OPTIONS,
  ...options
}
```

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
...
+function resolveOrigin(
+  originConfig: CorsOptions['origin'],
+  requestOrigin: string | null | undefined,
+  reqCtx: RequestContext
+): string {
+  const origin = requestOrigin || undefined;
+  
+  if (typeof originConfig === 'function') {
+    const result = originConfig(origin, reqCtx);
+    if (typeof result === 'boolean') {
```
svozza: This section feels like a code smell to me. What are we actually doing here and why are we doing these type checks and conditional logic? 

I'm also not sure about accepting a fucntion in the first place, I know Koa and Hono offer this but it adds complexity and I'm not sure how much value it adds. Thoughts @dreamorosi?

svozza (changes requested):

packages/event-handler/src/rest/middleware/cors.ts
dreamorosi: Yea, let's just keep `string | string[]` as input for now.

packages/event-handler/tests/unit/rest/Router/cors-integration.test.ts
```diff
@@ -0,0 +1,83 @@
+/**
+ * Integration tests for CORS middleware with Router
+ *
+ * @group unit/event-handler/rest/router/cors-integration
+ */
```
dreamorosi: These comments don't belong in this project, this is a feature from `jest-runner-groups` which we don't use.

packages/event-handler/tests/unit/rest/Router/cors-integration.test.ts
dreamorosi, line 1: I'd remove this entire file and keep only one test file.

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
```diff
@@ -0,0 +1,402 @@
...
+    await middleware({}, reqCtx, next);
+
+    expect(reqCtx.res.headers.get('Access-Control-Expose-Headers')).toBe(
+      'Content-Length, X-Custom-Header'
+    );
+    expect(next).toHaveBeenCalledTimes(1);
+  });
+});
+
+describe('Origin Resolution', () => {
```
dreamorosi: Please don't nest or add multiple `describe` statements - just keep one at the top and all the `it` cases within at the top level.

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
dreamorosi: +1, there's a lot of repetition here.

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
dreamorosi, line 1: Test cases should follow the style and conventions you see across the codebase.

Besides the comments on code duplication and test header wording left below, each test case should have defined section delimited by comments `// Prepare`, `// Act`, `// Assess` that contain the respective stages of the test.

packages/event-handler/src/rest/middleware/index.ts
```diff
@@ -0,0 +1,2 @@
+export { cors } from './cors.js';
+export type { CorsOptions } from './cors.js';
```
dreamorosi: We don't export runtime modules and type in the same file anywhere in the rest of the project, this type export doesn't belong here, please remove it and place it in the appropriate file.

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
...
+ *     const allowedOrigins = ['https://app.com', 'https://admin.app.com'];
+ *     return origin && allowedOrigins.includes(origin);
+ *   }
+ * }));
+ * ```
+ */
+export const cors = (options: CorsOptions = {}): Middleware => {
+  const config = resolveConfiguration(options);
+  
+  return async (_params: Record<string, string>, reqCtx: RequestContext, next: () => Promise<HandlerResponse | void>) => {
```
dreamorosi: The types in this function signature might be redundant or removable by using `satisfies Middleware` at the end of the statement.

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
...
+  if (config.credentials) {
+    res.headers.set('Access-Control-Allow-Credentials', 'true');
+  }
+}
+
+/**
+ * Creates a CORS middleware that adds appropriate CORS headers to responses
+ * and handles OPTIONS preflight requests.
+ * 
+ * @param options - CORS configuration options
```
dreamorosi: Please look at other functions and see how we document `@param` and where.

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
...
+ * @returns A middleware function that handles CORS
```
dreamorosi: Please remove this line.

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
...
+  return new Response(null, {
+    status: HttpErrorCodes.NO_CONTENT, // 204
+    headers,
+  });
+}
+
+/**
+ * Adds CORS headers to regular requests
+ */
+function addCorsHeaders(config: ResolvedCorsConfig, reqCtx: RequestContext): void {
```
dreamorosi: Please use arrow functions wherever possible.

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,219 @@
...
+  allowHeaders: string[];
+  exposeHeaders: string[];
+  credentials: boolean;
+  maxAge?: number;
+}
+
+/**
+ * Default CORS configuration matching Python implementation
+ */
+const DEFAULT_CORS_OPTIONS: Required<Omit<CorsOptions, 'maxAge'>> = {
```
dreamorosi: This should be exported in the `constants.ts` file.

packages/event-handler/src/rest/middleware/cors.ts
dreamorosi: Also we don't define types in the same file of runtime logic anywhere else in the project, this should be extracted in the appropriate file.

packages/event-handler/src/rest/middleware/cors.ts
dreamorosi: We don't, let's use what Stefano suggested above.

packages/event-handler/src/rest/middleware/cors.ts
dreamorosi: I think once we extract the types and constants to where they belong (aka dedicated files) and remove/change the helpers according to the comments we left under them, there will be very little left in this file, basically just the `cors` function and maybe at most one extra helper.

With this in mind, I think having a single file is enough.

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
```diff
@@ -0,0 +1,402 @@
+/**
+ * Test for CORS middleware
+ *
+ * @group unit/event-handler/rest/middleware/cors
+ */
```
dreamorosi: Same as the other file touched, this comment doesn't belong in this project.

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
dreamorosi: We don't need this section.

dreamorosi (changes requested):
Thank you for the PR.

There are still a few gaps to address before we can merge. In particular:
- The module’s execution flow and middleware order need to be aligned with our conventions (e.g., destructuring of `reqCtx`, middleware ordering, etc.).
- Please run the linter and fix notices/warnings/errors. Import order is unsorted in several files.
- Please follow project conventions and place constants, types, and modules in appropriate files; follow test style and conventions
- Remove the example files from this PR; we’ll add examples in a separate documentation-focused PR.

Meta/process note:
I’ve noticed several recent PRs landing with red CI (linting and tests). To keep reviews efficient, please ensure before requesting review that:
-  All tests pass locally and in CI.
-  Linting is clean (no warnings/errors).
-  Pre-commit hooks run successfully.

If CI is red, we’ll pause reviews until it’s green. If this remains an issue across PRs, we may close the PR and handle the changes internally to keep momentum.

If you’re running into setup issues, let us know - we’re happy to help you get the local environment/hooks configured.

dreamorosi:
Hi @dcabib, we would like to merge this today since we are on a tight schedule and we have other PRs in the pipeline that will impact areas of the code changed in this one.

Let us know if you're able to address the feedback or have questions. 

Thank you!

packages/event-handler/src/rest/middleware/cors.ts
dcabib: Changed to type to maintain consistency with the project conventions. The type is now properly defined in types/rest.ts.

packages/event-handler/src/rest/middleware/cors.ts
dcabib: I've removed the cast and properly handled the readonly arrays by spreading them

packages/event-handler/src/rest/middleware/cors.ts
dcabib: Fixed! Now mutating the existing res.headers object instead of creating new Headers. This ensures proper middleware composition.

packages/event-handler/src/rest/middleware/index.ts
dcabib: Removed the type export from middleware/index.ts. Types are now properly exported from types/index.ts

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
dcabib: Updated all test cases with // Prepare, // Act, // Assess sections following project conventions.

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
dcabib: Removed the 'Python Implementation Parity' test section as requested

dcabib:
Hey @svozza @dreamorosi! 👋 

Just pushed all the requested changes. Man, you guys caught some good stuff - totally missed that I was creating new Headers instead of mutating the existing ones 🤦‍♂️ 

Here's what I fixed:
- Moved all the types and constants where they belong (no more mixing runtime with types!)
- Fixed that middleware to properly mutate `res.headers` 
- Cleaned up all the tests with the Prepare/Act/Assess pattern
- Nuked those example files - they'll live in the docs PR
- Arrow functions everywhere now 🏹
- Linting is squeaky clean ✨

All 26 CORS tests are passing! The other test failures are from that missing testing-utils dependency that was already broken on main.

Build works great, both ESM and CJS are happy. Should be good to merge now!

Let me know if you spot anything else - ready to fix whatever's needed to get this in today 🚀

sdangol:
Hi @dcabib, there were some feedbacks remaining and some missing test coverage which I've now addressed due to our schedule.

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
```diff
@@ -1,480 +1,195 @@
...
-      };
-      const middleware = cors(options);
-      const reqCtx = createMockRequestContext('GET');
-      const next = vi.fn().mockResolvedValue(undefined);
+  it('uses default configuration when no options are provided', async () => {
+    // Prepare
+    const executionOrder: string[] = [];
+    app.get(
+      '/test',
+      [createTrackingMiddleware('middleware1', executionOrder)],
```
svozza: I don't think we need this tracking middelware, this stuff is not related to CORS and is tested elsewhere. 

What we might want though is to make sure that the CORS headers are added before `next` is called. That can be done by having a middleware added after the CORs one that inspects `res.headers` before it (i.e., the second middleware) calls `next` and verfies they are there.

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
```diff
@@ -1,480 +1,195 @@
...
+      [
+        cors({
+          origin: 'https://example.com',
+          allowMethods: ['GET', 'POST'],
+          allowHeaders: ['Authorization', 'Content-Type'],
+          credentials: true,
+          exposeHeaders: ['Authorization', 'X-Custom-Header'],
+          maxAge: 86400,
+        }),
+        createTrackingMiddleware('middleware1', executionOrder)
```
svozza: Again, same as above with the tracking middleware.

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
```diff
@@ -1,480 +1,195 @@
...
-  it('verifies default configuration values', async () => {
+  it('calls the next middleware if the Access-Control-Request-Method is not present', async () => {
     // Prepare
-    const middleware = cors();
-    const reqCtx = createMockRequestContext('OPTIONS');
-    const next = vi.fn().mockResolvedValue(undefined);
+    const executionOrder: string[] = [];
+    app.options(
+      '/test',
+      [createTrackingMiddleware('middleware1', executionOrder)],
```
svozza: Again, same as above with the tracking middleware.

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
```diff
@@ -1,480 +1,195 @@
...
-
-    it('handles missing request origin header', async () => {
-      // Prepare
-      const middleware = cors();
-      const reqCtx = createMockRequestContext('GET');
-      const next = vi.fn().mockResolvedValue(undefined);
+  it('merges user options with defaults', async () => {
+    // Prepare
+    const executionOrder: string[] = [];
+    const application = new Router();
```
svozza: We can still call this `app` because block scope takes prescendence or do you want to make it clear that it's different?

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -168,28 +50,51 @@ const addCorsHeaders = (
...
+    if (reqCtx.request.method === HttpVerbs.OPTIONS && reqCtx.request.headers.has('Access-Control-Request-Method')) {
+      return new Response(null, {
+        status: HttpErrorCodes.NO_CONTENT,
+        headers: reqCtx.res.headers,
+      });
     }
-    
-    // Continue to next middleware/handler first
+
+    // Continue to next middleware/handler
```
svozza: I don't think we need this comment

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
sdangol: I had used it for checking if the next() was called or not. I wasn't sure if I could spy if like before with this testing approach.

svozza: If `next` isn't called by a middleware then the handler won't be called either so you'll get a 500 error back. It's enough to check the status code for 200 to verfiy that.

sdangol: I've not implemented the header checking middleware as you suggested. Not sure if that is what you meant?

packages/event-handler/src/rest/middleware/cors.ts
```diff
@@ -0,0 +1,98 @@
...
+      reqCtx.res.headers.append('access-control-expose-headers', header);
+    });
+    reqCtx.res.headers.set('access-control-allow-credentials', config.credentials.toString());
+    if (config.maxAge !== undefined) {
+      reqCtx.res.headers.set('access-control-max-age', config.maxAge.toString());
+    }
+
+    // Handle preflight OPTIONS request
+    if (reqCtx.request.method === HttpVerbs.OPTIONS && reqCtx.request.headers.has('Access-Control-Request-Method')) {
+      return new Response(null, {
```
dreamorosi, line 91: What happens if the casing of the header is different?

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
dreamorosi, line 1: Forgot to document tests with `// Prepare`, `// Act`, and `// Assess` comments.

packages/event-handler/tests/unit/rest/middleware/cors.test.ts
svozza: I meant more as a one off test but what you have there works too.

packages/event-handler/src/rest/middleware/cors.ts
svozza: The `has` method in the `Headers` object is case-insenitive.

dreamorosi: TIL!

PR #4484, opened 2025-09-15
fix(batch): declare the @aws-lambda-powertools/commons dependency

### LINKED ISSUE #4485: Bug: @aws-lambda-powertools/commons is not declared in batch package
### Expected Behavior

Should be able to resolve `@aws-lambda-powertools/commons/utils/env` in the `batch` package.

### Current Behavior

✘ [ERROR] Could not resolve "@aws-lambda-powertools/commons/utils/env"

node_modules/.pnpm/@aws-lambda-powertools+batch@2.26.0/node_modules/@aws-lambda-powertools/batch/lib/esm/BasePartialBatchProcessor.js:1:33:
1 │ ...getStringFromEnv } from '@aws-lambda-powertools/commons/utils/env';

### Code snippet

N/A

### Steps to Reproduce

1. Upgrade to 2.26.0
2. Try to bundle this version with esbuild

### Possible Solution

_No response_

latest

22.x

### Packaging format used

npm

### Execution logs

```Shell

```

Grmiade:
## Summary

The `@aws-lambda-powertools/commons` dependency was not declared in the `batch` package.
It seems that this regression has been introduced with https://github.com/aws-powertools/powertools-lambda-typescript/pull/4465.
See https://github.com/aws-powertools/powertools-lambda-typescript/pull/4465/files#diff-179ff02baca1caba368814f8214b5dbdf33c3e9bf7cebe51014d0dfb54c59dadR1

### Changes

- We declare `@aws-lambda-powertools/commons` as dependency for `batch`
- We use the `commons` package instead of `../../commons/lib/esm/types/GenericLogger.js` imports.
- We delete the useless `BatchProcessor` import in types.

fixes #4485

---


packages/batch/src/types.ts
```diff
@@ -7,9 +8,7 @@ import type {
   StreamRecord,
 } from 'aws-lambda';
 import type { ZodType } from 'zod';
-import type { GenericLogger } from '../../commons/lib/esm/types/GenericLogger.js';
 import type { BasePartialBatchProcessor } from './BasePartialBatchProcessor.js';
-import type { BatchProcessor } from './BatchProcessor.js';
```
sdangol, line 12: @Grmiade This type is used in the docstring https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/batch/src/types.ts#L99 for reference to the class

sdangol (changes requested):
Thank you for the changes @Grmiade. Just a small comment.

packages/batch/src/types.ts
dreamorosi: And it's stripped when running `tsc build` - good catch.

@sdangol let's make a commit to put it back and move on with the release, I'll approve again the PR afterwards.

Grmiade: Hi @sdangol 👋 Normally we don't need to import the reference to make a link works in JSDoc.
Moreover, the syntax "{@link BatchProcessor | `BatchProcessor`}" seems wrong, it should be `{@link namepath}` or `{@link namepath|Link Text}`. I can fix that.

dreamorosi: Hi, really we need the import or the link will be broken in the IDE - see screenshot below:

Screenshot 2025-09-15 at 12 25 33

Regarding the backtick usage, that's an intentional style choice that we use to get symbols to be rendered with monospaced font.

Below is an example of how it looks with both the imported type and backtick:

Screenshot 2025-09-15 at 12 28 07

---

Having this type imported doesn't have any performance impact in production since TypeScript types are stripped away. We also have linting rules that flag actual unreferenced/unused types.

We'll revert the change and merge the PR.

PR #4495, opened 2025-09-15
feat(event-handler): added compress middleware for the REST API event handler

### LINKED ISSUE #4474: Feature request: Add compress middleware for Event Handler REST API
### Use case

We should have a middleware that compresses response bodies to reduce payload size and improve performance. The middleware should support common compression algorithms with sensible defaults that can be overridden when initializing the middleware.

### Solution/User Experience

The middleware will be applied like any other:

```ts
import { compress } from '@aws-lambda-powertools/event-handler/rest/middleware';

// use defaults
app.use(compress());

type options = {
  encoding?: 'gzip' | 'deflate'
  threshold?: number
}

// custom configuration
app.use(compress({
  encoding: 'gzip',
  threshold: 1024, // this will also be the default
}));
```

The middleware should:

- Only compress responses above the threshold size
- If `enocding` option is not present, infer from `Accept-Encoding` header with preference given to `gzip`
- Avoid loading the whole request into memory, use streams instead
  - Should we use the inbuilt [Nodejs compression](https://nodejs.org/api/zlib.html#class-zlibgunzip) or [CompressionStream](https://developer.mozilla.org/en-US/docs/Web/API/Compression_Streams_API). Explain your reasoning.
- Set appropriate Content-Encoding header
- Remove Content-Length header (since compressed size differs)
- Support gzip and deflate algorithms
- Check Accept-Encoding header to ensure client supports compression
- Skip compression for already compressed content types (images, videos, etc.)

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

sdangol:
## Summary

This PR adds a built in compress middleware which can be used along with the event handler for providing an easy way to compress response bodies to reduce payload size and improve performance.

### Changes

- Added a `middleware` directory in the rest directory to contain the built-in middlewares
- Exported the middleware path to be used
- Created the compress middleware which takes an encoding and a threshold and does the compression if needed

**Issue number:** closes #4474 

---

### DIFF
--- a/packages/event-handler/package.json
+++ b/packages/event-handler/package.json
@@ -77,6 +77,16 @@
         "types": "./lib/esm/rest/index.d.ts",
         "default": "./lib/esm/rest/index.js"
       }
+    },
+    "./experimental-rest/middleware": {
+      "require": {
+        "types": "./lib/cjs/rest/middleware/index.d.ts",
+        "default": "./lib/cjs/rest/middleware/index.js"
+      },
+      "import": {
+        "types": "./lib/esm/rest/middleware/index.d.ts",
+        "default": "./lib/esm/rest/middleware/index.js"
+      }
     }
   },
   "typesVersions": {
--- a/packages/event-handler/src/rest/middleware/compress.ts
+++ b/packages/event-handler/src/rest/middleware/compress.ts
@@ -21,7 +21,8 @@ const compress = (options?: CompressionOptions): Middleware => {
       reqCtx.request.method === 'HEAD' || // HEAD request
       (contentLength && Number(contentLength) < threshold) || // content-length below threshold
       !shouldCompress(reqCtx.res) || // not compressible type
-      !shouldTransform(reqCtx.res) // cache-control: no-transform
+      !shouldTransform(reqCtx.res) || // cache-control: no-transform
+      !reqCtx.res.body
     ) {
       return;
     }
@@ -33,9 +34,6 @@ const compress = (options?: CompressionOptions): Middleware => {
         acceptedEncoding?.includes(encoding)
       ) ??
       COMPRESSION_ENCODING_TYPES.GZIP;
-    if (!encoding || !reqCtx.res.body) {
-      return;
-    }
 
     // Compress the response
     const stream = new CompressionStream(encoding);
--- a/packages/event-handler/src/rest/middleware/compress.ts
+++ b/packages/event-handler/src/rest/middleware/compress.ts
@@ -1,0 +1,60 @@
+import type { Middleware } from '@aws-lambda-powertools/event-handler/types';
+import type { CompressionOptions } from 'src/types/rest.js';
+import {
+  CACHE_CONTROL_NO_TRANSFORM_REGEX,
+  COMPRESSIBLE_CONTENT_TYPE_REGEX,
+  COMPRESSION_ENCODING_TYPES,
+} from '../constants.js';
+
+const compress = (options?: CompressionOptions): Middleware => {
+  const threshold = options?.threshold ?? 1024;
+
+  return async (_, reqCtx, next) => {
+    await next();
+
+    const contentLength = reqCtx.res.headers.get('Content-Length');
+
+    // Check if response should be compressed
+    if (
+      reqCtx.res.headers.has('Content-Encoding') || // already encoded
+      reqCtx.res.headers.has('Transfer-Encoding') || // already encoded or chunked
+      reqCtx.request.method === 'HEAD' || // HEAD request
+      (contentLength && Number(contentLength) < threshold) || // content-length below threshold
+      !shouldCompress(reqCtx.res) || // not compressible type
+      !shouldTransform(reqCtx.res) // cache-control: no-transform
+    ) {
+      return;
+    }
+
+    const acceptedEncoding = reqCtx.request.headers.get('Accept-Encoding');
+    const encoding =
+      options?.encoding ??
+      Object.values(COMPRESSION_ENCODING_TYPES).find((encoding) =>
+        acceptedEncoding?.includes(encoding)
+      ) ??
+      COMPRESSION_ENCODING_TYPES.GZIP;
+    if (!encoding || !reqCtx.res.body) {
+      return;
+    }
+
+    // Compress the response
+    const stream = new CompressionStream(encoding);
+    reqCtx.res = new Response(reqCtx.res.body.pipeThrough(stream), reqCtx.res);
+    reqCtx.res.headers.delete('Content-Length');
+    reqCtx.res.headers.set('Content-Encoding', encoding);
+  };
+};
+
+const shouldCompress = (res: Response) => {
+  const type = res.headers.get('Content-Type');
+  return type && COMPRESSIBLE_CONTENT_TYPE_REGEX.test(type);
+};
+
+const shouldTransform = (res: Response) => {
+  const cacheControl = res.headers.get('Cache-Control');
+  // Don't compress for Cache-Control: no-transform
+  // https://tools.ietf.org/html/rfc7234#section-5.2.2.4
+  return !cacheControl || !CACHE_CONTROL_NO_TRANSFORM_REGEX.test(cacheControl);
+};
+
+export { compress };
--- a/packages/event-handler/src/rest/middleware/compress.ts
+++ b/packages/event-handler/src/rest/middleware/compress.ts
@@ -12,12 +12,12 @@ const compress = (options?: CompressionOptions): Middleware => {
   return async (_, reqCtx, next) => {
     await next();
 
-    const contentLength = reqCtx.res.headers.get('Content-Length');
+    const contentLength = reqCtx.res.headers.get('content-length');
 
     // Check if response should be compressed
     if (
-      reqCtx.res.headers.has('Content-Encoding') || // already encoded
-      reqCtx.res.headers.has('Transfer-Encoding') || // already encoded or chunked
+      reqCtx.res.headers.has('content-encoding') || // already encoded
+      reqCtx.res.headers.has('transfer-encoding') || // already encoded or chunked
       reqCtx.request.method === 'HEAD' || // HEAD request
       (contentLength && Number(contentLength) < threshold) || // content-length below threshold
       !shouldCompress(reqCtx.res) || // not compressible type
@@ -26,7 +26,7 @@ const compress = (options?: CompressionOptions): Middleware => {
       return;
     }
 
-    const acceptedEncoding = reqCtx.request.headers.get('Accept-Encoding');
+    const acceptedEncoding = reqCtx.request.headers.get('accept-encoding');
     const encoding =
       options?.encoding ??
       Object.values(COMPRESSION_ENCODING_TYPES).find((encoding) =>
@@ -40,18 +40,18 @@ const compress = (options?: CompressionOptions): Middleware => {
     // Compress the response
     const stream = new CompressionStream(encoding);
     reqCtx.res = new Response(reqCtx.res.body.pipeThrough(stream), reqCtx.res);
-    reqCtx.res.headers.delete('Content-Length');
-    reqCtx.res.headers.set('Content-Encoding', encoding);
+    reqCtx.res.headers.delete('content-length');
+    reqCtx.res.headers.set('content-encoding', encoding);
   };
 };
 
 const shouldCompress = (res: Response) => {
-  const type = res.headers.get('Content-Type');
+  const type = res.headers.get('content-type');
   return type && COMPRESSIBLE_CONTENT_TYPE_REGEX.test(type);
 };
 
 const shouldTransform = (res: Response) => {
-  const cacheControl = res.headers.get('Cache-Control');
+  const cacheControl = res.headers.get('cache-control');
   // Don't compress for Cache-Control: no-transform
   // https://tools.ietf.org/html/rfc7234#section-5.2.2.4
   return !cacheControl || !CACHE_CONTROL_NO_TRANSFORM_REGEX.test(cacheControl);
--- a/packages/event-handler/tests/unit/rest/helpers.ts
+++ b/packages/event-handler/tests/unit/rest/helpers.ts
@@ -66,3 +66,14 @@ export const createNoNextMiddleware = (
     // Intentionally doesn't call next()
   };
 };
+
+export const createSettingHeadersMiddleware = (headers: {
+  [key: string]: string;
+}): Middleware => {
+  return async (_params, _options, next) => {
+    await next();
+    Object.entries(headers).map(([key, value]) =>
+      _options.res.headers.set(key, value)
+    );
+  };
+};
--- a/packages/event-handler/tests/unit/rest/helpers.ts
+++ b/packages/event-handler/tests/unit/rest/helpers.ts
@@ -3,11 +3,12 @@ import type { Middleware } from '../../../src/types/rest.js';
 
 export const createTestEvent = (
   path: string,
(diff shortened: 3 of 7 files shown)


packages/event-handler/src/rest/middleware/compress.ts
```diff
@@ -0,0 +1,58 @@
+import type { Middleware } from '@aws-lambda-powertools/event-handler/types';
```
dreamorosi: This should be a local relative import.

The rule of thumb is:
- if it's under the same `packages/<name>/src` folder, it uses a relative import 
- if it's under a different one, it uses a full named import (like what you put above)

packages/event-handler/src/rest/middleware/compress.ts
```diff
@@ -0,0 +1,58 @@
...
+import type { CompressionOptions } from 'src/types/rest.js';
```
dreamorosi: This should also be a relative import starting with `../` - this is because the module resolution is always relative to the file that is importing, not the package root - despite what VSCode auto-suggests.

packages/event-handler/src/rest/middleware/compress.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  return async (_, reqCtx, next) => {
+    await next();
+
+    const contentLength = reqCtx.res.headers.get('content-length');
+
+    // Check if response should be compressed
+    if (
+      reqCtx.res.headers.has('content-encoding') || // already encoded
+      reqCtx.res.headers.has('transfer-encoding') || // already encoded or chunked
```
dreamorosi: Can we created local variables like `const isEncodedOrChunked = reqCtx.res.headers.has('content-encoding') || reqCtx.res.headers.has('transfer-encoding')` and use these to self-document the flow instead of comments? This way we can follow the meaning throughout the code.

sdangol:
@dreamorosi I've addressed the feedback and accepted the regex issue in Sonar

packages/event-handler/src/rest/middleware/compress.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    const contentLength = reqCtx.res.headers.get('content-length');
+    const isEncodedOrChunked =
+      reqCtx.res.headers.has('content-encoding') ||
+      reqCtx.res.headers.has('transfer-encoding');
+
+    // Check if response should be compressed
+    if (
+      isEncodedOrChunked ||
+      reqCtx.request.method === 'HEAD' || // HEAD request
```
svozza: Think all these code comments are redundant now.

packages/event-handler/tests/unit/rest/helpers.ts
```diff
@@ -65,3 +66,14 @@ export const createNoNextMiddleware = (
     // Intentionally doesn't call next()
   };
 };
+
+export const createSettingHeadersMiddleware = (headers: {
+  [key: string]: string;
+}): Middleware => {
+  return async (_params, _options, next) => {
```
svozza: Underscore normally means that the parameter is unused,we should remove it here

packages/event-handler/tests/unit/rest/middleware/compress.test.ts
```diff
@@ -0,0 +1,211 @@
...
+    );
+
+    // Act
+    const result = await app.resolve(event, context);
+
+    // Assess
+    expect(result.headers?.['content-encoding']).toEqual('gzip');
+  });
+
+  it('skips compression for non-compressible content types', async () => {
```
svozza: Is the large regex at the top of this responsible for whether this test passes or fails? I feel like we should make sure we test that more extensively by using `it.each` for multiple content types covered by the regex.

sdangol: Yeah, it's for that. I've added some a few content types in the test now

dreamorosi:
Code duplication is quite high, I suspect because of tests. Can we do anything to bring it down?

Also, and I suspect the answer is no, but is there any way to break down that regex into smaller chunks? Not just because Sonar complains (which I'd like to avoid if possible) but also because of complexity.

Can we, for example, make an educated guess about a subset of content types that are more common and put these in their own smaller (and faster?) regex, and then fall back to the second (or third) if the previous fail? Does this make sense?

sdangol:
> Code duplication is quite high, I suspect because of tests. Can we do anything to bring it down?

Yes, I'll try to optimize the tests

> Also, and I suspect the answer is no, but is there any way to break down that regex into smaller chunks? Not just because Sonar complains (which I'd like to avoid if possible) but also because of complexity.
> Can we, for example, make an educated guess about a subset of content types that are more common and put these in their own smaller (and faster?) regex, and then fall back to the second (or third) if the previous fail? Does this make sense?

Yeah, makes sense. I'll see if I can break it down

packages/event-handler/package.json
```diff
@@ -77,6 +77,16 @@
         "types": "./lib/esm/rest/index.d.ts",
         "default": "./lib/esm/rest/index.js"
       }
+    },
+    "./experimental-rest/middleware": {
```
dreamorosi, line 81: We need to add the new path to the `typeVersions` below as well.

packages/event-handler/src/rest/middleware/compress.ts
```diff
@@ -0,0 +1,120 @@
...
+ * - Content type is compressible
+ * - Cache-Control header doesn't contain no-transform
+ * - Response has a body
+ *
+ * **Basic compression with default settings**
+ *
+ * @example
+ * ```typescript
+ * import { Router } from '@aws-lambda-powertools/event-handler';
+ * import { compress } from '@aws-lambda-powertools/event-handler/rest/middleware';
```
dreamorosi: These paths are wrong, they should be:

```ts
import { compress } from '@aws-lambda-powertools/event-handler/experimental-rest/middleware';
```

packages/event-handler/src/rest/middleware/compress.ts
```diff
@@ -0,0 +1,120 @@
...
+ *   return { data: 'large response body...' };
+ * });
+ * ```
+ *
+ * **Custom compression settings**
+ *
+ * @example
+ * ```typescript
+ * import { Router } from '@aws-lambda-powertools/event-handler';
+ * import { compress } from '@aws-lambda-powertools/event-handler/rest/middleware';
```
dreamorosi: These package names are wrong - should be:

```ts
import { compress } from '@aws-lambda-powertools/event-handler/experimental-rest/middleware';
```

packages/event-handler/src/rest/middleware/compress.ts
```diff
@@ -0,0 +1,120 @@
...
+ * @param options - Configuration options for compression behavior
+ * @param options.threshold - Minimum response size in bytes to trigger compression (default: 1024)
+ * @param options.encoding - Preferred compression encoding to use when client supports multiple formats
+ */
+
+const compress = (options?: CompressionOptions): Middleware => {
+  const threshold = options?.threshold ?? 1024;
+
+  return async (_, reqCtx, next) => {
+    await next();
```
dreamorosi, line 71: The [Python docs](https://docs.powertools.aws.dev/lambda/python/latest/core/event_handler/api_gateway/#compress) suggest that even with `compress` enabled, if the client doesn't send the correct `Accept-Encoding` header, Event Handler won't return a compressed response.

The implementation also seems to point towards [the same behavior](https://github.com/aws-powertools/powertools-lambda-python/blob/4743b80c28dcc56df7af0a7dcc8f44bf97dde62f/aws_lambda_powertools/event_handler/api_gateway.py#L945-L957). 

Why are we not doing the same?

dreamorosi:
I think I found the issue: at [this point](https://github.com/aws-powertools/powertools-lambda-typescript/blob/87058a29c44dff0ed0f53104b058e8c8a334ee05/packages/event-handler/src/rest/converters.ts#L92-L98) when we convert the `Response` object to an API Gateway Proxy response we hardcode both `isBase64Encoded` and `body` respectively as `false` and text.

I applied a quick & dirty patch and it works:

```ts
// Check if response contains compressed/binary content
const isCompressed = response.headers.has('content-encoding');
let body: string;
let isBase64Encoded = false;

if (isCompressed) {
  // For compressed content, get as buffer and encode to base64
  const buffer = await response.arrayBuffer();
  body = Buffer.from(buffer).toString('base64');
  isBase64Encoded = true;
} else {
  // For text content, use text()
  body = await response.text();
  isBase64Encoded = false;
}

const result: APIGatewayProxyResult = {
  statusCode: response.status,
  headers,
  body,
  isBase64Encoded,
};
```

I don't know if there's a better way to do this, or if I'm missing any edge cases, so please triple check the code above if you decide to put it in the repo.

Also I'm not sure if compression is the only reason why the `isBase64Encoded` field should be set to `true` (probably one of the main ones) - but worth checking. 

@svozza - keen to hear your take on this as well.

packages/event-handler/src/rest/middleware/compress.ts
dreamorosi: @svozza what do you think about this? I saw at least another JS implementation doesn't do this, but I think the client should be able to bypass the API compression setting if they don't want to/can't support it, otherwise the request would be thrown away anyway.

This also matches the behavior from API Gateway as documented [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-make-request-with-compressed-payload.html) and [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-gzip-compression-decompression.html).

sdangol:
@dreamorosi Thanks for looking into this. I will see if there are any other edge cases we need to be aware of. I realized that the actual compressed response hasn't been tested in the code as well so missed this. I'll add the tests for that as well.

packages/event-handler/src/rest/middleware/compress.ts
svozza: No I agree, that header is there for a reason and we should respect it.

svozza:
> I think I found the issue: at [this point](https://github.com/aws-powertools/powertools-lambda-typescript/blob/87058a29c44dff0ed0f53104b058e8c8a334ee05/packages/event-handler/src/rest/converters.ts#L92-L98) when we convert the `Response` object to an API Gateway Proxy response we hardcode both `isBase64Encoded` and `body` respectively as `false` and text.
> 
> I applied a quick & dirty patch and it works:
> 
> ```ts
> // Check if response contains compressed/binary content
> const isCompressed = response.headers.has('content-encoding');
> let body: string;
> let isBase64Encoded = false;
> 
> if (isCompressed) {
>   // For compressed content, get as buffer and encode to base64
>   const buffer = await response.arrayBuffer();
>   body = Buffer.from(buffer).toString('base64');
>   isBase64Encoded = true;
> } else {
>   // For text content, use text()
>   body = await response.text();
>   isBase64Encoded = false;
> }
> 
> const result: APIGatewayProxyResult = {
>   statusCode: response.status,
>   headers,
>   body,
>   isBase64Encoded,
> };
> ```
> 
> I don't know if there's a better way to do this, or if I'm missing any edge cases, so please triple check the code above if you decide to put it in the repo.
> 
> Also I'm not sure if compression is the only reason why the `isBase64Encoded` field should be set to `true` (probably one of the main ones) - but worth checking.
> 
> @svozza - keen to hear your take on this as well.

Yes, the `base64` logic is something I [knew](https://github.com/aws-powertools/powertools-lambda-typescript/pull/4349#discussion_r2280500147) we'd have to come back to. I'm not sure if the presence of the `content-encoding` header is enough, I think we need to check for specific values in the header such as `deflate` and `gzip`. We also need to check the `content-type` headers for binary MIME types like `image/png`, `image/jpeg` etc.

dreamorosi:
Ok, thanks.

I think we should address that in a separate issue that possibly has its own spec with requirements and all.

For the sake of this PR, let's just make sure that when the `compress` middleware is enabled, `isBase64Encoded` is also set to `true` in the final response object.

sdangol:
@dreamorosi @svozza I've refactored the code a bit now.

It's now using the `accepted-encoding` header value to to find the preferred encoding and is disabled using the `identity` value.
I've removed the `content-type` checking to simplify things for now. I think if there are any requests to support additional types, we can add them later.
Also, the `isBase64Encoded` flag is set to true based on the value of the `content-encoding` header.

I've tried testing it out, I get the response already decompressed I think but the `content-length` seems to be different for the same request when enabling/disabling compression.

Screenshot 2025-09-17 at 6 42 18 pm

Screenshot 2025-09-17 at 6 43 30 pm

dreamorosi:
> I think but the content-length seems to be different for the same request when enabling/disabling compression

I think that's expected and the result of the content being compressed or not - I guess the only check would be to make sure it matches the expected value in the two cases.

packages/event-handler/tests/unit/rest/middleware/compress.test.ts
```diff
@@ -0,0 +1,169 @@
+import { gzipSync } from 'node:zlib';
+import context from '@aws-lambda-powertools/testing-utils/context';
+import { Router } from 'src/rest/Router.js';
+import { beforeEach, describe, expect, it } from 'vitest';
+import { compress } from '../../../../src/rest/middleware/index.js';
+import { createSettingHeadersMiddleware, createTestEvent } from '../helpers.js';
+
+describe('Compress Middleware', () => {
```
svozza, line 8: We should have assertions to check `isBase64Encoded` is set properly in these tests too.

packages/event-handler/src/rest/middleware/compress.ts
```diff
@@ -0,0 +1,120 @@
...
+  const isEncodedOrChunked =
+    response.headers.has('content-encoding') ||
+    response.headers.has('transfer-encoding');
+
+  const shouldEncode =
+    !acceptedEncoding.includes(COMPRESSION_ENCODING_TYPES.IDENTITY) &&
+    (acceptedEncoding.includes(preferredEncoding) ||
+      acceptedEncoding.includes(COMPRESSION_ENCODING_TYPES.ANY));
+
+  if (
```
svozza: Can this be simplified to:

```ts
return (
    !shouldEncode ||
    isEncodedOrChunked ||
    request.method === 'HEAD' ||
    (contentLength && Number(contentLength) < threshold) ||
    (cacheControl && CACHE_CONTROL_NO_TRANSFORM_REGEX.test(cacheControl)) ||
    !response.body
  )
```

sdangol:
> I think that's expected and the result of the content being compressed or not - I guess the only check would be to make sure it matches the expected value in the two cases.

The responses did match in both the cases

packages/event-handler/tests/unit/rest/middleware/compress.test.ts
sdangol: Added the checks in the tests

packages/event-handler/src/rest/middleware/compress.ts
sdangol: It should have returned false with that expression, so modified it a bit

dreamorosi (dismissed):

packages/event-handler/tests/unit/rest/converters.test.ts
```diff
@@ -396,6 +396,18 @@ describe('Converters', () => {
...
+    it('handles compressed response body', async () => {
+      const response = new Response('Hello World', {
+        status: 200,
+        headers: {
+          'content-encoding': 'gzip',
+        },
+      });
+
+      const result = await webResponseToProxyResult(response);
+      expect(result.isBase64Encoded).toBe(true);
```
svozza, line 499: We need to check the body here too make sure that the logic that converts it to a base64 string works.

svozza (approved):

PR #4498, opened 2025-09-16
fix(batch): fixed the build issue with Batch processor due to missing dependencies

### LINKED ISSUE #4496: Maintenance: Declare @standard-schema/spec as a dependency
### Summary

The import from @standard-schema/spec added [here](https://github.com/aws-powertools/powertools-lambda-typescript/pull/4408/files#diff-b498949319c62422801f330e47ec09a0b8fbfa1ea0ebd4bc182998ef11266574R1) was not declared as a dependency in the batch package.

PR with the new type import https://github.com/aws-powertools/powertools-lambda-typescript/pull/4408.

### Why is this needed?

I'm running into build errors without this dependency.

Batch Processing

### Solution

Adding @standard-schema/spec as a dependency in the batch package.

### Acknowledgment

### Future readers

sdangol:
## Summary

This PR fixes the build issue with the Batch Processor due to missing dependencies with Standard Schema.

### Changes

- Added `@standard-schema/spec` as a dependency of the Batch processor package
- Removed the optional on DynamoDB extended schemas to enforce the same schema to be used for both `OldImage` and `NewImage` and added a note in the docs about it
- Removed ZodType imports and replaced type representing Zod schema

**Issue number:** closes #4496 

### Screenshot

Screenshot 2025-09-16 at 4 00 40 pm

---


packages/batch/src/types.ts
```diff
@@ -239,6 +242,20 @@ type ParsedRecord<TRecord, TPayload, TOldPayload = TPayload> = TRecord extends {
...
+ * Parts of the parser integration within BatchProcessor rely on Zod for schema transformations,
+ * however some other parts also support other Standard Schema-compatible libraries.
+ *
+ * To avoid forcing a direct dependency on Zod, we use `any` here, which is not ideal but necessary.
+ *
+ * The vendor is checked at runtime to ensure Zod is being used when required using `StandardSchemaV1['~standard'].vendor`.
+ *
+ * biome-ignore lint/suspicious/noExplicitAny: using `any` to avoid direct dependency on 'zod'
+ */
+type RuntimeZodType = any;
```
svozza: Just for my own curiosity, why can't we use `unkonwn` here?

dreamorosi: I think I did an iteration with `unknown` and it didn't work in places where we expect a Zod type, but tbh I tried several combinations and it could've been something else.

@sdangol give it a try pls, if it works and there's no linting issues that's great.

Alternatively we can typecast or put `// @ts-expect-error` in these places.

sdangol: @svozza Yeah, it doesn't work when we assign to [StandardSchema](https://github.com/aws-powertools/powertools-lambda-typescript/pull/4498/files#diff-8164b2d4e3810362c41e8d01ebfd97fdaa3e1f11e4f6728219fc7cee9cb20c64R16) or `ZodType`

sdangol:
@dreamorosi This is ready for review now

packages/batch/src/types.ts
dreamorosi: If we want to make it `unknown` we'll have to:
- remove the type guard (which is incompatible with `unknown`)
- revert the runtime check at L142 of `parser.ts` ([here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/1a37d87a895bbb40c3653f11c26157ee28f4826b/packages/batch/src/parser.ts#L142)) to this `if (innerSchema['~standard'].vendor !== SchemaVendor.Zod) {`
- add `// @ts-expect-error` when we use the transformer helpers in `createExtendedSchema` like so
Screenshot 2025-09-16 at 23 57 16

I'm ambivalent about which one to pick with a slight inclination to actually do this instead of using `any`. From experience, as TS got better over time, some places where I had put an `@ts-expect-error` became redundant while places where we put `any` will always stay that way.

This would also remove the SonarQube issue - but up to you.

sdangol: > revert the runtime check at L142 of `parser.ts` ([here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/1a37d87a895bbb40c3653f11c26157ee28f4826b/packages/batch/src/parser.ts#L142)) to this `if (innerSchema['~standard'].vendor !== SchemaVendor.Zod) {`

Do we need to revert this? I didn't see any errors when I just removed the type guard from the function

packages/batch/src/parser.ts
```diff
@@ -13,7 +13,7 @@ import type {
  *
  * @param schema - The schema to check
  */
-const isZodSchema = (schema: StandardSchemaV1): schema is RuntimeZodType =>
+const isZodSchema = (schema: StandardSchemaV1) =>
   schema['~standard'].vendor === SchemaVendor.Zod;
```
dreamorosi: At this point, since it's not asserting the type, this is just extra lines - I'd just put the check in the `if` statement since it's the only occurrence, like I suggested.

packages/batch/src/types.ts
```diff
@@ -248,13 +248,11 @@ type ParsedRecord<TRecord, TPayload, TOldPayload = TPayload> = TRecord extends {
...
  *
- * To avoid forcing a direct dependency on Zod, we use `any` here, which is not ideal but necessary.
+ * To avoid forcing a direct dependency on Zod, we use `unknown` here, which is not ideal but necessary.
  *
  * The vendor is checked at runtime to ensure Zod is being used when required using `StandardSchemaV1['~standard'].vendor`.
- *
- * biome-ignore lint/suspicious/noExplicitAny: using `any` to avoid direct dependency on 'zod'
  */
-type RuntimeZodType = any;
+type RuntimeZodType = unknown;
```
dreamorosi: I believe we now use this alias in a single place - I think we can remove it and keep the description in the `createExtendedSchema` function description, so Sonar is also happy.

docs/features/batch.md
```diff
@@ -358,6 +358,9 @@ Available transformers by event type:
     --8<-- "examples/snippets/batch/samples/parser_SQS.json"
     ```
 
+!!! note
+    If `innerSchema` is used with DynamoDB streams, the schema will be applied to both the `NewImage` and the `OldImage` by default. If you want to have separate schema for both, you will need to extend the schema and use the full schema for parsing.
```
dreamorosi:
```suggestion
    If `innerSchema` is used with DynamoDB streams, the schema will be applied to both the `NewImage` and the `OldImage` by default. If you want to have dedicated schemas, see the section below.
```

PR #4511, opened 2025-09-19
feat(event-handler): throw error when middleware does not await next()

### LINKED ISSUE #4510: Feature request: Detect When Middleware Does Not Await Next Callback in REST API
### Use case

When authoring middleware, a common mistake is to forget to await the `next` callback that invokes the next middleware in the stack. This is problematic because it alters the execution order of the middleare leading to subtle and hard to diagnose bugs.

Consider the following example and note the incorrect order:

```ts
const app = new Router();

app.use(async (_params, _reqCtx, next) => {
    console.log('middleware 1 - before');
    await next();
    console.log('middleware 1 - after');
});

app.use(async (_params, _reqCtx, next) => {
    console.log('middleware 2 - before');
    next();
    console.log('middleware 2 - after');
});


app.use(async (_params, _reqCtx, next) => {
    console.log('middleware 3 - before');
    await next();
    console.log('middleware 3 - after');
});

// execution order
[
  "middleware 1 - before",
  "middleware 2 - before",
  "middleware 3 - before",
  "handler",
  "middleware 1 - after",
  "middleware 3 - after", // <= out of order
  "middleware 2 - after"
]
```

Once the order of middleware execution has been compromised like this, we are no longer following the onion layer pattern. As middleware can mutate requests and responses, this can cause unwanted mutliple side effects further down the chain.

### Solution/User Experience

We should detect when a middleware has not awaited the `next()` callback and throw a descriptive error, thus stopping the request.

### Alternative solutions

```Markdown
We could leave it as a responsibility of middleware authors to do this and allow incorrect middleware. This is unacceptable.
```

### Acknowledgment

### Future readers

svozza:
## Summary
Handles the case where middleware does not await the `next()` handler, which causes the the order of middleware to become interminate.

### Changes

- Add logic to `composeMiddleware` function to detect if `next` function has not been awaited.
- Unit test to cover this case

**Issue number:** closes #4510

---


packages/event-handler/src/rest/utils.ts
```diff
@@ -172,9 +172,22 @@ export const composeMiddleware = (middleware: Middleware[]): Middleware => {
...
+        nextPromise = dispatch(i + 1);
+        const result = await nextPromise;
+        nextAwaited = true;
+        return result;
+      };
+
+      const middlewareResult = await middlewareFn(params, reqCtx, nextFn);
+
+      if (nextPromise && !nextAwaited && i < middleware.length - 1) {
+        throw new Error(
```
dreamorosi, line 187: What do you think about logging a warning instead?

As the error message hints at, we don't know for sure if this will cause issues, so there's a chance we're throwing a 500 error and breaking the customer API prematurely.

If we log a warning instead, at best we are helping them solve these subtle bugs, at worst we just let them know and they can remove the warning by fixing it.

Also can we in any way include the middleware name (aka function name) in the message?

leandrodamascena: If a middleware can be used to control fine-grained authorization on a route/global, for example, and potentially executes out of order, allowing other middleware to execute even without proper authorization and introducing potential security issues, I believe this should thrown an error for the developer to fix in the DEV ENV. Also, make it super cleat in the doc.
If not and a middleware can only be used for other purposes, a warning makes more sense.

In Python, customers use middleware to execute custom authN and AuthZ.

svozza: I'm open to persuasion here but this is my reasoning:
- The execution order is a fundamental invariant of the composition algorithm, if this does not hold the system is in an invalid state regardless of whether it accidentally works.
- Throwing means this is far more likely to get caught at the unit test level before it ever gets to production. A warning is very easy to miss especially when you consider test frameworks such as Vitest suppress console output when running tests.
- _Any_ post processing middleware that interacts with `reqCtx.res.body` even something as innoucous as `reqCtx.res = new Response(reqCtx.res.body)` will throw an error. Which would you rather receive as an error in prod: `TypeError [ERR_INVALID_STATE]: Invalid state: The ReadableStream is locked` or `'Middleware called next() without awaiting. This may lead to unexpected behavior.`?

Regarding the middleware name, I had thought that we could add the number in the chain, ie., `the middleware at position ${i} in the middleware stack did not call next`. This might make it easier to find because a lot of middleware could just be anonymous functions but maybe we could do the function name and fall back to the location if the function is anonymous.

dreamorosi: Ok, you convinced me, let's throw an error.

Regarding the error message, I don't know if having the index helps at all since I might not necessarily keep track of the order.

Since we're throwing an error, it'll have a stack trace and that will be enough to identify the issue.

dreamorosi (approved):

PR #4523, opened 2025-09-22
feat(event-handler): implemented route prefixes in HTTP event handler

### LINKED ISSUE #4513: Feature request: route prefixes in HTTP Event Handler
### Use case

When defining multiple routes related to a specific resource, it's common to have a shared prefix. For example, you might have several routes that all start with `/todos`.

For example, if you have a custom domain `api.example.com` and you want to map it to the `/v1` base path of your API. In this case, all the requests will contain `/v1/<resource>` in the path, requiring you to repeat the `/v1` prefix in all your route definitions.

To avoid repeating the prefix in each route definition, you can use the `prefix` constructor parameter when creating a new `Router` instance, and we'll automatically strip it from the request path before matching routes. After mapping a path prefix, the new root path will automatically be mapped to the path argument of `/`.

### Solution/User Experience

```ts
import {
  Router,
  UnauthorizedError,
} from '@aws-lambda-powertools/event-handler/experimental-rest';

const app = new Router({ prefix: '/todos' });

app.post('/', async (_) => { // matches POST /todos
  const todos = await getUserTodos();
  return { todos };
});

  app.resolve(event, context);
```

This is also useful when splitting routes into separate files (see [Split routers](#split-routers) section) or when using [API mappings](https://docs.aws.amazon.com/apigateway/latest/developerguide/rest-api-mappings.html){target="_blank"} to map custom domains to specific base paths.

For example, when using `prefix: '/pay'`, there is no difference between a request path of `/pay` and `/pay/`; and the path argument would be defined as `/`.

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

sdangol:
## Summary

This PR implements the feature to allow the use of `prefix` so that we don't need to keep repeating the same shared prefix for every route path. 

### Changes

- Added `prefix` option to the `RestRouterOptions` in the constructor
- Added the logic to append the prefix if present

**Issue number:** closes #4513 

---

### DIFF
--- a/packages/event-handler/src/rest/Router.ts
+++ b/packages/event-handler/src/rest/Router.ts
@@ -56,6 +56,10 @@ class Router {
    * Whether the router is running in development mode.
    */
   protected readonly isDev: boolean = false;
+  /**
+   * The base prefix to be used for all routes registered using this Router.
+   */
+  protected readonly prefix?: Path;
 
   public constructor(options?: RestRouterOptions) {
     this.context = {};
@@ -73,6 +77,7 @@ class Router {
       logger: this.logger,
     });
     this.isDev = isDevMode();
+    this.prefix = options?.prefix;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


dreamorosi:
@sdangol please take a look at the SonarCloud findings and address them - based on a cursory read we should be able to resolve them.

PR #4528, opened 2025-09-22
chore: sanitize CI inputs via env var

### LINKED ISSUE #4529: Maintenance: change workflow inputs to always use `env`
### Summary

SonarCloud has surfaced a number of areas for improvement in our CI workflows where the workflow inputs are used directly inline in some comments rather than creating intermediate environment variables.

We should fix this and use env variables wherever possible.

### Why is this needed?

So that we can improve the input processing of our CI workflows.

Other

### Solution

_No response_

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR updates several workflows to use environment variables to temporarily store workflow inputs, before passing them to the command and expanding them as a string. For example, `${{ input.version }}` now becomes `$VERSION` with the variable being stored under `env`.

The PR also adds double quotes in multiple places to ensure string expansion works as intended.

**Issue number:** closes #4529

---


dreamorosi:
@sdangol you worked on the secrets access last, can you please check if Copilot's suggestions are appropriate? If not we can dismiss them together with the findings that might pop up after merging.

sdangol:
@dreamorosi Those are the same ones that we had discussed last time about the issue being mitigated by the use of environment specific secrets. So, the findings are a false positive.

dreamorosi:
Yes, I know, we had to revert them because the solution we came up with didn't work.

I was wondering if Copilot's solution is better and might work.

sdangol:
The suggested solution will need some adjustments to make it work like the checking for the `CONVERTED_REGION` value is wrong. Nonetheless, including the regions even just for the china and the gov cloud regions would make it very verbose.

dreamorosi:
Ok, thanks for checking - I remember you mentioning this in the previous iteration as well.

Happy to ignore the suggestions and merge whenever you're ready to approve (tomorrow after the release is fine).

Issue #4518, opened 2025-09-21
Feature request - consider changing the RequestContext abbreviated response name `res` to `response`


nateiler:
The `res` abbreviated property within the [RequestContext](https://github.com/aws-powertools/powertools-lambda-typescript/blob/1e72e57cabe410020f1df014a465cbfd0562cbd8/packages/event-handler/src/types/rest.ts#L21) seems out of place.  Recommending using the full `response` name instead.

Before
```
type RequestContext = {
  request: Request;
  event: APIGatewayProxyEvent;
  context: Context;
  res: Response;
};
```

After
```
type RequestContext = {
  request: Request;
  event: APIGatewayProxyEvent;
  context: Context;
  response: Response;
};
```

dreamorosi:
Hi @nateiler, thank you for opening this issue.

I think we picked that name primarily for brevity's sake, but I agree it's kind of out of place since all the others are spelled out.

Personally I'm not against the change, but I'll let @svozza decide.

nateiler:
Yeah, I figured as much.  Normally wouldn't raise such a picky item, but since it's still experimental, I figured we could start the discussion.  

Could also replace `request` with `req` but I would vote on the long form.

svozza:
I would be more in fsvour of changing `request` to `req`. I initially chose `res` because that's what most other frameworks although weirdly some also use `request` rather than `req` but I prefer the symmetry of using short form here,

dreamorosi:
In the next days we'll most likely add more properties to the object like `params` for path parameters, so shortening everything could make sense, but like I said I'm ambivalent.

@nateiler please keep the feedback coming; it's more than welcome!

dreamorosi:
We've decided to shorten to `req` so it's less characters to type.

We expect to have this changed by the next release.

PR #4530, opened 2025-09-22
improv(event-handler): changed the Middleware and RequestContext signatures

### LINKED ISSUE #4480: Maintenance: update middleware signature to object
### Summary

Currently the `Middleware` type for middleware functions describes a function with three parameters: `params`, `options`, `next`.

Since the function uses positional arguments, if I only want to use some of them I need to alias them to various combinations of `_` (underscore) or linters might complain that the parameter is unused.

This is particularly apparent in middleware functions like the one below (from our docs) that only need to use `next`, which appears as last argument:

```ts
const loggingMiddleware: Middleware = async (_, __, next) => {
  logger.info('Request processed');
  await next();
};
```

I wonder if we can change the signature to instead have an object, so that customers can pick only the parameters they want from it, like:

```ts
const loggingMiddleware: Middleware = async ({ next }) => {
  logger.info('Request processed');
  await next();
};
```

Will this cause the same issues described in the [destructuring pitfalls section](http://localhost:8000/lambda/typescript/latest/features/event-handler/rest/#avoiding-destructuring-pitfalls) of the docs?

### Why is this needed?

So that middleware functions have less boilerplate in the signature in most cases.

Event Handler

### Solution

_No response_

### Acknowledgment

### Future readers

sdangol:
## Summary

This PR updates the signatures of the Middleware function to use a single parameter object instead of multiple parameters. It also updates the `request` property of the `RequestContext` type to use the shortened version `req` to make it consistent with the response `res` 

### Changes

- Updated the middleware signatures
- Changed all occurence of `reqCtx.request` to `reqCtx.req`

**Issue number:** closes #4480 and #4518 

---

### DIFF
--- a/packages/event-handler/src/rest/Router.ts
+++ b/packages/event-handler/src/rest/Router.ts
@@ -164,7 +164,7 @@ class Router {
    *
    * @example
    * ```typescript
-   * const authMiddleware: Middleware = async (params, reqCtx, next) => {
+   * const authMiddleware: Middleware = async ({params, reqCtx, next}) => {
    *   // Authentication logic
    *   if (!isAuthenticated(reqCtx.request)) {
    *     return new Response('Unauthorized', { status: 401 });
@@ -231,7 +231,11 @@ class Router {
 
       const route = this.routeRegistry.resolve(method, path);
 
-      const handlerMiddleware: Middleware = async (params, reqCtx, next) => {
+      const handlerMiddleware: Middleware = async ({
+        params,
+        reqCtx,
+        next,
+      }) => {
         if (route === null) {
           const notFoundRes = await this.handleError(
             new NotFoundError(`Route ${path} for method ${method} not found`),
@@ -263,11 +267,11 @@ class Router {
         handlerMiddleware,
       ]);
 
-      const middlewareResult = await middleware(
-        route?.params ?? {},
-        requestContext,
-        () => Promise.resolve()
-      );
+      const middlewareResult = await middleware({
+        params: route?.params ?? {},
+        reqCtx: requestContext,
+        next: () => Promise.resolve(),
+      });
 
       // middleware result takes precedence to allow short-circuiting
       const result = middlewareResult ?? requestContext.res;
--- a/packages/event-handler/src/rest/Router.ts
+++ b/packages/event-handler/src/rest/Router.ts
@@ -166,7 +166,7 @@ class Router {
    * ```typescript
    * const authMiddleware: Middleware = async ({params, reqCtx, next}) => {
    *   // Authentication logic
-   *   if (!isAuthenticated(reqCtx.request)) {
+   *   if (!isAuthenticated(reqCtx.req)) {
    *     return new Response('Unauthorized', { status: 401 });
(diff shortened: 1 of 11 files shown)


packages/event-handler/src/rest/Router.ts
```diff
@@ -164,9 +164,9 @@ class Router {
    *
    * @example
    * ```typescript
-   * const authMiddleware: Middleware = async (params, reqCtx, next) => {
+   * const authMiddleware: Middleware = async ({params, reqCtx, next}) => {
```
svozza: I don't think the next callback should be in the `reqCtx` object. It's a positional argument in all other frameworks I can think of.

sdangol: The way we've used this so far in our tests and built-in middlewares, the `params` was usually unused. Wouldn't it make more sense to only pick what we need?

dreamorosi: I think there's a bit of confusion, and the lack of spaces in the example doesn't help (should be `async ({ params, reqCtx, next }) => {`).

The PR doesn't put `next` in the `RequestContext` type or `reqCtx` object but keeps it into the `Middleware` which now becomes an object:

```ts
// before
type Middleware = (
  params: Record<string, string>,
  reqCtx: RequestContext,
  next: NextFunction
) => Promise<void | HandlerResponse>;

// after
type Middleware = (args: {
  params: Record<string, string>;
  reqCtx: RequestContext;
  next: NextFunction;
}) => Promise<void | HandlerResponse>;
```

by doing this, when writing a middleware function you can now do this:

```ts
const myMiddleware: Middleware = async ({ reqCtx, next }) => {
  console.log('req.headers', reqCtx.req.headers);
  await next();
}
```

instead of having to do this:

```ts
const myMiddleware: Middleware = async (_, reqCtx, next) => {
  console.log('req.headers', reqCtx.req.headers);
  await next();
}
```

I understand other frameworks keep it positional, but I don't think we should do the same just for uniformity's sake.

Unless there's a foot gun or issue, I think we should pick whatever experience we think it's more ergonomic, and right now I think it being an object is better so I can grab values I care from without having to worry about unused args.

But I'm happy to be persuaded otherwise.

svozza: Ah sorry, I misunderstood. I'm happy to go with the object approach then.

sdangol:
@dreamorosi As per our conversation about making `params` as part of `reqCtx`, it might not be just those that would be affected. We might need to make changes to the [RouteHandler](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/event-handler/src/types/rest.ts#L63-L66) and the [resolve function](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/event-handler/src/rest/Router.ts#L250) as well?

dreamorosi:
Sure, let's do it, although I'm ok to do it in a separate PR if you think this is gonna balloon too much.

sdangol:
Yeah, I think there might be more changes than I had expected. Let's merge this one first.

PR #4536, opened 2025-09-23
improv(event-handler): made error handler responses versatile

### LINKED ISSUE #4479: Maintenance: make error handlers' response more versatile
### Summary

Currently error handlers must return an `ErrorResponse` object with type:

```ts
type ErrorResponse = {
  statusCode: HttpStatusCode;
  error: string;
  message: string;
};
```

While building with the new Event Handler I had a number of ideas that I'd like us to experiment:
- Can we change the return type of error handlers to `ErrorResponse | Response`? This would help in case customers want to have full control over what's returned by an error handler, similar to what they can do in regular handlers
- **For named error handlers only** (i.e. `notFound` and `methodNotAllowed` ) we should consider making the `statusCode` field from `ErrorResponse` optional, since these method already have a default one
- Right now both `error` and `message` are required, can we update the type so that at least one can be omitted? this reduces repetition in case I care about only one
- Finally, can I throw any of the built-in error classes (i.e. `NotFoundError`) from an error handler? Right now if I do that, the Event Handler interprets it as a runtime error and falls back to the default error handler (502)

### Why is this needed?

None of the ideas proposed above is required per se, especially if it doesn't make sense or complicates the overall API, but I'd like us to consider them as I think they'll make the API more ergonomic.

Event Handler

### Solution

_No response_

### Acknowledgment

### Future readers

sdangol:
## Summary

This PR allows Error Handler responses to be more so that it allows full control over what's returned by the error handler and improves the user experience when using the API.

### Changes

- Changed the return type of error handlers to `ErrorResponse | Response`
- `statusCode` omitted from ErrorResponse when using named error handler
- Updated the `ErrorResponse` so that either error or message is required
- Allow throwing of built in error classes from the error handler

**Issue number:** closes #4479 

---

### DIFF
--- a/.github/workflows/make-release.yml
+++ b/.github/workflows/make-release.yml
@@ -52,6 +52,7 @@ jobs:
         with:
           node-version: "22"
           cache: "npm"
+          registry-url: 'https://registry.npmjs.org'
       - name: Setup auth tokens
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -43,6 +43,7 @@ jobs:
         with:
           node-version: "22"
           cache: "npm"
+          registry-url: 'https://registry.npmjs.org'
       - name: Setup auth tokens
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/README.md
+++ b/README.md
@@ -5,7 +5,8 @@
 ![GitHub Release](https://img.shields.io/github/v/release/aws-powertools/powertools-lambda-typescript?style=flat)
 [![Quality Gate Status](https://sonarcloud.io/api/project_badges/measure?project=aws-powertools_powertools-lambda-typescript&metric=alert_status)](https://sonarcloud.io/summary/new_code?id=aws-powertools_powertools-lambda-typescript)
 [![Security Rating](https://sonarcloud.io/api/project_badges/measure?project=aws-powertools_powertools-lambda-typescript&metric=security_rating)](https://sonarcloud.io/summary/new_code?id=aws-powertools_powertools-lambda-typescript)
-[![OpenSSF Scorecard](https://api.securityscorecards.dev/projects/github.com/aws-powertools/powertools-lambda-typescript/badge)](https://api.securityscorecards.dev/projects/github.com/aws-powertools/powertools-lambda-typescript)
+[![OpenSSF Scorecard](https://api.securityscorecards.dev/projects/github.com/aws-powertools/powertools-lambda-typescript/badge)](https://scorecard.dev/viewer/?uri=github.com/aws-powertools/powertools-lambda-typescript)
+
 [![Discord](https://img.shields.io/badge/Discord-Join_Community-7289da.svg)](https://discord.gg/B8zZKbbyET)
 
 Powertools for AWS Lambda (TypeScript) is a developer toolkit to implement Serverless [best practices and increase developer velocity](https://docs.powertools.aws.dev/lambda/typescript/latest/#features).


packages/event-handler/src/rest/Router.ts
```diff
@@ -316,11 +316,27 @@ class Router {
       try {
         const { scope, ...reqCtx } = options;
         const body = await handler.apply(scope ?? this, [error, reqCtx]);
+        if (body instanceof Response) {
+          return body;
+        }
+        if (!body.statusCode) {
+          if (error instanceof NotFoundError) {
```
svozza, line 329: Could we not have this so you can throw any of our built-in errors? Sowe check if it's an instance of `ServiceError` instead.

sdangol: I had added them there to allow named error handlers to be used without having to specify the status codes

dreamorosi: If we throw an error - built-in or not - this code won't run and we'll go into the `catch` block (below L334-340).

I think this applies only if we return one of these two built-in errors.

svozza: Yeah, sorry, I put my comment in the wrong place, I mean to put it in the catch block.

packages/event-handler/src/rest/Router.ts
```diff
@@ -316,11 +316,24 @@ class Router {
...
+            body.statusCode = HttpErrorCodes.NOT_FOUND;
+          } else if (error instanceof MethodNotAllowedError) {
+            body.statusCode = HttpErrorCodes.METHOD_NOT_ALLOWED;
+          }
+        }
         return new Response(JSON.stringify(body), {
-          status: body.statusCode,
+          status: (body.statusCode as number) ?? HttpErrorCodes.OK,
```
svozza: I don't think this should be a 200 as a fallback.

packages/event-handler/src/rest/converters.ts
```diff
@@ -181,10 +181,12 @@ export const handlerResultToWebResponse = (
  * Handles APIGatewayProxyResult, Response objects, and plain objects.
  *
  * @param response - The handler response (APIGatewayProxyResult, Response, or plain object)
+ * @param statusCode - The response status code to return
  * @returns An API Gateway proxy result
  */
 export const handlerResultToProxyResult = async (
-  response: HandlerResponse
+  response: HandlerResponse,
+  statusCode: (typeof HttpErrorCodes)[keyof typeof HttpErrorCodes] = HttpErrorCodes.OK
```
svozza: This is already exported as a type called `HttpStatusCode`: https://github.com/aws-powertools/powertools-lambda-typescript/blob/d281d46cbdd2e813005c5b5174aa017ace8f0671/packages/event-handler/src/types/rest.ts#L70

packages/event-handler/tests/unit/rest/Router/error-handling.test.ts
```diff
@@ -393,4 +392,151 @@ describe('Class: Router - Error Handling', () => {
...
+        statusCode: HttpErrorCodes.NOT_FOUND,
+        error: 'NotFoundError',
+        message: 'This error is thrown from the error handler',
+      }),
+      headers: { 'content-type': 'application/json' },
+      isBase64Encoded: false,
+    });
+  });
+
+  it('handles throwing a built in MethodNotAllowedError error from the error handler', async () => {
```
svozza, line 513: Let's have a test case for throwing an error that's not `NotFound` or `MethodNotAllowedError`.

packages/event-handler/tests/unit/rest/Router/error-handling.test.ts
```diff
@@ -393,4 +392,179 @@ describe('Class: Router - Error Handling', () => {
...
+
+    app.get('/test', () => {
+      throw new BadRequestError('test error');
+    });
+
+    // Act
+    const result = await app.resolve(createTestEvent('/test', 'GET'), context);
+
+    // Assess
+    expect(result).toEqual({
```
svozza: Let's use dev mode (`vi.stubEnv('POWERTOOLS_DEV', 'true')`) here and use the `detail` field to check that the error message is correct: the way we're verifying now could mask unrelated errors.

PR #4547, opened 2025-09-24
style(batch): apply stricter linting

dreamorosi:
### Changes

This PR resolves all the linting issues resulting from the new rules being introduced in #4545.

While working on this, I also addressed a few SonarCloud issues that popped up over time.

> [!important]
> This PR is expected to have failing CI, focus only on the CI steps for this package (`batch`). It's safe to merge since it's not going on `main` but on `ci/fix_linting`.

**Issue number:** closes #<issue_number>

---

### DIFF
--- a/packages/batch/tests/helpers/handlers.ts
+++ b/packages/batch/tests/helpers/handlers.ts
@@ -39,7 +39,7 @@ const dynamodbRecordHandler = (record: DynamoDBRecord): object => {
   return body;
 };
 
-const asyncDynamodbRecordHandler = async (
+const asyncDynamodbRecordHandler = (
   record: DynamoDBRecord
 ): Promise<object> => {
   return Promise.resolve(dynamodbRecordHandler(record));
@@ -57,7 +57,7 @@ const handlerWithContext = (record: SQSRecord, context: Context): string => {
   return record.body;
 };
 
-const asyncHandlerWithContext = async (
+const asyncHandlerWithContext = (
   record: SQSRecord,
   context: Context
 ): Promise<string> => {
--- a/packages/batch/tests/helpers/handlers.ts
+++ b/packages/batch/tests/helpers/handlers.ts
@@ -1,4 +1,5 @@
 import type {
+  AttributeValue,
   Context,
   DynamoDBRecord,
   KinesisStreamRecord,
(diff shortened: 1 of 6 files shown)


packages/batch/tests/helpers/handlers.ts
```diff
@@ -8,50 +9,70 @@ import type {
...
 
 const asyncKinesisRecordHandler = async (
   record: KinesisStreamRecord
-): Promise<string> => Promise.resolve(kinesisRecordHandler(record));
+): Promise<string> => {
+  const body = record.kinesis.data;
+  if (body.includes('fail')) {
+    throw new Error('Failed to process record.');
+  }
+  await new Promise((resolve) => setTimeout(resolve, 1));
```
svozza: You can use the built in promise version of `setTimeout` rather than manually using the promise constuctor:

```ts
import {setTimeout} from 'node:timers/promises;

// ...

await setTimeout(1);
```

dreamorosi: Nice, didn't know this existed - changing it.

packages/batch/tests/unit/processPartialResponse.test.ts
```diff
@@ -36,7 +36,7 @@ describe('Function: processPartialResponse()', () => {
     context,
   };
 
-  const handlerWithSqsEvent = async (
+  const handlerWithSqsEvent = (
```
svozza, line 39: While we aren't actually using `await` in this function, it is still useful information to know that this function returns a promise so can we add a return value to the type signature here?

dreamorosi: Fair, added the `Promise<PartialItemFailureResponse>` return type

packages/batch/tests/unit/processPartialResponse.test.ts
```diff
@@ -51,7 +51,7 @@ describe('Function: processPartialResponse()', () => {
     return handler(event, context);
   };
 
-  const handlerWithKinesisEvent = async (
+  const handlerWithKinesisEvent = (
```
svozza, line 54: Can we do return types with the other two here so it's consistent?

dreamorosi: done!

PR #4566, opened 2025-09-24
style(testing): apply stricter linting

### LINKED ISSUE #4559: Maintenance: fix linting issues for `testing` package
### Summary

In #4549 we're introducing new and/or stricter linting rules. Because of this, we need to review existing code for the `testing` package and make sure it's passing the linting.

While working on it, we could also address some SonarCloud issues that have popped up over time.

### Why is this needed?

See rationale in parent issue (#4549).

_No response_

### Solution

_No response_

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR resolves all the linting issues resulting from the new rules being introduced in #4545.

While working on this, I also addressed a few SonarCloud issues that popped up over time.

> [!important]
> This PR is expected to have failing CI, and this package (`testing`) doesn't have a CI step that runs on PRs. It's safe to merge since it's not going on `main` but on `ci/fix_linting`.

**Issue number:** closes #4559

---

### DIFF
--- a/packages/testing/src/xray-traces-utils.ts
+++ b/packages/testing/src/xray-traces-utils.ts
@@ -16,7 +16,7 @@ import type {
 
 const retryOptions = {
   retries: 20,
-  minTimeout: 5_000,
+  minTimeout: 5000,
   maxTimeout: 10_000,
   factor: 1.25,
 };
--- a/packages/testing/src/TestStack.ts
+++ b/packages/testing/src/TestStack.ts
@@ -80,6 +80,8 @@ class TestStack {
          * Otherwise, we log messages that are either warnings or errors as well as periodic
          * updates on the stack creation and destruction process.
          *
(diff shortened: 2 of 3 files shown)


packages/testing/src/xray-traces-utils.ts
```diff
@@ -16,7 +16,7 @@ import type {
 
 const retryOptions = {
   retries: 20,
-  minTimeout: 5_000,
+  minTimeout: 5000,
   maxTimeout: 10_000,
```
svozza, line 20: Should this be `10000`? Or is once we get to numbers with 4 or more zeros that we want to use the `_` syntax?

dreamorosi: Sonar says that up to 4 digits it doesn't require `_`, but then beyond that the underscore is used as separator every 3 digits groups 🤷

PR #4573, opened 2025-09-24
feat(event-handler): added `includeRouter` method to split routes

### LINKED ISSUE #4481: Feature request: Split routes with Router - includeRouter method
### Use case

As applications grow and the number of routes a Lambda function handles increases, it becomes natural to either break into smaller Lambda functions or split routes into separate files to ease maintenance.

Currently, the TypeScript event-handler's `Router` class doesn't provide a way to compose multiple router instances, forcing developers to define all routes in a single file or manually merge route definitions.

This creates several challenges:

- **Maintainability**: Large route files become difficult to manage and navigate
- **Team collaboration**: Multiple developers working on different route groups can create merge conflicts
- **Code organization**: Related routes cannot be logically grouped in separate modules
- **Reusability**: Common route patterns cannot be easily shared across different Lambda functions

### Solution/User Experience

Implement an `includeRouter` method on the `Router` class that allows merging routes, context, and middleware from one router instance into another. This would mirror the functionality available in the Python version of Powertools for AWS.

### Basic usage

**routes/todos.ts** (separate route module):

```typescript

export const router = new Router();

const endpoint = 'https://jsonplaceholder.typicode.com/todos';

router.get('/todos', async (_, { request }) => {
  const apiKey = request.headers['x-api-key'];
  
  const response = await fetch(endpoint, {
    headers: { 'X-Api-Key': apiKey }
  });
  const todos = await response.json();
  
  return { todos: todos.slice(0, 10) };
});

router.get('/todos/{todoId}', async ({ todoId }, { request }) => {
  const apiKey = request.headers['x-api-key'];
  
  const response = await fetch(`${endpoint}/${encodeURIComponent(todoId)}`, {
    headers: { 'X-Api-Key': apiKey }
  });
  const todo = await response.json();
  
  return { todo };
});
```

**handler.ts** (main Lambda handler):

```typescript
import { router as todosRouter } from './routes/todos.js';

const logger = new Logger();
const app = new Router({ logger });

// Include routes from separate module
app.includeRouter(todosRouter);

  logger.addContext(context);
};
```

### Route prefix support

Allow prefixing routes when including a router to avoid repetitive path definitions:

```typescript
// Remove '/todos' prefix from route definitions
router.get('/', async (_, { request }) => { /* get all todos */ });
router.get('/{todoId}', async ({ todoId }, { request }) => { /* get specific todo */ });

// Include with prefix
app.includeRouter(todosRouter, { prefix: '/todos' });
```

### Context sharing with appendContext

Support sharing contextual data between router instances:

```typescript
  app.appendContext({ isAdmin: true, userId: 'user123' });
};
```

Routes in included routers can access this context:

```typescript
router.get('/todos', async (_, { request, reqCtx }) => {
  const isAdmin = context.get('isAdmin', false);
  // Use context in route logic
});
```

### Method signature

```typescript
class Router {
  includeRouter(router: Router, options?: { prefix?: string }): void;
  appendContext(data: Record<string, unknown>): void;
}
```

### Alternative solutions

N/A

## Open questions

The `appendContext` method allows adding arbitrary key-value pairs to the context.

This is not the Lambda function's `context` object but a shared request-specific context - hence the suggested `reqCtx` name.

Is this clear enough, or should we consider a different name? Should we also rename the `context` property currently used in the `RequestContext` to something like `lambdaContext` to avoid confusion?

Should we instead add the context directly to the `RequestContext` object, making it accessible in route handlers as:

```typescript
router.get('/todos', async (_, { isAdmin }) => {});
```

And if we do this, how do we handle potential naming conflicts with existing properties?

Also, I want to make sure this context is future proof in case as I want to make it easier to surface the context from [Lambda Authorizers](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-lambda-authorizer.html#http-api-lambda-authorizer.example-code) that is currently buried in the `request.requestContext.authorizer` property.

### Acknowledgment

### Future readers

sdangol:
## Summary

This PR adds an `includeRouter` method to allows merging routes, context, and middleware from one router instance into another. 

### Changes

- Added `merge` methods on the Route Handler and Error Handler registries to merge error handlers and routes
- Added `includeRouter` method on the Router to merge context, registries and middlewares

**Issue number:** closes #4481 

---

### DIFF
--- a/packages/event-handler/src/rest/ErrorHandlerRegistry.ts
+++ b/packages/event-handler/src/rest/ErrorHandlerRegistry.ts
@@ -76,6 +76,8 @@ export class ErrorHandlerRegistry {
    * Merges another {@link ErrorHandlerRegistry | `ErrorHandlerRegistry`} instance into the current instance.
    * It takes the handlers from the provided registry and adds them to the current registry.
    *
+   * Error handlers from the included router are merged with existing handlers. If handlers for the same error type exist in both routers, the included router's handler takes precedence.
+   *
    * @param errorHandlerRegistry - The registry instance to merge with the current instance
    */
   public merge(errorHandlerRegistry: ErrorHandlerRegistry): void {
--- a/packages/event-handler/src/rest/ErrorHandlerRegistry.ts
+++ b/packages/event-handler/src/rest/ErrorHandlerRegistry.ts
@@ -71,4 +71,19 @@ export class ErrorHandlerRegistry {
 
     return null;
   }
+
+  /**
+   * Merges another {@link ErrorHandlerRegistry | `ErrorHandlerRegistry`} instance into the current instance.
+   * It takes the handlers from the provided registry and adds them to the current registry.
+   *
+   * @param errorHandlerRegistry - The registry instance to merge with the current instance
+   */
+  public merge(errorHandlerRegistry: ErrorHandlerRegistry): void {
+    for (const [
+      errorConstructor,
+      errorHandler,
+    ] of errorHandlerRegistry.#handlers) {
+      this.register(errorConstructor, errorHandler);
+    }
+  }
 }
--- a/packages/event-handler/src/rest/RouteHandlerRegistry.ts
+++ b/packages/event-handler/src/rest/RouteHandlerRegistry.ts
@@ -8,7 +8,7 @@ import type {
   ValidationResult,
 } from '../types/rest.js';
 import { ParameterValidationError } from './errors.js';
-import type { Route } from './Route.js';
+import { Route } from './Route.js';
 import { compilePath, validatePathPattern } from './utils.js';
 
 class RouteHandlerRegistry {
@@ -192,6 +192,43 @@ class RouteHandlerRegistry {
     }
 
     return null;
+  }
+
+  /**
+   * Merges another {@link RouteHandlerRegistry | `RouteHandlerRegistry`} instance into the current instance.
+   * It takes the static and dynamic routes from the provided registry and adds them to the current registry.
+   *
+   * @param routeHandlerRegistry - The registry instance to merge with the current instance
+   * @param options.prefix - An optional prefix to be added to the paths defined in the router
+   */
+  public merge(
+    routeHandlerRegistry: RouteHandlerRegistry,
+    options?: { prefix: Path }
+  ): void {
+    for (const route of routeHandlerRegistry.#staticRoutes.values()) {
+      this.register(
+        new Route(
+          route.method as HttpMethod,
+          options?.prefix
+            ? route.path === '/'
+              ? options.prefix
+              : `${options.prefix}${route.path}`
+            : route.path,
+          route.handler,
+          route.middleware
+        )
+      );
+    }
+    for (const route of routeHandlerRegistry.#dynamicRoutes) {
+      this.register(
+        new Route(
+          route.method as HttpMethod,
+          options?.prefix ? `${options.prefix}${route.path}` : route.path,
+          route.handler,
+          route.middleware
+        )
+      );
+    }
   }
 }
 
--- a/packages/event-handler/src/rest/RouteHandlerRegistry.ts
+++ b/packages/event-handler/src/rest/RouteHandlerRegistry.ts
@@ -9,7 +9,11 @@ import type {
 } from '../types/rest.js';
 import { ParameterValidationError } from './errors.js';
 import { Route } from './Route.js';
-import { compilePath, validatePathPattern } from './utils.js';
+import {
+  compilePath,
+  resolvePrefixedPath,
+  validatePathPattern,
+} from './utils.js';
 
 class RouteHandlerRegistry {
   readonly #staticRoutes: Map<string, Route> = new Map();
@@ -199,6 +203,7 @@ class RouteHandlerRegistry {
    * It takes the static and dynamic routes from the provided registry and adds them to the current registry.
    *
    * @param routeHandlerRegistry - The registry instance to merge with the current instance
+   * @param options - Configuration options for merging the router
    * @param options.prefix - An optional prefix to be added to the paths defined in the router
    */
   public merge(
@@ -210,17 +215,10 @@ class RouteHandlerRegistry {
       ...routeHandlerRegistry.#dynamicRoutes,
     ];
     for (const route of routes) {
-      let path = route.path;
-      if (options?.prefix) {
-        path =
-          route.path === '/'
-            ? options.prefix
-            : `${options.prefix}${route.path}`;
-      }
       this.register(
         new Route(
           route.method as HttpMethod,
-          path,
+          resolvePrefixedPath(route.path, options?.prefix),
           route.handler,
           route.middleware
         )
--- a/packages/event-handler/src/rest/RouteHandlerRegistry.ts
+++ b/packages/event-handler/src/rest/RouteHandlerRegistry.ts
@@ -202,6 +202,8 @@ class RouteHandlerRegistry {
    * Merges another {@link RouteHandlerRegistry | `RouteHandlerRegistry`} instance into the current instance.
    * It takes the static and dynamic routes from the provided registry and adds them to the current registry.
    *
+   * Routes from the included router are added to the current router's registry. If a route with the same method and path already exists, the included router's route takes precedence.
+   *
    * @param routeHandlerRegistry - The registry instance to merge with the current instance
    * @param options - Configuration options for merging the router
    * @param options.prefix - An optional prefix to be added to the paths defined in the router
--- a/packages/event-handler/src/rest/RouteHandlerRegistry.ts
+++ b/packages/event-handler/src/rest/RouteHandlerRegistry.ts
@@ -205,25 +205,22 @@ class RouteHandlerRegistry {
     routeHandlerRegistry: RouteHandlerRegistry,
     options?: { prefix: Path }
   ): void {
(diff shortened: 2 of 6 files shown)


packages/event-handler/src/rest/Router.ts
```diff
@@ -551,6 +551,43 @@ class Router {
...
+   * @param options.prefix - An optional prefix to be added to the paths defined in the router
+   */
+  public includeRouter(router: Router, options?: { prefix: Path }): void {
+    this.context = {
+      ...this.context,
+      ...router.context,
+    };
+    this.routeRegistry.merge(router.routeRegistry, options);
+    this.errorHandlerRegistry.merge(router.errorHandlerRegistry);
+    this.middleware.push(...router.middleware);
```
dreamorosi, line 594: What happens if there are overlapping routes or middleware functions?

Is there any warning? Also whatever is the behavior, can we put a mention of it in the doc string of the method?

packages/event-handler/src/rest/Router.ts
```diff
@@ -551,6 +551,43 @@ class Router {
...
+   * });
+   *
+   * const app = new Router();
+   * app.includeRouter(todosRouter);
+   *
+   * export const handler = async (event: unknown, context: Context) => {
+   *   return app.resolve(event, context);
+   * };
+   * ```
+   * @param router - The Router from which to merge the routes, context and middleware
```
dreamorosi:
```suggestion
   * @param router - The `Router` from which to merge the routes, context and middleware
```

packages/event-handler/src/rest/Router.ts
```diff
@@ -551,6 +551,43 @@ class Router {
...
+   * @param options.prefix - An optional prefix to be added to the paths defined in the router
```
dreamorosi, line 585: I think we need a line for `@param options` alone above, or it won't have anything before you destructure it (same above)

packages/event-handler/tests/unit/rest/Router/basic-routing.test.ts
```diff
@@ -144,4 +144,47 @@ describe('Class: Router - Basic Routing', () => {
...
+
+    const taskRouter = new Router();
+    taskRouter.get('/', async () => ({ api: 'listTasks' }));
+    taskRouter.post('/create', async () => ({ api: 'createTask' }));
+    taskRouter.get('/:taskId', async () => ({ api: 'getTask' }));
+
+    const app = new Router();
+    app.includeRouter(baseRouter);
+    app.includeRouter(todoRouter, { prefix: '/todos' });
+    app.includeRouter(taskRouter, { prefix: '/todos/:id/tasks' });
```
dreamorosi: I feel like this test has a lot of redundant routes that only add noise, we don't want to re test the routing - I think having two routers with one route each and a middleware would would cover the logic we added.

dreamorosi: I feel like this test has a lot of redundant routes that only add noise, we don't want to re test the routing - I think having two routers with one route each and a middleware would would cover the logic we added.

dreamorosi (changes requested):
Thanks for the PR, left a couple comments and a question.

packages/event-handler/src/rest/Router.ts
svozza, line 594: This is going to cause issues if both routers have the same middleware. For example if both have `compress` then it will run twice. I'm not sure there's much we can do about that but we should at least call it out somewhere in the docs.

packages/event-handler/src/rest/RouteHandlerRegistry.ts
```diff
@@ -193,6 +193,40 @@ class RouteHandlerRegistry {
...
   }
+
+  /**
+   * Merges another {@link RouteHandlerRegistry | `RouteHandlerRegistry`} instance into the current instance.
+   * It takes the static and dynamic routes from the provided registry and adds them to the current registry.
+   *
+   * @param routeHandlerRegistry - The registry instance to merge with the current instance
+   * @param options.prefix - An optional prefix to be added to the paths defined in the router
+   */
+  public merge(
```
svozza, line 211: We should document that in the case of a route with the same path as an already existing one, that the new route wins. We will already get a warning in the logs because of how `register` is implemented but it does no harm to call it out up front.

packages/event-handler/tests/unit/rest/Router/basic-routing.test.ts
```diff
@@ -144,4 +144,47 @@ describe('Class: Router - Basic Routing', () => {
     expect(JSON.parse(createResult.body).actualPath).toBe('/todos');
     expect(JSON.parse(getResult.body).actualPath).toBe('/todos/1');
   });
+
+  it('routes to the included router when using split routers', async () => {
```
svozza, line 148: We need to have tests cases around the behaviour where a path already exists and is over-written.

packages/event-handler/src/rest/Router.ts
```diff
@@ -551,6 +551,43 @@ class Router {
...
+   */
+  public includeRouter(router: Router, options?: { prefix: Path }): void {
+    this.context = {
```
svozza, line 588: Not for this PR but we need to figure out how this relates to the `Store` abstraction we plan to implement.

packages/event-handler/src/rest/ErrorHandlerRegistry.ts
```diff
@@ -71,4 +71,19 @@ export class ErrorHandlerRegistry {
...
     return null;
   }
+
+  /**
+   * Merges another {@link ErrorHandlerRegistry | `ErrorHandlerRegistry`} instance into the current instance.
+   * It takes the handlers from the provided registry and adds them to the current registry.
+   *
+   * @param errorHandlerRegistry - The registry instance to merge with the current instance
+   */
+  public merge(errorHandlerRegistry: ErrorHandlerRegistry): void {
```
svozza, line 83: As with routes, we should call out that last one wins here in the case of error handlers associated with the same `Error` constructor that already exist in the registry.

packages/event-handler/src/rest/Router.ts
sdangol: In case of the routes and the error handlers, since we're reusing the `register` method to merge them, they would give warnings. But, currently merging the context and middleware wouldn't. Let me put the warning and mention in the doc string

packages/event-handler/tests/unit/rest/Router/basic-routing.test.ts
sdangol: [This part of the code](https://github.com/aws-powertools/powertools-lambda-typescript/pull/4573/files#diff-9dfa6eafe308fe2d374a90faf626f53f51bef1adade35eea55ed1f00fc0c4025R214-R219) has some conditions which led to have a combination of prefixed/non-prefixed, static/dynamic routes, route with/without `/`. Hence, the abomination of tests 😄 I'll see if I can improve it

packages/event-handler/src/rest/Router.ts
sdangol: Yeah, in case of middleware, it would be tricky to check for this. Maybe it could be intentional as well? I will call it out. What do you think about the order of the middleware, does it make sense?

sdangol: @dreamorosi @svozza About the warnings on merging the context, I think the context that would be passed would usually be a Lambda context. Do you think we should have the warnings on that?

sdangol: Sorry, my mistake, the context that we have here is different from the one passed during the resolve function

svozza: Yeah, it's confusing. As I said in my other comment, do we want to use the `Store` abstraction we're proposing. I think it's OK to punt on this for now and we can come back to it.

svozza: Because middleware are just functions, there's no reliable way to know what a middleware does so yeah, documentation is all we can do. Regarding the order, if routes are last one wins then it would follow that new middleware takes precendence too we should actually be prepending here not appending.

svozza: So basically we have two contexts, one that lives for the lifetime of the request and then one that lives for the lifetime of the router. Currently this second one is not used, it was just in the initial implementation for the router we had back in the beginning.

sdangol: Yeah ok, I'll address that when working on the `Store` abstraction for the context sharing

sdangol: I was thinking with the mental model of the onion pattern, it would be like the outer router wrapping the inner router inside it. So, incase of the middleware as well the precedence would be kind of like that? In case of routing, the inner ones would take precedence

packages/event-handler/tests/unit/rest/Router/basic-routing.test.ts
sdangol: I've updated the tests now by testing the prefixed path resolution behaviour separately

packages/event-handler/src/rest/Router.ts
svozza: Oh yeah, I get you because the new middleware is more specific so it should go last as you've done here.

dreamorosi: Yea, for now let's go with this - but I'll create an issue for the `Store` incorporating the discussion we had in the linked issue.

PR #4578, opened 2025-09-25
 style(event-handler): apply stricter linting

### LINKED ISSUE #4575: Maintenance: fix linting issues for the `event-handler` package
### Summary

In #4549 we're introducing new and/or stricter linting rules. Because of this, we need to review existing code for the `event-handler` package and make sure it's passing the linting.

While working on it, we could also address some SonarCloud issues that have popped up over time.

### Why is this needed?

See rationale in parent issue (#4549).

_No response_

### Solution

_No response_

### Acknowledgment

### Future readers

sdangol:
## Summary

This PR resolves all the linting issues resulting from the new rules being introduced in #4545

### Changes

- Removed `async` from functions not doing `await`

**Issue number:** closes #4575 

---

### DIFF
--- a/packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
+++ b/packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
@@ -174,7 +174,7 @@ class AppSyncGraphQLResolver extends Router {
     }
 
     return this.#withErrorHandling(
-      async () => await this.#executeSingleResolver(event, context, options),
+      () => this.#executeSingleResolver(event, context, options),
       event,
       options
     );
@@ -190,7 +190,7 @@ class AppSyncGraphQLResolver extends Router {
    * @param options - Optional resolve options for customizing resolver behavior.
    */
   async #withErrorHandling(
-    fn: () => Promise<unknown>,
+    fn: () => Promise<unknown> | unknown,
     event: AppSyncResolverEvent<Record<string, unknown>>,
     options?: ResolveOptions
   ) {
@@ -377,7 +377,7 @@ class AppSyncGraphQLResolver extends Router {
     event: AppSyncResolverEvent<Record<string, unknown>>,
     context: Context,
     options?: ResolveOptions
-  ) {
+  ): Promise<unknown> | unknown {
     const { fieldName, parentTypeName: typeName } = event.info;
 
     const resolverHandlerOptions = this.resolverRegistry.resolve(
--- a/packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
+++ b/packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
@@ -152,11 +152,7 @@ class AppSyncGraphQLResolver extends Router {
    * @param context - The AWS Lambda context object.
    * @param options - Optional parameters for the resolver, such as the scope of the handler.
    */
-  public async resolve(
-    event: unknown,
-    context: Context,
-    options?: ResolveOptions
-  ): Promise<unknown> {
+  public resolve(event: unknown, context: Context, options?: ResolveOptions) {
     if (Array.isArray(event)) {
       if (event.some((e) => !isAppSyncGraphQLEvent(e))) {
         this.logger.warn(
@@ -178,7 +174,7 @@ class AppSyncGraphQLResolver extends Router {
     }
 
     return this.#withErrorHandling(
-      () => this.#executeSingleResolver(event, context, options),
+      async () => await this.#executeSingleResolver(event, context, options),
       event,
       options
     );
@@ -197,7 +193,7 @@ class AppSyncGraphQLResolver extends Router {
     fn: () => Promise<unknown>,
     event: AppSyncResolverEvent<Record<string, unknown>>,
     options?: ResolveOptions
-  ): Promise<unknown> {
+  ) {
     try {
       return await fn();
     } catch (error) {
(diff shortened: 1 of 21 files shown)


sdangol:
@svozza There were some `any` being used in [this file](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/event-handler/tests/unit/rest/converters.test.ts#L86). There were some tests which was checking for handling null values. I wanted to check with you if those tests are needed if the `null` values aren't allowed for those parameters like `headers`, `multiValueHeaders`?

packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
```diff
@@ -178,7 +174,7 @@ class AppSyncGraphQLResolver extends Router {
     }
 
     return this.#withErrorHandling(
-      () => this.#executeSingleResolver(event, context, options),
+      async () => await this.#executeSingleResolver(event, context, options),
```
dreamorosi: Can we just pass the unresolved promise here?

packages/event-handler/src/types/rest.ts
```diff
@@ -57,7 +57,7 @@ type HandlerResponse = Response | JSONObject;
 
 type RouteHandler<TReturn = HandlerResponse> = (
   reqCtx: RequestContext
-) => Promise<TReturn>;
+) => Promise<TReturn> | TReturn;
```
dreamorosi, line 61: Nice

packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
svozza: Yeah, because as this is with this change, it slightly changes the behaviour from what it was.

packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
```diff
@@ -377,11 +373,11 @@ class AppSyncGraphQLResolver extends Router {
    * @param options - Optional parameters for the resolver, such as the scope of the handler.
    * @throws {ResolverNotFoundException} If no resolver is registered for the given field and type.
    */
-  async #executeSingleResolver(
+  #executeSingleResolver(
     event: AppSyncResolverEvent<Record<string, unknown>>,
     context: Context,
     options?: ResolveOptions
-  ): Promise<unknown> {
+  ) {
```
svozza: This function can still return a promise, the type signature should reflect that.

packages/event-handler/src/types/rest.ts
```diff
@@ -78,12 +78,16 @@ type RestRouteOptions = {
   middleware?: Middleware[];
 };
 
-type NextFunction = () => Promise<HandlerResponse | void>;
+type NextFunction = () =>
```
svozza: I'm worried this change is going to encourage people not to await the `next` function because they think it being async is optional. Failing to await `next` causes the an error to be thrown and the whole requst fails.

dreamorosi: I got an idea (haven't tested it tho):

What if whenever someone calls `app.use` we wrap their middleware arrow function in a function we control, before putting the wrapped middleware in the array.

Our function becomes the `next` and doesn't care if it's called with `await` or not, but it will always `await` the function it's wrapping before returning its result.

This way (assuming I'm not just saying something dumb), it doesn't matter if customers forget to call `await next` or even if the middleware is asynchronous or synchronous, we'll always treat it as async.

dreamorosi:
Hi @sdangol - can you please address the feedback and get this merged in the next couple hrs? Thank you.

packages/event-handler/src/types/rest.ts
svozza: So the issue is not with awaiting the middleware, it's the step after that, which is actually happening inside the the middlware handler. Not calling `await` returns immediately and messes up the execution order. Even when I tested capturing the Promise and forcing an await inside the `composeMiddleware` function, the problem was still there.

packages/event-handler/src/appsync-graphql/AppSyncGraphQLResolver.ts
sdangol: There are quite a few uses of `unknown` in the `AppSyncGraphQLResolver`. Do you have some suggestions on how we could we narrow down the return types for this?

svozza:
> @svozza There were some `any` being used in [this file](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/event-handler/tests/unit/rest/converters.test.ts#L86). There were some tests which was checking for handling null values. I wanted to check with you if those tests are needed if the `null` values aren't allowed for those parameters like `headers`, `multiValueHeaders`?

Ah yeah, that test is a hangover from a bug where we used to accept `null` as a value with headers but it can only ever be undefined (see the type [definition](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/bc685e16eb9e0c33df331e571993dd814aea9e20/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L111)) so you can get rid of it.

sdangol:
I'm struggling a bit to fix this one build issue in the cors middleware where we exit early without returning anything.

image

packages/event-handler/src/types/rest.ts
```diff
@@ -24,6 +24,7 @@ type ErrorHandler<T extends Error = Error> = (
 ) => Promise<HandlerResponse>;
 
 interface ErrorConstructor<T extends Error = Error> {
+  // biome-ignore lint/suspicious/noExplicitAny: this is a generic type that is intentionally open
```
sdangol, line 27: @svozza @dreamorosi Can we narrow this down to something? I wasn't sure if we could

packages/event-handler/src/types/rest.ts
dreamorosi: ok then - @sdangol I think this discussion is relevant to your question below.

packages/event-handler/src/types/rest.ts
svozza: I guess the issue here is supporting custom error types, which can have arbitrary parameters when constructed:

```ts
class CustomError extends Error {
  constructor(/* these parameters can be literally anything */) {
  }
}
```

sdangol:
@svozza @dreamorosi For the use of void in unions, I wasn't able to find a way through it so I added an ignore for the lint rule. Narrowing the `unknowns` on the `AppSyncResolver` might need more work and might not be suitable to block this PR. Rest of the issues, I've addressed them

dreamorosi (approved):

PR #4582, opened 2025-09-26
feat(event-handler): added support for catch all route

### LINKED ISSUE #4579: Feature request: Support for a catch all route in Event Handler REST API
### Use case

Currently, the Event Handler supports specific route patterns, but there's no built-in support for catch-all routes that can handle arbitrary paths. This is useful for:

- Creating fallback handlers for unmatched routes
- Building proxy-like functionality where you want to capture all requests under a certain path
- Implementing wildcard routing patterns

### Solution/User Experience

Add support for catch-all routes using regex patterns like `.+` that can match any path. This would allow developers to create flexible routing patterns that can handle multiple path segments.

```typescript
import { APIGatewayResolver } from '@aws-lambda-powertools/event-handler/resolvers';

const app = new APIGatewayResolver();

app.get('.+', () => {
  return {
    message: 'Hello, World!'
  };
});

export const handler = async (event, context) =>
  app.resolve(event, context);
```

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

sdangol:
## Summary

This PR adds the feature to support catch all routes in the Event Handler.

### Changes

- Added a `regexRoutes` property to store routes with regex
- Updated the register function to register the route to `regexRoutes` if the path had any non-alphanumeric characters except for `/`, `-` or `:`
- Updated the resolve function to match for the `regexRoutes` in the order of their registration. 

**Issue number:** closes #4579 

---


packages/event-handler/src/rest/RouteHandlerRegistry.ts
```diff
@@ -103,6 +104,18 @@ class RouteHandlerRegistry {
 
     const compiled = compilePath(route.path);
 
+    if (/[^a-zA-Z0-9/-:]/.test(compiled.path)) {
```
sdangol: @svozza @dreamorosi I'm using this test to check if the path is a regex path. Rather than this, would it make sense to actually use a `RegExp` instead?
So, we would have
```typescript
app.get(".+", () => {}) // Static/Dynamic Routes
app.get(/.+/, () => {}) // Regex Routes
```

dreamorosi: I think yes, especially if it allows us to make the `Path` type more strict again.

Before this PR it was `/{string}` but now we had to remove the `/`. If we can make it like below, it'd be great:

```ts
type Path =  `/${string}` | RegExp
```

svozza: Yeah, agreed, let's do this.

leandrodamascena:
I'll review this in the next few minutes.

packages/event-handler/src/rest/RouteHandlerRegistry.ts
```diff
@@ -103,6 +104,18 @@ class RouteHandlerRegistry {
...
+        this.#logger.warn(
+          `Handler for method: ${route.method} and path: ${route.path} already exists. The previous handler will be replaced.`
+        );
+      }
+      this.#regexRoutes.set(route.id, {
+        ...route,
+        ...compiled,
+      });
+      return;
+    }
```
leandrodamascena: Overall, this logic seems sound to me. This regular expression is capable of capturing routes like `/users`, `/api/v1/orders`, `/users/:id`, and others, and that's sufficient. One thing this code doesn't do, and Python doesn't either, is sanitize the input. I mean, someone using traversal paths like `/api/../../etc/passwd` could input data that shouldn't be accepted, but to be honest, I think that's just my brain overcomplicating things and thinking out loud. This should be handled at the WAF level, and we should probably improve the documentation with some warnings that customers should be aware of.

packages/event-handler/src/rest/RouteHandlerRegistry.ts
```diff
@@ -171,28 +184,10 @@ class RouteHandlerRegistry {
...
-          handler: route.handler,
-          params: processedParams,
-          rawParams: params,
-          middleware: route.middleware,
-        };
-      }
+    const routes = [...this.#dynamicRoutes, ...this.#regexRoutes.values()];
+    for (const route of routes) {
+      const result = this.#processRoute(route, method, path);
+      if (result) return result;
```
leandrodamascena, line 192: I don't know all of this codebase and it's easy to ask than read it all. Does this code mean that dynamic/static routes will be merged before regex routes and will take precedence in route resolution? If so, this is the right and expected behavior.

packages/event-handler/tests/unit/rest/Router/basic-routing.test.ts
```diff
@@ -190,4 +190,32 @@ describe('Class: Router - Basic Routing', () => {
...
+    app.get('/files/.+', async () => ({ api: 'serveFileOverride' }));
+    app.get('/api/v\\d+/.*', async () => ({ api: 'apiVersioning' }));
+    app.get('/users/:userId/files/.+', async (reqCtx) => ({
+      api: `dynamicRegex${reqCtx.params.userId}`,
+    }));
+    app.get('.+', async () => ({ api: 'getAnyRoute' }));
+    app.route(async () => ({ api: 'catchAllUnmatched' }), {
+      path: '.*',
+      method: [HttpVerbs.GET, HttpVerbs.POST],
+    });
```
leandrodamascena, line 213: I think this test is good enough to make sure we are covering the expected routes.

packages/event-handler/src/rest/RouteHandlerRegistry.ts
sdangol: Yes, we do exact matching for the static routes, then sort the dynamicRoutes in precedence based on specificity and do the matching. The regex routes will be evaluated at the end in the order of registration.

leandrodamascena:
I'll review this PR by EOD.

packages/event-handler/src/rest/RouteHandlerRegistry.ts
```diff
@@ -103,6 +105,18 @@ class RouteHandlerRegistry {
 
     const compiled = compilePath(route.path);
 
+    if (route.path instanceof RegExp) {
```
svozza: How come we're not using the new `isRegExp` function here?

packages/event-handler/src/types/rest.ts
```diff
@@ -79,14 +79,16 @@ type RestRouteOptions = {
   middleware?: Middleware[];
 };
 
-// biome-ignore lint/suspicious/noConfusingVoidType: To ensure next function is awaited
-type NextFunction = () => Promise<HandlerResponse | void>;
+type NextFunction = () => // biome-ignore lint/suspicious/noConfusingVoidType: To ensure next function is awaited
```
svozza: Is this change to allow this function be synchronous meant to be here? I thought we agreed in the other PR to leave this as strictly `async` so people would get feedback from the type system that they should always await the `next` function.

dreamorosi: Yea, I'm confused as well.

I still think we should investigate a long-term solution but it deserves its own discussion/issue. I'm advocating for this because Hono for example doesn't require an `await` in front of `next` and also allows sync handlers - so there must be a way.

I don't mind the async handlers specifically, since in practice most of the time - even with something as simple as `req.body.json()` - you end up using `async`, but the requirement to `await` the `next` is a major foot-gun imo.

The type system will only warn you to make the handler/middleware function `async`, but not to put `await` in front of `next`, so in practice you might still end up seeing the runtime error.

I understand this is not something we can address right now, but I'd still want to track this and look at it when time allows it, even if it requires some degree of internal refactoring of how the middleware stack works.

But yes, this doesn't belong in this PR.

svozza: Are you sure that Hono can support synchronous `next`? Their type is explicitly a `Promise`:
https://github.com/honojs/hono/blob/e3bf8c8913a707e49f374575bdf6c88f0ec7e3f6/src/types.ts#L36.

Every [example in the docs](https://hono.dev/docs/guides/middleware) uses `await` and they also call out that you await it:

> Middleware - should return nothing, will be proceeded to next middleware with `await` next()

sdangol: Yeah, agreed. I'll remove them from here. There were a few tests which were complaining which I had missed in the last PR. I'll address them in another PR later.

svozza: I think I fixed those in the stream PR so we should be OK.

PR #4584, opened 2025-09-29
ci: add `--delete` flag to docs publish command

### LINKED ISSUE #4414: Maintenance: docs still publish older pages like `/core/metrics` that no longer exist

When searching for `aws powertools typescript metrics` on Google, the first link I see is:

- https://docs.powertools.aws.dev/lambda/typescript/latest/core/metrics/

On the Utility Settings section of this page, there is a reference to an environment variable called `POWERTOOLS_METRICS_ENABLED`.

However, I found that the correct solution is to use `POWERTOOLS_METRICS_DISABLED` instead (the exact opposite). You can find this documented here:

- https://docs.powertools.aws.dev/lambda/typescript/latest/features/metrics/#utility-settings

It's look like the `core` section should have been deleted by https://github.com/aws-powertools/powertools-lambda-typescript/pull/3818 but the website still shows it.

https://docs.powertools.aws.dev/lambda/typescript/latest/core/metrics/#utility-settings

### How can we improve?

So maybe delete old `core` section on website would be enough.

### Got a suggestion in mind?

_No response_

### Acknowledgment

dreamorosi:
### Changes

This PR adds the `--delete` flag to the AWS CLI command that deploys the documentation site by means of `aws s3 sync`, this way older pages are removed in newer versions, especially `latest`.

**Issue number:** closes #4414

---

### DIFF
--- a/.github/workflows/reusable_publish_docs.yml
+++ b/.github/workflows/reusable_publish_docs.yml
@@ -109,7 +109,8 @@ jobs:
           set -euo pipefail
           aws s3 sync \
             site/ \
-            "s3://$AWS_DOCS_BUCKET/lambda-typescript/$VERSION/"
+            "s3://$AWS_DOCS_BUCKET/lambda-typescript/$VERSION/" \
+            --delete
       - name: Deploy Docs (Alias)
         env:
           VERSION: ${{ inputs.version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/reusable_publish_docs.yml
```diff
@@ -109,7 +109,8 @@ jobs:
           set -euo pipefail
           aws s3 sync \
             site/ \
-            "s3://$AWS_DOCS_BUCKET/lambda-typescript/$VERSION/"
+            "s3://$AWS_DOCS_BUCKET/lambda-typescript/$VERSION/" \
```
leandrodamascena, line 112: It's good if you need to redeploy docs for the same version. 👍

leandrodamascena (approved):

PR #4586, opened 2025-09-29
feat(event-handler): add streaming functionality

### LINKED ISSUE #4476: Feature request: Handle Response Streaming in Event Handler
### Use case

For Lambda function URLs it is [possible to stream response payloads](https://docs.aws.amazon.com/lambda/latest/dg/configuration-response-streaming.html) back to the client. While Event Handler does not currently support Lambda function URLs, we will in the future. 

We need to investigate how this mode of response impacts the API we plan to launch at GA.

### Solution/User Experience

We want an experience with as little friction as possible but there are a lot of unknowns here so this is by no means the definitive DX:

```ts

const app = new Router({streaming: true});

app.get('/stream', async () => {
  const myStream = createWriteStream()
  myStream.write("Hello ");
  myStream.write("world ");
  myStream.write("from ");
  myStream.write("Lambda!");
  myStream.end();
  return myStream;
});

};
```

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

svozza:
## Summary
This PR adds the logic for returning streaming HTTP responses as currently supported by Lambda Function URLs. As event-handler does not currently support Function URLs, we are not exposing the final `streamify` function that users will use to enable streaming:

```ts
import {Router, streamify} from 'from '@aws-lambda-powertools/event-handler/experimental-rest';'

const app = new Router(); 

// ...

app.get('/test', ({event}) => {
  return {
    headers: {
      'content-type': 'application/json'
    },
    statusCode: 200,
    body: Readable.from(Buffer.from(JSON.stringify(event)))
  }
});

// ...

export handler = streamify(app);
```

This means that we will not have to have a long lived branch with the logic for this feature while we wait on Function URL support.

### Changes

- Add new `resolveStream` method that takes a Lambda `ResponseStream` as one of its arguments
- Adds support for returning both Nodejs and web streams from middleware and route handlers but only when returned as a valid  `APIGatewayProxyResult` type.
- Unit tests for `resolveStream` method and assoicated helper functions.

**Issue number:** closes #4476

---


packages/event-handler/src/rest/converters.ts
```diff
@@ -189,8 +222,24 @@ const handlerResultToProxyResult = async (
   response: HandlerResponse,
   statusCode: HttpStatusCode = HttpStatusCodes.OK
 ): Promise<APIGatewayProxyResult> => {
-  if (isAPIGatewayProxyResult(response)) {
-    return response;
+  if (isExtendedAPIGatewayProxyResult(response)) {
+    if (isString(response.body)) {
+      return {
+        ...response,
+        body: response.body,
```
sdangol, line 245: We could just directly return the response if not changing anything

packages/event-handler/src/rest/converters.ts
```diff
@@ -189,8 +222,24 @@ const handlerResultToProxyResult = async (
...
+      isNodeReadableStream(response.body) ||
+      isWebReadableStream(response.body)
+    ) {
+      const nodeStream = bodyToNodeStream(response.body);
+      return {
+        ...response,
+        isBase64Encoded: true,
+        body: await nodeStreamToBase64(nodeStream),
+      };
+    }
```
sdangol, line 258: Are there any other conditions other than the two checks? Should we have a fallback? Could we skip the first `isString` check and `return response` as a fallback?

packages/event-handler/src/types/common.ts
```diff
@@ -1,5 +1,4 @@
-import type { AppSyncEventsResolver } from '../appsync-events/AppSyncEventsResolver.js';
-import type { AppSyncGraphQLResolver } from '../appsync-graphql/AppSyncGraphQLResolver.js';
```
sdangol, line 2: These types are still being used in the docstring comments, so shouldn't be removed

packages/event-handler/tests/unit/rest/Router/middleware.test.ts
```diff
@@ -192,6 +193,7 @@ describe('Class: Router - Middleware', () => {
 
       app.use(({ next }) => {
         next();
+        return Promise.resolve();
```
sdangol: This might not be needed now

packages/event-handler/package.json
```diff
@@ -125,7 +125,8 @@
     "url": "https://github.com/aws-powertools/powertools-lambda-typescript/issues"
   },
   "dependencies": {
-    "@aws-lambda-powertools/commons": "2.27.0"
+    "@aws-lambda-powertools/commons": "2.27.0",
+    "lambda-stream": "0.6.0"
```
sdangol: Should we have a hard dependency on this? If someone's not using streaming?

packages/event-handler/src/rest/converters.ts
svozza: I got a compiler error when I did that because the types are slightly different (`ExtendedAPIGatewayProxyResult` vs `APIGatewayProxyResult`)

packages/event-handler/tests/unit/rest/Router/middleware.test.ts
svozza: I was getting linting errors because there was no `await` in the async function.

packages/event-handler/package.json
svozza: I would rather avoid the complexity of the optional imports after all the trouble it caused in `BatchProcessor`, this dependency only has around 50-60 LOC. Although ultimately I think we might just end up rolling this ourselves for GA.

packages/event-handler/src/rest/converters.ts
svozza: As I said, because of the typing issue, we'd have to do a cast to `APIGatewayProxyResult`. It's safer to do it the way it is now because we're being explicit.

packages/event-handler/src/rest/utils.ts
```diff
@@ -209,3 +236,30 @@ export const resolvePrefixedPath = (path: Path, prefix?: Path): Path => {
   }
   return path;
 };
+
+export const HttpResponseStream =
+  globalThis.awslambda?.HttpResponseStream ??
+  class HttpResponseStream {
```
dreamorosi: Can we add this at the top of this class, so biome doesn't complain

> // biome-ignore lint/complexity/noStaticOnlyClass: required to polyfill AWS Lambda runtime API locally

packages/event-handler/src/types/common.ts
```diff
@@ -42,4 +43,9 @@ type ResolveOptions = {
   scope?: unknown;
 };
 
-export type { Anything, ResolveOptions };
+type ResolveStreamOptions = {
```
dreamorosi: Can we improve the docs of this method, similar to the one above? This is one of the outermost APIs, so it should be as descriptive as possible.

packages/event-handler/tests/unit/rest/Router/middleware.test.ts
dreamorosi: In this case I'd advocate for putting a `// @ ts-expect-error` with an explanation that we're explicitly doing something wrong (aka not awaiting the `next()`) to confirm it throws as expected.

Imo that'd be more descriptive and intentional, this way we're just silencing the linter but I could see myself removing it or questioning why the `Promise.resolve()` is here in 2 mo time.

packages/event-handler/src/types/common.ts
svozza: Ah yes, initially this type didn't exist and it was part of the `ResolveOptions`, I forgot to pull out the stream spcefic doc info when I decided to split them up.

packages/event-handler/package.json
dreamorosi: Should we just roll this up in our repo right now? We can copy/paste and attribute it right now like this and forget about it:

```ts
/**
 * Class to polyfill the AWS Lambda runtime streaming API for local and CI environments.
 * 
 * Credits: astuyve/lambda-stream - MIT
 */
export class ResponseStream extends Stream.Writable {
  private response: Buffer[]
  _contentType?: string
  _isBase64Encoded?: boolean

  constructor() {
    super()
    this.response = []
  }
  // @param chunk Chunk of data to unshift onto the read queue. For streams not operating in object mode, `chunk` must be a string, `Buffer`, `Uint8Array` or `null`. For object mode
  // streams, `chunk` may be any JavaScript value.
  _write(
    chunk: any,
    encoding: BufferEncoding,
    callback: (error?: Error | null) => void
  ): void {
    this.response.push(Buffer.from(chunk, encoding))
    callback()
  }

  getBufferedData(): Buffer {
    return Buffer.concat(this.response)
  }

  setContentType(contentType: string) {
    this._contentType = contentType
  }

  setIsBase64Encoded(isBase64Encoded: boolean) {
    this._isBase64Encoded = isBase64Encoded
  }
}
```

packages/event-handler/src/types/common.ts
svozza: Also, this shouldn't be in the `common` file as it's specific to the the HTTP API event handler.

packages/event-handler/package.json
svozza: Yeah, I'm happy to do that.

packages/event-handler/src/types/common.ts
```diff
@@ -42,4 +43,9 @@ type ResolveOptions = {
...
+  scope?: unknown;
+  responseStream: ResponseStream;
```
dreamorosi: If I try to use this type in a function together with the `aws-lambda` types, I get a TS incompatible type on `responseStream`.

```txt
Type 'HttpResponseStream' is not assignable to type 'ResponseStream'.
  Type 'HttpResponseStream' is missing the following properties from type 'ResponseStream': response, getBufferedData, setIsBase64Encodedts(2322)
common.d.ts(40, 5): The expected type comes from property 'responseStream' which is declared here on type 'ResolveStreamOptions'
(property) responseStream: ResponseStream
```

Try this code (`import type { StreamifyHandler } from 'aws-lambda';`):

```ts
export const handler: StreamifyHandler = awslambda.streamifyResponse(
  async (event, responseStream, context) =>
    app.resolveStream(convertV2ToV1(event), context, {
      responseStream,
    })
);
```

Can we try to align them? In the past we got several customer issues about mismatches between our types and these, and I agree that they're prevalent, especially if you don't want to type each param like I did, and assing a type to the handler.

svozza: Ah yes, I was using the type from the `lambda-stream` library, we should use the one from `aws-lambda`.

svozza: A nice benefit of using the built in lambda type has allowed us remove the `lambda-stream` dependency entirely.

packages/event-handler/src/types/common.ts
svozza: Done

packages/event-handler/tests/unit/rest/Router/middleware.test.ts
svozza: Done

dreamorosi (approved):
Amazing work, this is really nice.

We should write a small short blog post on Builder Hub about this.

PR #4643, opened 2025-10-13
fix(idempotency): add null check for idempotencyHandler before calling handleMiddyOnError

### LINKED ISSUE #4634: Bug: makeHandlerIdempotent error handler throws "Cannot read properties of undefined"
### Expected Behavior

The middy handler `makeHandlerIdempotent` should handle errors without throwing

### Current Behavior

The handler throws `Cannot read properties of undefined (reading 'handleMiddyOnError')` during error handling, when the following is true:
* there is middleware A that executes before powertools `makeHandlerIdempotent`
* while processing a request, middleware A throws an error. this happens before `makeHandlerIdempotent.before()` has executed
* middy invokes the `onError` handler for all middlewares, [even the ones that did not initialize](https://middy.js.org/docs/intro/handling-errors). [this diagram shows the flow for error handling](https://middy.js.org/docs/intro/how-it-works). not in the error example "3rd middleware" didn't execute the before() step, but its onError() is still invoked.
* `makeHandlerIdempotent` throws if `onError()` is executed when `before()` has not.

### Code snippet

```typescript
import middy from '@middy/core'
import middleware1 from 'sample-middleware1'
import middleware2 from 'sample-middleware2'
import { makeHandlerIdempotent } from '@aws-lambda-powertools/idempotency/middleware';

const lambdaHandler = (event, context) => {
  /* your business logic */
}

  .use(middleware1())
  .use(middleware2())
  .use(makeHandlerIdempotent())
  .handler(lambdaHandler)
```

### Steps to Reproduce

1. configure a middy handler as shown above
2. execute a request where middeware1 or middleware2 throw an error in the `before()` step
3. `makeHandlerIdempotent()` throws: `Cannot read properties of undefined (reading 'handleMiddyOnError')`

### Possible Solution

See the snippet below from `packages/idempotency/src/middleware/makeHandlerIdempotent.ts`. The last line does not check if `idempotencyHandler` is `undefined`. just need to add validation and return if it hasn't.

```typescript
  const onError = async (request: MiddyLikeRequest): Promise<void> => {
    if (shouldSkipIdempotency(request)) {
      return;
    }
    const idempotencyHandler =
      getIdempotencyHandlerFromRequestInternal(request);
    await idempotencyHandler.handleMiddyOnError();
  };
```

latest

22.x

### Packaging format used

npm

### Execution logs

```Shell

```

mdesousa:
## Summary

Fixes #4634 

----

Disclaimer: We value your time and bandwidth. As such, any pull requests created on non-triaged issues might not be successful.

### DIFF
--- a/.github/workflows/ossf_scorecard.yml
+++ b/.github/workflows/ossf_scorecard.yml
@@ -43,6 +43,6 @@ jobs:
 
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@e296a935590eb16afc0c0108289f68c87e2a89a5 # v4.30.7
+        uses: github/codeql-action/upload-sarif@f443b600d91635bebf5b0d9ebc620189c0d6fba5 # v4.30.8
         with:
           sarif_file: results.sarif
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,7 +32,7 @@
         "@types/node": "^24.7.1",
         "@vitest/coverage-v8": "^3.2.4",
         "husky": "^9.1.7",
-        "lint-staged": "^16.2.3",
+        "lint-staged": "^16.2.4",
         "markdownlint-cli2": "^0.18.1",
         "middy5": "npm:@middy/core@^5.4.3",
         "middy6": "npm:@middy/core@^6.0.0",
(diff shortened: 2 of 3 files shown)


svozza:
Thanks for the PR. Can you add the acknowledgement text into your issue, just paste it at the bottom:

> By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.
>
> Disclaimer: We value your time and bandwidth. As such, any pull requests created on non-triaged issues might not be successful.

Also, is there a way to have a unit test for when the value is null/undefined. Would be nice to be able to guard against regressions here for a bug that causes a runtime exception.

mdesousa:
hi @svozza unit test and disclamer added 👍

svozza:
LGTM. Thanks a lot!

PR #4663, opened 2025-10-17
feat(metrics): use async local storage for metrics

### LINKED ISSUE #4662: Feature request: Allow use of InvokeStore in Metrics
### Use case

The `InvokeStore` [pacakge](https://www.npmjs.com/package/@aws/lambda-invoke-store), allow for the use of [Asynchronous contexts ](https://nodejs.org/api/async_context.html#asynchronous-context-tracking) in lambda handlers. We should allow users to set metrics in async contexts across lambda executions so that they are isolated.

### Solution/User Experience

```ts
export const handler = async (event, context) => {
  // This metadata with this userId will be scoped only to this handler execution and will not be available in any other invocations
  metrics.addMetadata("userId", event.userId);

  return {
    requestId: InvokeStore.getRequestId(),
    userId,
  };
};

// ... 

  {
    customField: "value", // Optional custom fields
  },
  () => {
    return handler(event)
  }
);
```

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

svozza:
## Summary
This PR adds support for using an async context (specifically from the `InvokeStore` [package](https://www.npmjs.com/package/@aws/lambda-invoke-store)) in the Metrics utility. This allows users to emit metrics that are isolated specifically to the current lambda invocation, isolated from any other executions.

### Changes

- Added specific storage classes for metrics, metadata and dimensions:
  - `MetricsStore`
  - `MetadataStore`
  - `DimensionStore`
- The `Metrics` class only accesses the data in these stores through this interface and never touches the stored objects directly.
- These storage classes check whether they are running in the `InvokeStore` context: if they are then the metrics are stored in the current async context, otherwise we fallback to a plain instance wide object as per the current implementation.
- The storage class handle converting the data into the correct format now rather than the `Metrics` class, e.g., `setMetric` will check if the metric already exists and handle converting values into an array if the metric is already there. Likewise with setting timestamps.
- Added specific concurrency tests to ensure isolation is working as intended for all three stores and the `Metrics` class as a whole.

**Issue number:** closes #4662

---


packages/metrics/src/DimensionsStore.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  addDimensionSet(dimensionSet: Dimensions): void {
+    this.#getDimensionSets().push({ ...dimensionSet });
+  }
+
+  getDimensions(): Dimensions {
+    return { ...this.#getDimensions() };
+  }
+
+  getDimensionSets(): Dimensions[] {
```
svozza: I have possibly over-indexed on immutability here but my reasoning is that I don't want it to be possible for someone to get a reference to the underlying object in the store. Any mutations of stored objects should always be done through the interface and only the inteface.

packages/metrics/tests/unit/concurrency/dimensionsStore.test.ts
```diff
@@ -0,0 +1,156 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { DimensionsStore } from '../../../src/DimensionsStore.js';
+import { sequence } from '../helpers.js';
+
+describe('DimensionsStore concurrent invocation isolation', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it.each([
```
svozza, line 10: All these concurrency tests are the same. They consist of using the `sequence` function defined in `helpers.ts` to run a set of interleaving asynchronous operations such that the test without async context will return an incorrect result due to the shared mutable state it has and the one with aync context will be correct. 

These tests are deterministic and as the logic that runs in each pair of tests is the same except for the fact that they are run using `InvokeStore.run` or not, we can have a high degree of confidence that the classes are working as expected. 

Furthermore, as these tests test the _behaviour_ rather than the implementation, i.e., we are not mocking `InvokeStore` and checking that the correct methods are called, they will be far more resilient to change should we refactor the internals.

packages/metrics/tests/unit/helpers.ts
```diff
@@ -0,0 +1,130 @@
...
+ * const { promise, resolve, reject } = withResolvers<string>();
+ *
+ * // Later, from somewhere else:
+ * resolve('success');
+ *
+ * // Or:
+ * reject(new Error('failed'));
+ * ```
+ */
+const withResolvers = <T>() => {
```
svozza, line 32: Because we still have to support Node 20, I couldn't use the `Promise.withResolvers` built it as it is only in Node 22 and above.

packages/metrics/src/DimensionsStore.ts
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Manages storage of metrics dimensions with automatic context detection.
+ *
+ * This class abstracts the storage mechanism for metrics, automatically
+ * choosing between AsyncLocalStorage (when in async context) and a fallback
+ * object (when outside async context). The decision is made at runtime on
+ * every method call to support Lambda's transition to async contexts.
+ */
+class DimensionsStore {
```
svozza, line 12: I suspect we could create a single store abstraction and have each of these stores implement a facade pattern but for the moment I want to keep it simple. Premature abstractions can cause trouble.

packages/metrics/package.json
```diff
@@ -88,7 +88,8 @@
     "url": "https://github.com/aws-powertools/powertools-lambda-typescript/issues"
   },
   "dependencies": {
-    "@aws-lambda-powertools/commons": "2.27.0"
+    "@aws-lambda-powertools/commons": "2.27.0",
+    "@aws/lambda-invoke-store": "0.1.0"
```
dreamorosi, line 92: Since we added this as dependency of the `commons` pkg, do we need to declare it here again?

packages/metrics/src/DimensionsStore.ts
```diff
@@ -0,0 +1,104 @@
...
+  readonly #dimensionsKey = Symbol('powertools.metrics.dimensions');
```
dreamorosi, line 13: Nice symbol usage

packages/metrics/src/DimensionsStore.ts
```diff
@@ -0,0 +1,104 @@
...
+      | Dimensions[]
+      | undefined;
+    if (stored == null) {
+      stored = [];
+      InvokeStore.set(this.#dimensionSetsKey, stored);
+    }
+    return stored;
+  }
+
+  addDimension(name: string, value: string): string {
```
dreamorosi: Just to be extra explicit, can we add a `public` or `protected` access qualifier to all the methods based on how we want them to be used?

packages/metrics/src/DimensionsStore.ts
```diff
@@ -0,0 +1,104 @@
...
+    return value;
+  }
+
+  addDimensionSet(dimensionSet: Dimensions): Dimensions {
+    this.#getDimensionSets().push({ ...dimensionSet });
+    return dimensionSet;
+  }
+
+  getDimensions(): Dimensions {
+    return { ...this.#getDimensions() };
```
dreamorosi, line 59: Why are we spreading and creating a new object here?

packages/metrics/src/DimensionsStore.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  getDimensionSets(): Dimensions[] {
+    return this.#getDimensionSets().map((set) => ({ ...set }));
+  }
+
+  clearRequestDimensions(): void {
```
dreamorosi: Given we're using `AsyncLocalStorage`, do we need to manually clear up the context or can we expect it to be garbage collected once the handler promise is resolved/rejected? Also does the RIC do any cleanup after the request by chance?

packages/metrics/tests/unit/concurrency/dimensionsStore.test.ts
dreamorosi: Very nice and thought through - great work here!

packages/metrics/tests/unit/helpers.ts
dreamorosi: Can I suggest we put an explicit sentence like "remove when we drop support for Node.js 20"? This way when we do a find and replace for that work we'll find this one too

svozza: Yes, good shout cos I'll forget otherwise!

packages/metrics/package.json
svozza: Oh that's a good point, we may not. Let me play around with it.

packages/metrics/src/DimensionsStore.ts
svozza: Yep, will do!

packages/metrics/src/DimensionsStore.ts
svozza: Because I don't want it to be possible to get a reference to the underlying object and be able to mutate it without going through the store interface.

packages/metrics/src/DimensionsStore.ts
svozza: Yeah, according to the InvokeStore docs the RIC does some cleanup so I think we're OK in this regard.:

> 1. Initializes the InvokeStore context at the beginning of each Lambda invocation
> 2. Sets the requestId and [X-Ray traceId](https://docs.aws.amazon.com/xray/latest/devguide/xray->concepts.html#xray-concepts-traces) in the context
> 3. Ensures proper context isolation between concurrent invocations
> 4.  Cleans up the context after the invocation completes

svozza:
> Very nice work here, I haven't tested it and I won't be able to for a while's but the implementation makes a lot of sense - please if possible run some tests on your machine with the correct setup as we discussed.
>
> Other than that, do you have any idea of whether there's a significant performance impact for this new implementation when used on a Lambda on demand?

Yes, I plan to do some perf testing (vs the original version) in a real lambda function on Monday.

packages/metrics/package.json
svozza: Maybe showing my ignorance of npm workspaces but if it's used in multiple packages across the workspace, should it not be in the root `package.json` file rather than `commons` or `metrics` or whatever?

packages/metrics/tests/unit/helpers.ts
svozza: Done!

packages/metrics/src/DimensionsStore.ts
svozza: Done!

packages/metrics/package.json
dreamorosi: The main `package.json` doesn't get published to npm, only the package ones do. This means it won't be installed when the package gets installed.

In this case, it should be either in both or just in `commons`, since `metrics` depends on it.

When present in both, if the dependency version is aligned, it gets deduplicated, so only one copy gets installed.

In this case I was suggesting to bypass this risk entirely and just keep it as dependency for `commons`, but I don't feel too strongly about it since I don't expect this package to change very often.

svozza: Ah yes, I knew I was missing something. Honestly, I'd rather be explicit so you can see what dependencies a package uses when you look at the dependency file so I'd be in favour of keeping it here.

svozza:
So I did some performance tests, just a very simple lambda that adds a few metrics:

```ts
import { APIGatewayProxyEvent, APIGatewayProxyResult } from 'aws-lambda';

const metrics = new Metrics({ namespace: 'PerfTest', serviceName: 'lambda-perf' });

export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
  // Add dimensions
  metrics.addDimension('Environment', 'test');
  metrics.addDimension('Region', process.env.AWS_REGION || 'us-east-1');
  metrics.addDimension('Version', '1.0.0');
  
  // Add metadata
  metrics.addMetadata('requestId', event.requestContext?.requestId || 'unknown');
  metrics.addMetadata('userAgent', event.headers?.['User-Agent'] || 'unknown');
  metrics.addMetadata('sourceIp', event.requestContext?.identity?.sourceIp || 'unknown');
  
  // Emit 20 different metrics
  metrics.addMetric('RequestCount', 'Count', 1);
  metrics.addMetric('ProcessingTime', 'Milliseconds', Math.random() * 100);
  metrics.addMetric('MemoryUsed', 'Bytes', Math.random() * 1000000);
  metrics.addMetric('CpuUtilization', 'Percent', Math.random() * 100);
  metrics.addMetric('DatabaseConnections', 'Count', Math.floor(Math.random() * 10));
  metrics.addMetric('CacheHits', 'Count', Math.floor(Math.random() * 50));
  metrics.addMetric('CacheMisses', 'Count', Math.floor(Math.random() * 10));
  metrics.addMetric('ErrorRate', 'Percent', Math.random() * 5);
  metrics.addMetric('Throughput', 'Count/Second', Math.random() * 1000);
  metrics.addMetric('Latency', 'Milliseconds', Math.random() * 200);
  metrics.addMetric('NetworkIO', 'Bytes', Math.random() * 50000);
  metrics.addMetric('DiskIO', 'Bytes', Math.random() * 100000);
  metrics.addMetric('ActiveUsers', 'Count', Math.floor(Math.random() * 500));
  metrics.addMetric('QueueDepth', 'Count', Math.floor(Math.random() * 20));
  metrics.addMetric('ResponseSize', 'Bytes', Math.random() * 5000);
  metrics.addMetric('RequestSize', 'Bytes', Math.random() * 2000);
  metrics.addMetric('ConcurrentExecutions', 'Count', Math.floor(Math.random() * 100));
  metrics.addMetric('BusinessMetric1', 'Count', Math.floor(Math.random() * 1000));
  metrics.addMetric('BusinessMetric2', 'Count', Math.floor(Math.random() * 500));
  metrics.addMetric('CustomCounter', 'Count', Math.floor(Math.random() * 250));

  return {
  
  return {
    statusCode: 200,
    headers: {
      'Content-Type': 'application/json'
    },
    body: JSON.stringify({
      message: 'Perf test with 20 metrics emitted',
      timestamp: new Date().toISOString(),
      metricsEmitted: 20
    })
  };
};
```

I just invoked the lambda directly using the SDK. No noticeable performance impact:

```
Results:
Powertools Updated: Avg 111.94ms | Min 63ms | Max 852ms | P95 343ms
Powertools v2.27.0: Avg 111.48ms | Min 62ms | Max 856ms | P95 345ms
```

PR #4668, opened 2025-10-20
feat(logger): use async local storage for logger

### LINKED ISSUE #4667: Feature request: Allow use of InvokeStore in Logger
### Use case

The `InvokeStore` [pacakge](https://www.npmjs.com/package/@aws/lambda-invoke-store), allow for the use of [Asynchronous contexts ](https://nodejs.org/api/async_context.html#asynchronous-context-tracking) in lambda handlers. We should allow users to emit log statements in async contexts across lambda executions so that they are isolated.

### Solution/User Experience

```ts
const logger = new Logger({
  serviceName: 'serverlessAirline',
});

const processTransaction = async (customerId: string): Promise<void> => {
  try {
   // This log key with customerId will be scoped only to this handler execution and will not be available in any other invocations
    logger.appendKeys({
      customerId,
    });

    // ... your business logic

    logger.info('transaction processed');
  } finally {
    logger.resetKeys(); 

  }
};

export const handler = async (
  event: { customerId: string },
  _context: unknown
): Promise<void> => {
  await processTransaction(event.customerId);

  // .. other business logic

  logger.info('other business logic processed');
};

// ... 

  {
    customField: "value", // Optional custom fields
  },
  () => {
    return handler(event)
  }
);
```

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

svozza:
## Summary

This PR adds support for using an async context (specifically from the InvokeStore [package](https://www.npmjs.com/package/@aws/lambda-invoke-store)) in the Logger utility. This allows users to emit log statements that are isolated specifically to the current lambda invocation, isolated from any other executions.

### Changes

- Added a specific storage class for log attributes: `LogAttributesStore`
- The Logger class only accesses the data in this store through this interface and never touches the stored objects directly.
- Updated all methods that stored log attributes and log keys to use the new storage class.
- The storage class checks whether it is running in the `InvokeStore` context: if they are then the log keys are stored in the current async context, otherwise we fallback to a plain instance wide object as per the current implementation.
- Moved the `sequence` testing function to the common testing package.
- Added specific concurrency tests to ensure isolation is working as intended for the `LogAttributesStore` and the`Logger` class as a whole.

**Issue number:** closes #4667 

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10793,6 +10793,7 @@
       "license": "MIT-0",
       "dependencies": {
         "@aws-lambda-powertools/commons": "2.28.1",
+        "@aws/lambda-invoke-store": "0.1.1",
         "lodash.merge": "^4.6.2"
       },
       "devDependencies": {
@@ -10810,6 +10811,15 @@
         "@middy/core": {
           "optional": true
         }
+      }
+    },
+    "packages/logger/node_modules/@aws/lambda-invoke-store": {
+      "version": "0.1.1",
+      "resolved": "https://registry.npmjs.org/@aws/lambda-invoke-store/-/lambda-invoke-store-0.1.1.tgz",
+      "integrity": "sha512-RcLam17LdlbSOSp9VxmUu1eI6Mwxp+OwhD2QhiSNmNCzoDb0EeUXTD2n/WbcnrAYMGlmf05th6QYq23VqvJqpA==",
+      "license": "Apache-2.0",
+      "engines": {
+        "node": ">=18.0.0"
       }
     },
     "packages/metrics": {
--- a/packages/logger/package.json
+++ b/packages/logger/package.json
@@ -99,6 +99,7 @@
   },
   "dependencies": {
     "@aws-lambda-powertools/commons": "2.28.1",
+    "@aws/lambda-invoke-store": "0.1.1",
     "lodash.merge": "^4.6.2"
   },
   "keywords": [


packages/logger/src/Logger.ts
```diff
@@ -687,7 +656,14 @@ class Logger extends Utility implements LoggerInterface {
...
    */
   public setPersistentLogAttributes(attributes: LogKeys): void {
-    this.persistentLogAttributes = attributes;
+    this.#attributesStore.setPersistentAttributes(attributes);
+  }
+
+  /**
+   * @deprecated Use getPersistentLogAttributes() instead
+   */
+  public get persistentLogAttributes(): LogKeys {
```
svozza, line 666: We need to add this getter because  the `persistentLogAttributes` instance variable does not exist anymore.

packages/logger/src/LogAttributesStore.ts
```diff
@@ -0,0 +1,152 @@
...
+ */
+class LogAttributesStore {
+  readonly #temporaryAttributesKey = Symbol(
+    'powertools.logger.temporaryAttributes'
+  );
+  readonly #keysKey = Symbol('powertools.logger.keys');
+
+  #fallbackTemporaryAttributes: LogAttributes = {};
+  readonly #fallbackKeys: Map<string, 'temp' | 'persistent'> = new Map();
+  #persistentAttributes: LogAttributes = {};
```
sdangol: Does the `#persistentAttributes` not need to be stored in the Async Storage?

packages/logger/src/Logger.ts
```diff
@@ -875,18 +853,7 @@ class Logger extends Utility implements LoggerInterface {
...
-      if (!this.#checkReservedKeyAndWarn(key)) {
-        attributes[key] =
-          type === 'persistent'
-            ? this.persistentLogAttributes[key]
-            : this.temporaryLogAttributes[key];
-      }
-    }
-
-    return attributes;
+    return this.#attributesStore.getAllAttributes();
```
sdangol: Do we need to check `#checkReservedKeyAndWarn` here as it was before?

sdangol (dismissed):
Just a small question. Rest looks good from my side.

packages/logger/src/Logger.ts
svozza: Ah good catch!

svozza: So looking into this a bit more, I was surprised that we didn't have a failing test case in light of this removal. 

Upon investigation, I have found that `#appendKeys` already does this check but the deprecated `setPersistentLogAttributes` does not. However, we do not have a unit test that uses `setPersistentLogAttributes` to set a reserved keyword. Upon adding a test case that does this I now had a failing unit test.

Initially I thought that rather than adding reserved key filtering logic to `setPersistentLogAttributes`  that I would instead centralise it in `#createAdditionalAttributes` as this method  is called whenever _any_ log statement is emitted in the `createAndPopulateLogItem` method. This centralisation would allow us to remove the check from `#appendKeys` too.

However, when I implemented this, the tests that check for reserved keys started failing with stack overflow errors. This makes sense because the `#checkReservedKeyAndWarn` function itself has a log statement in it. If you try to use it to do a check in `createAndPopulateLogItem` you will trigger an infinite loop of log statements.
 
What I have done instead is add a new function called `filterReservedAttributeKeys` and used it in both `#appendKeys` and `setPersistentLogAttributes` (even though the function is deprecated, it is still part of the public API for now) so that reserved keys will not be added to the store. I have also removed the `#createAdditionalAttributes` method entirely and we just call `this.#attributesStore.getAllAttributes()` directly now.

PR #4684, opened 2025-10-27
docs: move the should i use this or Parser to the top of the validation

### LINKED ISSUE #4633: Docs: Move the `Should I use this or Parser?` to the top of the Validation docs

Documentation regarding validation

https://docs.aws.amazon.com/powertools/typescript/latest/features/validation/#should-i-use-this-or-parser

### How can we improve?

Right now, I would only know that the `Parser` package is recommended over the `Validation` package after reading through the `Validation` documentation and reaching at the end where I read the `Should I use this or Parser?` 

### Got a suggestion in mind?

The `Should I use this or Parser?`  section should be at the top so that we know the recommended package is the `Parser`. Also would be nice to have something in the Parser docs which tells when when the `Validation` package should be used over `Parser`

### Acknowledgment

kawaaaas:
### Changes

This PR improves the documentation structure for the Validation utility by moving the “Should I use this or Parser?” section from the bottom to the top of the page.
In addition, a copy of the same explanatory section is added to the Parser documentation to provide mutual reference between both utilities.

Please let me know if you have any suggestions regarding the added content or its placement.

**Issue number:** closes #4633

---

### DIFF
--- a/docs/features/parser.md
+++ b/docs/features/parser.md
@@ -375,9 +375,9 @@ This also works when using `safeParse` option.
 
 One of Powertools for AWS Lambda [tenets](../index.md#tenets) is to be progressive. This means that our utilities are designed to be incrementally adopted by customers at any stage of their serverless journey.
 
-For new projects, especially those using TypeScript, we recommend using the [Parser](parser.md) utility. Thanks to its integration with [Zod](http://zod.dev), it provides an expressive and type-safe way to validate and parse payloads.
+For new projects, especially those using TypeScript, we recommend using the Parser utility. Thanks to its integration with [Zod](http://zod.dev), it provides an expressive and type-safe way to validate and parse payloads.
 
-If instead you are already using JSON Schema, or simply feel more comfortable with it, the Validation utility is a great choice. It provides an opinionated thin layer on top of the popular [ajv](https://ajv.js.org) library, with built-in support for JMESPath and AWS service envelopes.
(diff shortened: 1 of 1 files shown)


docs/features/parser.md
```diff
@@ -370,3 +370,17 @@ This also works when using `safeParse` option.
...
     --8<-- "examples/snippets/parser/schema.ts"
     ```
+
+## Should I use this or Validation?
+
+One of Powertools for AWS Lambda [tenets](../index.md#tenets) is to be progressive. This means that our utilities are designed to be incrementally adopted by customers at any stage of their serverless journey.
+
+For new projects, especially those using TypeScript, we recommend using the [Parser](parser.md) utility. Thanks to its integration with [Zod](http://zod.dev), it provides an expressive and type-safe way to validate and parse payloads.
+
+If instead you are already using JSON Schema, or simply feel more comfortable with it, the Validation utility is a great choice. It provides an opinionated thin layer on top of the popular [ajv](https://ajv.js.org) library, with built-in support for JMESPath and AWS service envelopes.
```
sdangol: @kawaaaas Could you also add a link to the validation utility

sdangol (changes requested):
Thanks @kawaaaas for the changes. Just added a small feedback

docs/features/parser.md
kawaaaas: @sdangol 
Thank you for the feedback! I’ve addressed the comment and updated the parser link accordingly.

sdangol (approved):

PR #4693, opened 2025-10-28
improv(commons): Make trace ID access more robust

### LINKED ISSUE #4657: Maintenance: Leverage new InvokeStore module to keep track of X-RAY trace id
### Summary

The AWS Lambda team has released a new `InvokeStore` NPM [module](https://www.npmjs.com/package/@aws/lambda-invoke-store) that allows us to retriece the X-RAY trace id from `AsyncLocalStorage`. It means there is now another source which we need to consider when computing the X-ray trace ID. We should use it as well.

### Why is this needed?

This is needed to robustly compute the X-ray trace ID.

Other

### Solution

Use `InvokeStore` module and update the common trace id calculation logic.

### Acknowledgment

### Future readers

svozza:
## Summary

Last week we had to revert several PRs dies to an issue with the `InvokeStore` [module](https://www.npmjs.com/package/@aws/lambda-invoke-store). This module has now been patched.

This PR re-introduces the module, which means we now fetch the X-ray trace id from `AsyncLocalStorage` if it is available and an environment variable if not.. 

### Changes
Updated the `getXrayTraceDataFromEnv` function in `envUtils` to check `AsyncLocalStorage` first and fall back to the `_X_AMZN_TRACE_ID` environment variable otherwise.

Also updated the `envUtils` tests to use `unstubAllEnvs` rather than directly mutate `process.env`.

**Issue number:** closes #4657 

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10687,16 +10687,16 @@
       "version": "2.28.1",
       "license": "MIT-0",
       "dependencies": {
-        "@aws/lambda-invoke-store": "0.1.0"
+        "@aws/lambda-invoke-store": "0.1.1"
       },
       "devDependencies": {
         "@aws-lambda-powertools/testing-utils": "file:../testing"
       }
     },
     "packages/commons/node_modules/@aws/lambda-invoke-store": {
-      "version": "0.1.0",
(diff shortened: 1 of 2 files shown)


sdangol (approved):
Works now! Thanks for the PR and also fixing the library

PR #4700, opened 2025-10-29
feat(batch): use async local storage for batch processing

### LINKED ISSUE #4676: Feature request: Allow use of InvokeStore in Batch Processor
### Use case

The `InvokeStore` [pacakge](https://www.npmjs.com/package/@aws/lambda-invoke-store), allows for the use of [Asynchronous contexts ](https://nodejs.org/api/async_context.html#asynchronous-context-tracking) in lambda handlers. We should allow users to process SQS, Kinesis and DynamoDB streams in async contexts across lambda executions so that they are isolated.

### Solution/User Experience

```ts
import { BatchProcessor, EventType } from '@aws-lambda-powertools/batch';
import type { SQSHandler, SQSRecord } from 'aws-lambda';

const processor = new BatchProcessor(EventType.SQS);
const logger = new Logger();

const recordHandler = (record: SQSRecord): void => {
  const payload = record.body;
  if (payload) {
    const item = JSON.parse(payload);
    logger.info('Processed item', { item });
  }
};

export const handler: SQSHandler = async (event, context) => {
  const batch = event.Records; 
  
  // The records registered to the processor instance will be scoped only to this handler 
  // execution and will not be available in any other invocations
  processor.register(batch, recordHandler, { context }); 

  const processedMessages = await processor.process();

  for (const message of processedMessages) {
    const [status, error, record] = message;

    logger.info('Processed record', { status, record, error });
  }

  return processor.response();
};

// ... 

  {
  },
  () => {
    return handler(event)
  }
);
```

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

svozza:
## Summary
This PR adds support for using an async context (specifically from the InvokeStore [package](https://www.npmjs.com/package/@aws/lambda-invoke-store)) in the Batch utility. This allows users to process batches that are isolated specifically to the current lambda invocation, isolated from any other executions.

### Changes
- Added storage classes to abstract the use of `InvokeStore` to persist state across invocations.
- Added a storage class for the state (records, handlers, failures etc) associated with batch processing: `BatchProcessingStore`
- Add a storage class for SQS FIFO processing to track group IDs: `SqsFifoProcessorStore`
- The Batch Processor class are highly stateful with mutiple subclasses that mutate instance variables across the class hierarchy. As such I have taken a slightly different approach to incorporating the `BatchProcessingStore` class: if we followed the pattern of replacing all in place mutation with specific methods that guard against mutating the underlying resources,  this would entail far more changes to the code across the mutliple subclasses. Instead, I have chose to use getters and setters to proxy updates to the store to ensure the the amount of changes to the code are minimised.
- Added specific concurrency tests for both stores and also the `BatchProcessor` and `SqsFifoPartialProcessor` classes that use them.

**Issue number:** closes #4676 

---

### DIFF
--- a/packages/batch/src/SqsFifoPartialProcessor.ts
+++ b/packages/batch/src/SqsFifoPartialProcessor.ts
@@ -142,11 +142,7 @@ class SqsFifoPartialProcessor extends BatchProcessorSync {
     const remainingRecords = this.records.slice(firstFailureIndex);
 
     for (const record of remainingRecords) {
-      const result = this.#processFailRecord(
-        record,
-        new SqsFifoShortCircuitError()
-      );
-      processedRecords.push(result);
+      this.#processFailRecord(record, new SqsFifoShortCircuitError());
     }
 
     this.clean();
--- a/packages/batch/src/SqsFifoPartialProcessorAsync.ts
+++ b/packages/batch/src/SqsFifoPartialProcessorAsync.ts
@@ -141,11 +141,7 @@ class SqsFifoPartialProcessorAsync extends BatchProcessor {
     const remainingRecords = this.records.slice(firstFailureIndex);
 
     for (const record of remainingRecords) {
-      const result = this.#processFailRecord(
-        record,
-        new SqsFifoShortCircuitError()
-      );
-      processedRecords.push(result);
+      this.#processFailRecord(record, new SqsFifoShortCircuitError());
     }
 
     this.clean();
--- a/packages/batch/tests/unit/concurrency/SqsFifoPartialProcessor.test.ts
+++ b/packages/batch/tests/unit/concurrency/SqsFifoPartialProcessor.test.ts
@@ -8,6 +8,7 @@ import {
 import type {
   BatchProcessingOptions,
   FailureResponse,
+  PartialItemFailureResponse,
   SuccessResponse,
 } from '../../../src/types.js';
 import { sqsRecordFactory } from '../../helpers/factories.js';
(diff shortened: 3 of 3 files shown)


packages/batch/src/BasePartialBatchProcessor.ts
```diff
@@ -200,9 +195,9 @@ abstract class BasePartialBatchProcessor extends BasePartialProcessor {
    * Set up the processor with the initial state ready for processing
    */
   public prepare(): void {
-    this.successMessages.length = 0;
-    this.failureMessages.length = 0;
-    this.errors.length = 0;
+    this.successMessages = [];
```
svozza, line 198: I've changed this because it seemed odd to be just setting the length to 0 rather than just resetting the variable as is done elsewhere.

packages/batch/src/SqsFifoPartialProcessor.ts
```diff
@@ -142,7 +142,11 @@ class SqsFifoPartialProcessor extends BatchProcessorSync {
     const remainingRecords = this.records.slice(firstFailureIndex);
 
     for (const record of remainingRecords) {
-      this.#processFailRecord(record, new SqsFifoShortCircuitError());
+      const result = this.#processFailRecord(
+        record,
+        new SqsFifoShortCircuitError()
+      );
+      processedRecords.push(result);
```
sdangol: Did we miss having this previously?

packages/batch/src/SqsFifoPartialProcessorAsync.ts
```diff
@@ -141,7 +141,11 @@ class SqsFifoPartialProcessorAsync extends BatchProcessor {
     const remainingRecords = this.records.slice(firstFailureIndex);
 
     for (const record of remainingRecords) {
-      this.#processFailRecord(record, new SqsFifoShortCircuitError());
+      const result = this.#processFailRecord(
+        record,
+        new SqsFifoShortCircuitError()
+      );
+      processedRecords.push(result);
```
sdangol: Same in here, was there a bug because of this?

svozza: I initally thought this was a bug but now I'm not so sure, so I reverted the change and left it as it was.

Issue #4514, opened 2025-09-20
Feature request: first-class support for binary responses


dreamorosi:
### Use case

As a customer, when I want to return binary responses, I need to: 1/ manually construct the API Gateway response object, 2/ base64 encode the body, 3/ set the `isBase64Encoded: true`, and 4/ set the appropriate headers.

```ts
import { readFile } from 'node:fs/promises';
import { Router } from '@aws-lambda-powertools/event-handler/experimental-rest';
import { compress } from '@aws-lambda-powertools/event-handler/experimental-rest/middleware';
import type { Context } from 'aws-lambda';

const app = new Router();

app.get('/logo', [compress()], async () => {
  const logoFile = await readFile(`${process.env.LAMBDA_TASK_ROOT}/logo.png`);
  return {
    body: logoFile.toString('base64'),
    isBase64Encoded: true,
    headers: {
      'Content-Type': 'image/png',
    },
    statusCode: 200,
  };
});

export const handler = async (event: unknown, context: Context) =>
  app.resolve(event, context);
```

This means I can't use the Web API `Response` object, nor get the same automatic encoding behavior I get when returning JSON objects.

### Solution/User Experience

Customers should be able to return a `Buffer` or `ArrayBufferLike` or equivalent readable together with a content type header and Event Handler should base64 encode it (or skip this step if already a string) and set the `isBase64Encoded: true` field in the response:

```ts
import { readFile } from 'node:fs/promises';
import { Router } from '@aws-lambda-powertools/event-handler/experimental-rest';
import { compress } from '@aws-lambda-powertools/event-handler/experimental-rest/middleware';
import type { Context } from 'aws-lambda';

const app = new Router();

app.get('/logo', [compress()], async () => {
  const logoFile = await readFile(`${process.env.LAMBDA_TASK_ROOT}/logo.png`);
  return {
    body: logoFile, // Buffer<ArrayBufferLike>
    headers: {
      'Content-Type': 'image/png',
    }
  }
});

export const handler = async (event: unknown, context: Context) =>
  app.resolve(event, context);
```

Additionally, the current transformer function that converts `Response` objects to API Gateway responses currently doesn't take in account the actual encoding of the body as set in the headers and hardcodes `isBase64Encoded` to `false` unless compression is enabled.

### Alternative solutions

Continue doing this manually and only with API Gateway response objects - which is not great.

### Acknowledgment

[...]

PR #4714, opened 2025-11-05
feat(event-handler): Add support for HTTP APIs (API Gateway v2)

### LINKED ISSUE #4713: Feature request: Add support for HTTP APIs (API Gateway v2) in Event Handler
### Use case

Currently event handler only supports API Gateway Rest APIs. We should add support for HTTP APIs.

### Solution/User Experience

The event source should be transparent to users, they should be able to define handlers that return any valid API Gateway response we support (JSON object, `APIGatewayProxyEvent`, `APIGatewayProxyEventV2`, Web `Response` object) and event handler will automatically detect the input event and convert to the appropriate response.

```ts
const app = new Router();

// this returns APIGatewayProxyEventV2
app.get('/test', () => ({
    statusCode: 200,
    cookies: ['cookie1', 'cookie2'],
}));

const v1Result = await app.resolve(aV1Event, context);

/**
v1Result = {
    "statusCode": 200,
    "headers": {
    },
    "body": "{\"message\":\"success\"}",
    "isBase64Encoded": false,
    "multiValueHeaders": {
        "set-cookie": [
            "cookie1",
            "cookie2"
        ]
    }
}
 */

const v2Result = await app.resolve(aV2Event, context);

/**
v2Result = {
    "statusCode": 200,
    "headers": {
    },
    "body": "{\"message\":\"success\"}",
    "isBase64Encoded": false,
    "cookies": [
        "cookie1",
        "cookie2"
    ]
}
 **/
    
```

### Alternative solutions

```Markdown
Alternative solution would be for customers to have specific resolvers for specific event sources:


const httpApiRouter = new HttpApiRouter()
// will only work with HttpApi events
```

### Acknowledgment

### Future readers

svozza:
## Summary
This PR adds support for HTTP API (API Gateway v2) to the event handler utility. It also adds the framework for adding other event sources like function URLs, ALB and VPC lattice.

Edit: after some testing, I've realised that as function URLs and HTTP APIs share the same event type, which means function URLs also work.

### Changes
- The public `resolve` method now has overloads to support both `APIGatewayProxyEvent` and `APIGatewayProxyEventV2` events. New overloads can be added for new event types. I attempted to do this with conditional types but was unable to make it it work, furthermore, the type signatures were much harder to read than simpler overrides.
- The `#resolve` method in the Router class now returns the entire request context object. We use `reqCtx.res` now as the canonical representation of the response to be returned. As the request is now always represented as a web `Response` object, converting the response required for different event sources is simpler. We also save the type of response required in the request context using the `ResponseType` type.
- The `handlerResultToProxyResult` function has been removed now that all response are converted to Web `Response` objects.
- I have removed the ad hoc base64 encoding functionality. The change to the `#resolve` method will allow us to handle this is a more principled way by using the request context to indicate whether a response should be encoded or not. I plan to do this work when implementing https://github.com/aws-powertools/powertools-lambda-typescript/issues/4514.
- A new type guard to identify `APIGatewayProxyEventV2` objects has been added.
- A function to convert `APIGatewayProxyEventV2` events to Web `Request` objects has been added.
- A function to convert Web `Respsonse` objects to `APIGatewayProxyReultV2` objects has been added.

# Testing
- Most of the tests now run using both v1 events and v2 events. The pattern established is easily extended to use additional event types. One suite I have not done so is the `middleware` tests. These use the exact same mechanisms as the other tests so I have decided to only add some additional  `v2` specific tests. I am open to persusion as to whether these should follow the other tests.
- I have used the same lambda function to integrate with an API Gateway REST API and an HTTP API and verified using `curl` that requests will be fulfilled by the same code with no changes required.
- Tested a lambda function URL with streaming enabled and it also worked.

**Issue number:** closes #4713 

---

### DIFF
--- a/packages/event-handler/src/rest/Router.ts
+++ b/packages/event-handler/src/rest/Router.ts
@@ -24,6 +24,7 @@ import type {
   RequestContext,
   ResolveStreamOptions,
   ResponseStream,
+  ResponseType,
   RestRouteOptions,
   RestRouterOptions,
   RouteHandler,
@@ -217,7 +218,9 @@ class Router {
       throw new InvalidEventError();
     }
 
-    const responseType = isAPIGatewayProxyEventV2(event) ? 'v2' : 'v1';
+    const responseType: ResponseType = isAPIGatewayProxyEventV2(event)
+      ? 'ApiGatewayV2'
+      : 'ApiGatewayV1';
 
     let req: Request;
     try {
--- a/packages/event-handler/src/rest/converters.ts
+++ b/packages/event-handler/src/rest/converters.ts
@@ -198,13 +198,17 @@ const webHeadersToApiGatewayV2Headers = (webHeaders: Headers) => {
 const webHeadersToApiGatewayHeaders = <T extends ResponseType>(
   webHeaders: Headers,
   responseType: T
-): T extends 'v1' ? V1Headers : { headers: Record<string, string> } => {
-  if (responseType === 'v1') {
-    return webHeadersToApiGatewayV1Headers(webHeaders) as T extends 'v1'
+): T extends 'ApiGatewayV1'
+  ? V1Headers
+  : { headers: Record<string, string> } => {
+  if (responseType === 'ApiGatewayV1') {
+    return webHeadersToApiGatewayV1Headers(
+      webHeaders
+    ) as T extends 'ApiGatewayV1'
       ? V1Headers
       : { headers: Record<string, string> };
   }
-  return webHeadersToApiGatewayV2Headers(webHeaders) as T extends 'v1'
+  return webHeadersToApiGatewayV2Headers(webHeaders) as T extends 'ApiGatewayV1'
     ? V1Headers
     : { headers: Record<string, string> };
 };
@@ -317,7 +321,7 @@ const webResponseToProxyResult = <T extends ResponseType>(
   response: Response,
   responseType: T
 ): Promise<ResponseTypeMap[T]> => {
-  if (responseType === 'v1') {
+  if (responseType === 'ApiGatewayV1') {
     return webResponseToProxyResultV1(response) as Promise<ResponseTypeMap[T]>;
   }
   return webResponseToProxyResultV2(response) as Promise<ResponseTypeMap[T]>;
--- a/packages/event-handler/src/types/rest.ts
+++ b/packages/event-handler/src/types/rest.ts
@@ -15,11 +15,11 @@ import type { Route } from '../rest/Route.js';
 import type { HttpResponseStream } from '../rest/utils.js';
 import type { ResolveOptions } from './common.js';
 
-type ResponseType = 'v1' | 'v2';
+type ResponseType = 'ApiGatewayV1' | 'ApiGatewayV2';
 
 type ResponseTypeMap = {
-  v1: APIGatewayProxyResult;
-  v2: APIGatewayProxyStructuredResultV2;
+  ApiGatewayV1: APIGatewayProxyResult;
+  ApiGatewayV2: APIGatewayProxyStructuredResultV2;
 };
 
 type RequestContext = {
--- a/packages/event-handler/tests/unit/rest/converters.test.ts
+++ b/packages/event-handler/tests/unit/rest/converters.test.ts
@@ -549,7 +549,7 @@ describe('Converters', () => {
       });
 
       // Act
-      const result = await webResponseToProxyResult(response, 'v1');
+      const result = await webResponseToProxyResult(response, 'ApiGatewayV1');
 
       // Assess
       expect(result.statusCode).toBe(200);
@@ -566,7 +566,7 @@ describe('Converters', () => {
       });
 
       // Act
-      const result = await webResponseToProxyResult(response, 'v1');
+      const result = await webResponseToProxyResult(response, 'ApiGatewayV1');
 
       // Assess
       expect(result.statusCode).toBe(201);
@@ -587,7 +587,7 @@ describe('Converters', () => {
       });
 
       // Act
-      const result = await webResponseToProxyResult(response, 'v1');
+      const result = await webResponseToProxyResult(response, 'ApiGatewayV1');
 
       // Assess
       expect(result.headers).toEqual({ 'content-type': 'application/json' });
@@ -607,7 +607,7 @@ describe('Converters', () => {
       });
 
       // Act
-      const result = await webResponseToProxyResult(response, 'v1');
+      const result = await webResponseToProxyResult(response, 'ApiGatewayV1');
 
       // Assess
       expect(result.headers).toEqual({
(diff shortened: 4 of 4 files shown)


packages/event-handler/tests/unit/rest/Router/basic-routing.test.ts
```diff
@@ -1,15 +1,18 @@
...
-  InternalServerError,
   Router,
 } from '../../../../src/rest/index.js';
 import type { HttpMethod, RouteHandler } from '../../../../src/types/rest.js';
-import { createTestEvent } from '../helpers.js';
+import { createTestEvent, createTestEventV2 } from '../helpers.js';
 
-describe('Class: Router - Basic Routing', () => {
+describe.each([
+  { version: 'V1', createEvent: createTestEvent },
```
svozza, line 13: As mentioned, these tests can be extended just by adding a new test event creation function to the `describe.each` block.

packages/event-handler/src/rest/converters.ts
```diff
@@ -285,26 +447,10 @@ const bodyToNodeStream = (body: ExtendedAPIGatewayProxyResultBody) => {
...
 };
 
-/**
- * Converts a Node.js Readable stream to a base64 encoded string.
- * Handles both Buffer and string chunks by converting all to Buffers.
- *
- * @param stream - The Node.js Readable stream to convert
- * @returns A Promise that resolves to a base64 encoded string
- */
-async function nodeStreamToBase64(stream: Readable) {
```
svozza, line 295: We will need to add this back when we implement #4514.

packages/event-handler/src/rest/Router.ts
```diff
@@ -202,38 +209,45 @@ class Router {
...
+      req = proxyEventToWebRequest(event);
+    } catch (err) {
+      if (err instanceof InvalidHttpMethodError) {
+        this.logger.error(err);
+        // We can't throw a MethodNotAllowedError outside the try block as it
+        // will be converted to an internal server error by the API Gateway runtime
+        return {
+          event,
+          context,
+          req: new Request('https://invalid'),
```
svozza, line 236: I'm not fond of creating this dummy `Request` object but the alternative is to make `req` optional, which will make the ergonomics of writing middleware much worse: users would need to do a `null` check any time they want to interact with a request that will always be there apart from this particular corner case.

packages/event-handler/tests/unit/rest/Router/error-handling.test.ts
```diff
@@ -567,3 +588,32 @@ describe('Class: Router - Error Handling', () => {
...
 });
+describe('Class: Router - proxyEventToWebRequest Error Handling', () => {
+  beforeEach(() => {
+    vi.resetModules();
+  });
+
+  it('re-throws non-InvalidHttpMethodError from proxyEventToWebRequest', async () => {
+    // Prepare
+    vi.doMock('../../../../src/rest/converters.js', async () => {
+      const actual = await vi.importActual<
```
svozza, line 599: I very much dislike tests like this. The reason we have to do this is because the type guards before this code path are so strict that the error code is virtually unreachable. I have to mock the `proxyEventToWebRequest` function to throw an exception that it will almost certainly never throw in real use.

packages/event-handler/src/rest/converters.ts
```diff
@@ -81,8 +89,68 @@ const proxyEventToWebRequest = (event: APIGatewayProxyEvent): Request => {
...
+  }
+
+  if (Array.isArray(event.cookies)) {
+    headers.set('Cookie', event.cookies.join('; '));
+  }
+
+  const hostname = headers.get('Host') ?? domainName;
+  const protocol = headers.get('X-Forwarded-Proto') ?? 'https';
+
+  const url = rawQueryString
```
sdangol, line 117: Should we explicitly check for an empty string?

packages/event-handler/tests/unit/rest/Router/middleware.test.ts
```diff
@@ -595,8 +596,8 @@ describe('Class: Router - Middleware', () => {
 
       // Assess
       expect(result.statusCode).toBe(200);
-      expect(result.isBase64Encoded).toBe(true);
-      expect(result.body).toBe(Buffer.from(testData).toString('base64'));
+      expect(result.isBase64Encoded).toBe(false);
```
sdangol, line 599: Was this because of removing `nodeStreamToBase64`? Does this mean compressed response aren't supported now?

svozza: No, this was the ad hoc compression I mentioned in the summary. We were automatically compressing a response if you returned a stream but it only worked for a middleware that returned a response. The compression response is handled in `webResponseToProxyResultV1` and `webResponseToProxyResultV2`, so is unaffected. I want to ensure the logic for compression is more centralised and this is the first phase of that.

sdangol: Ok, got it. Thanks for explaining

packages/event-handler/src/rest/converters.ts
svozza: I think this is one of the few times we actually want falsy behaviour, we want to skip on `null`, `undefined` and empty string here. I think the type in `aws-lambda` is wrong here, it says this field can't be `undefined` but if you look at the Rust SDK it's marked as optional.

sdangol: Ok, makes sense

svozza: Actually nvm the Rust bit it has lots of things marked optional such as even `rawPath` so I think they've been over-zealous in their definition. But I don't think it does any harm to be defensive here.

svozza:
Nice!

svozza:
@swopnildangol I made a change to how we calculate what response to return. Rather than passing the event around and using type guards, I save the response type once in the request context with a type called `ResponseType` that can only have the value `v1` or `v2`. This allows us to remove a bunch of overloads in the conversion functions and just use conditional types.

sdangol:
> @swopnildangol I made a change to how we calculate what response to return. Rather than passing the event around and using type guards, I save the response type once in the request context with a type called `ResponseType` that can only have the value `v1` or `v2`. This allows us to remove a bunch of overloads in the conversion functions and just use conditional types.

So, with this, as we add on support for more handlers, we'd be adding options to the ResponseType? Would it make more sense to have maybe ApiGatewayV1 or ApiGatewayV2 rather than just v1 and v2?

svozza:
Good point, I've changed it.

.github/workflows/make-release.yml
```diff
@@ -5,7 +5,7 @@ name: Make Release
 # === Automated activities ===
 # 1. [Quality check] run unit tests, linting, examples, layer, doc snippets
 # 2. [Release] publish all packages to npmjs.org using the latest git commit, ensure provenance with NPM_CONFIG_PROVENANCE=true
-# 3. [Create tag] create a new git tag using released version, i.e. v1.13.1
+# 3. [Create tag] create a new git tag using released version, i.e. ApiGatewayV1.13.1
```
sdangol: Probably an unintended change

svozza: Yeah, find and replace went wrong. 😓 Fixed now.

PR #4723, opened 2025-11-08
feat(event-handler): add first-class support for binary responses

### LINKED ISSUE #4514: Feature request: first-class support for binary responses

svozza:
## Summary
This PR adds support for base64 encoding of responses. If a handler returns an `APIGatewayProxyResult` or `APIGatewayProxyStructuredResultV2` where the `body` field is a stream or an `ArrayBuffer` we will automatically base64 encode it. Handlers can now also just return a stream or `ArrayBuffer` and we will base64 encode it and set the status code to 200 and the content type to `application/json`.

### Changes

- Handler functions can now return `ArrayBuffer`s and streams, either on their own or as part of a response event, these will automatically be base64 encoded.
- Handler resolution logic detects whether to base64 encode a response and stores decision in the `requestContext` object in a field called `isBase64Encoded`. This decision is made by whether the `content-encoding` header is set ot the the `content-type` header is image, audio, or video.
- `isBase64Encoded` field will only be set once, this means that users can over-ride the default detection behaviour by setting `base64Encoded` in the handler and it will not be changed.
- Added unit tests to test new functionality.

**Issue number:** closes #4514 

---

### DIFF
--- a/packages/event-handler/src/rest/converters.ts
+++ b/packages/event-handler/src/rest/converters.ts
@@ -10,11 +10,13 @@ import type {
   ExtendedAPIGatewayProxyResult,
   ExtendedAPIGatewayProxyResultBody,
   HandlerResponse,
+  HttpStatusCode,
   ResponseType,
   ResponseTypeMap,
   V1Headers,
   WebResponseToProxyResultOptions,
 } from '../types/rest.js';
+import { HttpStatusCodes } from './constants.js';
 import { InvalidHttpMethodError } from './errors.js';
 import {
   isAPIGatewayProxyEventV2,
@@ -345,13 +347,15 @@ function addProxyEventHeaders(
  * Handles APIGatewayProxyResult, Response objects, and plain objects.
  *
  * @param response - The handler response (APIGatewayProxyResult, Response, or plain object)
- * @param resHeaders - Optional headers to be included in the response
+ * @param options - Optional configuration with statusCode and resHeaders
  * @returns A Web API Response object
  */
 const handlerResultToWebResponse = (
   response: HandlerResponse,
-  resHeaders?: Headers
+  options?: { statusCode?: HttpStatusCode; resHeaders?: Headers }
 ): Response => {
+  const statusCode = options?.statusCode ?? HttpStatusCodes.OK;
+  const resHeaders = options?.resHeaders;
   if (response instanceof Response) {
     if (resHeaders === undefined) return response;
     const headers = new Headers(resHeaders);
@@ -373,7 +377,7 @@ const handlerResultToWebResponse = (
         : response;
 
     return new Response(body, {
-      status: 200,
+      status: statusCode,
       headers,
     });
   }
@@ -393,7 +397,7 @@ const handlerResultToWebResponse = (
       headers,
     });
   }
-  return Response.json(response, { headers });
+  return Response.json(response, { headers, status: statusCode });
 };
 
 /**
--- a/packages/event-handler/src/rest/utils.ts
+++ b/packages/event-handler/src/rest/utils.ts
@@ -11,6 +11,7 @@ import type {
   ExtendedAPIGatewayProxyResult,
   HandlerResponse,
   HttpMethod,
+  HttpStatusCode,
   Middleware,
   Path,
   ResponseStream,
@@ -18,6 +19,7 @@ import type {
 } from '../types/rest.js';
 import {
   COMPRESSION_ENCODING_TYPES,
+  HttpStatusCodes,
   HttpVerbs,
   PARAM_PATTERN,
   SAFE_CHARS,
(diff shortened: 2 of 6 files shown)


packages/event-handler/src/rest/utils.ts
```diff
@@ -318,3 +330,47 @@ export const HttpResponseStream =
...
+    contentEncoding != null &&
+    [
+      COMPRESSION_ENCODING_TYPES.GZIP,
+      COMPRESSION_ENCODING_TYPES.DEFLATE,
+    ].includes(contentEncoding)
+  ) {
+    return true;
+  }
+
+  const contentType = headers.get('content-type');
```
svozza, line 361: An alternative approach here is to check if we have don't a text-based content type, i.e, if the content type is not one of  `text/*`, `application/json`, `application/xml`, `application/javascript` or `application/x-www-form-urlencoded` then we base64 encode it. My worry is that there are other non binary types I might be missing here.

packages/event-handler/src/rest/converters.ts
```diff
@@ -385,6 +365,19 @@ const handlerResultToWebResponse = (
...
   const headers = new Headers(resHeaders);
+
+  if (isBinaryResult(response)) {
+    const body =
+      response instanceof Readable
+        ? (Readable.toWeb(response) as ReadableStream)
+        : response;
+
+    return new Response(body, {
+      status: 200,
```
sdangol: Is this always 200

packages/event-handler/src/rest/utils.ts
```diff
@@ -318,3 +330,47 @@ export const HttpResponseStream =
...
+export const getBase64EncodingFromResult = (result: HandlerResponse) => {
+  if (isBinaryResult(result)) {
+    return true;
+  }
+  if (isExtendedAPIGatewayProxyResult(result)) {
+    return (
+      result.body instanceof ArrayBuffer ||
+      isNodeReadableStream(result.body) ||
+      isWebReadableStream(result.body)
+    );
```
sdangol: Could this be `isBinaryResult(result.body)`?

packages/event-handler/src/rest/converters.ts
svozza: Ah very good catch!

packages/event-handler/src/rest/utils.ts
svozza: Oh yes, it could.

packages/event-handler/src/rest/converters.ts
svozza: Fixed!

packages/event-handler/src/rest/utils.ts

packages/event-handler/src/rest/utils.ts
svozza: What do you think about this @swopnildangol?

sdangol: Yeah, there might be chances of missing some binary types with both approaches. With the current implementation we might be missing some binary types such as pdf, doc, excel, archives, executables?

svozza: Yeah, I wasn't sure how far to go so I just picked the most popular ones. I remember we had that giant regex in the compress middleware before that we removed. Ultimately, customers can still set `isBase64Encoded` themselves and we can add more types in the future based on demand.

sdangol (approved):

PR #4725, opened 2025-11-10
fix(event-handler): allow event handler response to return array

### LINKED ISSUE #4593: Bug: Unable to return an array in the event handler response
### Expected Behavior

I should be able to return an array when using event handlers.

### Current Behavior

I get a typescript error when trying to return an array from the handler

### Code snippet

```typescript

const ddbDocClient = DynamoDBDocument.from(new DynamoDBClient());
const app = new Router();

app.get('/orders', async () => {
  const response = await ddbDocClient.send(
    new ScanCommand({
      TableName: 'OrdersWorkshop',
    })
  );
  if (response.Items && response.Items.length > 0) {
    return response.Items;
  }
  return { message: 'No orders found' };
});

export const lambdaHandler = (
  event: APIGatewayProxyEvent,
  context: Context
) => {
  app.resolve(event, context);
};
```

### Steps to Reproduce

1. Create a lambda function with the code provided
2. There's an typescript error on the handler for get orders

### Possible Solution

We have defined the `HandlerResponse` type as  `type HandlerResponse = Response | JSONObject;`
Instead, it should be `type HandlerResponse = Response | JSONValue;` to support both array and objects

latest

22.x

### Packaging format used

npm

### Execution logs

```Shell

```

arnabrahman:
## Summary
Previously, we couldn't return an array as an event handler response. This PR fixes that.
### Changes

- update `HandlerResponse` type, replace `JSONObject` with `JSONValue`
- update error handling function to handle the new type
- Add unit tests to cover array response

Example:

```typescript
import type { APIGatewayProxyEvent, Context } from 'aws-lambda';
import { DynamoDBClient } from '@aws-sdk/client-dynamodb';
import { ScanCommand, DynamoDBDocument } from '@aws-sdk/lib-dynamodb';

const ddbDocClient = DynamoDBDocument.from(new DynamoDBClient());
const app = new Router();

app.get('/orders', async () => {
  const response = await ddbDocClient.send(
    new ScanCommand({
      TableName: 'OrdersWorkshop',
    })
  );
  if (response.Items && response.Items.length > 0) {
    return response.Items;
  }
  return { message: 'No orders found' };
});

export const lambdaHandler = (
  event: APIGatewayProxyEvent,
  context: Context
) => {
  app.resolve(event, context);
};
```

**Issue number:** closes #4593

---


packages/event-handler/src/rest/Router.ts
```diff
@@ -488,6 +480,48 @@ class Router {
...
+   * Converts an error handler's response body to an HTTP Response object.
+   *
+   * If the body is a record object without a status code, sets the status code for
+   * NotFoundError (404) or MethodNotAllowedError (405). Uses the status code from
+   * the body if present, otherwise defaults to 500 Internal Server Error.
+   *
+   * @param body - The response body returned by the error handler, of type JSONValue
+   * @param error - The Error object associated with the response
+   */
+  #errorBodyToResponse(body: JSONValue, error: Error): Response {
```
svozza: Let's call this `errorBodyToWebResponse`.

packages/event-handler/tests/unit/rest/Router/decorators.test.ts
```diff
@@ -67,74 +67,111 @@ const createStreamHandler =
   (event: unknown, _context: Context, responseStream: MockResponseStream) =>
     app.resolveStream(event, _context, { scope, responseStream });
 
+const createTestLambdaClass = (app: Router, expectedResponse: unknown) => {
```
svozza: Let's move this function to the `event-handler/tests/unit/rest/helpers.ts` file.

svozza:
Two very minor comments but other than that, this is pretty much good to go.

packages/event-handler/tests/unit/rest/Router/decorators.test.ts
arnabrahman: Then we’ll need to move `createHandler` inside `helpers.ts` as well since it’s used by `createTestLambdaClass`. I think that should be fine.

Should we also move the other helper handlers from the `decorator.test` file to `helpers.ts` too? @svozza

svozza: Yeah, I think that's fine. I had been meaning to tidy up the decorator tests anyway so now is a good a time as any.

svozza (approved):

svozza:
Great stuff @arnabrahman , merged now!

PR #4736, opened 2025-11-11
feat(event-handler): add validation support for REST router

### LINKED ISSUE #4516: Feature request: first class support for data validation in Event Handler
### Use case

As a customer I want to use Standard Schema-compatible libraries to validate requests and/or responses when using Event Handler.

When enabled, things like body and query parameters (at minimum) should be strongly typed in route handlers and possibly middleware functions.

### Solution/User Experience

We will add a new options argument to the route handler registration methods such as `app.get`, `app.posts` etc. Users will be able to validate elements of both the request and the response. The registartion methods will habe a new overload that will allow users to use gernics to type the handlers. The solution will support the stadard schema protocol, this will allow users to bring their own validation library rather than us choosing for them.

#### Validation Options

```ts
/**
 * Configuration for request validation
 */
type RequestValidationConfig<T = unknown> = {
  body?: StandardSchemaV1<unknown, T>;
  headers?: StandardSchemaV1<unknown, Record<string, string>>;
  path?: StandardSchemaV1<unknown, Record<string, string>>;
  query?: StandardSchemaV1<unknown, Record<string, string>>;
};

/**
 * Configuration for response validation
 */
type ResponseValidationConfig<T extends HandlerResponse = HandlerResponse> = {
  body?: StandardSchemaV1<HandlerResponse, T>;
  headers?: StandardSchemaV1<Record<string, string>, Record<string, string>>;
};

/**
 * Validation configuration for request and response
 */
type ValidationConfig<
  TReqBody = unknown,
  TResBody extends HandlerResponse = HandlerResponse
> = {
  req?: RequestValidationConfig<TReqBody>;
  res?: ResponseValidationConfig<TResBody>;
};
```

#### Example Usage (with Zod)

```ts
import { z } from 'zod';

const createUserRequestSchema = z.object({ 
  name: z.string(), 
  email: z.string().email() 
});

const createUserResponseSchema = z.object({ 
  id: z.string(), 
  name: z.string(),
  email: z.string()
});

// Infer types from schemas
type CreateUserRequest = zodInfer<typeof createUserRequestSchema>;
type CreateUserResponse = zodInfer<typeof createUserResponseSchema>;

app.post<CreateUserRequest, CreateUserResponse>('/users', [], async (reqCtx) => {
  return { id: '123', name: body.name };
}, {
  validation: {
    req: {
      body: createUserRequestSchema,
      headers: z.object({ 'x-api-key': z.string() }),
      path: z.object({ userId: z.string() }),
      query: z.object({ filter: z.string().optional() })
    },
    res: {
      body: createUserResponseSchema,
      headers: z.object({ 'x-request-id': z.string() })
    }
  }
});
```

#### Open question

Should we add a new field to `RequestContext` to allow users access validated parameters in a type safe way rather than just validating them and throwing an error if they are not correctly:

```ts
app.post('/users', [], async (reqCtx) => {
  // reqCtx.valid.req.body is typed as { name: string; email: string }
  // reqCtx.valid.req.headers is typed as {  'x-api-key': string }
  // etc
  return { id: '123', name: body.name }; // type error if this does not conform to CreateUserResponse type
}, {
  validation: {
    // ...
  }
});
```

#### Validation Errors

We should create a custom validation error type that wraps the schema errors in a uniform interface. This will also allow users to catch this specific error and do custom error handling:

```ts
class RequestValidationError extends HttpError {

  constructor(message, errors} {
    // ...
  }

}

// ...


app.errorHandler(RequestValidationError, (error) => {
 // do something with error
 throw new BadRequestError('custom message');
});
```

### Implementation

**Note:** These implementations are rough sketches that are meant to be a guideline. If the the final implementation differs from this, that is perfectly fine.

Under the hood, this can just be another middleware, Using the API from the open question section, we could do this:

```ts
/**
 * Validated request data
 */
type ValidatedRequest<TBody = unknown> = {
  body: TBody;
  headers?: Record<string, string>;
  path?: Record<string, string>;
  query?: Record<string, string>;
};

/**
 * Validated response data
 */
type ValidatedResponse<TBody extends HandlerResponse = HandlerResponse> = {
  body?: TBody;
  headers?: Record<string, string>;
};

type RequestContext = {
  req: Request;
  res: Response
  // ...
  valid?: {
    req: ValidatedRequest
    res: ValidatedResponse
  }
};

function validation<TReqBody, TResBody extends HandlerResponse>(
  validation: ValidationConfig<TReqBody, TResBody>
): Middleware {
  return async ({ reqCtx, next }) => {
    // Initialize valid object
    reqCtx.valid = {
      req: {},
      res: {}
    };

    // Validate request body
    if (validation.req?.body) {
      const body = await reqCtx.req.json();
      const result = await validation.req.body['~standard'].validate(body);
      
      if (!result.success) {
        throw new RequestValidationError('Validation failed!, result.issues );
      }
      
      reqCtx.valid.req.body = result.value;
    }
   
    // ...headers, path etc
    
    await next();

    // Validate response body
    if (validation.res?.body && reqCtx.res.ok) {
      const responseBody = await reqCtx.res.json();
      const result = await validation.res.body['~standard'].validate(responseBody);
      
      if (!result.success) {
        throw new RequestValidationError('Validation failed!, result.issues );
      }
    }
  };
}
```

We will also need to update  the existing handler registration methods to thread the new types through.

```ts
#handleHttpMethod<TReqBody = never, TResBody extends HandlerResponse = HandlerResponse>(
  method: HttpMethod,
  path: Path,
  middlewareOrHandler?: Middleware[] | RouteHandler,
  handlerOrOptions?: RouteHandler | TypedRouteHandler<TReqBody, TResBody>,
  options?: { validation: ValidationConfig<TReqBody, TResBody> }
): MethodDecorator | undefined {
  if (Array.isArray(middlewareOrHandler)) {
    if (typeof handlerOrOptions === 'function') {
      // Check if options has validation
      if (options?.validation) {
        const validationMiddleware = this.#createValidationMiddleware(options.validation);
        this.route(handlerOrOptions, { 
          method, 
          path, 
          middleware: [...middlewareOrHandler, validationMiddleware] 
        });
        return;
      }
      // Existing: app.post(path, [middleware], handler)
      this.route(handlerOrOptions, { method, path, middleware: middlewareOrHandler });
      return;
    }
    // Decorator case...
  }
  // ... rest of existing logic
}

// Existing overloads (unchanged)
public post(path: Path, handler: RouteHandler): void;
public post(path: Path, middleware: Middleware[], handler: RouteHandler): void;
public post(path: Path): MethodDecorator;
public post(path: Path, middleware: Middleware[]): MethodDecorator;

// New overload with validation
public post<TReqBody = never, TResBody extends HandlerResponse = HandlerResponse>(
  path: Path,
  middleware: Middleware[],
  handler: TypedRouteHandler<TReqBody, TResBody>,
  options: { validation: ValidationConfig<TReqBody, TResBody> }
): void;

// Implementation
public post<TReqBody = never, TResBody extends HandlerResponse = HandlerResponse>(
  path: Path,
  middlewareOrHandler?: Middleware[] | RouteHandler,
  handler?: RouteHandler | TypedRouteHandler<TReqBody, TResBody>,
  options?: { validation: ValidationConfig<TReqBody, TResBody> }
): MethodDecorator | undefined {
  return this.#handleHttpMethod<TReqBody, TResBody>(
    HttpVerbs.POST,
    path,
    middlewareOrHandler,
    handler,
    options
  );
}

```

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

sdangol:
> **Edit (by @svozza):** I have updated the description to reflect the current state of this PR. Any comments before 2026-02-19 refer to the previous implementation.

## Summary

This PR adds request and response validation middleware to the HTTP event handler, along with type-safe route handler support and extended response body handling.

Closes #4516 

---

## Changes

### Validation middleware                                                                                                 
                                                                                                                       
A new `validate()` middleware validates request and response data against [Standard Schema v1](https://standardschema.dev/) schemas. Standard Schema is a vendor-neutral interface supported by Zod, Valibot, ArkType, and other popular validation libraries.                                                                          
                                                                                                                            
You can validate any combination of request body, headers, path parameters, and query parameters. Request validation errors return a structured `422 Unprocessable Entity` response. Response validation errors return `500 Internal Server Error`.

When you provide a validation config, TypeScript infers the handler's context type automatically. The handler receives a `valid` property containing the parsed and typed data: **only the fields you configured schemas for are accessible**. As such, accessing an unvalidated field (e.g., `reqCtx.valid.req.headers` when no headers schema was configured) is a compile-time error:

  ```ts
  import { z } from 'zod';

  const bodySchema = z.object({ name: z.string() });

  app.post('/users', (reqCtx) => {
    const { name } = reqCtx.valid.req.body; // typed as { name: string }
    // reqCtx.valid.req.headers             // compile error — no headers schema configured
    return { id: 1, name };
  }, {
    validation: {
      req: { body: bodySchema },
    },
  });
  ```

When a response schema is also configured, `reqCtx.valid.res` becomes available in post-handler middleware. Without a response schema, `valid.res` does not exist on the type:

  ```ts
  const resSchema = z.object({ id: z.number(), name: z.string() });

  app.post('/users', (reqCtx) => {
    const { name } = reqCtx.valid.req.body;
    // reqCtx.valid.res                     // compile error — only available when res schema is configured
    return { id: 1, name };                 // checked against resSchema at runtime
  }, {
    validation: {
      req: { body: bodySchema },
      res: { body: resSchema },
    },
  });
  ```

  You can also pass validation alongside a middleware array:

  ```ts
  app.post('/users', [authMiddleware], handler, {
    validation: {
      req: { body: bodySchema },
    },
  });
  ```

### Type-safe route handlers

This provides compile-time checking only — TypeScript raises an error if the handler returns a shape that does not match `User`, but no runtime validation takes place. To validate the response shape at compile time and runtime, use the `res` option in the `validation` config instead:

```ts
const resSchema = z.object({ id: z.number(), name: z.string() });

app.get('/users', () => {
  return { id: 1, name: 'Alice' }; // validated at runtime against resSchema
}, {
  validation: { res: { body: resSchema } },
});
```

You can also combine request validation with a compile-time response type by annotating the handler's return type directly:

```ts
app.post('/users', (reqCtx): User => {
  const { name } = reqCtx.valid.req.body;
  return { id: 1, name }; // TypeScript enforces the return shape here
}, { validation: { req: { body: bodySchema } } });
```

### New error classes

| Class | Status |
|---|---|
| `RequestValidationError` | 422 Unprocessable Entity |
| `ResponseValidationError` | 500 Internal Server Error |

Both classes extend `HttpError` and include structured issue details in the response body.

### Extended response body

`ExtendedAPIGatewayProxyResult` now accepts any JSON-serialisable value as the `body` field, and correctly handles responses with no body (such as `204 No Content`). Previously, only `string`, `ArrayBuffer`, and streams were accepted, and omitting `body` would cause the response object itself to be serialised as JSON with a 200 status.

```ts
// JSON object body — previously required JSON.stringify()
app.get('/users', () => ({
  statusCode: 200,
  body: { id: 1, name: 'Alice' },
}));

// JSON array body
app.get('/items', () => ({
  statusCode: 200,
  body: [1, 2, 3],
}));

// No body
app.delete('/users/:id', () => ({
  statusCode: 204,
}));
```

### New dependency

[`@standard-schema/spec`](https://github.com/standard-schema/standard-schema) is added as an optional peer dependency. It provides only TypeScript types and a small interface definition with no runtime overhead, and is only required when using the `validate` middleware.

### Testing
- Unit tests for all new features have been added
- E2E tests have been added for new extended response body change and are all green: https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/22164580643

---

### DIFF
--- a/packages/event-handler/src/http/middleware/index.ts
+++ b/packages/event-handler/src/http/middleware/index.ts
@@ -1,4 +1,3 @@
 export { compress } from './compress.js';
 export { cors } from './cors.js';
 export { tracer } from './tracer.js';
-export { validate } from './validation.js';


svozza:
Given that this is just middleware under the hood, I think we can simplify the API here. We already allow to users to pass middleware into their routes, I don't see a reason to make the type signature of the HTTP verb methods more complicated when we can do this instead:

```ts
import {validation} from '@aws-lambda-powertools/event-handler/experimental-rest/middleware';

app.post('/users', [validation({ req: { body: z.object({ name: z.string() }) } })], async () => {
  return { id: '123', name: 'John' };
});
```

packages/event-handler/src/rest/middleware/validation.ts
```diff
@@ -0,0 +1,100 @@
...
+  config: RestRouteOptions['validation']
+): Middleware => {
+  const reqSchemas = config?.req;
+  const resSchemas = config?.res;
+
+  return async ({ reqCtx, next }) => {
+    // Validate request
+    if (reqSchemas) {
+      if (reqSchemas.body) {
+        let bodyData: unknown = reqCtx.event.body;
```
svozza: We  should avoid using the raw event in middleware because they are specific to the particualr lambda integration. This means that the middleware may only work with say API Gateway or function URLs etc. We should use the `req` and `res` objects where bossible. Bearing in mind that if we wish to use the body or request or a response, we need to clone it.

packages/event-handler/src/rest/Router.ts
```diff
@@ -551,170 +557,247 @@ class Router {
     method: HttpMethod,
     path: Path,
     middlewareOrHandler?: Middleware[] | RouteHandler,
-    handler?: RouteHandler
+    handlerOrOptions?: RouteHandler | Omit<RestRouteOptions, 'method' | 'path'>,
```
svozza: As mentioned in https://github.com/aws-powertools/powertools-lambda-typescript/pull/4736#issuecomment-3523631276 we won't need to change anything in all these HTTP methods if we treat validation as just another middleware.

packages/event-handler/src/rest/middleware/validation.ts
```diff
@@ -0,0 +1,100 @@
...
+    // Validate response
+    if (resSchemas) {
+      const response = reqCtx.res;
+
+      if (resSchemas.body) {
+        const clonedResponse = response.clone();
+        const contentType = response.headers.get('content-type');
+
+        let bodyData: unknown;
+        if (contentType?.includes('application/json')) {
```
svozza: This can be simplified:

```ts
const bodyData = contentType?.includes('application/json') 
  ? await clonedResponse.json() : await clonedResponse.text();;
```

packages/event-handler/src/rest/middleware/validation.ts
```diff
@@ -0,0 +1,100 @@
...
+      }
+    }
+  };
+};
+
+async function validateComponent(
+  schema: StandardSchemaV1,
+  data: unknown,
+  component: 'body' | 'headers' | 'path' | 'query',
+  isRequest: boolean
```
svozza: Magic booleans in APIs are a code smell imo.

dreamorosi:
This was implemented this way because of the RFC.

It's true that it's just a middleware, but I'd like us to at least consider the option of having a special treatment for this one.

I'm suggesting this because of two aspects:
1. We want whatever schemas are defined there to strongly type the handler and its arguments. I am not sure how to write a generic that picks the `z.infer<>` types from an array of unknown middleware functions and transposes it to the handler parameter. Having a named argument like the PR suggests makes it a bit more tenable.
2. This is mainly stylistic, but considering that this feature is expected to be the foundation for the OpenAPI schema one, these schemas might become quite large. Having them mixed into a middleware array and above the handler can make things a bit messy to read. Instead, putting them under a named argument might be a bit better.

Arguably the first one might be a skill issue on my side, if there's a way to mitigate it, happy to remove the concern. For the second, it's primarily preference so again, I just wanted to have a discussion, not trying to force any other direction.

----

Also @sdangol, I'd consider moving all the AIDLC files into a [public gist](https://gist.github.com) and linking that one in the PR body, the sheer amount of markdown generated is a bit unwieldy when it comes to looking at the diff.

packages/event-handler/src/types/rest.ts
```diff
@@ -241,6 +243,40 @@ type WebResponseToProxyResultOptions = {
   isBase64Encoded?: boolean;
 };
 
+/**
+ * Configuration for request validation
+ */
+type RequestValidationConfig<T = unknown> = {
```
svozza: Seems odd to me that these are all optional. Is there a way to express through the type system that at least one of these should be present?

svozza:
I'll answer your points in reverse order @dreamorosi.

> This is mainly stylistic, but considering that this feature is expected to be the foundation for the OpenAPI schema one, these schemas might become quite large. Having them mixed into a middleware array and above the handler can make things a bit messy to read. Instead, putting them under a named argument might be a bit better.

I wouldn't recommend embedding a large validation, espeically OpenAPI, in the handler signature either way. I would have a separate place for the validation definitions, something like this:

```ts
// validators.ts
import {validation} from '@aws-lambda-powertools/event-handler/experimental-rest/middleware';

export userValidation = validiation(/** very large validation object **/);

// index.ts
import {userValidation} from './validators.js';

app.post('/users', [userValidation], async () => {
  return { id: '123', name: 'John' };
});
```

> We want whatever schemas are defined there to strongly type the handler and its arguments. I am not sure how to write a generic that picks the `z.infer<>` types from an array of unknown middleware functions and transposes it to the handler parameter. Having a named argument like the PR suggests makes it a bit more tenable.

While I would be disappointed to see the simplicity of treating everything as middleware go, this is a very compelling argument. Trying to infer types from things that may or may not be in an array sounds like a nightmare, especially when the elements in that array are just functions.

My main objection to adding this extra argument is the compelxity it creates because we now have these HTTP methods that can take, 2, 3 or 4 arguments depending on the use case. This is also compounded when you throw decorators into the mix.  I think maybe the way forward then is to have a options object so that we can reduce the combinations of arguments. This way the function will only ever have 2 or 3 arguments.

```ts
app.post('/users', async () => {
  return { id: '123', name: 'John' };
}, {middleware: [ /** ... */], validation: {/** ... */} });
```

I don't love the way the middleware is after the handler here but having the optional argument last simplifies things more imo. Obviously this is a decision that needs to be made before GA as it is a breaking change.

svozza:
After a discussion with @sdangol I've decided we should keep the proposed signature in this PR:

```
app.post('/users', [[ /** middleware */]], async () => {
  return { id: '123', name: 'John' };
}, {validation: {/** ... */} });
```

This is on the condition that this function will no have any more arguments added to it, any new additions will always go in to the final options object where the `validation` field is. I will write this up formally in the validation implementation issue I am going to write this week.

svozza:
I updated the issue with more requirements a few days ago: https://github.com/aws-powertools/powertools-lambda-typescript/issues/4516.

sdangol:
@svozza I've updated the PR now addressing the updated requirements.

packages/event-handler/src/http/middleware/validation.ts
```diff
@@ -0,0 +1,191 @@
...
+
+/**
+ * Creates a validation middleware from the provided validation configuration.
+ *
+ * @param config - Validation configuration for request and response
+ * @returns Middleware function that validates request/response
+ */
+export const createValidationMiddleware = <
+  TReqBody = unknown,
+  TResBody extends HandlerResponse = HandlerResponse,
```
sdangol: We currently only have the type safety for the body of the req/res, not for the headers, path, query

packages/event-handler/tests/unit/http/middleware/validation.test.ts
```diff
@@ -0,0 +1,488 @@
...
+
+  it('validates response body successfully', async () => {
+    // Prepare
+    const responseSchema = z.object({ id: z.string(), name: z.string() });
+    type ResponseType = z.infer<typeof responseSchema>;
+
+    app.get<never, ResponseType>(
+      '/users/:id',
+      () => {
+        return { id: '123', name: 'John' };
```
sdangol, line 296: Right now, this only works when we return the structure described in the schema and not when using `Response` object

packages/event-handler/src/http/middleware/validation.ts
svozza: Does that mean if someone does `valid.req.headers` they won't get autocomplete?

packages/event-handler/tests/unit/http/middleware/validation.test.ts
svozza: Yeah, that makes sense.

packages/event-handler/src/http/middleware/validation.ts
sdangol: @svozza I've now updated this to support type safety for the header, path and query too

packages/event-handler/src/http/middleware/index.ts
```diff
@@ -1,3 +1,4 @@
 export { compress } from './compress.js';
 export { cors } from './cors.js';
 export { tracer } from './tracer.js';
+export { validate } from './validation.js';
```
dreamorosi: If this is used internally only, should we avoid exporting it? Or in other words, can/should this be used directly by customers?

svozza: Yes, I was in two minds about this. my thinking here initially was perhaps people might want a middleware only solution where they care more about the runtime validation that the compile type guarantees. However, is that likely? Probably not, I'm in favour of removing the export, if there is demand we can expose it at a later date.

dreamorosi: Ok, let's remove it for now, then we can merge this PR

PR #4740, opened 2025-11-12
fix(logger): not passing persistent keys to children

### LINKED ISSUE #4719: Bug: persistentKeys are not persisted across children
### Expected Behavior

When creating a child logger I expect `persistentKeys` from the parent to be merged into the child's.

### Current Behavior

When calling `logger.createChild` with the `persistentKeys` object, existing persistent keys on the parent are completely cleared out instead of merged together.

### Code snippet

```typescript

const logger = new Logger({
  persistentKeys: {
    foo: 'hello',
    overridable: 1,
  },
})

logger.appendKeys({
  resettableKey: 'some-id',
})

logger.appendPersistentKeys({
  dynamic: 'stays',
})

const child = logger.createChild({
  persistentKeys: {
    bar: 'world',
    overridable: 2,
  },
})

child.info('Log Message')
child.resetKeys()
child.info('Log Message 2')

```

**Expected Output:**

```json
{
  "level": "INFO",
  "message": "Log Message",
  "timestamp": "2025-11-06T18:54:49.202Z",
  "sampling_rate": 0,
  "foo": "hello",
  "overridable": 2,
  "dynamic": "stays",
  "bar": "world",
  "resettableKey": "some-id"
}
{
  "level": "INFO",
  "message": "Log Message 2",
  "timestamp": "2025-11-06T18:54:49.204Z",
  "sampling_rate": 0,
  "foo": "hello",
  "overridable": 2,
  "dynamic": "stays",
  "bar": "world"
}
```

**Current Output:**

```json
{
  "level": "INFO",
  "message": "Log Message",
  "timestamp": "2025-11-06T18:57:06.315Z",
  "sampling_rate": 0,
  "bar": "world",
  "overridable": 2,
  "resettableKey": "some-id"
}
{
  "level": "INFO",
  "message": "Log Message 2",
  "timestamp": "2025-11-06T18:57:06.315Z",
  "sampling_rate": 0,
  "bar": "world",
  "overridable": 2
}
```

### Steps to Reproduce

1. Create file with code snippet from above
2. Run the code and see the console output

### Possible Solution

I believe this is due to [this line](https://github.com/aws-powertools/powertools-lambda-typescript/blob/b387db063388f3046ba2435eb7042319a18ed326/packages/logger/src/Logger.ts#L338-L339) where creating a child relies on setting the deprecated `persistentLogAttributes` instead of the newer `persistentKeys` attribute. Which is ignored during instantiation if both are used [seen here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/b387db063388f3046ba2435eb7042319a18ed326/packages/logger/src/Logger.ts#L1272-L1276). This usage of the deprecated api also contributes to why when creating a child and setting `persistentKeys` on it, the 

```
Both persistentLogAttributes and persistentKeys options were provided.
``` 

warning is displayed regardless of callers actually using that deprecated attribute or not. [Related issue](https://github.com/aws-powertools/powertools-lambda-typescript/issues/4624)

I can help contribute this if it is a desired fix

latest

22.x

### Packaging format used

npm

### Execution logs

```Shell

```

iamgerg:
### Changes

This PR fixes the bug where `persistentKeys` passed into a parent logger would be overwritten instead of merged by any value for `persistentKeys` being passed using the `logger.createChild` method. See the issue linked below for more details.

**Issue number:** closes #4719 

---

### DIFF
--- a/examples/app/package.json
+++ b/examples/app/package.json
@@ -29,8 +29,8 @@
     "#errors": "./functions/commons/errors.js"
   },
   "devDependencies": {
-    "@types/aws-lambda": "^8.10.157",
-    "@types/node": "24.10.0",
+    "@types/aws-lambda": "^8.10.158",
+    "@types/node": "24.10.1",
     "aws-cdk-lib": "^2.223.0",
     "constructs": "^10.4.3",
     "source-map-support": "^0.5.21",
@@ -48,8 +48,8 @@
     "@aws-sdk/client-ssm": "^3.928.0",
     "@aws-sdk/lib-dynamodb": "^3.928.0",
     "@middy/core": "^4.7.0",
-    "@types/aws-lambda": "^8.10.157",
-    "@types/node": "24.10.0",
+    "@types/aws-lambda": "^8.10.158",
+    "@types/node": "24.10.1",
     "aws-cdk": "^2.1031.2",
     "constructs": "^10.4.3",
     "esbuild": "^0.27.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,8 +28,8 @@
       ],
       "devDependencies": {
         "@biomejs/biome": "^2.3.5",
-        "@types/aws-lambda": "^8.10.157",
-        "@types/node": "^24.10.0",
+        "@types/aws-lambda": "^8.10.158",
+        "@types/node": "^24.10.1",
         "@vitest/coverage-v8": "^3.2.4",
         "husky": "^9.1.7",
         "lint-staged": "^16.2.5",
@@ -59,16 +59,16 @@
         "@aws-sdk/client-ssm": "^3.928.0",
         "@aws-sdk/lib-dynamodb": "^3.928.0",
         "@middy/core": "^4.7.0",
-        "@types/aws-lambda": "^8.10.157",
-        "@types/node": "24.10.0",
+        "@types/aws-lambda": "^8.10.158",
+        "@types/node": "24.10.1",
         "aws-cdk": "^2.1031.2",
         "constructs": "^10.4.3",
         "esbuild": "^0.27.0",
         "typescript": "^5.9.3"
       },
       "devDependencies": {
-        "@types/aws-lambda": "^8.10.157",
-        "@types/node": "24.10.0",
+        "@types/aws-lambda": "^8.10.158",
+        "@types/node": "24.10.1",
         "aws-cdk-lib": "^2.223.0",
         "constructs": "^10.4.3",
         "source-map-support": "^0.5.21",
@@ -5015,9 +5015,9 @@
       "license": "MIT"
     },
     "node_modules/@types/aws-lambda": {
-      "version": "8.10.157",
-      "resolved": "https://registry.npmjs.org/@types/aws-lambda/-/aws-lambda-8.10.157.tgz",
-      "integrity": "sha512-ofjcRCO1N7tMZDSO11u5bFHPDfUFD3Q9YK9g4S4w8UDKuG3CNlw2lNK1sd3Itdo7JORygZmG4h9ZykS8dlXvMA==",
+      "version": "8.10.158",
+      "resolved": "https://registry.npmjs.org/@types/aws-lambda/-/aws-lambda-8.10.158.tgz",
+      "integrity": "sha512-v/n2WsL1ksRKigfqZ9ff7ANobfT3t/T8kI8UOiur98tREwFulv9lRv+pDrocGPWOe3DpD2Y2GKRO+OiyxwgaCQ==",
       "dev": true,
       "license": "MIT"
     },
@@ -5103,9 +5103,9 @@
       "license": "MIT"
     },
     "node_modules/@types/node": {
-      "version": "24.10.0",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-24.10.0.tgz",
-      "integrity": "sha512-qzQZRBqkFsYyaSWXuEHc2WR9c0a0CXwiE5FWUvn7ZM+vdy1uZLfCunD38UzhuB7YN/J11ndbDBcTmOdxJo9Q7A==",
+      "version": "24.10.1",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-24.10.1.tgz",
+      "integrity": "sha512-GNWcUTRBgIRJD5zj+Tq0fKOJ5XZajIiBroOF0yvj2bSU1WvNdYS/dn9UxwsujGW4JX06dnHyjV2y9rRaybH0iQ==",
       "license": "MIT",
       "dependencies": {
         "undici-types": "~7.16.0"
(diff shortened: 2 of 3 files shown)


packages/logger/tests/e2e/childLogger.manual.test.FunctionCode.ts
```diff
@@ -4,13 +4,13 @@ import type { LogLevel } from '../../src/types/index.js';
 import type { TestEvent, TestOutput } from '../helpers/types.js';
 
 const PERSISTENT_KEY = process.env.PERSISTENT_KEY || 'persistentKey';
-const PERSISTENT_VALUE = process.env.ERSISTENT_VALUE || 'persistentValue';
+const PERSISTENT_VALUE = process.env.PERSISTENT_VALUE || 'persistentValue';
 const PARENT_LOG_MSG = process.env.PARENT_LOG_MSG || 'parent-only-log-msg';
```
iamgerg, line 8: I noticed this was misspelled and probably the reason the values weren't asserted against in the actual assertions. I did not spend the time to set up an e2e environment to test this though and will remove if deemed unnecessary or the tests fail.

packages/logger/src/Logger.ts
```diff
@@ -335,8 +335,7 @@ class Logger extends Utility implements LoggerInterface {
           logFormatter: this.getLogFormatter(),
           customConfigService: this.getCustomConfigService(),
           environment: this.powertoolsLogData.environment,
-          persistentLogAttributes:
-            this.#attributesStore.getPersistentAttributes(),
+          persistentKeys: this.#attributesStore.getPersistentAttributes(),
```
svozza: Even though it's deprecated, I think we still need to keep `persistentLogAttributes` as well, as it's still part of the public API until we remove it in the next major version.

packages/logger/tests/e2e/childLogger.manual.test.FunctionCode.ts
svozza: Let's revert this change and I will raise a separate PR with the fix, that way I can run the E2E tests GitHub action because my branch has been created directly in the Powertools repo.

svozza:
Thank you for the quick turnaround! I only have two minor comments but other than that this looks great.

packages/logger/src/Logger.ts
iamgerg: Makes sense, I can add it back. Just to clarify though do you want something like this?

```suggestion
          persistentLogAttrubtes: this.#attributesStore.getPersistentAttributes(),
          persistentKeys: this.#attributesStore.getPersistentAttributes(),
```

packages/logger/tests/e2e/childLogger.manual.test.FunctionCode.ts
iamgerg: Sounds good!

packages/logger/src/Logger.ts
iamgerg: Actually after looking into this I'm pretty sure having both is what causes the warning message `Both persistentLogAttributes and persistentKeys options were provided.` and `persistentLogAttributes` is overridden in the constructor already if `persistentKeys` is passed in [see here](https://github.com/aws-powertools/powertools-lambda-typescript/blob/dd84f940c1b002012be9717bbddd39d88a933d27/packages/logger/src/Logger.ts#L1275). 

So maybe this should also do a check for which key the user passes in and merge the parent's `#attributesStore` with that key to preserve state. For the case where both are passed in when creating a child I believe ignoring `persistentLogAttributes` and logging a warning similar to the one in `setOptions` makes sense but let me know if you have other thoughts and if you have any other cases I may have missed.

**Examples**

```ts

const loggerA = new Logger({
  persistentKeys: { ... } // should always persist
})

const loggerB = new Logger({
  persistentLogAttributes: { ... } // should always persist
})

const loggerC = new Logger({
  persistentKeys: { ... }, // should always persist
  persistentLogAttributes: { ... } // will get overridden by persistentKeys and log warning
})

const childA = loggerA.createChild({
  persistentKeys: { ... } // will get merged with loggerA's persistentKeys
})

const childB = loggerA.createChild({
  persistentLogAttributes: { ... } // will get merged with loggerA's persistentKeys
// but set as `persistentLogAttributes` so no warning message logged
})

const childC = loggerB.createChild({
  persistentKeys: { ... } // will get merged with loggerB's persistentLogAttributes
// and set as persistentKeys so no warning message logged
})

const childD = loggerB.createChild({
  persistentKeys: { ... }, // will get merged with loggerB's persistentLogAttributes
  persistentLogAttributes: { ... } // will get ignored since persistentKeys is preferred
// will log warning message
})

```

svozza: Ah yes, that's a good spot! I agree with both your suggestion so let's do that.

iamgerg: Okay made the changes, ended up adding an extra block of tests in the `workingWithKeys.test.ts` file since, while the fix was simple, this logic and intended behavior is not.

packages/logger/src/Logger.ts
```diff
@@ -323,6 +323,19 @@ class Logger extends Utility implements LoggerInterface {
    * @param options - The options to initialize the child logger with.
    */
   public createChild(options: ConstructorOptions = {}): Logger {
+    // Determine proper key to use when merging persistent options with
+    // parent. Preferring persistentKeys over persistentLogAttributes.
+    // NOTE: when both are passed in for options warning message in
+    // setOptions is still triggered
+    const persistentKeyName: keyof Pick<
```
svozza, line 330: Nice

packages/logger/tests/unit/workingWithkeys.test.ts
```diff
@@ -816,4 +826,200 @@ describe('Working with keys', () => {
       );
     }
   );
+
+  describe('deprecated persistentLogAttributes usage', () => {
+    it('should set #attributesStore on the logger', () => {
```
svozza: We prefer not to use the should construction for test names in the codebase:

`it('sets #attributesStore on the logger`

packages/logger/src/Logger.ts
svozza: Nice work, looks very good!

packages/logger/tests/unit/workingWithkeys.test.ts
```diff
@@ -1,3 +1,4 @@
+import { log } from 'node:console';
```
svozza: I think this got added by mistake.

packages/logger/tests/unit/workingWithkeys.test.ts
iamgerg: Ahh I see, thanks for the suggestion will update!

packages/logger/tests/unit/workingWithkeys.test.ts
iamgerg: Good catch! Dang autoimport, will remove

svozza (approved):

svozza:
Really thorough job here @iamgerg, thank you! Merging now.

PR #4741, opened 2025-11-13
fix(logger): infinite loop on log buffer when item size is max bytes

### LINKED ISSUE #4737: Bug: Log buffer would run an infinite loop when the size of the log is exactly the same as the maxBytes size specified.
### Expected Behavior

It should not keep attempting to remove from the log buffer (if it's currently empty).

### Current Behavior

It keeps attempting to remove from the buffer even if the buffer is empty.

There are checks to make sure that the item size is less than or equal to the maximum size however, subsequent checks check that the `item size + buffer size` is less than the maximum size.

So in an execution where:
`maxBytes` is `20000`
`item size`  is `20000`
`buffer size` is `0`

there would be an infinite loop because while it attempts to remove from the map.

if item size was greater than `item size`, it would have thrown an error instead.

### Code snippet

link to code:

https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/logger/src/logBuffer.ts#L135

### Steps to Reproduce

1. Enable buffered logs
2. get log data that would be exactly same as the maxBytes specified

### Possible Solution

Either check that the item size is definitely less than max bytes (not less than or equal)
or only attempt to remove from the `buffer` if there is something in the buffer

```javascript
while (buffer.currentBytesSize + item.byteSize >= this.#maxBytesSize && buffer.currentBytesSize) {
```

latest

22.x

### Packaging format used

npm

### Execution logs

```Shell

```

fidelisojeah:
### Changes

This PR fixes a bug where an infinite loop occurs when a new buffered log's byte size is equal to the maxBytes size (`logBufferOptions.maxBytes`) specified while log buffering is enabled (`logBufferOptions.enabled = true`).

See issue for more details.

**Issue number:** closes #4737

---

### DIFF
--- a/packages/logger/src/logBuffer.ts
+++ b/packages/logger/src/logBuffer.ts
@@ -110,8 +110,12 @@ export class CircularMap<V> extends Map<string, SizedSet<V>> {
 
     const buffer = this.get(key) || new SizedSet<V>();
 
-    if (buffer.currentBytesSize + item.byteSize >= this.#maxBytesSize) {
+    if (
+      buffer.currentBytesSize !== 0 &&
+      buffer.currentBytesSize + item.byteSize >= this.#maxBytesSize
+    ) {
       this.#deleteFromBufferUntilSizeIsLessThanMax(buffer, item);
+
       if (this.#onBufferOverflow) {
         this.#onBufferOverflow();
       }
@@ -132,7 +136,10 @@ export class CircularMap<V> extends Map<string, SizedSet<V>> {
     buffer: SizedSet<V>,
     item: SizedItem<V>
   ) {
-    while (buffer.currentBytesSize + item.byteSize >= this.#maxBytesSize) {
+    while (
+      buffer.size !== 0 &&
+      buffer.currentBytesSize + item.byteSize >= this.#maxBytesSize
+    ) {
       buffer.shift();
       buffer.hasEvictedLog = true;
     }
--- a/packages/logger/tests/unit/logBufferStructures.test.ts
+++ b/packages/logger/tests/unit/logBufferStructures.test.ts
@@ -144,4 +144,74 @@ describe('CircularMap', () => {
       options.maxBytesSize
     );
   });
+
+  it('stores item in buffer when item size is equal to maxBytesSize', () => {
+    // Prepare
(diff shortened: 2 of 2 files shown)


dreamorosi:
Thank you for the PR, we'll review it soon.

packages/logger/src/logBuffer.ts
```diff
@@ -132,7 +136,10 @@ export class CircularMap<V> extends Map<string, SizedSet<V>> {
     buffer: SizedSet<V>,
     item: SizedItem<V>
   ) {
-    while (buffer.currentBytesSize + item.byteSize >= this.#maxBytesSize) {
+    while (
+      buffer.size &&
```
fidelisojeah: choosing to use the `set.size` here rather than checking against `currentBytesSize`

dreamorosi (commented):
Set @svozza as reviewer

packages/logger/src/logBuffer.ts
```diff
@@ -110,8 +110,12 @@ export class CircularMap<V> extends Map<string, SizedSet<V>> {
 
     const buffer = this.get(key) || new SizedSet<V>();
 
-    if (buffer.currentBytesSize + item.byteSize >= this.#maxBytesSize) {
+    if (
+      buffer.currentBytesSize &&
```
svozza: Can we be explicit here and check exactly rather than relying on type coercion: `!== 0`.

packages/logger/src/logBuffer.ts
svozza: Same comment regarding type coercion.

svozza:
Nice work @fidelisojeah! I only had two very minor comments.

packages/logger/src/logBuffer.ts
fidelisojeah: updated

packages/logger/src/logBuffer.ts
fidelisojeah: Updated

svozza: Great, everything looks good to me.

svozza:
Great stuff @fidelisojeah , merging now!

PR #4747, opened 2025-11-14
chore: upgrade vitest & update coverage exceptions

### LINKED ISSUE #4746: Maintenance: upgrade vitest to next major
### Summary

Vitest released a new major version (4.x) and introduced some breaking changes around code coverage in the process.

These changes make our CI fail as shown in #4679. We should fix this.

### Why is this needed?

So that the same coverage is maintained and tests continue to pass.

_No response_

### Solution

_No response_

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR upgrades vitest and its coverage module to the latest major version and revisits all the code paths that were previously ignored by either updating them [to the new ignore syntax](https://vitest.dev/guide/coverage.html#ignoring-code) or removing the ignore comment entirely.

In the process, I also had to update a couple tests that were not being awaited correctly.

Thanks to #4679 and @sdangol for the prework on this, I opted for a new PR and branch since the other had conflicts and this was faster. 

**Issue number:** closes #4746

---


packages/parameters/src/base/BaseProvider.ts
```diff
@@ -46,6 +46,7 @@ abstract class BaseProvider implements BaseProviderInterface {
     awsSdkV3ClientPrototype,
   }: BaseProviderConstructorOptions) {
     this.store = new Map();
+    /* v8 ignore else -- @preserve */
```
svozza, line 49: Is this going to ignore this whole if-block?

packages/parameters/src/base/GetOptions.ts
```diff
@@ -19,6 +19,7 @@ class GetOptions implements GetOptionsInterface {
   public constructor(options: GetOptionsInterface = {}) {
     Object.assign(this, options);
 
+    /* v8 ignore else -- @preserve */
```
svozza, line 22: I'm surprised this code isn't being hit by at least one test.

packages/parser/src/schemas/dynamodb.ts
```diff
@@ -57,6 +57,7 @@ const unmarshallDynamoDBTransform = (
   // @ts-expect-error - We are intentionally mutating the object
   result.Keys = unmarshalledKeys;
 
+  /* v8 ignore next -- @preserve */
```
svozza, line 60: Again, I would expect this to be hit by at least one test

dreamorosi: I'm gonna explain here, but this answer is valid for both your other comments.

In previous versions of vitest implicit else branches were not included in coverage reports, while now they are.

Take a code like this (sorry I can't think of a less trivial example):

```ts
function sumFirstPositive(a, b) {
  if (a < 0) {
    throw new Error('Operator a should be positive');
  } else {
    void; // this would please the coverage
  }
  return a + b;
}
```

While Has an implicit `else` statement that we're not necessarily covering or testing for, at least according to the AST.

With these `/* v8 ignore next -- @preserve */` we're telling the coverage reporter to ignore the `else` statement of these conditional blocks since there's nothing useful for us there.

I was also surprised by some of these, but I think it's either some implicit code path we don't care in the generated output (remember the code is run through `esbuild` before testing) or the new coverage logic not being 100% good yet.

svozza: Ah I see. Makes sense!

svozza (approved):

PR #4753, opened 2025-11-14
test(maintenance): Expose ResponseStream object to simplify testing in event handler

### LINKED ISSUE #4752: Maintenance: Simplify Stream Testing in Event Handler HTTP APIs
### Summary

Currently, to test the stream functionality we create we need to implement a mock streaming class in our tests. However, this helper mock file will not be available to users who wish to test their streaming functions so they will have to implement this mock themselves. We should create a `ResponseStream` class that is used by the fallback version of `streamifyHandler` that can be used in our tests in `utils` module. Customers will also be able to use this class in their own tests too.

### Why is this needed?

This simplifies testing as we are using the same implementation in the fallabck and the tests.

Event Handler

### Solution

Unfortunately, the types in `aws-lambda` package are [not correct](https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/73792) so we will have to do some workarounds to make everything type check. But ultimately we should be able to do this:

```ts
import {ResponseStream} from "../../../../src/rest/utils.js";

it('streams a simple JSON response', async () => {
    // Prepare
    const app = new Router();
    app.get('/test', async () => ({ message: 'Hello, World!' }));

    const handler = streamify(app);
    const responseStream = new ResponseStream();

    // Act
    const result = await handler(
      createEvent('/test', 'GET'),
      responseStream,
      context
    );

    // Assess
    expect(JSON.parse(result.body)).toEqual({ message: 'Hello, World!' });
  });
```

### Acknowledgment

### Future readers

svozza:
## Summary
This PR removes the need for us to implement a custom response stream mock when testing handlers that use the `streamify` or `resolveStream` functions.

### Changes
- Creates `ResponseStream` class that matches the actiual type used by lambda functions when streaming
- Updated the fallback `streamifyResponse` function to use the `ResponseStream` class

#### Testing
- Updated the streaming tests to use the new `ResponseStream` class
- Removed the `MockResponseStream` class

**Issue number:** closes #4752 

---

### DIFF
--- a/packages/event-handler/src/rest/utils.ts
+++ b/packages/event-handler/src/rest/utils.ts
@@ -1,4 +1,4 @@
-import { Duplex, Readable, Writable } from 'node:stream';
+import { Duplex, Readable } from 'node:stream';
 import {
   isRecord,
   isRegExp,
@@ -18,9 +18,9 @@ import type {
   HandlerResponse,
   HttpMethod,
   HttpStatusCode,
-  ResponseStream as IResponseStream,
   Middleware,
   Path,
+  ResponseStream,
   ValidationResult,
 } from '../types/rest.js';
 import {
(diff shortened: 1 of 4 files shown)


packages/event-handler/src/rest/utils.ts
```diff
@@ -401,22 +426,21 @@ const streamifyResponse =
...
-
-        return {
-          body: bodyBuffer.toString(),
-          headers: prelude.headers,
-          statusCode: prelude.statusCode,
-        } as TResult;
-      }
+      /* v8 ignore next -- @preserve */
+      const output: Buffer =
+        (responseStream as IResponseStream).getBuffer?.() ?? Buffer.from([]);
```
svozza: This cast is necessary because of the mismatch between the types in the  `aws-lambda`.types package.

packages/event-handler/src/types/rest.ts
```diff
@@ -140,9 +139,11 @@ type ValidationResult = {
   issues: string[];
 };
 
-type ResponseStream = InstanceType<typeof HttpResponseStream> & {
-  _onBeforeFirstWrite?: (write: (data: Uint8Array | string) => void) => void;
-};
+interface ResponseStream extends Writable {
```
svozza, line 142: This interface is necessary because of the mismatch between the types in the aws-lambda types package.

dreamorosi (commented):
I've left a comment in the linked issue about the general idea of exporting this, however everything else already looks good 👍

svozza:
I've update the PR based on your feedback.

Issue #4751, opened 2025-11-14
Bug: compression middleware ignores threshold when content-length header is missing


dreamorosi:
### Expected Behavior

The `compress` middleware should only compress responses that exceed the configured threshold (default: 1024 bytes, configurable via options). When a response is below the threshold, it should NOT be compressed, regardless of whether the `content-length` header is present.

For example, with a threshold of 100 bytes:
- A 20-byte response should NOT be compressed
- A 200-byte response SHOULD be compressed

### Current Behavior

The `compression` middleware **always compresses** responses when the `content-length` header is not set, completely ignoring the threshold setting.

Since JSON responses generated by the Event Handler do NOT have content-length headers set by default, **all JSON responses are compressed regardless of size**, defeating the purpose of the threshold configuration.

In `packages/event-handler/src/rest/middleware/compress.ts` line 109 we have: `(!contentLength || Number(contentLength) > threshold)`

This condition means:
- If `content-length` is NOT set → compress (regardless of threshold)
- If `content-length` IS set AND greater than threshold → compress

The logic should only compress when we can determine the size AND it exceeds the threshold.

### Code snippet

Failing E2E test:

```ts
// packages/event-handler/tests/e2e/restEventHandler.test.ts
it('does not compress small responses below threshold', async () => {
  const response = await fetch(`${apiUrl}/compress/small`, {
    headers: { 'Accept-Encoding': 'gzip' },
  });
  const data = await response.json();

  expect(response.status).toBe(200);
  expect(data.message).toBe('Small');
  // Small response (~20 bytes) is below 100 byte threshold, should not be compressed
  expect(response.headers.get('content-encoding')).toBeNull(); // FAILS - gets 'gzip'
});
```

Handler code:
```ts
// packages/event-handler/tests/e2e/routers/compressRouter.ts
const compressRouter = new Router();
compressRouter.use(compress({ threshold: 100 })); // 100 byte threshold

compressRouter.get('/small', () => ({
  message: 'Small', // ~20 bytes when stringified
}));
```

Failing unit test demonstrating the bug:

```ts
// Add to packages/event-handler/tests/unit/rest/middleware/compress.test.ts after line 63
it('skips compression when content is below threshold and content-length is not set', async
() => {
  // Prepare
  // This test simulates real-world usage where JSON responses don't have
  // content-length headers set. The middleware should NOT compress small responses
  // even when content-length is missing.
  const application = new Router();
[...]

Issue #4750, opened 2025-11-14
Bug: duplicate query string parameters when multiple occurrences are present


dreamorosi:
### Expected Behavior

When using the Event Handler with REST API Gateway V1 and multi-value query parameters (e.g., `?filter=active&filter=published`), the handler should:
1. Extract all parameter values in the correct order: `['active', 'published']`
2. Not duplicate any values
3. Match the behavior documented in AWS API Gateway where `multiValueQueryStringParameters` contains all values in order

### Current Behavior

The Event Handler duplicates multi-value query parameters.

For example, with the URL `?filter=active&filter=published`:
- Expected: `['active', 'published']`
- Actual: `['published', 'active', 'published']`

This happens, I think, because the converter in `packages/event-handler/src/rest/converters.ts` (lines 74-86) processes BOTH `queryStringParameters` AND `multiValueQueryStringParameters` from the API Gateway V1 event, but according to AWS documentation:
- `queryStringParameters` contains the last value of each parameter
- `multiValueQueryStringParameters` contains all values for each parameter

When both fields contain the same parameter, the current implementation appends values from both, causing duplication.

### Code snippet

Failing E2E test:

```ts
// packages/event-handler/tests/e2e/restEventHandler.test.ts (line 206)
it('handles array query parameters', async () => {
  const searchQuery = 'test';
  const filters = ['active', 'published'];

  const response = await fetch(
    `${apiUrl}/params/search?q=${searchQuery}&filter=${filters[0]}&filter=${filters[1]}`
  );
  const data = await response.json();

  expect(response.status).toBe(200);
  expect(data.query).toBe(searchQuery);
  expect(data.filters).toEqual(['active', 'published']); // FAILS
});
```

Handler implementation:

```ts
// packages/event-handler/tests/e2e/routers/paramsRouter.ts
paramsRouter.get('/search', ({ req }) => {
  const url = new URL(req.url);
  const filters = url.searchParams.getAll('filter');
  return { filters: filters.length > 0 ? filters : undefined };
});
```

Failing unit test demonstrating the bug:

```ts
// Add to packages/event-handler/tests/unit/rest/converters.test.ts after line 299
it('handles same parameter in both queryStringParameters and multiValueQueryStringParameters without duplication', () => {
  // Prepare
  // This simulates real API Gateway V1 behavior where multi-value query params
  // appear in BOTH fields:
  // - queryStringParameters contains the LAST value
[...]

PR #4754, opened 2025-11-15
test(event-handler): add e2e tests

### LINKED ISSUE #4502: Maintenance: Add e2e Tests for Event Handler REST API
### Summary

We should add end to end tests against a real API Gateway endpoint to test the REST API Event Handler.

### Why is this needed?

The REST API Handler handler has several interacting components, especially between API Gateway and the lambda function; unit tests alone are not enough to give us confidence that everything is functioning as expected.

Event Handler

### Solution

At a very high level, this would involve creating an API Gateway backed by a lambda function and then invoking the API by making an HTTP request to a test endpoint once it is deployed and ensuring we receive the expected response.

We should have negative and positive tests:
- Test the happy path where a 200 request is returned
- Test the 404 case when a route is not present
- Test the 405 case when a HTTP method is disallowed
- Test the custom error case, where we return a specific expected error type for a particular scenario, e.g., a 400 error for an input validation failure

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR adds e2e tests for the HTTP Router part of the Event Handler utility.

The end to end test deploys a single CloudFormation stack with one lambda function that includes dozens of routes split between a few routers that are then merged into a single app.

The routes are intended to cover most of the use cases I could think of (with LLM help for the spec) and the tests are done by making direct requests using the `fetch` module. 

**Issue number:** closes #4502

---

### DIFF
--- a/packages/event-handler/tests/e2e/httpRouter.test.ts
+++ b/packages/event-handler/tests/e2e/httpRouter.test.ts
@@ -407,8 +407,6 @@ describe('REST Event Handler E2E tests', () => {
     });
   });
 
-  describe('Custom Middleware', () => {});
-
   describe('Nested Router', () => {
     it('handles GET request to nested router', async () => {
       // Prepare
--- a/packages/event-handler/tests/e2e/routers/binaryRouter.ts
+++ b/packages/event-handler/tests/e2e/routers/binaryRouter.ts
@@ -10,7 +10,7 @@ binaryRouter.post('/upload', async ({ req }) => {
     received: true,
     contentType,
     bodyLength: body.length,
-    isBase64Encoded: /^[A-Za-z0-9+/]+=*$/.test(body.replace(/\s/g, '')),
+    isBase64Encoded: /^[A-Za-z0-9+/]+=*$/.test(body.replaceAll(/\s/g, '')),
   };
 });
 


dreamorosi:
Before I can remove the `.skip` statements & `TODO`s from the tests we need to merge #4750 and address #4751.

packages/event-handler/tests/e2e/httpRouter.test.ts
```diff
@@ -0,0 +1,810 @@
...
+    it('applies middleware that modifies response', async () => {
+      // Prepare
+      const response = await fetch(`${apiUrl}/middleware`);
+      const data = await response.json();
+
+      expect(response.status).toBe(200);
+      expect(data.message).toBe('Middleware applied');
+      expect(data.responseTime).toBeGreaterThanOrEqual(0);
+      expect(typeof data.responseTime).toBe('number');
+    });
```
dreamorosi: @svozza I think I need your help.

This is the last test that I can't make pass, and I suspect it's because of a subtle design issue - or me misunderstanding how things should work.

The test fails like this:

```sh
TypeError: terminated
Caused by: Error: incorrect header check
Serialized Error: { errno: -3, code: 'Z_DATA_ERROR' }
```

which basically means there's some issue with the compression of the response (according to Google & LLMs).

I've tried to track the issue down and _I think_ this is the request flow:

1. Request to `/middleware`
2. Global compress middleware runs (inherited from `compressRouter.includeRouter()`)
3. Compress calls `next()` -> `loggingMiddleware` runs -> handler executes
4. `loggingMiddleware` returns plain object `{ message: 'Middleware applied', responseTime: duration }`
5. Object converted to Response with `Response.json()`
6. Control returns to compress middleware
7. Compress middleware attempts to pipe `reqCtx.res.body` through `CompressionStream` and sets `content-encoding: gzip`
8. Response sent with `content-encoding: gzip` header but body is either:
   - Already consumed from `Response.json()` causing compression to fail
   - Not actually compressed but header claims it is
9. Client `fetch().json()` tries to decompress, which causes the `Z_DATA_ERROR`

This happens because when router with middleware is included via `includeRouter()`, its middleware becomes global for entire parent app. In this case, the `compression` middleware from `compressRouter` now runs globally for ALL routes, including `/middleware`.


This happens because in line 762 of `Router.ts` we merge middleware functions like this `this.middleware.push(...router.middleware);`.

I guess my questions to you are:
- did I get the request flow right?
- if so, and the issue is what I said, was this intended? Should we create a more complex merging logic so that global middleware functions of a leaf router stay scoped to that router when merged?

svozza: The flow looks correct to me exception of this:

`Already consumed from `Response.json()` causing compression to fail`

That method stringifies the playload and then converts it to a stream before set it to be the `body`; it also set the correct headers for the `Response`. There is definitely something weird going on here, the compression is not being done properly when middleware returns early. I was able to create a unit test that fails iwth `Error: incorrect header check` when I try to unzip the response:

```ts
  it('compresses early return middleware', async () => {
    // Prepare
    const application = new Router();
    application.get(
        '/test',
        [
          compress({
            encoding: 'gzip',
          }),
          async () => {
            return {message: 'Middleware applied'}
            // return { test: 'y'.repeat(2000) }
          }
        ],
        () => {
          return body;
        }
    );

    // Act
    const result = await application.resolve(event, context);

    // Assess
    expect(result.isBase64Encoded).toBe(true);
    expect(result.headers).toEqual({
      "content-encoding": "gzip",
      "content-type": "application/json"
    });
    expect(
        gunzipSync(Buffer.from(result.body, 'base64')).toString('utf8'))
        .toEqual(JSON.stringify({message: 'Middleware applied'}))
  });
```

Regarding the middleware behaviour, yes, I'm almost sure it's intentional (correct me if I'm wrong here @sdangol as you implemented the feature) because that's how Python does it:

https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/event_handler/api_gateway.py#L2843

Regarding the more complex middleware merging behaviour, I'm in two minds, I can imagine some customers wanting the current behaviour and some wanting a more granular behaviour. Personally I think you should only apply global middleware before/after you have merged all your routers because trying to keep track of global things across multiple files is difficult. Unfortunately, we can't enforce that in code:

```ts
import someRouter from './someRouter.js'; //
import otherRouter from './someRouter.js';

const app = new Router();
app.includeRouter(someRouter);
app.includeRouter(otherRouter);

app.use(golablMiddleware1);
// ...
```

dreamorosi: Ok, what do you recommend we do?

I'd really want to get this merged but I also don't want to keep track of that weird behavior.

svozza: Oops, sorry, I meant to add that. I think regarding middleware merging we should keep the same behaviour as Python but document the footgun. Regarding the compression thing, I need to figure out what's going wrong, I will do it after lunch.

svozza: Actually, given that we now have a failing unit test case, I think we can remove this testcase for the moment and add it back in when it's fixed.

sdangol: I think we are currently doing the compression of the `reqCtx.res.body` which seems like it doesn't get mutated when we have multiple middlewares being composed and one of them shortciruits. This seems to be a bug in the `composeMiddleware` function. I'll create an issue for it and fix it.

dreamorosi:
Thanks for the engagement in the discussion - I have removed the tests and noted them down in their respective issues.

The current test suite is passing: https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/19507043601

PR #4759, opened 2025-11-16
feat(event-handler): add support for ALB

### LINKED ISSUE #4758: Feature request: Add support for ALB integration
### Use case

Currently event handler does not support the ALB lambda integration. We should add support for Application Load Balancer (ALB) events.

### Solution/User Experience

The event source should be transparent to users, they should be able to define handlers that return any valid response we support (JSON object, `APIGatewayProxyResult`, `ALBResult`, Web `Response` object) and event handler will automatically detect the input event and convert to the appropriate response.

```ts
const app = new Router();

// this returns APIGatewayProxyResult
app.get('/test', () => ({
    statusCode: 200,
    multiValueHeaders: {
        'set-cookie': ['cookie1=value1', 'cookie2=value2']
    }
}));

const v1Result = await app.resolve(aV1Event, context);

/**
v1Result = {
    "statusCode": 200,
    "headers": {
    },
    "body": "{\"message\":\"success\"}",
    "isBase64Encoded": false,
    "multiValueHeaders": {
        "set-cookie": [
            "cookie1=value1",
            "cookie2=value2"
        ]
    }
}
 */

const albResult = await app.resolve(anALBEvent, context);

/**
albResult = {
    "statusCode": 200,
    "statusDescription": "200 OK",
    "headers": {
    },
    "body": "{\"message\":\"success\"}",
    "isBase64Encoded": false,
    "multiValueHeaders": {
        "set-cookie": [
            "cookie1=value1",
            "cookie2=value2"
        ]
    }
}
 **/
```    

### Alternative solutions

```Markdown
Alternative solution would be for customers to have specific resolvers for specific event sources:


const albiRouter = new AlbRouter()
// will only work with ALB events
```

### Acknowledgment

### Future readers

svozza:
## Summary

This PR adds support for ALB lambda integrations, as with the other implementations, the router autodetects the event type and returns the appropriate response.

### Changes

- Added ALB event type guard
- Added ALB event to `Request` conversion function
- Added ALB response to `Response` conversion function
- Added look up map for HTTP codes to set `descriptionStatus` field correctly

#### Testing
- Added ALB tests to `describe.each` block in routing tests
- Added unit tests to specifically test `statusDescription` field
- Deployed to lambda function and tested with ALB event

**Issue number:** closes #4758 

---

### DIFF
--- a/packages/event-handler/src/rest/utils.ts
+++ b/packages/event-handler/src/rest/utils.ts
@@ -46,7 +46,7 @@ export function compilePath(path: Path): CompiledRoute {
     PARAM_PATTERN,
     (_match, paramName) => {
       paramNames.push(paramName);
-      return `(?<${paramName}>[${SAFE_CHARS}${UNSAFE_CHARS}\\w]+)`;
+      return String.raw`(?<${paramName}>[${SAFE_CHARS}${UNSAFE_CHARS}\w]+)`;
     }
   );
 
--- a/packages/event-handler/tests/unit/rest/helpers.ts
+++ b/packages/event-handler/tests/unit/rest/helpers.ts
@@ -41,6 +41,17 @@ export const createTestEvent = (
   resource: '',
 });
 
+const createAlbBody = (httpMethod: string, body: JSONValue): string | null => {
+  // ALB events represent GET and PATCH request bodies as empty strings
+  if (httpMethod === HttpVerbs.GET || httpMethod === HttpVerbs.PATCH) {
+    return '';
+  }
+  if (body === null) {
+    return null;
+  }
+  return JSON.stringify(body);
+};
+
 export const createTestEventV2 = (
   rawPath: string,
   method: string,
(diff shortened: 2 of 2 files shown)


packages/event-handler/src/rest/constants.ts
```diff
@@ -116,9 +116,78 @@ const COMPRESSION_ENCODING_TYPES = {
   ANY: '*',
 } as const;
 
+const HttpStatusText: Record<number, string> = {
+  // 2xx Success
+  200: 'OK',
```
sdangol, line 121: Should we use `HttpStatusCodes` as the keys?

packages/event-handler/src/rest/converters.ts
```diff
@@ -154,14 +160,45 @@ const proxyEventV2ToWebRequest = (event: APIGatewayProxyEventV2): Request => {
...
+  const hostname = headers.get('Host') ?? 'localhost';
+  const protocol = headers.get('X-Forwarded-Proto') ?? 'https';
+
+  const url = new URL(path, `${protocol}://${hostname}/`);
+  populateV1QueryParams(url, event);
+
+  // ALB events represent GET and PATCH request bodies as empty strings
+  const body =
+    httpMethod === HttpVerbs.GET || httpMethod === HttpVerbs.PATCH
+      ? null
```
sdangol, line 183: Shouldn't this be an empty string if it represents GET and PATCH bodies as empty strings?

svozza: You can't create a a `Request` object with a body, it throws an error.

svozza:
The SonarQube warning is from a piece of code I haven't touched.

dreamorosi:
> The SonarQube warning is from a piece of code I haven't touched.

You're right - it checks the file I think.

If possible, still address it please.

packages/event-handler/src/rest/constants.ts
svozza: The issue is that the `Request` object doesn't support `1xx` codes, it is possible to use `HttpStatusCodes` to create a new type that exludes those keys using the `Exclude` type utility but I'm not sure the complexity is worth it.

svozza:
TIL about `string.raw`. Have pushed the fix.

PR #4773, opened 2025-11-19
fix(event-handler): moved the response mutation logic to the `composeMiddleware` function

### LINKED ISSUE #4772: Bug: Response does not get mutated when Middleware returns early
### Expected Behavior

The Response in `reqCtx.res` should also get mutated when a middleware returns early when there are multiple middlewares. The middleware that would do the post processing should be able to access the response returned by the earlier middleware

### Current Behavior

The Response in `reqCtx.res` doesn't get mutated and the post processing middleware can't access the response

### Code snippet

This test fails currently as the response is not being mutated

```typescript
it('allows post processing middleware to access the response returned early by a pre processing middleware', async () => {
  // Prepare
  const app = new Router();
  let message = "";
  app.get('/test', [
    async ({ reqCtx, next }) => {
      await next();
      const clonedRes = reqCtx.res.clone();
      message = (await clonedRes.json()).message;
    },
    () => {
      return Promise.resolve({ message: 'Middleware applied' })
    }
  ], () => {
    return { message: 'Handler applied' }
  });

  // Act
  await app.resolve(
    createTestEvent('/test', 'GET'),
    context
  );

  // Assess
  expect(message).toEqual('Middleware applied');
})
```

### Steps to Reproduce

1. Run the above unit test in the project
2. The message is empty and hasn't been mutated

### Possible Solution

Move the Response mutation logic in [Router.ts](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/event-handler/src/rest/Router.ts#L316C7-L322C8) to the [`composeMiddleware` ](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/event-handler/src/rest/utils.ts#L298)helper function

latest

22.x

### Packaging format used

npm

### Execution logs

```Shell

```

sdangol:
## Summary

This MR fixes the bug when the Response didn't get mutated when a middleware returns early when there are multiple middlewares. 

### Changes

- Moved the Response mutation logic to the `composeMiddleware` function

**Issue number:** closes #4772 

---

### DIFF
--- a/.github/workflows/ossf_scorecard.yml
+++ b/.github/workflows/ossf_scorecard.yml
@@ -43,6 +43,6 @@ jobs:
 
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@014f16e7ab1402f30e7c3329d33797e7948572db # v4.31.3
+        uses: github/codeql-action/upload-sarif@e12f0178983d466f2f6028f5cc7a6d786fd97f4b # v4.31.4
         with:
           sarif_file: results.sarif
--- a/.github/workflows/run-e2e-tests.yml
+++ b/.github/workflows/run-e2e-tests.yml
@@ -31,6 +31,7 @@ jobs:
             packages/logger,
             packages/metrics,
             packages/parameters,
+            packages/event-handler,
             packages/tracer,
             layers,
           ]
--- a/examples/app/package.json
+++ b/examples/app/package.json
@@ -29,7 +29,7 @@
     "#errors": "./functions/commons/errors.js"
   },
   "devDependencies": {
-    "@types/aws-lambda": "^8.10.158",
+    "@types/aws-lambda": "^8.10.159",
     "@types/node": "24.10.1",
     "aws-cdk-lib": "^2.225.0",
     "constructs": "^10.4.3",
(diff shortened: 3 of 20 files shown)


packages/event-handler/tests/unit/rest/Router/middleware.test.ts
```diff
@@ -777,5 +777,33 @@ describe('Class: Router - Middleware', () => {
       expect(result.statusCode).toBe(403);
       expect(result.body).toBe('Route middleware response');
     });
+
+    it('allows post processing middleware to access the response returned early by a pre processing middleware', async () => {
```
svozza:
```suggestion
    it('allows post processing middleware to access the response returned early by a pre-processing middleware', async () => {
```

packages/event-handler/src/rest/utils.ts
```diff
@@ -294,8 +295,13 @@ export const composeMiddleware = (middleware: Middleware[]): Middleware => {
         );
       }
 
+      // middleware result takes precedence to allow short-circuiting
       if (middlewareResult !== undefined) {
         result = middlewareResult;
+        reqCtx.res = handlerResultToWebResponse(middlewareResult, {
```
svozza, line 306: Let's update the JSDoc to explain this behaviour, which is not necessarily obvious for a `compose` function.

Issue #4749, opened 2025-11-14
Docs: review Event Handler docs


dreamorosi:
### What were you searching in the docs?

Over the past few weeks we've made updates to the Event Handler ahead of the GA release, but in some cases we've let the docs lag behind as reported by internal customers.

### How can we improve?

Before the next release we should review the docs and make sure they're up to date.

### Acknowledgment

- [x] I understand the final update might be different from my proposed suggestion, or refused.

dreamorosi:
I've used Q CLI to do a couple passes at assessing the docs against the implementation in the codebase and these are the issues that were flagged. I think most of them look sensible and kind of match changes we've made since the last update but feel free to disregard any that are off - hope this helps!

### 1. Split Routers - Incorrectly Marked as "Coming Soon" ✅
- **Documentation (lines 566-576)**: Says "Coming soon" and claims "Router class doesn't provide a way to compose multiple router instances"
- **Reality**: `includeRouter()` method is fully implemented in `Router.ts` (lines 755-763)
- **Evidence**: Method exists with full implementation including context merging, route registry merging, error handler merging, and middleware merging
- **Impact**: Users don't know this feature exists and is ready to use

### 2. Response Streaming - Incorrectly Marked as "Coming Soon"
- **Documentation (lines 520-524)**: Says "Coming soon" with issue #4476
- **Reality**: `resolveStream()` method exists and is fully implemented in `Router.ts`, and `streamify()` utility is exported
- **Evidence**: Full implementation with `HttpResponseStream` support
- **Impact**: Users don't know streaming is available

### 3. Missing Exported Utilities Documentation
- **Documentation**: Doesn't list utility exports
- **Reality**: Package exports these utilities from `index.ts`:
  - `composeMiddleware` - mentioned in docs but import not shown
  - `isAPIGatewayProxyEventV1` - type guard
  - `isAPIGatewayProxyEventV2` - type guard
  - `isExtendedAPIGatewayProxyResult` - type guard
  - `isHttpMethod` - type guard
  - `streamify` - streaming helper
- **Impact**: Users don't know about helpful utilities

### 4. CORS `maxAge` Default Value is Wrong ✅
- **Documentation (line 446)**: Says default is `0`
- **Reality**: `DEFAULT_CORS_OPTIONS` in `constants.ts` does NOT include `maxAge` property (it's `undefined`)
- **Evidence**: The middleware checks `if (config.maxAge !== undefined)` before setting the header

[...]

PR #4779, opened 2025-11-20
docs(event-handler): updated docs for split router, catch all routes and error handler

sdangol:
## Summary

This PR updates the documentation for some of the features that have been released recently for the event handler utility

### Changes

- Update documentation for the Split Router feature
- Add documentation for the Catch all route feature
- Updated the CORS documentation
- Update the Built-in error handler documentation

**Issue number:** #4749 

---


docs/features/event-handler/rest.md
```diff
@@ -91,6 +91,35 @@ All dynamic route parameters will be available as typed object properties in the
...
 
+#### Catch-all routes
+
+For scenarios where you need to handle arbitrary or deeply nested paths, you can use regex patterns directly in your route definitions. These are particularly useful for proxy routes or when dealing with file paths.
+
+**We recommend** having explicit routes whenever possible; use catch-all routes sparingly.
+
+##### Using Regex Patterns
+
+You can use standard [Regular Expressions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions){target="_blank" rel="nofollow"} in your route definitions, for example:
```
svozza:
```suggestion
You can use standard [regular expressions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions){target="_blank" rel="nofollow"} in your route definitions, for example:
```

docs/features/event-handler/rest.md
```diff
@@ -91,6 +91,35 @@ All dynamic route parameters will be available as typed object properties in the
...
+
+    ```python hl_lines="7 10 13 20"
+    --8<-- "examples/snippets/event-handler/rest/gettingStarted_dynamic_routes_catch_all.ts"
+    ```
+
+???+ warning "Route Matching Priority"
+    - Routes are matched in **order of specificity**, not registration order
+    - More specific routes (exact matches) take precedence over regex patterns
+    - Among regex routes, the first registered matching route wins
+    - Always place catch-all routes (`.*`) last
```
svozza: This seems to contradict point 1, surely it doesn't matter where you place `.*` if what matters is specificity, the specificity algorithm will always rank this last. Or am I misunderstanding what this item is saying?

docs/features/event-handler/rest.md
```diff
@@ -565,15 +595,43 @@ Please [check this issue](https://github.com/aws-powertools/powertools-lambda-ty
...
+    ```typescript hl_lines="8"
+    --8<-- "examples/snippets/event-handler/rest/split_route_index.ts"
+    ```
+
+#### Route Prefix
+
+In the previous example, `split_route.ts` routes had a `/todos` prefix. This might grow over time and become repetitive.
 
-Once this feature is available, you will be able to define routes in separate files and import them into a main router file, improving code organization and maintainability.
+When necessary, you can set a prefix when including a `Router` instance. This means you could remove `/todos` prefix altogether.
```
svozza:
```suggestion
When necessary, you can set a prefix when including a `Router` instance. This means you can remove `/todos` prefix altogether.
```

docs/features/event-handler/rest.md
sdangol: The "non-regex" routes are matched in order of specificity, but the regex routes will be matched based on the order of registration.

```typescript
app.get(/.*/, () => ("Catch all");     // <----- This will be matched for all routes which don't have any specific matches. Eg: /files/xyz

app.get(/\/files\/.+/, () => ("Catch all files prefix");     // <----- This will not be matched for /files/xyz since it is already preceded by a matching regex route

app.get("/files/specific",() => ("Specific file");     // <----- This will be matched for /files/specific even if it's last
```

svozza: Can we say something like `Among regex routes, registration order detemrines matching precidence, therefore, always place catch-all routes (`.*`) last`

PR #4794, opened 2025-11-21
chore(deps): upgrade InvokeStore to v0.2.1

### LINKED ISSUE #4787: Maintenance: Update InvokeStore to v0.2.1
### Summary

The InvokeStore package release a new version (v.0.2.1) last week that didn't export types correctly for a particular `tsconfig.json` setup. This affected Powertools so we didn't not upgrade. Yesterday a new version was released that fix the issue. 

### Why is this needed?

v0.2.x contains important changes to the API that we need to account for:

```ts
// old
import InvokeStore from '@aws/lambda-invoke-store';

// new
import InvokeStore from '@aws/lambda-invoke-store';

 //.. 
export const handler = () => {
  const invokeStore = InvokeStore.getInstanceAsync();
}

```

Other

### Solution

We must upgrade to the new version and rather than using `InvokeStore` directly from the import, we must use `globalThis.awslambda.InvokeStore`.

### Acknowledgment

### Future readers

svozza:
## Summary
This PR updates InvokeStore to v0.2.1, which contained several breaking changes including async initialisatioon. 

### Changes

- use the `globalThis.awslambda.InvokeStore` to avoid the async initialisation.
- check for the `AWS_LAMBDA_MAX_CONCURRENCY` variable to decide whether to use `InvokeStore` because the `getContext` method no longer calls the underlying store.
- Update, logger, metrics, and batch utilities to use `globalThis.awslambda.InvokeStore`
- updated all the tests

**Issue number:** closes #4787 

---


packages/batch/src/BatchProcessingStore.ts
```diff
@@ -35,124 +36,152 @@ class BatchProcessingStore {
   #fallbackErrors: Error[] = [];
 
   public getRecords(): BaseRecord[] {
-    if (InvokeStore.getContext() === undefined) {
+    if (!shouldUseInvokeStore()) {
       return this.#fallbackRecords;
     }
-    return (InvokeStore.get(this.#recordsKey) as BaseRecord[]) ?? [];
+
+    if (globalThis.awslambda?.InvokeStore === undefined) {
```
svozza, line 43: I have made the pragmstic decision to only do this check in `getRecoords` and `serRecords` rather than litter 10 functions with this check. We know that these methods will always be called when processing records so that is suffiecient to be sure this is not not null.

packages/batch/src/types.ts
```diff
@@ -13,6 +14,12 @@ import type { parser } from './parser.js';
 import type { SqsFifoPartialProcessor } from './SqsFifoPartialProcessor.js';
 import type { SqsFifoPartialProcessorAsync } from './SqsFifoPartialProcessorAsync.js';
 
+declare global {
```
svozza, line 17: This is the workaround for the build issue I was seeing

dreamorosi: Wonder why we're seeing this only here and not elsewhere- any chance we can open an issue to check back later?

dreamorosi (approved):
e2e tests are passing, let's go

packages/batch/src/types.ts
svozza: Yes, will do.

svozza: Added here: #4796

PR #4834, opened 2025-12-02
feat(idempotency): Allow durable function to replay

### LINKED ISSUE #4831: Feature request: add support for durable function in Idempotency Utility
### Expected Behavior

Idempotency Utility works correctly when a durable function is suspended and replayed

### Current Behavior

- Idempotency utility breaks in some Lambda durable executions
- Will raise IdempotencyAlreadyInProgressError if a durable execution is suspended and then replayed, such as after a `context.wait()` call
- Identified as a duplicate of an existing payload, rather than a replay of the original payload (which should be allowed)

### Code snippet

```typescript
import {
  withDurableExecution,
  DurableContext,
} from "aws-durable-execution-sdk-js";

import {
  IdempotencyConfig,
  makeIdempotent,
} from "@aws-lambda-powertools/idempotency";

  tableName:
    process.env.IDEMPOTENCY_TABLE_NAME || "powertools-idempotency-table",
});

const innerHandler = async (event, context: DurableContext) => {
  config.registerLambdaContext(context.lambdaContext);

  try {
    console.log({ event });

    // This breaks the idempotency function as
    // the replay detects the existing "in progress" execution,
    // so throws an IdempotencyAlreadyInProgressError error.
    await context.wait("wait", 60);
    const processedData = await context.step(async () => {
      // Simulating idempotent operation
      return {
        id: "unique-id",
        timestamp: Date.now(),
      };
    });

    context.step(async () => {
      // Additional processing step
      console.log("Processing data:", processedData);
    });

    return {
      statusCode: 200,
      body: JSON.stringify({
        message: "Success",
        data: processedData,
      }),
    };
  } catch (error) {
    console.error(error);
    console.trace();
    return {
      statusCode: 500,
      body: JSON.stringify({
        message: "Error",
        error: error instanceof Error ? error.message : "Unknown error",
      }),
    };
  }
};

export const handler = withDurableExecution(
  makeIdempotent(innerHandler, {
    persistenceStore,
  })
);
```

### Steps to Reproduce

1. Create a lambda function with a durable config specified, and the handler above
2. Invoke the function
3. Observe an error is raised

### Possible Solution

_No response_

latest

24.x

### Packaging format used

npm

### Execution logs

```Shell

```

ConnorKirk:
## Summary

This PR adds support for Durable Functions to the Idempotency Utility.
In testing, we discovered that the Idempotency utility did not work with Durable Functions when they suspended and replayed. This is because the replay appeared like a duplicate, so raised an `IdempotencyAlreadyInProgressError`

We use the `DurableMode` property on the `DurableContext` object to track if this is a new execution, or a replay of an prior execution.

The behaviour of the handler is the same, except when it's a previously seen payload "in progress" and the `DurableMode` is `REPLAY_MODE`. Then the execution will continue, where previously it would have raised the error `IdempotencyAlreadyInProgressError`

### Changes

* Check the context type in `makeIdempotent`. If it's a durable context, extract the lambda context and the `DurableMode`.
* Add a new optional parameter to the `IdempotentHandler.handler()` method, `durableMode`.
* Update `IdempotentHandler` to check if the execution is in `REPLAY_MODE` when `IdempotencyAlreadyInProgressError` is thrown.
* Add tests to `IdempotentHandler.test.js`
* Add test to `makeIdempotent.test.js`

### Notes
* It's missing e2e tests. These will be added in a future PR.
* We use context.getTimeInMillis() to calculate the expiry time of the INPROGRESS status. Duplicate payloads received after that expiry time will be allowed. In a durable execution, the length of the workflow can be up to 1 year, across multiple lambda invocations. This is longer than the maximum 15 minutes that an INPROGRESS status can currently have. Idempotency records for long running workflows will not be protected from duplicate invokes beyond the first 15 minutes of their life.

**Issue number:** closes #4831 

---


packages/idempotency/tests/unit/IdempotencyHandler.test.ts
```diff
@@ -239,6 +239,76 @@ describe('Class IdempotencyHandler', () => {
...
+            )
+          );
+
+        // Act & Assess
+        await expect(idempotentHandler.handle({ isReplay: false })).rejects.toThrow(IdempotencyAlreadyInProgressError);
+    })
+
+    it("returns the result of the original durable execution when another durable execution with the same payload is invoked", async () => {
+
+      // Arrange
```
dreamorosi: Keyword is Prepare

dreamorosi (changes requested):

packages/idempotency/tests/unit/IdempotencyHandler.test.ts
```diff
@@ -239,6 +239,76 @@ describe('Class IdempotencyHandler', () => {
...
+
+      // Act
+      await idempotentHandler.handle({isReplay: true})
+
+      // Assess
+      expect(mockFunctionToMakeIdempotent).toBeCalled()
+    })
+
+    it("raises an IdempotencyAlreadyInProgressError error when isReplay is false and there is an IN PROGRESS record", async ()=> {
+        // Arrange
```
dreamorosi: The correct keyword is Prepare

dreamorosi (approved):
Thank you

ConnorKirk:
I've opened #4839 to track the E2E test task

PR #4837, opened 2025-12-02
fix(idempotency): Update durableExecutionMode string

### LINKED ISSUE #4831: Feature request: add support for durable function in Idempotency Utility

ConnorKirk:
## Summary

The Idempotency utility uses the `DurableContext.durableExecutionMode` value to determine if a function is replaying or executing. This allows us to discriminate between duplicate invokes (bad), and replays of the same function (good).

The value of the string changed, from `REPLAY_MODE` to `ReplayMode`. The earlier PR https://github.com/aws-powertools/powertools-lambda-typescript/pull/4834 didn't work, as it would never see the expected string.

## Changes

Check for `ReplayMode` instead of `REPLAY_MODE`

There also seems to be some formatting changes - I'm not sure why these didn't format correctly in the staging environment.

**Issue number:** closes #4831 (again)

---

### DIFF
--- a/.github/workflows/bootstrap_region.yml
+++ b/.github/workflows/bootstrap_region.yml
@@ -43,11 +43,11 @@ jobs:
     environment: layer-${{ inputs.environment }}
     steps:
       - name: checkout
-        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3  # v6.0.0
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
         with:
           ref: ${{ github.sha }}
       - name: Setup Node.js
-        uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
+        uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0
         with:
           node-version: 24
       - name: Setup dependencies
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -17,6 +17,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: 'Checkout Repository'
-        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
       - name: 'Dependency Review'
         uses: actions/dependency-review-action@3c4e3dcb1aa7874d2c16be7d79418e9b7efd6261 # v4.8.2
--- a/.github/workflows/make-release.yml
+++ b/.github/workflows/make-release.yml
@@ -44,11 +44,11 @@ jobs:
       RELEASE_VERSION: ${{ steps.set-release-version.outputs.RELEASE_VERSION }}
     steps:
       - name: Checkout code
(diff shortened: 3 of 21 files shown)


packages/idempotency/src/makeIdempotent.ts
```diff
@@ -146,7 +148,7 @@ function makeIdempotent<Func extends AnyFunction>(
       }
     }
 
-    const isReplay = args[1]?.durableExecutionMode === "REPLAY_MODE"
+    const isReplay = args[1]?.durableExecutionMode === 'ReplayMode';
```
svozza, line 151: How come we didn't have to update any unit tests for this?

ConnorKirk: Great question - I had the same thought yesterday.
We don't have any E2E tests yet, and the unit tests focus on the IdempotencyHandler class behaviour, rather than the decorator (where this logic happens). 

E2E tests are in a follow up PR once CDK support is out

ConnorKirk: This is probably a gap - I will add a test. Thanks for the question @svozza

ConnorKirk:
My pre-commit has clobbered the formatting for the file. I added two tests to the end of the `makeIdempotent.test.js` file.
I'm reluctant to skip the pre-commit formatting, but this seems odd?

ConnorKirk:
#4839 - Task for adding E2E tests

packages/idempotency/src/makeIdempotent.ts
svozza: Yeah, I think E2E are only way to truly test this but it doesn't do any harm to have a unit test to make sure that we are testing the right string here.

PR #4840, opened 2025-12-03
fix(logger): merged temp keys with same keys when appending to the logger 

### LINKED ISSUE #4838: Bug: Temp keys no longer being merged when appended to logger
### Expected Behavior

Say we have an instance of `logger` and we've used `appendKeys()` to add an object `obj` to it.

In `2.28.1`, if we then append `obj` to `logger` again, the existing `obj` is merged with the one being added. For example:

```js
logger.appendKeys({ obj: { key1: 'value' } })
logger.appendKeys({ obj: { key2: 'value' } })
logger.critical('')

// -> { ... "obj":{"key1":"value","key2":"value"}}
```

### Current Behavior

In `2.29.0`, the existing `obj` is replaced by the one being added. For example:

```js
logger.appendKeys({ obj: { key1: 'value' } })
logger.appendKeys({ obj: { key2: 'value' } })
logger.critical('')

// -> { ... "obj":{"key2":"value"}}
```

Note that this does _not_ happen with persistent keys in either `2.28.1` or `2.29.0`:

```js
logger.appendPersistentKeys({ obj: { key1: 'value' } })
logger.appendPersistentKeys({ obj: { key2: 'value' } })
logger.critical('')

// -> { ... "obj":{"key1":"value","key2":"value"}}
```

### Code snippet

```js
import { Logger } from '@aws-lambda-powertools/logger'
const logger = new Logger()

logger.appendKeys({ obj: { key1: 'value' } })
logger.appendKeys({ obj: { key2: 'value' } })
logger.critical('')
```

### Steps to Reproduce

See above.

### Possible Solution

_No response_

latest

22.x

### Packaging format used

Lambda Layers

### Execution logs

```Shell

```

sdangol:
## Summary

This PR fixes the issue with the temporary keys not being merged when appending to the Logger. When we try to append a temporary key with an object value when there's a key with the same name, it was getting overridden instead of merging. The keys would get merged previously before [version 2.29.0](https://github.com/aws-powertools/powertools-lambda-typescript/releases/tag/v2.29.0) was released. There were some changes in the logic which caused this change in behavior.

### Changes

- Merged the temporary keys instead of overriding them
- Added tests to verify the behavior

**Issue number:** closes #4838 

---

### DIFF
--- a/.github/workflows/bootstrap_region.yml
+++ b/.github/workflows/bootstrap_region.yml
@@ -43,11 +43,11 @@ jobs:
     environment: layer-${{ inputs.environment }}
     steps:
       - name: checkout
-        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3  # v6.0.0
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
         with:
           ref: ${{ github.sha }}
       - name: Setup Node.js
-        uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
+        uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0
         with:
           node-version: 24
       - name: Setup dependencies
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -17,6 +17,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: 'Checkout Repository'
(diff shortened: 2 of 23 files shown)


packages/logger/src/LogAttributesStore.ts
```diff
@@ -1,5 +1,6 @@
 import '@aws/lambda-invoke-store';
 import { shouldUseInvokeStore } from '@aws-lambda-powertools/commons/utils/env';
+import merge from 'lodash.merge';
```
svozza, line 3: Btw, not for this PR but I would be in favour of us writing our own deep merge utility function in the commons package so we could remove this dependency.

sdangol: Yeah, makes sense. Will create an issue to track that

PR #4892, opened 2025-12-20
refactor(logger): replace lodash.merge with internal deepMerge utility

### LINKED ISSUE #4846: Maintenance: Replace `lodash.merge` with a custom deep merge utility function
### Summary

We're using a third party dependency, `lodash.merge` on our Logger utility. We would want to replace the dependency on this library and implement our own deep merge utility function.

### Why is this needed?

We would ideally want minimal dependency on other third party libraries

Logger

### Solution

_No response_

### Acknowledgment

### Future readers

jaimellamasi:
### Changes

* Removes `lodash.merge` dependency from logger package
* Adds new `deepMerge` utility to commons package
* Includes prototype pollution and circular reference protection

**Issue number:** closes #4846

---

### DIFF
--- a/packages/commons/src/deepMerge.ts
+++ b/packages/commons/src/deepMerge.ts
@@ -14,11 +14,14 @@ const isPlainObject = (value: unknown): value is Record<string, unknown> => {
 };
 
 /**
- * Merge source array into target array by index.
+ * Merge source array items into target array by index.
+ *
+ * When both source and target items at the same index are plain objects,
+ * they are merged recursively. Otherwise, the source item replaces the target.
  *
  * @internal
  */
-const mergeArraysByIndex = (
+const mergeArrayItemsByIndex = (
   targetArray: unknown[],
   sourceArray: unknown[],
   seen: WeakSet<object>
@@ -63,7 +66,7 @@ const handleArrayMerge = (
     return;
   }
 
-  mergeArraysByIndex(targetValue, sourceArray, seen);
+  mergeArrayItemsByIndex(targetValue, sourceArray, seen);
 };
 
 /**


packages/commons/src/deepMerge.ts
```diff
@@ -0,0 +1,164 @@
...
+    const tgtItem = targetArray[i];
+
+    const isSrcPlainObject = isPlainObject(srcItem);
+
+    // Skip already-seen objects to prevent circular references
+    if (isSrcPlainObject && seen.has(srcItem)) {
+      continue;
+    }
+
+    // Merge nested plain objects recursively
```
svozza, line 40: This method is called `mergeArraysByIndex` but it also seems to handle merging objects too. Is this intentional?

jaimellamasi: Good catch! Somehow the name can be misleading. I've renamed it to mergeArrayItemsByIndex and updated the JSDoc to clarify the behavior. The term "items" better conveys that we're working with the contents of the arrays, which can include objects that get merged recursively. Does it make sense?

svozza: Ah yes, that's definitely clearer!

svozza:
This is great work @jaimellamasi. Thank you for the contribution!

dreamorosi:
I'm seeing some transient issues on the end to end tests that _I think_ are unrelated to this PR, but since I want to make sure these tests are green before merging I'd ask to please hold on with merging this PR until I have fixed them.

svozza:
Hi @jaimellamasi, I checked out your changes and ran the updated E2E tests from #4909 on the branch and everything worked so we are good to go. Could you please rebase against `main` to resolve the merge conflict? It's just a `package-lock.json` issue so you should be able to fix it with `npm install`.

sdangol (approved):

svozza:
@jaimellamasi we are releasing today so I have done the rebase on your behalf, all your other commits will still be credited in the git history.

faberchri:
@jaimellamasi and @sdangol  Just a heads up: `deepMerge` is not exactly equivalent to `lodash.merge`.
See:
```JS
Welcome to Node.js v24.13.0.
Type ".help" for more information.

{ deepMerge } = require('@aws-lambda-powertools/commons')
_ = require('lodash')

_.merge({a:'x'}, {a: undefined})
{ a: 'x' }
deepMerge({a:'x'}, {a: undefined})
{ a: undefined }
```

Not sure if the difference in behaviour between `lodash.merge` and `deepMerge` is intentional. However, for us the difference resulted in losing the `message` props in all log statements, because we essentially did (without being aware of it)
```JS
    const logItem = new LogItem({
      attributes: {
        message: 'our message',
      },
    });
   ...
   logItem.addAttributes({ message: undefined, other: 'stuff' });
```
.

svozza:
Interesting, it looks like Lodash `deepMerge` has special behaviour for handling `undefined`. We followed the semantics of `Object.assign`, which imo is more strictly correct:

```js
let obj = {a: 100};
let obj1 = {a: undefined}
Object.assign(obj, obj1);
// => { a: undefined }
```

Is there a reason why you're doing `logItem.addAttributes({ message: undefined, other: 'stuff' });` rather than `logItem.addAttributes({ other: 'stuff' });`?

faberchri:
> Interesting, it looks like Lodash deepMerge has special behaviour for handling undefined.

Yes and it's clearly documented: [Source properties that resolve to undefined are skipped if a destination value exists.](https://lodash.com/docs/4.17.23#merge)

> Is there a reason why you're doing logItem.addAttributes({ message: undefined, other: 'stuff' }); rather than logItem.addAttributes({ other: 'stuff' });?

No, we weren't aware of the fact that we always add `message: undefined` to the input of `addAttributes`. Actually, we did `logItem.addAttributes(redactedAdditionalAttrs)` where `redactedAdditionalAttrs` is the output of `'@amaabca/sensitive-param-filter'` with some custom handling for a prop with name `message`. This custom handling resulted in always adding `{message: undefined}`. IMO in general it shouldn't make a difference whether a prop is missing or undefined, here it does.

Still, it's kind of a (undocumented?) breaking change of version [v2.30.1](https://github.com/aws-powertools/powertools-lambda-typescript/releases/tag/v2.30.1).

svozza:
Yep, agreed this is a breaking change and we should fix it, I was just curious as to your use case. Could you open a bug report please? We will try to get this into the next release. If you would like to contribute a PR to fix this too, that's most welcome as well.

faberchri:
@svozza Added an issue: https://github.com/aws-powertools/powertools-lambda-typescript/issues/5138

svozza:
Thank you!

PR #4918, opened 2026-01-06
feat: allow @middy/core v7

### LINKED ISSUE #4920: Feature request: Support @middy/core v7
### Use case

@middy/core v7 has been released: https://github.com/middyjs/middy/releases/tag/7.0.0

### Solution/User Experience

Powertools cannot be used in projects that use @middy/core v7 when using NPM 11.6.2. Currently this fails when installing it using NPM:

```bash
npm warn Found: @middy/core@6.4.5
npm warn node_modules/@middy/core
npm warn   @middy/core@"7.x" from the root project
npm warn   1 more (@aws-lambda-powertools/metrics)
npm warn
npm warn Could not resolve dependency:
npm warn peerOptional @middy/core@"4.x || 5.x || 6.x" from @aws-lambda-powertools/metrics@2.30.1
npm warn node_modules/@aws-lambda-powertools/metrics
npm warn   @aws-lambda-powertools/metrics@"2.30.1" from the root project
```

### Acknowledgment

### Future readers

coderbyheart:
See https://github.com/middyjs/middy/releases/tag/7.0.0

Closes #4920

## Summary

This adds `@middy/core@v7` as an allowed peer dependency.

Ideally we also bump the min. Node.js version to 22, but since AWS lambda still supports Node.js 20 until July 2026 [^runtimes] I opted not to introduce this breaking change.

[^runtimes]: https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

### DIFF
--- a/docs/getting-started/lambda-layers.md
+++ b/docs/getting-started/lambda-layers.md
@@ -19,41 +19,41 @@ We publish the Lambda Layer for Powertools for AWS Lambda in all commercial regi
 
 | Region           | Layer ARN                                                                                                  |
 | ---------------- | ---------------------------------------------------------------------------------------------------------- |
-| `us-east-1`      | [arn:aws:lambda:us-east-1:094274105915:layer:AWSLambdaPowertoolsTypeScriptV2:42](#){: .copyMe}             |
-| `us-east-2`      | [arn:aws:lambda:us-east-2:094274105915:layer:AWSLambdaPowertoolsTypeScriptV2:42](#){: .copyMe}             |
-| `us-west-1`      | [arn:aws:lambda:us-west-1:094274105915:layer:AWSLambdaPowertoolsTypeScriptV2:42](#){: .copyMe}             |
-| `us-west-2`      | [arn:aws:lambda:us-west-2:094274105915:layer:AWSLambdaPowertoolsTypeScriptV2:42](#){: .copyMe}             |
-| `ap-south-1`     | [arn:aws:lambda:ap-south-1:094274105915:layer:AWSLambdaPowertoolsTypeScriptV2:42](#){: .copyMe}            |
-| `ap-south-2`     | [arn:aws:lambda:ap-south-2:094274105915:layer:AWSLambdaPowertoolsTypeScriptV2:42](#){: .copyMe}            |
-| `ap-east-1`      | [arn:aws:lambda:ap-east-1:094274105915:layer:AWSLambdaPowertoolsTypeScriptV2:42](#){: .copyMe}             |
-| `ap-northeast-1` | [arn:aws:lambda:ap-northeast-1:094274105915:layer:AWSLambdaPowertoolsTypeScriptV2:42](#){: .copyMe}        |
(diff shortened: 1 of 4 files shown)


svozza:
Thank you for the PR. As you can see, one of the GitHub checks is failing. Could you please create an issue for this and link to it in your PR?

coderbyheart:
@svozza Done!

dreamorosi:
Before we can merge this we need to add this new version to the test matrix in the e2e tests. I don't remember if we have dropped support for an older version of Middy.js yet, but we should keep only versions that align with Node.js 20, 22, and 24.

Once we have Middy.js in our test suite and it's passing, we can merge this.

svozza:
Yes, I've been looking at the E2E tests and it looks like we are currently only running them against Middy v5 for most tests.

coderbyheart:
Makes sense. I dug a bit into it, and from what I understand how it could be achieved that the lambda test function is bundled with the `@middy/core` version that is specified in the build matrix is through an esbuild plug-in that influences how esbuild resolves the `@middy/core` peer dependency at build time.

dreamorosi:
We don't use any esbuild plugin, the `NodejsFunction CDK` construct that we use in this project doesn't support them.

We just have a number of Middy.js versions installed under different aliases, and import one or the other using that alias.

I think we should look at Middy.js release notes, figure out which one maps with each Node.js version, and remove the ones for versions older than 20 (if any).

As long as we have 1-2 test fixtures with v7 I think that's enough for now.

svozza:
It looks like Middy 4 targeted Node 18, so we could remove that and replace with v7: https://middy.js.org/docs/upgrade/4-5. We don't have any Middy 4 tests so we'll need to move some of the v5 tests to v7.

svozza:
The Tracer tests might make sense, because that means we would have Logger -> v5,  Idempotency -> v6, and Tracer -> v7:

https://github.com/aws-powertools/powertools-lambda-typescript/blob/f7c1001fc7cd4aafb5253d0a714cddde79ab5d89/packages/tracer/tests/e2e/middy.test.functionCode.ts#L3

That said, there's only one Middy test in that utility so we might want to change one of the logger ones too, to give a bit more coverage.

dreamorosi:
I'm ok to remove Middy.js v4, my preference would be to add v7 to the Idempotency tests since most of the v7 changes are related to Durable Functions, which we also made changes for in that module.

With that said, I don't expect breaking changes so either works.

coderbyheart:
Added middy v7 and bumped all alias to respective latest version.

svozza:
I've ran the E2E tests and the idempotency ones are passing. However, I've realised that this means we now have no coverage of Middy v6 now. I propose we change two of the tests in the Logging utility to use v6 instead of v5.

coderbyheart:
I've changed the tracer to v6:

https://github.com/aws-powertools/powertools-lambda-typescript/pull/4918/changes#diff-211a56102c6dd0645f32ff9b3eb27d9fa0b79c041738d94965b67f836b65b978R3

svozza:
Ah I missed that. Great stuff!

PR #4978, opened 2026-01-28
test: add e2e tests for batch processing

### LINKED ISSUE #4489: Maintenance: add e2e tests for Batch Processing
### Summary

Following the `v2.26.0` release we introduced a regression that couldn't be detected in unit tests and existing e2e tests. 

Because of this, we should add isolated and dedicated test cases for the Batch Processing utility.

### Why is this needed?

So that we can increase the confidence that new code in this utility stays backwards compatible and with less chances of regressions.

Other

### Solution

_No response_

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR adds an integration test case for the Batch Processing utility, which was not previously covered. The test purposefully includes only one test case with standard SQS queue since the main goal of the test is to verify that the utility works in Lambda together with Parser, rather than covering the entire spectrum of features of the utility, which rely on similar implementations anyway.

To create this test I had to implement a new utility function that tails CloudWatch logs since this is also the first test that we add that can't be observed by querying logs or by collecting the response of a specific function invocation. This is due to the fact that we only publish messages to a SQS queue and we don't control how many or when the function invocations will happen.

This new testing utility lives under the packages/testing package and handles the tailing of a CloudWatch Log Group, collects the logs, and stops tailing after a given idle time.

**Issue number:** closes #4489

---

### DIFF
--- a/packages/testing/src/LogTailer.ts
+++ b/packages/testing/src/LogTailer.ts
@@ -231,22 +231,31 @@ class LogTailer {
   }
 
   #collectSessionResults(
-    sessionResults: SessionResult[] | undefined,
+    sessionResults: SessionResult[],
     requestLogs: Map<string, ParsedLog[]>
-  ): boolean {
-    if (!sessionResults?.length) return false;
+  ): Map<string, ParsedLog[]> {
+    const mergedLogs = new Map<string, ParsedLog[]>();
+    for (const [requestId, logs] of requestLogs.entries()) {
+      mergedLogs.set(requestId, [...logs]);
+    }
+
+    if (!sessionResults.length) return mergedLogs;
 
     for (const logEvent of sessionResults) {
       const { requestId, logObj } = this.#parseLogEvent(logEvent);
       if (!requestId || !logObj) continue;
 
-      if (!requestLogs.has(requestId)) {
-        requestLogs.set(requestId, []);
+      if (!mergedLogs.has(requestId)) {
+        mergedLogs.set(requestId, []);
       }
-      requestLogs.get(requestId)?.push(logObj);
+      mergedLogs.get(requestId)?.push(logObj);
     }
 
-    return true;
+    for (const logs of mergedLogs.values()) {
+      this.#sortLogs(logs);
+    }
+
+    return mergedLogs;
   }
 
   #shouldAutoStop(
@@ -292,7 +301,7 @@ class LogTailer {
       throw new Error('No response available');
     }
 
-    const requestLogs: Map<string, ParsedLog[]> = new Map();
+    let requestLogs: Map<string, ParsedLog[]> = new Map();
     let aborted = false;
     let idleTimer: NodeJS.Timeout | undefined;
 
@@ -319,12 +328,12 @@ class LogTailer {
       for await (const event of this.#response.responseStream || []) {
         if (aborted) break;
 
-        const hadLogData = this.#collectSessionResults(
-          event.sessionUpdate?.sessionResults,
-          requestLogs
-        );
+        const sessionResults = event.sessionUpdate?.sessionResults ?? [];
+        if (sessionResults.length > 0) {
+          refreshIdleTimer();
+        }
 
-        if (hadLogData) refreshIdleTimer();
+        requestLogs = this.#collectSessionResults(sessionResults, requestLogs);
 
         if (this.#shouldAutoStop(requestLogs, cooldownState)) {
           markAborted();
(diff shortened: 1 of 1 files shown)


dreamorosi:
Integration [tests - including the new ones - are green](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/21472169288).

packages/testing/src/LogTailer.ts
```diff
@@ -0,0 +1,349 @@
...
+    };
+
+    refreshIdleTimer();
+
+    try {
+      for await (const event of this.#response.responseStream || []) {
+        if (aborted) break;
+
+        const hadLogData = this.#collectSessionResults(
+          event.sessionUpdate?.sessionResults,
```
svozza: If we have a fallback value here we can remove `undefined` from the type signature of the `sessionResults` param in `collectSessionResults`:

```ts
event.sessionUpdate?.sessionResults ?? [];
```

packages/testing/src/LogTailer.ts
```diff
@@ -0,0 +1,349 @@
...
+          requestLogs
```
svozza: Rather than passing an empty `map` in here and then mutating it, could the map  creation and population be in `collectSessionResults` and have this function return the map instead? Then the check here just becomes a check to see if the map is empty or not. It's confusing to me to have a function called `collectX` that returns a boolean rather than a collection.

packages/testing/src/LogTailer.ts
```diff
@@ -0,0 +1,349 @@
...
+    if (!sessionResults?.length) return false;
+
+    for (const logEvent of sessionResults) {
+      const { requestId, logObj } = this.#parseLogEvent(logEvent);
+      if (!requestId || !logObj) continue;
+
+      if (!requestLogs.has(requestId)) {
+        requestLogs.set(requestId, []);
+      }
+      requestLogs.get(requestId)?.push(logObj);
```
svozza: Do we need this `?`, we've already checked above with `requestLogs.has(requestId)` so we know the array will be here.

packages/testing/src/LogTailer.ts
```diff
@@ -0,0 +1,349 @@
...
+
+        if (this.#shouldAutoStop(requestLogs, cooldownState)) {
+          markAborted();
+          break;
+        }
+      }
+
+      if (idleTimer) clearTimeout(idleTimer);
+
+      for (const logs of requestLogs.values()) {
```
svozza: Am I following the logic correctly here: the code creates an iterator of arrays from a map, then mutates each of those arrays in place from the original map that was passed in, and then returns that map whose values been mutated? 

To make this a bit easier to follow, could this sorting logic also be moved to `collectSessionResults` so all the mutation is confined to a single a function that just returns a map of sorted logs.

packages/testing/src/LogTailer.ts
dreamorosi: I agree and dislike it, but this is a long standing limitation of TypeScript (see microsoft/TypeScript#13086) that causes it to not be able to "follow" the flow analysis. Removing the `?` would mean either adding a `!` plus a linting suppression, or adding a `ts-ignore`, so I went for `?`.

svozza: Ah yes, this has caught me before too, `?` is definitely preferable imo.

dreamorosi:
Thanks for the review @svozza - good shouts on the simplification, I think I addressed most/all of them, let me know if this is what you had in mind.

packages/testing/src/LogTailer.ts
```diff
@@ -0,0 +1,354 @@
...
+    try {
+      for await (const event of this.#response.responseStream || []) {
+        if (aborted) break;
+
+        const sessionResults = event.sessionUpdate?.sessionResults ?? [];
+        if (sessionResults.length > 0) {
+          refreshIdleTimer();
+        }
+
+        requestLogs = this.#collectSessionResults(sessionResults, requestLogs);
```
svozza, line 336: Do we need to pass this map in now? It will always empty because of the definition on L304. I think `sessionResults` as the only param is sufficient now that the map is constucted from this array.

svozza:
> Thanks for the review @svozza - good shouts on the simplification, I think I addressed most/all of them, let me know if this is what you had in mind.

One tiny comment but other than that LGTM.

packages/testing/src/LogTailer.ts
dreamorosi: It's not always empty because we can process multiple log streams in the same tailing session - if we drop it, the next stream will always replace the previous I think.

svozza: Ah I get you!

PR #4982, opened 2026-02-01
feat(event-handler): add tracer middleware for HTTP routes

### LINKED ISSUE #4821: Feature request: Add Event Handler Tracer Middleware
### Use case

We should have a way of getting traces from HTTP routes in event handler so that the customer does not have to manually instrument their routes.

### Solution/User Experience

We could create a middleware that behaves similarly to the Middy tracer middlware we currently have. A sample implementation could look like this:

```ts
export const tracerMiddleware = (tracer: Tracer, options?: {captureResponse?: boolean}): Middleware => {
    const {captureResponse = true} = options ?? {};
    
    return async ({ reqCtx, next }) => {
        const url = new URL(reqCtx.req.url);
        const subsegment = tracer.getSegment()?.addNewSubsegment(url.pathname);
        subsegment && tracer.setSegment(subsegment);
        tracer.annotateColdStart();
        tracer.addServiceNameAnnotation();
        try {
            await next();
        } catch(err) {
            tracer.addErrorAsMetadata(err as Error);
            subsegment?.close();
            subsegment && tracer.setSegment(subsegment.parent);
            throw err;
        }
        if(captureResponse && reqCtx.res.headers.get('Content-Type') === 'application/json') {
            const responseBody = await reqCtx.res.clone().json() ?? {};
            tracer.addResponseAsMetadata(responseBody);
        }
        subsegment?.close();
        subsegment && tracer.setSegment(subsegment.parent);
    }
}
```

Open questions:
1. Should the middleware take multiple tracer instances like our Middy one does?
2. Should the logic around capturing responses be more complex, i.e., should we disallow it when we are in streaming mode? This would require us to add data to` reqCtx` to indicate when we are in HTTP streaming mode.

### Alternative solutions

```Markdown
We could tell users to use the Middy tracer middleware.
```

### Acknowledgment

### Future readers

arnabrahman:
## Summary
Add tracer middleware for tracing HTTP requests automatically
### Changes

- Create tracing middleware to trace HTTP routes
- Capture cold start, service name annotation
- Capture JSON response as metadata; other responses are not captured
- When tracing is disabled & during streaming, the middleware is skipped
- Capture route error during tracing
- Keep the tracer package in peer-dependency & expose the tracer middleware as a subpath for avoiding tracer dependency
- 100% unit tests are covered
- Docs are added for tracer middleware

**Issue number:** closes #4821 

---


packages/event-handler/src/http/middleware/tracer.ts
```diff
@@ -0,0 +1,91 @@
...
+      throw err;
+    } finally {
+      if (segment && subSegment) {
+        try {
+          subSegment.close();
+        } catch (error) {
+          console.warn(
+            'Failed to close or serialize segment %s. Data might be lost.',
+            subSegment.name,
+            error
```
arnabrahman: Does the middleware have logger instance as an option so that we can use it here?

svozza: No, ultimately middleware are just functions. I guess we could add the ability to pass a logger in with the `TracerOptions`?

packages/event-handler/tests/unit/http/middleware/tracer.test.ts
```diff
@@ -0,0 +1,383 @@
+import context from '@aws-lambda-powertools/testing-utils/context';
+import { Tracer } from '@aws-lambda-powertools/tracer';
+import { Segment, Subsegment } from 'aws-xray-sdk-core';
+import { afterAll, beforeEach, describe, expect, it, vi } from 'vitest';
+import { tracer as tracerMiddleware } from '../../../../src/http/middleware/tracer.js';
+import { Router } from '../../../../src/http/Router.js';
+import { createTestEvent, ResponseStream } from '../helpers.js';
+
+describe('Tracer Middleware', () => {
+  const ENVIRONMENT_VARIABLES = process.env;
```
svozza: Let's use the `vitest` environment variable utilities here rather than managing this manually: https://vitest.dev/api/vi.html#vi-stubenv.

packages/event-handler/tests/unit/http/middleware/tracer.test.ts
```diff
@@ -0,0 +1,383 @@
...
+  let app: Router;
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+    vi.resetModules();
```
svozza: I don't think we're doing anything in our tests that requires`resetModules`.

packages/event-handler/src/http/middleware/tracer.ts
arnabrahman: Oh yes, no need for that. But i agree on having logger options in `TracerOptions`

packages/event-handler/tests/unit/http/middleware/tracer.test.ts
arnabrahman: Actually, we don't need env variable handling either.

packages/event-handler/src/http/middleware/tracer.ts
svozza: Cool. Let's add a unit test for this configuration option too.

dreamorosi:
Hi @svozza - do you think this is ready to be merged?

svozza:
Oh yes, sorry, I thought I'd approved it already!

svozza:
Forget to say yesterday, great work as always @arnabrahman!

PR #4990, opened 2026-02-04
fix(event-handler): handle set-cookie header values with multiple attributes

### LINKED ISSUE #4986: Bug: Event Handler Response Headers incorrectly handling Set-Cookie values
### Expected Behavior

The current [implementation allows a semicolon](https://github.com/aws-powertools/powertools-lambda-typescript/blob/180a9d109a00425a809054a040e04eed9a277dd5/packages/event-handler/src/http/converters.ts#L232) to delimit multiple header values.  However this conflicts with the the syntax used with the `set-cookie` header with attributes.

### Current Behavior

When a cookie is set with [attribute syntax](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#syntax) it's treated as a separate cookie.  This creates incorrect and unwanted cookies.

### Code snippet

```typescript
response.headers.append(
	"Set-Cookie",
	"foo=bar; Max-Age=3600" // This turns into two cookies
);

response.headers.append(
	"Set-Cookie",
	"bar=baz"
);
```

When the response is returned from the handler, the header parsing produces three 'set-cookie' headers instead of two.

### Steps to Reproduce

Create a route similar to this:

```typescript
router.get("/foo", async () => {
	const response = Response.json({ foo: "bar" });

	response.headers.append(
		"Set-Cookie",
		"foo=bar; Max-Age=3600",
	);

	response.headers.append("Set-Cookie", "bar=baz");

	return response;
});
```

### Possible Solution

It seems [getSetCookie()](https://developer.mozilla.org/en-US/docs/Web/API/Headers/getSetCookie) might be a more reliable way to get this multi-value header.  In which case we probably want to ignore the `set-cookie` when looping through the headers 

latest

24.x

### Packaging format used

npm

### Execution logs

```Shell

```

nateiler:
## Summary
This fixes the scenario where a single `set-cookie` header value `foo=bar; Max-Age=3600` is treated as two `set-cookie` values because multiple values may be separated by a `;`.

### Changes
When parsing the headers, `set-cookie` headers are handled using the [getSetCookie](https://developer.mozilla.org/en-US/docs/Web/API/Headers/getSetCookie) method.

Logic to detect multiple header entries has been removed as multiple header values are appended to a single header (via comma).

**Issue number:** closes #4986

---


packages/event-handler/src/http/converters.ts
```diff
@@ -228,13 +228,30 @@ const webHeadersToApiGatewayV1Headers = (webHeaders: Headers) => {
...
+  }
+
+  if (allCookies.length > 1) {
+    multiValueHeaders['set-cookie'] = allCookies;
+  } else if (allCookies.length === 1) {
+    headers['set-cookie'] = allCookies[0];
+  }
+
   for (const [key, value] of webHeaders.entries()) {
+    // Skip set-cookie as it's already handled above
```
dreamorosi: Let's remove these comments please

packages/event-handler/src/http/converters.ts
```diff
@@ -228,13 +228,30 @@ const webHeadersToApiGatewayV1Headers = (webHeaders: Headers) => {
...
+      continue;
+    }
+
     const values = value.split(/[;,]/).map((v) => v.trimStart());
 
-    if (headers[key]) {
-      multiValueHeaders[key] = [headers[key], ...values];
-      delete headers[key];
-    } else if (values.length > 1) {
+    if (values.length > 1) {
```
dreamorosi, line 250: Why are we removing this logic?

packages/event-handler/tests/unit/http/Router/basic-routing.test.ts
dreamorosi, line 1: There's a lot of diffs that seem to be unrelated to this PR - any chance you could revert?

dreamorosi (changes requested):
Thank you for the PR @nateiler - appreciate it.

I have left a couple comments on the diff, would like to also get @svozza's review on the PR before we merge/

packages/event-handler/tests/unit/http/converters.test.ts
```diff
@@ -1247,7 +1246,7 @@ describe('Converters', () => {
     it('handles multi-value headers split by semicolon', () => {
       // Prepare
       const headers = new Headers({
-        'set-cookie': 'session=abc123; theme=dark',
+        'cache-control': 'no-cache; no-store',
```
svozza: How come we're changing these?

packages/event-handler/tests/unit/http/Router/basic-routing.test.ts
```diff
@@ -273,7 +271,10 @@ describe('Class: Router - V1 Multivalue Headers Support', () => {
       statusCode: 200,
       body: JSON.stringify({ message: 'success' }),
       headers: { 'content-type': 'application/json' },
-      multiValueHeaders: { 'set-cookie': ['session=abc123', 'theme=dark'] },
+      multiValueHeaders: {
+        'cache-control': ['no-cache', 'no-store'],
```
svozza, line 275: Where are these `cache-control` headers coming from?

packages/event-handler/tests/unit/http/converters.test.ts
nateiler: `set-cookie` is a header that shouldn't be separated by a semicolon.  Keeping it as it were would result in a single `set-cookie` header with an unsupported ['theme' attribute](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#attributes) which would cause the test to fail and ultimately seems to defeat the purpose of the test.  `cache-control` represents a normal header that can be separated by a semicolon.   Other tests cover the nuances of 'set-cookie'.

packages/event-handler/src/http/converters.ts
```diff
@@ -228,13 +228,30 @@ const webHeadersToApiGatewayV1Headers = (webHeaders: Headers) => {
...
+    headers['set-cookie'] = allCookies[0];
+  }
+
   for (const [key, value] of webHeaders.entries()) {
+    // Skip set-cookie as it's already handled above
+    if (key.toLowerCase() === 'set-cookie') {
+      continue;
+    }
+
     const values = value.split(/[;,]/).map((v) => v.trimStart());
```
nateiler: Ultimately, splitting on a semicolon seems out of place and is the root of this entire issue.  If a goal wasn't to keep backwards compatibility I would suggest removing it and using a comma only.  Headers by default a separated by a comma.

```typescript
const headers = new Headers({
  'accept': 'application/json'
})

headers.append('accept', 'text/html');

const acceptHeaders = headers.get('accept'); // Results in "application/json, text/html"
```

packages/event-handler/src/http/converters.ts
nateiler: To my knowledge `set-cookie` is the only header that can have multiple keys.  Other headers, when appended to, will append the value to the end via a comma.  This logic would never be evaluated (which was verified through the code coverage)

packages/event-handler/tests/unit/http/Router/basic-routing.test.ts
nateiler: Ah, yeah.  Looks like my own biome got the best of me.

nateiler: Actually, it looks like these changes were automatically made via `lint-staged` commit hook.  After reverting them and attempting to commit, my revert gets stashed and I'm left with an empty commit which is prevented.

packages/event-handler/tests/unit/http/Router/basic-routing.test.ts
nateiler: I added them to illustrate a 'simple' multi-value header pair along with the set-cookie multi-value header which contains a semicolon.

packages/event-handler/src/http/converters.ts
svozza: So the reason I wrote it this way was because I was under the impression that cookies were an exception in that they could be separated by a semicolon. I got this from the [RFC 6265](https://www.rfc-editor.org/rfc/rfc6265#section-4.2.1):

> The user agent sends stored cookies to the origin server in the
   Cookie header.  If the server conforms to the requirements in
   [Section 4.1](https://www.rfc-editor.org/rfc/rfc6265#section-4.1) (and the user agent conforms to the requirements in
   [Section 5](https://www.rfc-editor.org/rfc/rfc6265#section-5)), the user agent will send a Cookie header that conforms to
   the following grammar:
>
>   cookie-header = "Cookie:" OWS cookie-string OWS
>   cookie-string = cookie-pair *( ";" SP cookie-pair )

It looks like I made a mistake and thought this referred to the `Set-Cookie` header rather than the `Cookie` header as indicated in the quote above.

svozza: What this means though is that to handle the `Cookie` header case correctly we do need to have special handling for the semicolon. Btw, regarding this:

> If a goal wasn't to keep backwards compatibility 

We don't need to preserve backward compatibility for what was a bug. If we change this so only `Cookie` gets this handling and that's fine.

nateiler: Thanks for the references to the RFC.  In this particular circumstance the [webHeadersToApiGatewayV1Headers](https://github.com/aws-powertools/powertools-lambda-typescript/blob/664901232ecd35bedeab96feb0f917acacf77c0d/packages/event-handler/src/http/converters.ts#L227) function is converting from server to user agent (not also user agent to server) so we should be able to omit special handling of the `Cookie` header as per [RFC6265 4.2](https://www.rfc-editor.org/rfc/rfc6265#section-4.2) the `Cookie` header is a user agent to server header.

It seems like the semicolon was included to handle the special usage of the 'Cookie' header which we don't actually need here.

> We don't need to preserve backward compatibility for what was a bug. If we change this so only Cookie gets this handling and that's fine.

I see your point here; however I think only `set-cookie` with a semicolon is a bug.  Since the current implementation accepts _any_ header with a semicolon separated value (while it may not be popular) removing it might break current implementations.

I'm unsure of the best path here.  Ultimately it seems like dropping the semicolon would simplify and unify the usage across the board (all headers are treated the same).

svozza: The issue here is larger though, splitting on `;` causes worse problems. The accepted way to inidicate parameters in a header is with a semi colon, e.g., `Accept: text/html; q=0.5, text/plain; q=0.2`. With the current implementation, this ends up with a multivalue header value which is clearly wrong:

```js
// ...
multiValueHeaders: {
  accept: [
    "text/html",
    "q=0.5",
    "text/plain",
    "q=0.2",
  ]
}
// ...
```

It's far more important that we handle this case correctly than headers that aren't `Cookie` erroneously delimited by `;`.

nateiler: Yep, removing the semicolon support is the way to go.

svozza (approved):

svozza:
Thank you @nateiler for your work on this. Approved.

PR #5024, opened 2026-02-16
fix(event-handler): default error handler returns a web Response correctly

### LINKED ISSUE #5013: Bug: Event Handler default error handling does not handle a Response gracefully
### Expected Behavior

Looking at [HttpError.toJSON()](https://github.com/aws-powertools/powertools-lambda-typescript/blob/78091aaa6c27447d6404cafa3e3bf410b8d73ec9/packages/event-handler/src/http/errors.ts#L38) it can technically return a [Response](https://github.com/aws-powertools/powertools-lambda-typescript/blob/78091aaa6c27447d6404cafa3e3bf410b8d73ec9/packages/event-handler/src/types/http.ts#L88) object.  In which case, I would expect that response object to be returned (which may contain header values, etc.).

I 'm aware a custom error handler could be registered, however I feel if we're working with the first party `HttpError` a custom handler _shouldn't_ be required. 

### Current Behavior

The [default error handler](https://github.com/aws-powertools/powertools-lambda-typescript/blob/78091aaa6c27447d6404cafa3e3bf410b8d73ec9/packages/event-handler/src/http/Router.ts#L492) creates and return a new `Response` .

### Code snippet

Extend the `HttpError` and to include some `Header` values and return a `Response` object.

```typescript
import {
	HttpError,
	HttpStatusCodes,
} from "@aws-lambda-powertools/event-handler/http";
import type { HandlerResponse } from "@aws-lambda-powertools/event-handler/types";

export class AuthenticationRequiredError extends HttpError {
	readonly statusCode: 307 = HttpStatusCodes.TEMPORARY_REDIRECT;
	readonly errorType = "login_required";

	readonly headers: Headers;

	constructor(
		redirectTo: URL,
		message?: string,
		options?: ErrorOptions,
		details?: Record<string, unknown>,
	) {
		super(message || "Failed to grant access", options, details);

		this.headers = new Headers([["Location", redirectTo.toString()]]);
	}

	toJSON(): HandlerResponse {
		return Response.json(
			{
				statusCode: this.statusCode,
				error: this.errorType,
				message: this.message,
				...(this.details && {
					details: this.details,
				}),
			},
			{
				status: this.statusCode,
				headers: this.headers,
			},
		);
	}
}
```

### Steps to Reproduce

Simply throw this extended `AuthenticationRequiredError` within the route handler.

### Possible Solution

The fact that [HttpError.toJson()](https://github.com/aws-powertools/powertools-lambda-typescript/blob/78091aaa6c27447d6404cafa3e3bf410b8d73ec9/packages/event-handler/src/http/errors.ts#L38) can return a Response object seems a little strange.  However, a simple instance check here could determine if we already have a `Response` or need to create a new one.

latest

24.x

### Packaging format used

npm

### Execution logs

```Shell

```

nateiler:
## Summary
The default router error handling was not handling a `Response` object correctly when returned from the `toJSON` method.

### Changes
The base `HttpError` now has a correct `toJSON` response type and an added `toWebResponse` method which is responsible for constructing a web `Response` for the default error handler.  Errors that extend the `HttpError` can easily modify the `Response` (ex: adding additional headers, etc).

**Issue number:** closes #5013

---

### DIFF
--- a/packages/kafka/src/consumer.ts
+++ b/packages/kafka/src/consumer.ts
@@ -196,7 +196,11 @@ const deserializeRecord = async (
     },
     originalKey: key,
     get value() {
+      if (value === undefined) {
+        return undefined;
+      }
       if (isNull(value)) return null;
+
       const deserializedValue = deserialize({
         value: value,
         deserializer: deserializerValue,
--- a/packages/kafka/src/types/types.ts
+++ b/packages/kafka/src/types/types.ts
@@ -201,7 +201,7 @@ type Record = {
   /**
    * Base64-encoded value of the record
    */
-  value: string | null;
+  value?: string | null;
   /**
    * Array of record headers
    */
--- a/packages/kafka/tests/unit/consumer.test.ts
+++ b/packages/kafka/tests/unit/consumer.test.ts
@@ -609,4 +609,32 @@ describe('Kafka consumer', () => {
     // Assess
     expect(result).toBeNull();
   });
+
+  it('handles tombstone events without a message value', async () => {
+    // Prepare
+    const event = structuredClone(jsonTestEvent);
+    delete event.records['mytopic-0'][0].value;
+
+    const handler = kafkaConsumer<string, unknown>(
+      async (event) => {
+        await setTimeout(1); // simulate some processing time
+        const firstRecord = event.records[0];
+        if (firstRecord) {
+          return firstRecord.value;
+        }
+        return undefined;
+      },
+      {
+        value: {
+          type: SchemaType.JSON,
+        },
+      }
+    );
+
+    // Act
+    const result = await handler(event, context);
+
+    // Assess
+    expect(result).toBeUndefined();
+  });
 });


packages/event-handler/src/http/errors.ts
```diff
@@ -45,6 +45,13 @@ abstract class HttpError extends Error {
       }),
     };
   }
+
+  toWebResponse(): Response {
+    return new Response(JSON.stringify(this.toJSON()), {
```
svozza: Could we use the `Response.json` static method here??

packages/event-handler/tests/unit/http/errors.test.ts
```diff
@@ -174,6 +178,111 @@ describe('HTTP Error Classes', () => {
...
+      statusCode,
+      message,
+    }) => {
+      const error = new ErrorClass(message);
+      const response = error.toWebResponse();
+
+      expect(response.status).toEqual(statusCode);
+      expect(response.headers.get('Content-Type')).toEqual('application/json');
+
+      expect(await response.json()).toEqual({
```
svozza: Let's use the `resolves` assertions for these async verifications: https://vitest.dev/api/expect.html#resolves.

packages/event-handler/src/http/errors.ts
nateiler: Good catch!  Fixed

packages/event-handler/tests/unit/http/errors.test.ts
nateiler: Learn something new everyday!  Fixed

svozza:
Nice work @nateiler, merging now!

PR #5052, opened 2026-02-28
fix(event-handler): add overloads to route() for typed validation context

### LINKED ISSUE #5051: Bug: Type error when using request context in event handler route method
### Expected Behavior

When you define a route using the `app.route` with a handler that uses the  `reqCtx` object it should be of type of `RequestContext` or `TypedRequestContext`.

### Current Behavior

Currently when you try to access the context object you get a typescript error: `Parameter 'reqCtx' implicitly has an 'any' type`.

### Code snippet

```ts
const app = new Router();
app.route(({req}) => { // <- type error
      return req;
    }, {
      path: '/test',
      method: [HttpVerbs.GET, HttpVerbs.POST],
    });
```

### Steps to Reproduce

Create a simple handler like the code example above that interacts with the request context object in any way and you will see the error.

### Possible Solution

The issue is that we are missing a method overload on the `route` method that the other HTTP handlers have that tell the compiler that `reqCtx` is of type `RequestContext` when there is no validation options object and `TypedRequestContext` when it is present.

latest

24.x

### Packaging format used

npm

### Execution logs

```Shell

```

svozza:
## Summary

`Router.route()` did not correctly infer handler parameter types — when destructuring (e.g., `{req}`), TypeScript inferred `any` because the handler was typed as a union of `RouteHandler | TypedRouteHandler`. This adds overloads so the handler type is determined by whether `validation` is present in the options.

### Changes

- Added two overloads to `route()`: without `validation` the handler receives `RequestContext`; with `validation` it receives `TypedRequestContext` (with the `valid` property)
- Extracted the implementation into a private `#registerRoute` method so internal callers bypass the public overloads
- Added unit tests for `route()` with validation and type-level tests

**Issue number:** closes #5051

---


packages/event-handler/src/http/Router.ts
```diff
@@ -453,7 +453,26 @@ class Router {
     }
   }
 
-  public route<
+  public route(
+    handler: RouteHandler,
+    options: Omit<HttpRouteOptions, 'validation'> & { validation?: never }
```
svozza, line 458: We need to do this `'validation'> & { validation?: never }` stuff because both both overloads have the same number of arguments. With the other HTTP methods we could figure it out based on the arity of the method.

PR #5081, opened 2026-03-10
feat(event-handler): add type-safe Store API for request and shared state 

### LINKED ISSUE #5072: Feature request: Type-safe Store API for sharing state across the request lifecycle
### Use case

There is currently no official way to share state across the request lifecycle. Middleware, route handlers, and after-middleware all receive the same `reqCtx` object, but there is no typed mechanism for one to pass data to another. For example, an authentication middleware that extracts a user ID from a JWT has no clean way to make that available to the route handler, and a route handler has no way to surface data to after-middleware further down the chain.

Today, users must resort to workarounds like module-scoped variables, which are not request-scoped and break under concurrency, or mutating the `reqCtx` object with arbitrary properties, which provides no type safety.

Similarly, there is no mechanism for sharing data that persists across the lifetime of a Router instance (e.g., a database connection pool or application config initialized at cold start). The existing `context` property on the Router is untyped and not exposed to handlers.

This was originally discussed in #4481 and is related to #4602 where a user specifically requested the ability to `context.set('username', username)` from middleware.

### Solution/User Experience

Introduce a type-safe `Store` API with two scopes:

1. **Request store** - created fresh per request, flows through the middleware -> handler -> after-middleware lifecycle, then discarded
2. **Shared store** - lives on the Router instance, persists across all requests, merged during `includeRouter`

Both stores are configured via a single generic parameter on the `Router` class. The generic shape uses a `store` wrapper to namespace store types and leave room for future type-level configuration.

#### Defining store types

```typescript

type AppEnv = {
  store: {
    request: { userId: string; isAuthenticated: boolean };
    shared: { db: Database; config: AppConfig };
  };
};

const app = new Router<AppEnv>();
```

Both `request` and `shared` are optional - `new Router()` continues to work with no generics.

#### Request store - middleware to handler data flow

The request store is accessed via convenience methods (`set`, `get`, `has`, `delete`) directly on `reqCtx`. This discourages destructuring `reqCtx` (which has known gotchas with `reqCtx.res`).

```typescript
const app = new Router<AppEnv>();

// Global middleware - types flow automatically from Router generic
app.use(async ({ reqCtx, next }) => {
  const db = reqCtx.shared.get('db');
  const user = await db.authenticate(reqCtx.req);

  reqCtx.set('userId', user.id);              // typed from store.request
  reqCtx.set('isAuthenticated', true);         // typed from store.request
  reqCtx.set('randomKey', 'value');            // TypeScript error

  await next();
});

// Route handler - typed automatically
app.get('/me', async (reqCtx) => {
  const userId = reqCtx.get('userId');         // typed as string | undefined
  const isAuth = reqCtx.get('isAuthenticated'); // typed as boolean | undefined
  return { userId };
});

};
```

#### Shared store - data that persists across requests

The shared store is exposed on `reqCtx.shared` as a `Store` instance. It is mutable and persists for the lifetime of the Router.

```typescript
const app = new Router<AppEnv>();

// Set shared data at cold start
app.shared.set('db', new Database(connectionString));
app.shared.set('config', loadConfig());

// Access in middleware
app.use(async ({ reqCtx, next }) => {
  const db = reqCtx.shared.get('db');         // typed as Database
  const config = reqCtx.shared.get('config'); // typed as AppConfig
  await next();
});
```

#### Standalone middleware with explicit types

When defining middleware in a separate file with no reference to the Router instance, the store type must be provided explicitly:

```typescript
// middleware/auth.ts
import type { Middleware } from '@aws-lambda-powertools/event-handler/http';
import type { AppEnv } from '../types.js';

export const authMiddleware: Middleware<AppEnv> = async ({ reqCtx, next }) => {
  reqCtx.set('userId', '123');                // typed
  reqCtx.set('unknownKey', 'value');          // TypeScript error
  await next();
};
```

#### Split routers with includeRouter

When merging routers, both request and shared store types are combined via intersection. If two routers define the same key with incompatible types, TypeScript will flag this at the `includeRouter` call site.

```typescript
type AuthEnv = {
  store: {
    request: { userId: string };
    shared: { db: Database };
  };
};

type MetricsEnv = {
  store: {
    request: { startTime: number };
    shared: { metrics: MetricsClient };
  };
};

const authRouter = new Router<AuthEnv>();
const metricsRouter = new Router<MetricsEnv>();

// Parent router sees the union of both store types
const app = new Router();
app.includeRouter(authRouter);
app.includeRouter(metricsRouter);
// app's store type: { userId: string; startTime: number } for request,
//                   { db: Database; metrics: MetricsClient } for shared
```

#### Store class

A generic `Store` class backed by a `Map`. This class lives at the top level of the package (not http-specific) so it can be reused by other event handler implementations (AppSync Events, AppSync GraphQL, Bedrock Agent).

```typescript
class Store<T extends Record<string, unknown> = Record<string, unknown>> {
  #data = new Map<keyof T, T[keyof T]>();

  set<K extends keyof T>(key: K, value: T[K]): void {
    this.#data.set(key, value);
  }

  get<K extends keyof T>(key: K): T[K] | undefined {
    return this.#data.get(key) as T[K] | undefined;
  }

  has<K extends keyof T>(key: K): boolean {
    return this.#data.has(key);
  }

  delete<K extends keyof T>(key: K): boolean {
    return this.#data.delete(key);
  }
}
```

#### RequestContext changes

`RequestContext` remains a plain object (not a class) to avoid breaking changes with spreading, `Object.keys()`, or `JSON.stringify()`. The request store convenience methods are added via closure when the `RequestContext` is created in `#resolve()`:

```typescript
const store = new Store<TRequest>();
const requestContext: RequestContext = {
  event,
  context,
  req,
  res: new Response('', { ... }),
  params: {},
  responseType,
  shared: this.sharedStore,
  set: (key, value) => store.set(key, value),
  get: (key) => store.get(key),
  has: (key) => store.has(key),
  delete: (key) => store.delete(key),
};
```

The request store itself is not exposed as a property - users interact with it exclusively through the convenience methods on `reqCtx`.

#### Type propagation

The `TEnv` generic flows from `Router<TEnv>` down to `Middleware<TEnv>`, `RouteHandler<TEnv>`, and `RequestContext<TEnv>`, so inline handlers and middleware registered via `app.use()`, `app.get()`, etc. are typed automatically without additional annotations.

### Alternative solutions

**Expose a `Map` directly on `RequestContext`**: Rather than introducing a `Store` abstraction, we could add a plain `Map<string, unknown>` to `reqCtx` that users mutate directly. This is simpler but sacrifices type safety entirely - users would need to cast values on every `get()` call and there would be no compile-time checks on keys or values. The `Store` class adds minimal overhead while providing full type inference.

**Make `RequestContext` a class**: This would allow adding `set`/`get`/`has`/`delete` as class methods. However, it introduces potential breaking changes with code that spreads `reqCtx`, uses `Object.keys()`, or serializes it. The closure approach achieves the same ergonomics without these risks.

### Open questions

- Exact file location for the `Store` class (top-level `src/store/` vs `src/common/`)
- How the `TEnv` generic propagates through the existing overloads on Router HTTP method helpers (`get`, `post`, etc.) and `route()`
- Whether `includeRouter` should enforce type compatibility at the method signature level or just rely on intersection producing `never` for conflicts

### Acknowledgment

- [] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with :+1: and your use case to help us understand customer demand.

svozza:
## Summary

Add a Store API to the HTTP event handler that provides type-safe, scoped state management for route handlers. This gives users two storage scopes:

- **Request store** — per-invocation state (set in middleware, read in handlers). Cleared automatically between requests.
- **Shared store** — router-scoped state (set at cold start, read everywhere). Persists across invocations.

### Changes

- Add `Store` class backed by a `Map` with generic key/value typing via an `Env` type parameter
- Make `Router` generic over `TEnv extends Env` so that `reqCtx.get()`, `reqCtx.set()`, and `reqCtx.shared.get()` are fully typed
- Add `includeRouter` overload that returns `Router<MergeEnv<[TEnv, TOther]>>`, enabling a fluent chaining style that accumulates store types across sub-routers
- Export `MergeEnv` utility type for users who prefer declaring the full environment upfront
- Add `store` subpath export (`@aws-lambda-powertools/event-handler/store`)
- Unit tests, type-level tests, and e2e tests for both store scopes
- Documentation with examples for untyped usage, typed usage, and split-router patterns

**Issue number:** closes #5072

---


packages/event-handler/src/http/Router.ts
```diff
@@ -78,13 +84,18 @@ import {
   resolvePrefixedPath,
 } from './utils.js';
 
-class Router {
+class Router<TEnv extends Env = Env> {
   protected context: Record<string, unknown>;
```
svozza, line 91: Technically would be breaking change but I think we should remove this as it's just an untyped version of what the shared store.

packages/event-handler/tests/e2e/routers/nestedRouter.ts
```diff
@@ -2,9 +2,9 @@ import { Router } from '@aws-lambda-powertools/event-handler/http';
 
 const nestedRouter = new Router();
 
-nestedRouter.get('/info', ({ event }) => ({
+nestedRouter.get('/info', ({ req }) => ({
   nested: true,
-  path: event.path,
+  path: new URL(req.url).pathname,
```
svozza, line 7: This is not related to the changes but was a tpye error because `path` does not exist on all proxy event types, specifically in v2 it's `rawPath`.

docs/features/event-handler/http.md
```diff
@@ -922,6 +922,94 @@ When necessary, you can set a prefix when including a `Router` instance. This me
...
+    2. Access shared values inside route handlers via `reqCtx.shared`.
+
+#### Typed stores
+
+By default, both stores accept any string key and return `unknown` values. When you know the shape of your data ahead of time, you can define an `Env` type to get full type safety on store keys and values.
+
+=== "index.ts"
+
+    ```ts hl_lines="4-9 14 20-21 27-28 31"
+    --8<-- "examples/snippets/event-handler/http/advanced_store_typed.ts:5"
```
sdangol:
```suggestion
    --8<-- "examples/snippets/event-handler/http/advanced_store_typed.ts:8"
```

sdangol: The highlighting seems to be off due to the starting line:

Image

packages/event-handler/src/http/Router.ts
sdangol: Yeah, I agree that this now would be redundant. Should we mark it as deprecated?

packages/event-handler/tests/types/http.test-d.ts
```diff
@@ -0,0 +1,250 @@
...
+  it('accepts a sub-router whose Env is a subset of the parent', () => {
+    type ParentEnv = {
+      store: {
+        request: { userId: string; featureFlags: string[] };
+        shared: { db: string; cache: Map<string, unknown> };
+      };
+    };
+    const app = new Router<ParentEnv>();
+    const authRouter = new Router<AuthEnv>();
+    app.includeRouter(authRouter);
```
sdangol, line 241: Missing expectations on the test

packages/event-handler/tests/types/http.test-d.ts
```diff
@@ -0,0 +1,250 @@
...
+  });
+
+  it('accepts an untyped Router (backward compat)', () => {
+    const app = new Router();
+    const legacyRouter = new Router();
+    app.includeRouter(legacyRouter);
```
sdangol, line 254: Missing expectations on the test

packages/event-handler/tests/types/http.test-d.ts
sdangol, line 1: Missing the test comment convention

packages/event-handler/tests/types/store.test-d.ts
sdangol, line 1: Missing test comment convention

packages/event-handler/tests/unit/Store.test.ts
sdangol, line 1: Missing test comment convention

packages/event-handler/src/http/Router.ts
svozza: Done

packages/event-handler/tests/types/http.test-d.ts
svozza: Fixed all the comments around these type tests.

PR #5086, opened 2026-03-12
feat(event-handler): add metrics middleware for HTTP routes

### LINKED ISSUE #4602: Feature request: Event Handler Request Logging
### Use case

I'm attempting to emit a request log (also known as canonical log) per request. Using global middleware I've got something working but I have a few bits of feedback.

Per request, I'd like to be able to emit key metadata like source ip, user-agent, requestId, http method, route, path, status code, error message, error type, error stack and useful information I pick out of the request.

I've hacked something together that kind of works, but it's not quite 100%.  The code below has 3 routes, `/ping` returns a 200. `/error` throws a 400 BadRequestError and `/fault` throws an uncaught Error resulting in a 500 Internal Server Error.

```javascript
import { BadRequestError, Router } from '@aws-lambda-powertools/event-handler/experimental-rest';
import {
  correlationPaths,
  search,
} from '@aws-lambda-powertools/logger/correlationId';

let startTime: number;

const logger = new Logger({
  correlationIdSearchFn: search,
});
const metrics = new Metrics({
  namespace: 'testing',
  serviceName: 'powertools-event-handler',
});

const app = new Router({ logger });
app.use(async ({ reqCtx, next }) => {
  try {
    metrics.addMetadata('route', reqCtx.route); //note route doesn't exist on context
    metrics.addMetadata('path', reqCtx.event.path);
    metrics.addMetadata('method', reqCtx.req.method);
    metrics.addMetadata('ip', reqCtx.event.headers['X-Forwarded-For'].split(',')[0]);
    metrics.addMetadata('userAgent', reqCtx.event.headers['User-Agent']);
    metrics.addMetadata('apiGRequestId', reqCtx.event.requestId);
    metrics.addMetadata('apiGExtendedRequestId', reqCtx.event.extendedRequestId);
    metrics.addMetadata('cfRequestId', reqCtx.event.headers['X-Amz-Cf-Id']);
    metrics.addMetadata('requestId', reqCtx.context.awsRequestId);
    metrics.addMetadata('functionVersion', reqCtx.context.functionVersion);
    metrics.addMetadata('traceId', reqCtx.event.headers['X-Amzn-Trace-Id']?.replace('Root=', ''));
  }catch(error){
    logger.error('Error adding metadata', { error });
  }
  try {
    await next();
  } catch(error) {
    if(error){
      metrics.addMetadata('errorDetails', error.message);
      metrics.addMetadata('errorType', error.name);
      metrics.addMetadata('errorStack', error.cause ? error.cause.stack : error.stack);
      if (reqCtx.res.status >= 500) {
        throw error;
      }
    }
  }
  finally {
    metrics.addMetadata('responseCode', reqCtx.res.status + "");
    metrics.addMetric('fault', MetricUnit.Count, reqCtx.res.status >= 500 ? 1 : 0);
    metrics.addMetric('error', MetricUnit.Count, (reqCtx.res.status >= 400 && reqCtx.res.status < 500) ? 1 : 0);
    metrics.addMetric('latency', MetricUnit.Milliseconds, Date.now() - startTime);
  }
});

app.get('/ping', async () => {
  return { message: 'pong' };
});

app.get('/error', async () => {
  throw new BadRequestError("This is an error");
});

app.get('/fault', async () => {
  throw new Error('This is a fault');
});

  startTime = Date.now();
  logger.addContext(context);
  logger.setCorrelationId(event, correlationPaths.API_GATEWAY_REST);
};
```

Below are some sample log outputs for the exposed routes:

`/ping`
```json
{
    "_aws": {
        "Timestamp": 1759372476245,
        "CloudWatchMetrics": [
            {
                "Namespace": "testing",
                "Dimensions": [
                    [
                        "service"
                    ]
                ],
                "Metrics": [
                    {
                        "Name": "fault",
                        "Unit": "Count"
                    },
                    {
                        "Name": "error",
                        "Unit": "Count"
                    },
                    {
                        "Name": "latency",
                        "Unit": "Milliseconds"
                    }
                ]
            }
        ]
    },
    "service": "powertools-event-handler",
    "fault": 0,
    "error": 0,
    "latency": 241,
    "path": "/ping",
    "method": "GET",
    "ip": "71.212.29.6",
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:143.0) Gecko/20100101 Firefox/143.0",
    "apiGRequestId": "8300e8f6-45ca-467c-92ee-fac812d54f1d",
    "apiGExtendedRequestId": "RzCtYHR8iYcEu2A=",
    "cfRequestId": "viQrbfwMYaFhUEEkozh1ykmtBbixdN0JzRBMx-oNMpAH5sLzcszLaA==",
    "requestId": "cf60e110-fe37-4a6b-8ab4-ac251c3f7271",
    "functionVersion": "39",
    "traceId": "1-68dde4bb-44b8df72029d7fff1208aba3",
    "responseCode": "200"
}
```

`/error`
```json
{
    "_aws": {
        "Timestamp": 1759372479479,
        "CloudWatchMetrics": [
            {
                "Namespace": "testing",
                "Dimensions": [
                    [
                        "service"
                    ]
                ],
                "Metrics": [
                    {
                        "Name": "fault",
                        "Unit": "Count"
                    },
                    {
                        "Name": "error",
                        "Unit": "Count"
                    },
                    {
                        "Name": "latency",
                        "Unit": "Milliseconds"
                    }
                ]
            }
        ]
    },
    "service": "powertools-event-handler",
    "fault": 1,
    "error": 0,
    "latency": 2,
    "path": "/error",
    "method": "GET",
    "ip": "71.212.29.6",
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:143.0) Gecko/20100101 Firefox/143.0",
    "apiGRequestId": "5385f574-dc12-476a-a3b2-616f39e83b41",
    "apiGExtendedRequestId": "RzCt-HVCCYcEv5Q=",
    "cfRequestId": "XLdWKQSJAsw1-6E4_qKfeghpxlixMVL6MNcUz_h2TnNnBlTE8w2dYg==",
    "requestId": "894ee350-9444-4b49-bf2c-9f095ee8b875",
    "functionVersion": "39",
    "traceId": "1-68dde4bf-4bc7c50f7cd809de016d0508",
    "errorDetails": "This is an error",
    "errorType": "BadRequestError",
    "errorStack": "BadRequestError: This is an error\n    at /var/task/index.js:6903:9\n    at handlerMiddleware (/var/task/index.js:1683:39)\n    at dispatch (/var/task/index.js:1019:38)\n    at nextFn (/var/task/index.js:1014:23)\n    at /var/task/index.js:6881:11\n    at dispatch (/var/task/index.js:1019:38)\n    at /var/task/index.js:1030:11\n    at Router.resolve (/var/task/index.js:1694:38)\n    at Runtime.handler (/var/task/index.js:6912:14)\n    at Runtime.handleOnceNonStreaming (file:///var/runtime/index.mjs:1306:29)",
    "responseCode": "500"
}
```

Note this has the responseCode set to 500 even though the request returns a 400 so the fault and error counters are wrong.

`/fault`
```json
{
    "_aws": {
        "Timestamp": 1759372482624,
        "CloudWatchMetrics": [
            {
                "Namespace": "testing",
                "Dimensions": [
                    [
                        "service"
                    ]
                ],
                "Metrics": [
                    {
                        "Name": "fault",
                        "Unit": "Count"
                    },
                    {
                        "Name": "error",
                        "Unit": "Count"
                    },
                    {
                        "Name": "latency",
                        "Unit": "Milliseconds"
                    }
                ]
            }
        ]
    },
    "service": "powertools-event-handler",
    "fault": 1,
    "error": 0,
    "latency": 5,
    "path": "/fault",
    "method": "GET",
    "ip": "71.212.29.6",
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:143.0) Gecko/20100101 Firefox/143.0",
    "apiGRequestId": "6b762937-0ca9-4406-b2fe-2f93c856cd2d",
    "apiGExtendedRequestId": "RzCueEv6iYcEFlg=",
    "cfRequestId": "p-ODGZvBOTfvfYIfjXcOvwbu6X7FWgpG0viM8rjjEo4wcyoCs9kvZA==",
    "requestId": "1af4b006-09b7-4770-823a-15ddab8af6e3",
    "functionVersion": "39",
    "traceId": "1-68dde4c2-7a71c6d017b69c45550b64b7",
    "errorDetails": "This is a fault",
    "errorType": "Error",
    "errorStack": "Error: This is a fault\n    at /var/task/index.js:6906:9\n    at handlerMiddleware (/var/task/index.js:1683:39)\n    at dispatch (/var/task/index.js:1019:38)\n    at nextFn (/var/task/index.js:1014:23)\n    at /var/task/index.js:6881:11\n    at dispatch (/var/task/index.js:1019:38)\n    at /var/task/index.js:1030:11\n    at Router.resolve (/var/task/index.js:1694:38)\n    at Runtime.handler (/var/task/index.js:6912:14)\n    at Runtime.handleOnceNonStreaming (file:///var/runtime/index.mjs:1306:29)",
    "responseCode": "500"
}
```

/notfound
```json
{
    "_aws": {
        "Timestamp": 1759372493850,
        "CloudWatchMetrics": [
            {
                "Namespace": "testing",
                "Dimensions": [
                    [
                        "service"
                    ]
                ],
                "Metrics": [
                    {
                        "Name": "fault",
                        "Unit": "Count"
                    },
                    {
                        "Name": "error",
                        "Unit": "Count"
                    },
                    {
                        "Name": "latency",
                        "Unit": "Milliseconds"
                    }
                ]
            }
        ]
    },
    "service": "powertools-event-handler",
    "fault": 0,
    "error": 1,
    "latency": 9,
    "path": "/notfound",
    "method": "GET",
    "ip": "71.212.29.6",
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:143.0) Gecko/20100101 Firefox/143.0",
    "apiGRequestId": "a3ccaaef-1673-453e-9886-f414ead740e7",
    "apiGExtendedRequestId": "RzCwOFQ-CYcEpvQ=",
    "cfRequestId": "6Cc8LZyzOOhGVAbndbYXYGAFN8d6LawxHNSw-9IIQaWRlOrhUtDgRQ==",
    "requestId": "287cbfd3-a533-4c5f-b78d-eec34b26d56a",
    "functionVersion": "39",
    "traceId": "1-68dde4cd-656de8ab1e1bdfcf550e9f06",
    "responseCode": "404"
}
```

### Solution/User Experience

After doing this exercise I have 4 requests.

1. Normalize common metadata and expose it in a consistent way regardless of source, i.e. regardless of whether it's APIG or a Function URL, expose things like ip, path and user-agent with context.ip or some other mechanism so the user doesn't need to extract them from headers. [lambda-api](https://github.com/jeremydaly/lambda-api?tab=readme-ov-file#request) has a good example of this
2. Expose the matched route to middleware on the context, I didn't see it surfaced.
3. In global middleware, when throwing pre-defined errors, the status on the response should not be 500, it should be the status code for the actual error.  When the response is returned to the browser it is the correct value, just in the global middleware it is 500 (see /error log above).  Note, this is not the case on /notfound so I'm not sure whether this is user error. I did read the [errorHandling docs](https://docs.aws.amazon.com/powertools/typescript/latest/features/event-handler/rest/#error-handling) but I don't want a separate handler per error type. I just want to throw the predefined error from the route and have the default propogate to the end user.
1. Expose a way to set values on the context from middleware. For example if I had a JWT middleware, I might want to add username to the context.  Both hono and lambda-api allow you to do something like `context.set('username', username)` so you can access it later in the request.

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

svozza:
## Summary

Add metrics middleware for emitting per-request metrics from the Event Handler HTTP router using the Powertools Metrics utility.

### Changes

- Add `route` field to `RequestContext` populated with the matched route pattern (e.g. `GET /users/:id`), or `null` when no route matches
- Create metrics middleware that emits `latency`, `fault`, and `error` metrics per request
- Use the matched route as a metric dimension (uses `NOT_FOUND` when no route matches to prevent dimension explosion)
- Make `RequestContext` a discriminated union on `responseType`, enabling type-safe access to event-specific properties (e.g. `requestContext.requestId` on API Gateway events but not ALB)
- Add request metadata to metrics:
  - `httpMethod` and `path` for all requests
  - `statusCode` for all requests
  - `ipAddress` from `requestContext.identity.sourceIp` (V1), `requestContext.http.sourceIp` (V2), or `X-Forwarded-For` header (ALB fallback)
  - `userAgent` from request headers
  - `apiGwRequestId` and `apiGwApiId` for API Gateway V1 and V2 events
  - `apiGwExtendedRequestId` for API Gateway V1 events when available
- Remove unused `getResponseType` function (replaced by type guard branching in `#buildRequestContext`)
- Add `@aws-lambda-powertools/metrics` as an optional peer dependency, following the same pattern as the tracer middleware
- Expose the metrics middleware as a subpath export to avoid requiring the metrics dependency
- Unit tests are covered

**Issue number:** closes #4602

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10327,10 +10327,14 @@
         "@aws-lambda-powertools/testing-utils": "file:../testing"
       },
       "peerDependencies": {
+        "@aws-lambda-powertools/metrics": ">=2.32.0",
         "@aws-lambda-powertools/tracer": ">=2.32.0",
         "@standard-schema/spec": "^1.0.0"
       },
       "peerDependenciesMeta": {
+        "@aws-lambda-powertools/metrics": {
+          "optional": true
+        },
         "@aws-lambda-powertools/tracer": {
           "optional": true
         }


packages/event-handler/src/http/middleware/metrics.ts
```diff
@@ -0,0 +1,100 @@
+import type { Metrics } from '@aws-lambda-powertools/metrics';
+import { MetricUnit } from '@aws-lambda-powertools/metrics';
+import type { Middleware, RequestContext } from '../../types/http.js';
+import { HttpError } from '../errors.js';
+
+const getHeaderMetadata = (req: Request): Record<string, string> => {
+  const metadata: Record<string, string> = {};
+
+  const xForwardedFor = req.headers.get('X-Forwarded-For');
+  if (xForwardedFor) {
```
perpil: Although this usually works, if the client sets the header, it might be [wrong](https://speedrun.nobackspacecrew.com/blog/2026/03/06/obtaining-caller-ip-address-with-lambda-furls.html)

It might be worth asking Lambda if they plan on making any changes for furls, or at least documenting that due to limitations this may not be the true client ip. If my understanding is correct you want the 2nd to last ip address for api gateway or to use the `sourceIp` in the event.

svozza: Nice info. What about using `event.requestContext.identity.sourceIp` (V1) / `event.requestContext.http.sourceIp` (V2) and falling back to this header if they don't exist?

perpil: That makes some sense. It's worth having a discussion with the Lambda team about it if you can. I think `event.requestContext.http.sourceIp` is the cloudfront edge server ip and not the client ip if you front your furl with cloudfront. It's possible that Lambda can fix that to be caller ip as it's less a breaking change than changing what they are doing to the `x-forwarded-for` header.

packages/event-handler/src/http/middleware/metrics.ts
```diff
@@ -0,0 +1,100 @@
...
+    } finally {
+      const metadata = {
+        ...getHeaderMetadata(reqCtx.req),
+        ...getEventMetadata(reqCtx),
+      };
+      for (const [key, value] of Object.entries(metadata)) {
+        metrics.addMetadata(key, value);
+      }
+      metrics
+        .addDimension('route', reqCtx.route)
```
perpil: I think the way you are setting route here:

```javascript
requestContext.route = route?.route ?? `${method} ${path}\`;
```

will result in a lot of metric bloat for bad paths. I think you want to change that to some catch-all route. I don't know what the best value is for that but maybe `/*`.

svozza: If we change this then we won't get metrics for 404 requests, which I'm not sure we want to ignore.

perpil: We need metrics on 404 requests. What I'm saying is they shouldn't be a different dimension per path when there is no route. Path makes sense to have as metadata. If route is set to path when there is no route, it becomes high cardinality when someone starts hitting random urls probing for vulnerabilities. I believe you are charged by dimension and can only have up to 30 per metric.

svozza: Ah I see what you mean, so should we check the status code in the middleware and do something different if it's 404? What would that look like?

perpil: There are a couple ways of solving it, but I don't think you want to use status code to decide what to set the route dimension to. Off the top of my head, you could leave the `reqCtx.route = undefined` i.e. `requestContext.route = route?.route`. If it is undefined in your addDimension, use `/*` or `/:` as the default value. Alternatively you could set some static default value for the route as a catchall: `requestContext.route = route?.route ?? '/*'`.

svozza: I still want `route` in the request context to reflect reality but what I meant about the 404 was that we change the dimension in the middleware if we get a 404:

```ts
const url = new URL(reqCtx.req.url);
const is404 = status === 404;
const metadata = {
  httpMethod: reqCtx.req.method,
  path: url.pathname,
  ...getHeaderMetadata(reqCtx.req),
  ...getEventMetadata(reqCtx),
};
for (const [key, value] of Object.entries(metadata)) {
  metrics.addMetadata(key, value);
}

metrics
  .addDimension('route', is404 ? 'NOT_FOUND' : reqCtx.route)
// ...
```

packages/event-handler/src/http/middleware/metrics.ts
svozza: Yes, of course, in the meantime, I think this will work for the vast majority of cases.

packages/event-handler/src/http/middleware/metrics.ts
perpil: The route not being found and a 404 can be two different things. It's possible the route `/profile/:username` returns a 404 due to business logic (user not found), but that route is valid. I'd recommend you don't overload the meaning of route and set it to method+path if there is no matching route. If route is not found, I'd leave it `undefined` since you can easily concatenate method + path if that is actually what you need. By leaving route `undefined`, it's clear in the middleware there is no route and you can set it to `NOT_FOUND`.

svozza: Ah i see what you mean, that makes sense.

svozza: Done!

packages/event-handler/package.json
```diff
@@ -158,11 +172,15 @@
   },
   "peerDependencies": {
     "@aws-lambda-powertools/tracer": ">=2.32.0",
+    "@aws-lambda-powertools/metrics": ">=2.32.0",
```
sdangol, line 175: @svozza Shouldn't the lock file be updated too?

svozza: Ah I forgot to add it back after my rebase

svozza: Good catch! Added.

PR #5143, opened 2026-03-29
feat(data-masking): add Data Masking utility

### LINKED ISSUE #4960: Feature request: Add DataMasking utility for encrypting and masking sensitive data
### Use case

AWS Lambda functions often process sensitive data (PII, credentials, financial information) that needs to be protected when:

- Storing data in logs, databases, or checkpoints (e.g., AWS Lambda Durable Functions)
- Passing data between services
- Meeting compliance requirements (GDPR, HIPAA, PCI-DSS)

Current challenges:

1. **No built-in masking utilities** - Developers manually implement masking logic, leading to inconsistent approaches
2. **Full payload encryption loses visibility** - Encrypting entire objects makes debugging, querying, and monitoring impossible
3. **Field-level encryption is complex** - Requires significant boilerplate to encrypt specific fields while preserving structure
4. **No integration with AWS Encryption SDK** - Developers must configure envelope encryption, key caching, and KMS integration from scratch

**Real-world scenarios:**

- **Durable Functions checkpointing**: Store workflow state with encrypted PII while keeping workflow metadata visible for queries
- **Multi-tenant applications**: Encrypt customer data with tenant-specific encryption context for isolation
- **Compliance logging**: Mask sensitive fields before logging while preserving log structure for analysis
- **API responses**: Redact sensitive fields before returning data to clients

Powertools for Python has a DataMasking utility that solves these problems. TypeScript needs equivalent functionality.

### Solution/User Experience

Add a new `@aws-lambda-powertools/data-masking` package that provides:

### 1. Core DataMasking Class

```typescript
import { DataMasking } from '@aws-lambda-powertools/data-masking';
import { AWSEncryptionSDKProvider } from '@aws-lambda-powertools/data-masking/providers';

// Initialize with encryption provider
const provider = new AWSEncryptionSDKProvider({ keys: [KMS_KEY_ARN] });
const dataMasker = new DataMasking({ provider });
```

### 2. Three Primary Operations

**Erase (Irreversible Masking):**

```typescript
const masked = dataMasker.erase(data, { 
  fields: ['email', 'address.street', 'customer.ssn'] 
});

// Result:
// {
//   "email": "*****",
//   "address": { "street": "*****", "city": "Anytown" },
//   "customer": { "ssn": "*****" }
// }
```

**Encrypt (Full Payload):**

```typescript
const encrypted = await dataMasker.encrypt(data, {
  context: { tenantId: 'acme-corp', dataType: 'pii' }
});
// Returns: base64 encrypted string
```

**Encrypt (Field-Level with Structure Preservation):**

```typescript
const encrypted = await dataMasker.encrypt(data, {
  fields: ['customer.ssn', 'payment.creditCard'],
  context: { tenantId: 'acme-corp' }
});

// Result:
// {
//   "orderId": "12345",  // Visible for queries
//   "customer": {
//     "name": "John",  // Visible
//     "ssn": { "__encrypted": "customer.ssn" }  // Placeholder
//   },
//   "payment": {
//     "creditCard": { "__encrypted": "payment.creditCard" },
//     "amount": 99.99  // Visible
//   },
//   "__powertools_encrypted_data": "AQICAHh8...",  // Encrypted blob
//   "__powertools_encryption_context": { "tenantId": "acme-corp" }
// }
```

**Decrypt:**

```typescript
const decrypted = await dataMasker.decrypt(encrypted);
// Automatically detects format (full or field-level) and restores original data
```

### 3. Advanced Masking Options

```typescript
const masked = dataMasker.erase(data, {
  maskingRules: {
    email: { 
      regexPattern: /(.)(.*)(@.*)/, 
      maskFormat: '$1****$3'  // j****@example.com
    },
    age: { dynamicMask: true },  // Maintains length: "30" -> "**"
    'address.zip': { customMask: 'XXXXX' }
  }
});
```

### 4. Lambda Handler Integration

```typescript
import { DataMasking } from '@aws-lambda-powertools/data-masking';
import { AWSEncryptionSDKProvider } from '@aws-lambda-powertools/data-masking/providers';

const logger = new Logger();
const provider = new AWSEncryptionSDKProvider({ keys: [process.env.KMS_KEY_ARN!] });
const dataMasker = new DataMasking({ provider });

export const handler = async (event: any) => {
  const orderData = event.body;
  
  // Mask before logging
  logger.info('Processing order', { 
    order: dataMasker.erase(orderData, { fields: ['creditCard'] })
  });
  
  // Encrypt for storage
  const encrypted = await dataMasker.encrypt(orderData, {
    fields: ['customer.ssn', 'payment.creditCard'],
    context: { orderId: orderData.orderId }
  });
  
  await dynamodb.put({ Item: encrypted });
  
  return { statusCode: 200 };
};
```

### 5. Durable Functions Integration (Future)

```typescript
// Custom SerDes for encrypted checkpoints (when SDK supports it)
class EncryptedFieldsSerDes implements SerDes<any> {
  constructor(
    private fields: string[],
    private dataMasker: DataMasking
  ) {}
  
  async serialize(value: any): Promise<string> {
    const encrypted = await this.dataMasker.encrypt(value, { 
      fields: this.fields 
    });
    return JSON.stringify(encrypted);
  }
  
  async deserialize(data: string): Promise<any> {
    const encrypted = JSON.parse(data);
    return this.dataMasker.decrypt(encrypted);
  }
}
```

### 6. Provider System

Support AWS Encryption SDK with KMS:

```typescript
const provider = new AWSEncryptionSDKProvider({
  keys: [KMS_KEY_ARN],
  // Optional: Configure caching
  localCacheCapacity: 100,
  maxCacheAgeSeconds: 300,
  maxMessagesEncrypted: 4294967296,
  maxBytesEncrypted: 9223372036854775807
});
```

### 7. Key Features

- **JSONPath field selection**: `'customer.ssn'`, `'orders[*].payment'`, `'$.items[?(@.price > 100)]'`
- **Encryption context**: Additional authenticated data for KMS operations
- **Automatic format detection**: Decrypt handles both full and field-level formats
- **Batched encryption**: Single KMS call encrypts all specified fields together
- **Type safety**: Full TypeScript types for all operations
- **AWS Encryption SDK integration**: Envelope encryption with data key caching
- **Structure preservation**: Non-sensitive fields remain queryable/debuggable

### 8. API Parity with Python

Match the existing Python DataMasking API (with TypeScript idioms):

- `erase(data, options)` → `data_masker.erase(data, fields=[])`
- `encrypt(data, options)` → `data_masker.encrypt(data, fields=[])`
- `decrypt(data)` → `data_masker.decrypt(data)`
- Encryption providers → Same provider system
- Field selection → Same JSONPath syntax

### Alternative solutions

```Markdown
### 1. Manual Implementation

Developers can manually:

- Traverse objects to mask fields
- Configure AWS Encryption SDK directly
- Implement custom serialization logic

**Drawbacks:**

- Significant boilerplate (100+ lines per use case)
- Inconsistent implementations across teams
- Error-prone field traversal
- No reusable patterns

### 2. Third-party Libraries

Generic encryption libraries exist but:

- Not designed for Lambda/serverless patterns
- No AWS Encryption SDK integration
- No KMS key caching optimization
- Missing field-level encryption with structure preservation

### 3. AWS Services

- **AWS Secrets Manager**: For static secrets, not dynamic data masking
- **Amazon Macie**: For data discovery/classification, not runtime masking
- **KMS Encryption SDK**: Low-level, requires significant integration work

### 4. Redaction-only Solutions

Libraries like `redact-pii` provide masking but:

- No encryption support
- No reversible operations
- Limited field selection
- No AWS integration

**Why a Powertools utility is better:**

- ✅ Integrated with Powertools ecosystem (Logger, Tracer, Metrics)
- ✅ Optimized for Lambda (cold start, memory, caching)
- ✅ Follows Powertools patterns and conventions
- ✅ Tested, documented, and maintained by AWS
- ✅ Type-safe with full TypeScript support
- ✅ Parity with Python implementation
```

### Acknowledgment

### Future readers

svozza:
### Changes

This PR is an experiment in delivering a full feature, end to end, using spec-driven development and agentic coding. As such, I have set it as a draft. We may or may not merge this, but if we do, only after a thorough review by the team. The purpose of this PR is as much to provoke discussion as it is to implement a feature. I would appreciate if @dreamorosi and @sdangol could look at the code and give their opinions.

From my perspective, I think this was a very successful experiment: I am happy with the code quality and I also took the opportunity to add property tests, which are a perfect fit for this sort of logic.

Something I would note is that this probably worked so well because of how well-defined the issue was by @walmsles, and also that we could use the Python implementation as a reference.

One place I differed from the proposed implementation was that I don't batch the calls to KMS. This simplifies the API and means we mirror the Python implementation exactly. While ordinarily this would be a performance concern, we use the caching feature in the AWS Cryptography library to ensure that we only ever make one call to KMS when encrypting multiple fields.

#### What's included

- `@aws-lambda-powertools/data-masking` package with erase, encrypt, and decrypt operations
- `AWSEncryptionSDKProvider` using KMS envelope encryption (`@aws-crypto/client-node` as optional peer dep)
- Field selection via dot notation, `[*]` array wildcards, and `.*` object wildcards
- Custom masking rules (regex, dynamic length, custom strings)
- Encryption context (AAD) for integrity and authenticity
- Prototype pollution protection
- Unit tests with property-based testing via `fast-check`
- E2e test scaffolding (CDK stack + Lambda handlers)
- Documentation mirroring the Python Powertools data masking docs

#### ~~A note on field path resolution~~

~~The Python implementation uses `jsonpath_ng` for field selection, which natively supports both querying and path extraction for write-back. We considered using a JavaScript JSONPath library (e.g. [`jsonpath-plus`](https://www.npmjs.com/package/jsonpath-plus)) to match this approach, but decided against it for a few reasons:~~

~~- `jsonpath-plus` (9.8m weekly downloads) is marked as unmaintained by its maintainers~~
~~- We already have `@aws-lambda-powertools/jmespath` in-house~~

~~Instead, we use JMESPath to validate expressions and a small ((approx. 20 line)) custom walker to resolve wildcards (`[*]` and `*`) into concrete paths for write-back. JMESPath is read-only by design so it can't be used for path extraction directly, but the walker is simple, well-tested, and avoids any new dependencies.~~

~~There is however another library, [jsonpath](https://www.npmjs.com/package/jsonpath), that has 3.8m weekly downloads. I am always hesitant to introduce new dependencies to the project but I think if we want feature parity with Python we will need to take the dependency on. I would like to hear the maintainers thoughts before committing to this course of action though.~~

**Update:** The JMESPath dependency has been removed. It was unnecessary since we only used it for reading, and keeping it could mislead users into thinking we support the full JMESPath spec for writes. Instead, we use a small custom walker that supports:

- dot notation (`obj.key`)
- wildcard for object keys (`obj.*.key`)
- wildcard for arrays (`obj.[*].key`)

These three cases are validated with property tests, which are much more thorough than traditional unit tests.

**Issue number:** closes #4960

---


svozza:
A note here: I wasted a couple of hours getting the end to end tests to work because the linter wouldn't allow us to create an async function without the `await` keyword and the LLM ended up accidentally making the lambda function sync because of this. There are perfectly valid reason to want to use the `async` keyword in a function that doesn't use `await`, e.g., the value we return is a call to an async function: `return someAsyncFunction(value)`. This has caused me issues multiple times in the past and I think we should disable this rule. I want a developer or an LLM to be able to know a function is async and the best way to do that is with the `async` keyword, not that `await` has been used there. In fact, the Biome [docs](https://biomejs.dev/linter/rules/use-await/) already say that this rule is not recommended:

> ### Summary
>
>    - Rule available since: v1.4.0
>    - Diagnostic Category: [lint/suspicious/useAwait](https://biomejs.dev/reference/diagnostics#diagnostic-category)
>    - **This rule isn’t recommended, so you need to enable it.**

I will raise a separate issue and PR to handle this.

svozza:
I have removed the JMESPath dependency here, it's an unnecesaary depencendency because we only use it for reading but could mislead users into thinking we support the whole JMESPath spec for writes. In fact, for this MVP, we only support the following write paths.

- dot notation (`obj.key`)
- wildcard for object keys (`obj.*.key`)
- wildcard for arrays (`obj.[*].key`)

I am very confident in the implementation of these three cases as they are validated with property tests, which are much more thorough than traditional unit tests.

mkdocs.yml
dreamorosi, line 32: Need to add new pkg here

dreamorosi, line 199: Need to add new pkg here

packages/data-masking/package.json
```diff
@@ -0,0 +1,100 @@
...
+  ],
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/aws-powertools/powertools-lambda-typescript.git"
+  },
+  "bugs": {
+    "url": "https://github.com/aws-powertools/powertools-lambda-typescript/issues"
+  },
+  "dependencies": {
+    "@aws-lambda-powertools/commons": "2.32.0"
```
dreamorosi: refresh this

packages/data-masking/package.json
```diff
@@ -0,0 +1,100 @@
+{
+  "name": "@aws-lambda-powertools/data-masking",
+  "version": "2.32.0",
```
dreamorosi: refresh this

packages/data-masking/src/index.ts
```diff
@@ -0,0 +1,15 @@
...
+  DataMaskingEncryptionError,
+} from './errors.js';
+export type {
+  EncryptionProvider,
+  EraseOptions,
+  EncryptOptions,
+  DecryptOptions,
+  MaskingRule,
+  DataMaskingConstructorOptions,
+} from './types.js';
```
dreamorosi, line 16: Sub path export for `/types` where we export only types

packages/data-masking/src/index.ts
```diff
@@ -0,0 +1,15 @@
+export { DataMasking } from './DataMasking.js';
+export {
+  DataMaskingFieldNotFoundError,
+  DataMaskingUnsupportedTypeError,
+  DataMaskingEncryptionError,
+} from './errors.js';
```
dreamorosi, line 7: Sub path export for `/error` where we export only errors

packages/data-masking/src/errors.ts
```diff
@@ -0,0 +1,23 @@
...
+export class DataMaskingFieldNotFoundError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = 'DataMaskingFieldNotFoundError';
+  }
+}
+
+/** Thrown when input data contains an unsupported type for the operation. */
+export class DataMaskingUnsupportedTypeError extends Error {
+  constructor(message: string) {
```
dreamorosi: Is the optional `options?: ErrorOptions` missing intentional?

packages/data-masking/src/errors.ts
```diff
@@ -0,0 +1,23 @@
+/** Thrown when a field path does not match any value in the data. */
+export class DataMaskingFieldNotFoundError extends Error {
+  constructor(message: string) {
```
dreamorosi: Is the optional `options?: ErrorOptions` missing intentional?

packages/data-masking/src/constants.ts
```diff
@@ -0,0 +1 @@
+export const DEFAULT_MASK_VALUE = '*****';
```
dreamorosi, line 4: Let's add a comment to explain what this is

packages/data-masking/src/DataMasking.ts
```diff
@@ -0,0 +1,270 @@
...
+ *
+ * const masker = new DataMasking();
+ * const masked = masker.erase(data, { fields: ['customer.ssn'] });
+ * ```
+ */
+export class DataMasking {
+  readonly #provider?: EncryptionProvider;
+  readonly #throwOnMissingField: boolean;
+
+  constructor(options?: DataMaskingConstructorOptions) {
```
dreamorosi: Need to set `public` on the constructor

packages/data-masking/src/DataMasking.ts
```diff
@@ -0,0 +1,270 @@
...
+ * ```typescript
+ * import { DataMasking } from '@aws-lambda-powertools/data-masking';
+ *
+ * const masker = new DataMasking();
+ * const masked = masker.erase(data, { fields: ['customer.ssn'] });
+ * ```
+ */
+export class DataMasking {
+  readonly #provider?: EncryptionProvider;
+  readonly #throwOnMissingField: boolean;
```
dreamorosi, line 36: Add comments documenting the two props

packages/data-masking/src/constants.ts
dreamorosi, line 4: consider if needs to be `as const`

packages/data-masking/src/DataMasking.ts
```diff
@@ -0,0 +1,270 @@
...
+   *
+   * @example
+   * ```typescript
+   * const masked = masker.erase(data, { fields: ['email', 'customer.ssn'] });
+   * ```
+   */
+  erase<T>(data: T, options?: EraseOptions): T {
+    if (data === null || data === undefined) return data;
+    if (!options?.fields && !options?.maskingRules) {
+      return DEFAULT_MASK_VALUE as unknown as T;
```
dreamorosi: This `as unknown as T` needs looking into it - right now I think it's needed because `DEFAULT_MASK_VALUE` is a string and we're casting it to `T` but the bigger question is: does the return type need to stay `T` or should it be `string`? Can it ever not be a `string`?

dreamorosi: Ok this is a recursive function that can return objects - so this `as unknown as T` is basically "the leaf state" of the recursion which is always a string but we're telling it to be `T`. So because it's a public method we need to keep `T` but it could've been probably `T | string`.

packages/data-masking/src/DataMasking.ts
```diff
@@ -0,0 +1,270 @@
...
+  /**
+   * Irreversibly mask fields in a data object. Returns a deep copy.
+   *
+   * @example
+   * ```typescript
+   * const masked = masker.erase(data, { fields: ['email', 'customer.ssn'] });
+   * ```
+   */
+  erase<T>(data: T, options?: EraseOptions): T {
+    if (data === null || data === undefined) return data;
```
dreamorosi: Might want to use the "isNullorUndefined" (look up exact name) helper from `commons` here

packages/data-masking/src/DataMasking.ts
```diff
@@ -0,0 +1,270 @@
...
+
+  /**
+   * Irreversibly mask fields in a data object. Returns a deep copy.
+   *
+   * @example
+   * ```typescript
+   * const masked = masker.erase(data, { fields: ['email', 'customer.ssn'] });
+   * ```
+   */
+  erase<T>(data: T, options?: EraseOptions): T {
```
dreamorosi: Need to document args & options in the comment above.

packages/data-masking/src/DataMasking.ts
```diff
@@ -0,0 +1,270 @@
...
+    if (data === null || data === undefined) return data;
+    if (!options?.fields && !options?.maskingRules) {
```
dreamorosi: This looks odd - the way I understand it right now is that we replace the entire value/object with `DEFAULT_MASK_VALUE` if I don't specify any option - which feels odd. I don't know for sure, but at the moment I'd expect instead to replace ALL values in the object instead?

packages/data-masking/src/DataMasking.ts
```diff
@@ -0,0 +1,270 @@
...
+      return DEFAULT_MASK_VALUE as unknown as T;
+    }
+
+    const copy = this.#deepCopy(data);
+
+    if (options.maskingRules) {
+      this.#applyMaskingRules(copy, options.maskingRules);
+    } else {
+      /* v8 ignore next -- @preserve fallback unreachable: line 46 returns early when fields is falsy */
```
dreamorosi: Is this really unreachable or is the agent lazy?

dreamorosi: Ok I got it now, maybe we want to consider changing `options.fields` in the signature to not optional and make it default to`[]` so that the below never happens.

packages/data-masking/src/DataMasking.ts
```diff
@@ -0,0 +1,270 @@
...
+   */
+  erase<T>(data: T, options?: EraseOptions): T {
+    if (data === null || data === undefined) return data;
+    if (!options?.fields && !options?.maskingRules) {
+      return DEFAULT_MASK_VALUE as unknown as T;
+    }
+
+    const copy = this.#deepCopy(data);
+
+    if (options.maskingRules) {
```
dreamorosi: Can we consider destructuring `options` at the top of this method.

packages/data-masking/src/DataMasking.ts
```diff
@@ -0,0 +1,270 @@
...
+  let current = data as Record<string, unknown>;
+  for (const key of path) {
+    if (RESERVED_KEYS.has(key)) return undefined;
+    current = current[key] as Record<string, unknown>;
+  }
+
+  return current;
+};
+
+const setAtPath = (data: unknown, path: string[], value: unknown): void => {
```
dreamorosi, line 346: This function needs a comment explaining what it does as it's a bit hard to reason about

packages/data-masking/src/DataMasking.ts
```diff
@@ -0,0 +1,270 @@
...
+    for (const field of fields) {
+      const paths = this.#resolveFieldPaths(
+        copy as Record<string, unknown>,
+        field
+      );
+      if (paths.length === 0 && this.#throwOnMissingField) {
+        throw new DataMaskingFieldNotFoundError(`Field not found: '${field}'`);
+      }
+      for (const path of paths) {
+        setAtPath(copy, path, DEFAULT_MASK_VALUE);
```
dreamorosi: Should this always be the default mask value or should the customer be able to override it? Current behavior always sets the default

packages/data-masking/src/DataMasking.ts
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  #applyMaskingRules<T>(copy: T, rules: Record<string, MaskingRule>): void {
+    for (const [field, rule] of Object.entries(rules)) {
+      for (const path of this.#resolveFieldPaths(
+        copy as Record<string, unknown>,
+        field
+      )) {
+        const value = getAtPath(copy, path);
+        if (typeof value !== 'string') {
```
dreamorosi: might want to use `isString` helper from commons

packages/data-masking/src/DataMasking.ts
```diff
@@ -0,0 +1,270 @@
...
+  if (!lastKey || RESERVED_KEYS.has(lastKey)) return;
+  current[lastKey] = value;
+};
+
+const applyMaskingRule = (value: string, rule: MaskingRule): string => {
+  if (rule.regexPattern && rule.maskFormat) {
+    return value.replace(rule.regexPattern, rule.maskFormat);
+  }
+  if (rule.dynamicMask) return '*'.repeat(value.length);
+  if (rule.customMask !== undefined) return rule.customMask;
```
dreamorosi: What happens if I set a custom mask (i.e. `+`) and also want it dynamic at the same time?

packages/data-masking/src/types.ts
```diff
@@ -0,0 +1,56 @@
...
+/** Per-field custom masking configuration for {@link DataMasking.erase}. */
+export interface MaskingRule {
+  /** Regex pattern to match within the field value. */
+  regexPattern?: RegExp;
+  /** Replacement format string (e.g., `'$1****$3'`). */
+  maskFormat?: string;
+  /** If true, mask length matches original value length. */
+  dynamicMask?: boolean;
+  /** Fixed replacement string. */
+  customMask?: string;
```
dreamorosi: These seem to be mutually exclusive in the implementation and meaning - if that's the case the type should reflect it.

packages/data-masking/src/types.ts
```diff
@@ -0,0 +1,56 @@
...
+/** Options for the {@link DataMasking} constructor. */
+export interface DataMaskingConstructorOptions {
+  /** Encryption provider for encrypt/decrypt operations. */
+  provider?: EncryptionProvider;
+  /** Whether to throw when a field path doesn't match. Default: `true`. */
+  throwOnMissingField?: boolean;
+}
+
+/** Per-field custom masking configuration for {@link DataMasking.erase}. */
+export interface MaskingRule {
```
dreamorosi: Review all the options to make sure that the mutual exclusion is reflecting the implementation

packages/data-masking/src/DataMasking.ts
```diff
@@ -0,0 +1,270 @@
...
+   *   fields: ['customer.ssn'],
+   *   context: { tenantId: 'acme' },
+   * });
+   * ```
+   */
+  async encrypt<T>(data: T, options?: EncryptOptions): Promise<T | string> {
+    const provider = this.#requireProvider();
+
+    if (!options?.fields) {
+      return provider.encrypt(JSON.stringify(data), options?.context);
```
dreamorosi: Maybe this could be its own named function in the method

```
const encrypt = provider.encrypt(JSON.stringify(data), options?.context);
```

packages/data-masking/src/DataMasking.ts
```diff
@@ -0,0 +1,270 @@
...
+      )) {
+        operations.push(
+          transform(getAtPath(data, path)).then((result) =>
+            setAtPath(data, path, result)
+          )
+        );
+      }
+    }
+
+    await Promise.all(operations);
```
dreamorosi, line 276: This might need to be `allSettled` - check what happens if one of them fails

dreamorosi: Maybe this is fine because as you said it's an all or nothing op - I'm just curious to see how the error is bubbled up.

packages/data-masking/src/DataMasking.ts
```diff
@@ -0,0 +1,270 @@
...
+    const provider = this.#requireProvider();
+
+    if (typeof data === 'string') {
+      return JSON.parse(await provider.decrypt(data, options?.context)) as T;
+    }
+
+    const copy = this.#deepCopy(data);
+    if (options?.fields) {
+      await this.#transformFields(copy, options.fields, async (value) => {
+        if (typeof value !== 'string') return value;
```
dreamorosi: Should we consider emitting a warning here? It's good that we prevent the customer from shooting themselves in the foot, but they should know so they can fix the field paths or the data.

packages/data-masking/src/provider/AWSEncryptionSDKProvider.ts
```diff
@@ -0,0 +1,108 @@
...
+type EncryptFn = ReturnType<typeof buildEncrypt>['encrypt'];
+type DecryptFn = ReturnType<typeof buildDecrypt>['decrypt'];
+type CacheManager = Parameters<EncryptFn>[0];
+
+interface SDKClients {
+  encryptFn: EncryptFn;
+  decryptFn: DecryptFn;
+  cacheManager: CacheManager;
+}
+
```
dreamorosi: Consider splitting these in their own types file under `provider`

packages/data-masking/src/provider/AWSEncryptionSDKProvider.ts
```diff
@@ -0,0 +1,108 @@
+import type { buildDecrypt, buildEncrypt } from '@aws-crypto/client-node';
+import type { EncryptionProvider } from '../types.js';
+
+export interface AWSEncryptionSDKProviderOptions {
```
dreamorosi: All these need documenting

packages/data-masking/src/provider/AWSEncryptionSDKProvider.ts
```diff
@@ -0,0 +1,108 @@
...
+
+interface SDKClients {
+  encryptFn: EncryptFn;
+  decryptFn: DecryptFn;
+  cacheManager: CacheManager;
+}
+
+export class AWSEncryptionSDKProvider implements EncryptionProvider {
+  readonly #options: AWSEncryptionSDKProviderOptions;
+  #clients?: SDKClients;
```
dreamorosi: Can we call this `client` (singular?) I get that it's an object we compose from functions coming from the encryption sdk but it reads weird and a client class is an object after all

packages/data-masking/src/provider/AWSEncryptionSDKProvider.ts
```diff
@@ -0,0 +1,108 @@
...
+  constructor(options: AWSEncryptionSDKProviderOptions) {
+    this.#options = options;
+  }
+
+  async #init(): Promise<SDKClients> {
+    if (this.#clients) return this.#clients;
+
+    // Dynamic import — @aws-crypto/client-node is an optional peer dep,
+    // so we only load it when encryption is actually used.
+    const {
```
dreamorosi: Consider making this a regular import at the top using the same pattern that we use in Idempotency for the persistence layers - which looks something like this:

```ts
import { DataMasking } from '@aws-lambda-powertools/data-masking'; // this does not include any provider
import { AWSEncryptionSDKProvider } from '@aws-lambda-powertools/data-masking/aws-provider';

const datamasking = new DataMasking({ provider: new AWSEncryptionSDKProvider() });
```

This way the provider gets its own sub-path export which functionally makes it opt-in.

dreamorosi: This might also allow you to not have `#init` to be async - which then in turn means you could do it in the constructor.

packages/data-masking/src/provider/AWSEncryptionSDKProvider.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    this.#clients = {
+      encryptFn: encrypt,
+      decryptFn: decrypt,
+      cacheManager: new NodeCachingMaterialsManager({
+        backingMaterials: keyring,
+        cache,
+        maxAge: (this.#options.maxCacheAgeSeconds ?? 300) * 1000,
+        maxMessagesEncrypted:
+          this.#options.maxMessagesEncrypted ?? 4294967296,
```
dreamorosi: Check that the defaults are aligned with the Python implementation.

packages/data-masking/src/provider/AWSEncryptionSDKProvider.ts
```diff
@@ -0,0 +1,108 @@
...
+    const ciphertext = Buffer.from(data, 'base64');
+    const { plaintext, messageHeader } = await decryptFn(
+      cacheManager,
+      new Uint8Array(ciphertext)
+    );
+
+    if (context) {
+      for (const [key, value] of Object.entries(context)) {
+        if (messageHeader.encryptionContext[key] !== value) {
+          throw new Error(
```
dreamorosi: Should we make this a custom package-managed error like all others

docs/features/data-masking.md
```diff
@@ -0,0 +1,693 @@
...
+            "address": {
+                "street": "123 Main St",
+                "postcode": 12345
+            }
+        }
+        ```
+
+    === "Result"
+
+        ```json hl_lines="6"
```
dreamorosi, line 276: Consider extracting all these in separate files

docs/features/data-masking.md
```diff
@@ -0,0 +1,693 @@
...
+
+#### AWS Encryption SDK
+
+You can modify the following values when initialising the `AWSEncryptionSDKProvider` to best accommodate your security and performance thresholds.
+
+| Parameter | Default | Description |
+| --- | --- | --- |
+| `localCacheCapacity` | `100` | The maximum number of entries that can be retained in the local cryptographic materials cache |
+| `maxCacheAgeSeconds` | `300` | The maximum time (in seconds) that a cache entry may be kept in the cache |
+| `maxMessagesEncrypted` | `4294967296` | The maximum number of messages that may be encrypted under a cache entry |
```
dreamorosi: Why is this magic number so special?

packages/data-masking/tests/e2e/dataMasking.test.ts
```diff
@@ -0,0 +1,122 @@
...
+  });
+
+  const lambdaFunctionCodeFilePath = join(
+    __dirname,
+    'dataMasking.test.FunctionCode.ts'
+  );
+
+  const kmsKey = new Key(testStack.stack, 'TestKmsKey', {
+    description: 'KMS key for DataMasking e2e tests',
+  });
```
dreamorosi, line 47: Let's make sure that the KMS Key actually gets cleaned up after the test

packages/data-masking/src/constants.ts
svozza: A `const` primitive is already narrowed to the literal type `'*****'` — `as const` only changes anything for object/array literals, so it would be a no-op here. Resolving.

packages/data-masking/src/DataMasking.ts
svozza: This matches Python's behaviour: `erase(data)` with no fields replaces the whole payload with `DATA_MASKING_STRING` (see provider/base.py `erase()`). We did adopt one refinement from Python while looking into this: arrays are now masked element-wise preserving length (`['a', 'b']` → `['*****', '*****']`) instead of collapsing to a single string — added in f9e33a18 along with erase() overloads so the return type is honest about this (`MaskedPayload<T>` for no-options calls, `T` otherwise).

packages/data-masking/src/DataMasking.ts
svozza: Superseded by the type rework in f9e33a18: `EraseOptions` is now a union (`{ fields: string[] } | { maskingRules: ... }`), so `fields` is required in its branch and the unreachable `?? []` fallback question goes away. Defaulting to `[]` would also have changed semantics — `erase(data)` (mask everything) and `erase(data, { fields: [] })` (mask nothing) mean different things.

packages/data-masking/src/DataMasking.ts
svozza: Per-field overrides are already covered by `maskingRules` (`customMask`, `dynamicMask`, regex). What's missing vs Python is the *top-level* options (`dynamic_mask`/`custom_mask`/`regex_pattern`/`mask_format` applied across all listed fields in one call). Should we aim for Python parity here?

For illustration, this is what the top-level form looks like in Python — one rule applied to every listed field:

```python
data_masker.erase(data, fields=["ssn", "card"], dynamic_mask=True)
# {"ssn": "***********", "card": "*******************", ...}
```

vs the per-field form we have today, where the rule must be repeated:

```typescript
masker.erase(data, {
  maskingRules: {
    ssn: { dynamicMask: true },
    card: { dynamicMask: true },
  },
});
```

The TS equivalent would be:

```typescript
masker.erase(data, { fields: ["ssn", "card"], dynamicMask: true });
```

Design notes if we go ahead:

- Per-field `maskingRules` should take precedence over the top-level rule when both are present, so the top level sets the default and individual fields can override it.
- Like Python, the top-level options should also work *without* `fields`: the rule is applied recursively to every leaf value in the payload. This doesn't change the no-options behaviour discussed above — `erase(data)` with no rule still collapses the whole payload to `*****`; the recursive walk only kicks in when a rule is given (a length-matched mask of an entire object makes no sense, so Python applies the rule per leaf instead).

packages/data-masking/src/DataMasking.ts
svozza: Kept `Promise.all` deliberately: encrypt/decrypt is all-or-nothing, and fail-fast means we never hand back a partially-encrypted payload (which would be easy to persist by accident). Since we operate on a copy, the caller's original is untouched on failure. Added tests in 0ca27079 demonstrating both: a single failing field rejects the whole operation with the provider's error, and the input object is left unmodified.

packages/data-masking/src/DataMasking.ts
svozza: Agreed, added in 0ca27079. We now `console.warn` when decrypt skips a non-string value, and also when a field path matches nothing with `throwOnMissingField: false` (Python warns in both places too).

packages/data-masking/src/provider/AWSEncryptionSDKProvider.ts
svozza: Verified against Python's data_masking/constants.py:
- localCacheCapacity 100 - same
- maxCacheAgeSeconds 300 - same
- maxMessagesEncrypted 4294967296 (2^32) - same
- maxBytesEncrypted: intentionally diverges — Python uses 2^63-1, which isn't representable as a safe JS integer, so we use Number.MAX_SAFE_INTEGER (2^53-1). Both are "effectively unlimited" for Lambda-scale payloads.

docs/features/data-masking.md
svozza: It's 2^32 — the AWS Encryption SDK's maximum number of messages allowed under a single cached data key. Added an explanation to the defaults table in e11edd25.

mkdocs.yml
svozza: Done — data-masking is in the nav now.

svozza: Done — data-masking is in the nav now.

packages/data-masking/package.json
svozza: Addressed in 2425dbb3 — version and commons dep refreshed to 2.33.1.

packages/data-masking/package.json
svozza: Addressed in 2425dbb3 — version and commons dep refreshed to 2.33.1.

packages/data-masking/src/index.ts
svozza: Addressed in 36276e76 — `@aws-lambda-powertools/data-masking/types` added.

packages/data-masking/src/index.ts
svozza: Addressed in 36276e76 — `@aws-lambda-powertools/data-masking/errors` added.

packages/data-masking/src/errors.ts
svozza: Addressed in 2425dbb3 — all custom errors now accept `options?: ErrorOptions`.

packages/data-masking/src/errors.ts
svozza: Addressed in 2425dbb3 — all custom errors now accept `options?: ErrorOptions`.

packages/data-masking/src/constants.ts
svozza: Addressed in 2425dbb3 — docblock added.

packages/data-masking/src/DataMasking.ts
svozza: Addressed in 2425dbb3.

packages/data-masking/src/DataMasking.ts
svozza: Addressed in 2425dbb3.

packages/data-masking/src/DataMasking.ts
svozza: Addressed in f9e33a18 — erase() overloads: no-options calls return `MaskedPayload<T>`, calls with options return `T`; cast removed.

packages/data-masking/src/DataMasking.ts
svozza: Addressed in 2425dbb3.

packages/data-masking/src/DataMasking.ts
svozza: Addressed in 2425dbb3, split per overload in 36276e76.

packages/data-masking/src/DataMasking.ts
svozza: Genuinely unreachable (early return above guarantees fields is set); comment reworded in 2425dbb3 and the question is moot after the union types in f9e33a18.

packages/data-masking/src/DataMasking.ts
svozza: Addressed in 2425dbb3.

packages/data-masking/src/DataMasking.ts
svozza: Addressed in 2425dbb3.

packages/data-masking/src/DataMasking.ts
svozza: Addressed in 2425dbb3.

packages/data-masking/src/DataMasking.ts
svozza: Addressed in f9e33a18 — union type makes the combination unrepresentable in TS; runtime precedence for JS callers matches Python (regex+format → custom → dynamic). Question: should we be supporting plain-JS callers defensively like this at runtime, or do we treat the types as the contract and drop guards the type system already enforces? This affects more than just this function, so let's be consistent either way.

packages/data-masking/src/types.ts
svozza: Addressed in f9e33a18 — `MaskingRule` and `EraseOptions` are now unions.

packages/data-masking/src/types.ts
svozza: Addressed in f9e33a18 — `MaskingRule` and `EraseOptions` are now unions.

packages/data-masking/src/DataMasking.ts
svozza: Addressed in 2425dbb3.

packages/data-masking/src/provider/AWSEncryptionSDKProvider.ts
svozza: Addressed in d889ee54 — `providers/kms/types.ts` with its own sub-path export.

packages/data-masking/src/provider/AWSEncryptionSDKProvider.ts
svozza: Addressed in 2425dbb3.

packages/data-masking/src/provider/AWSEncryptionSDKProvider.ts
svozza: Addressed in 2425dbb3.

packages/data-masking/src/provider/AWSEncryptionSDKProvider.ts
svozza: Addressed in d889ee54 — provider moved to `/providers/kms`, static imports, `#init` folded into the constructor.

svozza: Addressed in d889ee54 — `#init` removed; client assembly now happens synchronously in the constructor.

packages/data-masking/src/provider/AWSEncryptionSDKProvider.ts
svozza: Addressed in 2425dbb3 — throws `DataMaskingEncryptionError`.

docs/features/data-masking.md
svozza: Addressed in e11edd25 — all snippets extracted to `examples/snippets/data-masking/`.

packages/data-masking/tests/e2e/dataMasking.test.ts
svozza: Addressed in e1b4d47c — `RemovalPolicy.DESTROY` with 7-day pending window; keys were indeed being retained before.

svozza:
All review feedback addressed — replies posted on every thread with the commit that handles it.

Two threads carry **open questions**:

1. [Top-level mask options](https://github.com/aws-powertools/powertools-lambda-typescript/pull/5143#discussion_r3397020991) — should we aim for Python parity (`dynamic_mask`/`custom_mask`/etc. applied across all listed fields in one call)?
2. [TS-only vs JS-defensive runtime](https://github.com/aws-powertools/powertools-lambda-typescript/pull/5143#discussion_r3397075560) — do we keep runtime guards the type system already enforces, for plain-JS callers? Let's pick a stance and be consistent.

packages/data-masking/src/DataMasking.ts
dreamorosi: Good points:
- per-field `maskingRules` do take precedence yes
- top-level w/o fields are applied to leaves - agreed
- since we don't have positional named arguments, we need to make them props of the same object where we pass `fields` and `maskingRules` so we don't have `undefined positional arguments

```typescript
masker.erase(data, {
  dynamicMask: boolean,
  // other "top level"
  maskingRules: {
    ssn: { dynamicMask: true },
    card: { dynamicMask: true },
  },
});
```

packages/data-masking/src/DataMasking.ts
dreamorosi: Historically we've not been very consistent with this topic in the codebase. When we started, mostly we relied on the type system but over time, in certain areas, we added this defensive programming.

My suggestion would be to basically evaluate on a case by case basis, and see how much complexity vs risk of runtime misbehavior you see and decide accordingly.

packages/data-masking/src/constants.ts
dreamorosi: Gotcha - TIL

packages/data-masking/src/DataMasking.ts
dreamorosi: Let's make sure to document this in the docs

packages/data-masking/src/DataMasking.ts
svozza: Implemented in ff5c2f8e. `EraseOptions` is now `MaskingRule & { fields?; maskingRules? }`, so a top-level rule, `fields`, and `maskingRules` compose in one call exactly as sketched:

```typescript
masker.erase(data, {
  fields: ['ssn', 'card'],
  dynamicMask: true,                              // default for both fields
  maskingRules: { card: { customMask: 'XXXX' } }, // overrides card
});
```

The top-level rule applies to every listed field, or to every leaf when `fields` is omitted; per-field `maskingRules` take precedence for the paths they name. Added unit + type tests and a docs section with a snippet.

packages/data-masking/src/DataMasking.ts
svozza: Went with trusting the type system here, per your case-by-case steer: `applyMaskingRule` now relies on the `MaskingRule` union rather than re-checking conflicting combinations at runtime, and I dropped the two JS-caller precedence tests (ff5c2f8e).

Separately, while aligning the new top-level options against the Python utility I found our per-field rules were *throwing* `DataMaskingUnsupportedTypeError` on non-string values, which Python doesn't do — its provider stringifies (`provider.erase(str(value), ...)`) and only skips `None`. I changed both the per-field and top-level paths to stringify so TS matches Python; `null`/`undefined` still pass through unchanged.

dreamorosi (approved):
Great work on this one @svozza - let's chat before the next release so we can do the pre-req steps on npm.

packages/data-masking/src/DataMasking.ts
dreamorosi: Good one

svozza:
Sweet!

PR #5146, opened 2026-03-30
style: remove useAwait linting rule

### LINKED ISSUE #5144: Maintenance: disable useAwait biome lint rule
### Summary

We should disable the `useAwait` biome lint rule (`suspicious/useAwait`), which is currently explicitly set to `"error"` in our `biome.json`.

### Why is this needed?

Biome themselves mark this rule as [not recommended](https://biomejs.dev/linter/rules/use-await/) and it is not enabled in any of their default presets. In practice, the rule causes real issues:

- Lambda ESM handlers **must** return a Promise for the runtime to serialise the return value. A sync handler that returns a plain value will result in `null` being returned to the caller. The `useAwait` rule forces us to use workarounds like `Promise.resolve()` instead of simply marking the handler as `async`, which is less readable and obscures the actual intent.
- There are many legitimate cases where a function is `async` for interface conformance (e.g. implementing a provider interface, middleware signatures) without needing an `await` in the body.
- The rule adds friction during development without catching real bugs.

Automation

### Solution

Remove `"useAwait": "error"` from `biome.json`. Since the rule is not in any default preset, removing it is sufficient — it will default to off.

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR removes the `useAwait` linting rule from the project because as discussed in the linked issue, it only added unneeded overhead. The PR also removes most - if not all - the `setTimeout` and `Promise.resolve` occurrences that were added because of this rule.

Finally, as part of working on the PR I ran `npm run lint:fix` on the project and several export orders changed as result. Similarly, I ran `npm audit fix` to remove a few CVEs from the dev environment rather than waiting for Dependabot to pick things up. 

**Issue number:** closes #5144

---

### DIFF
--- a/packages/tracer/tests/unit/Tracer.test.ts
+++ b/packages/tracer/tests/unit/Tracer.test.ts
@@ -1,3 +1,4 @@
+import { setTimeout } from 'node:timers/promises';
 import type { LambdaInterface } from '@aws-lambda-powertools/commons/types';
 import context from '@aws-lambda-powertools/testing-utils/context';
 import type { Context } from 'aws-lambda';
@@ -1278,7 +1279,9 @@ describe('Class: Tracer', () => {
 
       class Lambda implements LambdaInterface {
         @tracer.captureMethod()
-        public async dummyMethod(): Promise<void> {}
+        public async dummyMethod(): Promise<void> {
+          setTimeout(1);
+        }
 
         public async handler(
           _event: unknown,


dreamorosi:
The Dependency Review CI check fails because of the `yaml` package having [`GHSA-48c2-rrv3-qjmp`](https://github.com/advisories/GHSA-48c2-rrv3-qjmp).

The package comes from multiple libs belonging to AWS CDK:

```sh
npm ls yaml
aws-lambda-powertools-typescript@0.0.1 /Users/aamorosi/Codes/aws-lambda-powertools-typescript
├─┬ @aws-lambda-powertools/testing-utils@2.32.0 -> ./packages/testing
│ ├─┬ @aws-cdk/toolkit-lib@1.19.1
│ │ └── yaml@1.10.3
│ └─┬ aws-cdk-lib@2.245.0
│   └── yaml@1.10.3
├─┬ lint-staged@16.4.0
│ └── yaml@2.8.3
├─┬ typedoc@0.28.18
│ └── yaml@2.8.3 deduped
└─┬ vitest@4.1.2
  └─┬ vite@8.0.3
    └── yaml@2.8.3 deduped
```

I tried manually updating them but nothing changed - since they're not related to this PR and are development dependencies, I'd recommend merging the PR anyway even though I don't like having a failing CI while doing it.

PR #5157, opened 2026-04-01
fix(commons): keep AWS_SDK_UA_APP_ID idempotent and <=50 chars

### LINKED ISSUE #5082: Bug: `AWS_SDK_UA_APP_ID` can exceed 50-character limit during initialisation
### Expected Behavior

The `AWS_SDK_UA_APP_ID` should remain within the 50-character limit and should not contain redundant or duplicated Powertools version strings, ensuring compatibility with the AWS SDK.

### Why this matters
As shown in the logic flow of how User Agents are constructed, the App ID is a suffix that helps AWS identify the origin of the request. If it is malformed or too long, it risks breaking the integration with AWS services that enforce strict header validation.

### Current Behavior

In `packages/commons/src/index.ts`, the logic used to append Powertools metadata to the AWS_SDK_UA_APP_ID environment variable does not account for the AWS SDK's 50-character limit or the possibility of multiple initializations (e.g., when multiple versions of commons exist in the dependency tree).

The current implementation blindly appends strings:

```
if (process.env.AWS_SDK_UA_APP_ID) {
  process.env.AWS_SDK_UA_APP_ID = `${process.env.AWS_SDK_UA_APP_ID}/PT/NO-OP/${PT_VERSION}/PTEnv/${env}`;
} else {
  process.env.AWS_SDK_UA_APP_ID = `PT/NO-OP/${PT_VERSION}/PTEnv/${env}`;
}
```

If a user already has an App ID defined, or if two different versions of the library are loaded (common in monorepos or nested dependencies), the string can quickly look like this:
MyCustomAppID/PT/NO-OP/2.0.0/PTEnv/prod/PT/NO-OP/2.1.0/PTEnv/prod

This exceeds the 50-character limit defined in the [AWS SDK Shared Configuration documentation](https://docs.aws.amazon.com/sdkref/latest/guide/feature-appid.html), which may lead to SDK initialization errors or truncated User-Agent headers.

### Code snippet

```typescript
/**
 * Reproduction Script: AWS SDK UA App ID Overflow
 * This simulates how 'packages/commons/src/index.ts' handles the environment variable.
 */

const PT_VERSION = "2.0.4";
const env = "production";
const LIMIT = 50;

function simulateCommonsInit() {
    // This is the exact logic currently in Powertools commons
    if (process.env.AWS_SDK_UA_APP_ID) {
        process.env.AWS_SDK_UA_APP_ID = `${process.env.AWS_SDK_UA_APP_ID}/PT/NO-OP/${PT_VERSION}/PTEnv/${env}`;
    } else {
        process.env.AWS_SDK_UA_APP_ID = `PT/NO-OP/${PT_VERSION}/PTEnv/${env}`;
    }
}

// --- SCENARIO 1: User already has a custom App ID ---
process.env.AWS_SDK_UA_APP_ID = "MyCompany-ECommerce-Ordering-Service"; // 36 chars
console.log(`Initial Length: ${process.env.AWS_SDK_UA_APP_ID.length}`);

simulateCommonsInit();

console.log(`\n--- After 1st Init ---`);
console.log(`Value:  ${process.env.AWS_SDK_UA_APP_ID}`);
console.log(`Length: ${process.env.AWS_SDK_UA_APP_ID.length}`);
console.log(`Status: ${process.env.AWS_SDK_UA_APP_ID.length > LIMIT ? '❌ FAILED (Too Long)' : '✅ OK'}`);

// --- SCENARIO 2: Multiple versions of Powertools loaded ---
// (Simulating a second 'commons' package from a different version/location being loaded)
simulateCommonsInit();

console.log(`\n--- After 2nd Init (Multiple Versions) ---`);
console.log(`Value:  ${process.env.AWS_SDK_UA_APP_ID}`);
console.log(`Length: ${process.env.AWS_SDK_UA_APP_ID.length}`);
console.log(`Status: ${process.env.AWS_SDK_UA_APP_ID.length > LIMIT ? '❌ FAILED (Too Long)' : '✅ OK'}`);
```

### Steps to Reproduce

1. Set an initial environment variable: `export AWS_SDK_UA_APP_ID="MyExistingApplicationID"` (24 chars)
2. Import any Powertools utility that triggers the `commons` index initialization
3. Observe that the new value is ~55+ characters
4. If a second version of `commons` is initialised, the string grows indefinitely

### Possible Solution

1. Idempotency: Check if the string already contains `PT/NO-OP` before appending
2. Bounds Checking: Ensure the final string is sliced to a maximum of 50 characters

latest

24.x

### Packaging format used

npm

### Execution logs

```Shell
WARN The provided userAgentAppId exceeds the maximum length of 50 characters.
```

saschabuehrle:
Fixes #5082

This patch makes commons user-agent app-id initialization safe for repeated module loads and compliant with the AWS SDK 50-char limit:
- avoid appending the PT/NO-OP marker when it already exists
- cap the final AWS_SDK_UA_APP_ID value to 50 chars
- add regression tests for idempotency and max-length behavior

Greetings, saschabuehrle

### DIFF
--- a/packages/commons/src/index.ts
+++ b/packages/commons/src/index.ts
@@ -1,11 +1,20 @@
 import { PT_VERSION } from './version.js';
 
 const env = process.env.AWS_EXECUTION_ENV || 'NA';
-if (process.env.AWS_SDK_UA_APP_ID) {
-  process.env.AWS_SDK_UA_APP_ID = `${process.env.AWS_SDK_UA_APP_ID}/PT/NO-OP/${PT_VERSION}/PTEnv/${env}`;
-} else {
-  process.env.AWS_SDK_UA_APP_ID = `PT/NO-OP/${PT_VERSION}/PTEnv/${env}`;
-}
+const POWETOOLS_NOOP_UA = `PT/NO-OP/${PT_VERSION}/PTEnv/${env}`;
+const AWS_SDK_UA_APP_ID_MAX_LENGTH = 50;
+
+const currentUserAgentAppId = process.env.AWS_SDK_UA_APP_ID?.trim();
+const nextUserAgentAppId = currentUserAgentAppId?.includes('PT/NO-OP')
+  ? currentUserAgentAppId
+  : currentUserAgentAppId
+    ? `${currentUserAgentAppId}/${POWETOOLS_NOOP_UA}`
+    : POWETOOLS_NOOP_UA;
+
+process.env.AWS_SDK_UA_APP_ID = nextUserAgentAppId.slice(
+  0,
+  AWS_SDK_UA_APP_ID_MAX_LENGTH
+);
 
 export { addUserAgentMiddleware, isSdkClient } from './awsSdkUtils.js';
 export { deepMerge } from './deepMerge.js';
--- a/packages/commons/tests/unit/awsSdkUtils.test.ts
+++ b/packages/commons/tests/unit/awsSdkUtils.test.ts
@@ -89,6 +89,32 @@ describe('Helpers: awsSdk', () => {
     );
   });
 
+  it('does not append PT no-op marker more than once', async () => {
+    // Prepare
+    process.env.AWS_SDK_UA_APP_ID = `test/PT/NO-OP/${version}/PTEnv/NA`;
+
+    // Act
+    vi.resetModules();
+    await import('../../src/index.js');
+
+    // Assess
+    expect(process.env.AWS_SDK_UA_APP_ID).toEqual(
+      `test/PT/NO-OP/${version}/PTEnv/NA`
+    );
+  });
+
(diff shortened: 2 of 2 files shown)


saschabuehrle:
Thanks for the detailed review. Agreed on the concern about doing normalization work at import time.

I’m going to rework this around an explicit opt-in/guard path so we avoid repeated module-load mutations and keep behavior predictable. I’ll post an updated patch here.

Greetings, saschabuehrle

saschabuehrle:
Thanks for the review. I agree this needs a different approach than import-time normalization. I’m parking this one while I work out a cleaner opt-in path and will come back with a revised patch.\n\nGreetings, saschabuehrle

saschabuehrle:
Closing this out for now since the current approach isn't the right direction. I'll reopen with a new patch once I have a cleaner opt-in design that matches the maintainer guidance.\n\nGreetings, saschabuehrle

PR #5165, opened 2026-04-06
docs(commons): fix isStrictEqual and areArraysEqual JSDoc array comparison description

### LINKED ISSUE #5162: Docs: isStrictEqual and areArraysEqual JSDoc incorrectly describes array comparison as order-independent
### Changes

**What were you searching in the docs?**
I was reading the API reference for `isStrictEqual` in the commons package to understand how it compares arrays.

**Is this related to an existing documentation section?**
https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/commons/src/typeUtils.ts

**How can we improve?**
The JSDoc for `isStrictEqual` and `areArraysEqual` says array comparison is "regardless of order", but the implementation is order-dependent. It compares elements at the same index: `left.every((value, i) => isStrictEqual(value, right[i]))`.

The existing unit test confirms this: `isStrictEqual([1, 2, 3], [1, 3, 2])` is expected to return `false`.

Two things in the JSDoc are wrong:

1. `areArraysEqual` says "This function compares each element in the arrays, regardless of order." It should say something like "at the same position".
2. `isStrictEqual` has the same issue. The code example shows `isStrictEqual([1, 2, 3], [3, 2, 1]) // true`, but the actual return value is `false`.

**Suggestion:**
- Change "regardless of order" to "at the same position" in both JSDoc blocks.
- Update the `isStrictEqual` example: either change `[3, 2, 1]` to `[1, 2, 3]`, or change the comment from `// true` to `// false`.

---

### Acknowledgment

Zelys-DFKH:
## Summary

The JSDoc for `areArraysEqual` and `isStrictEqual` described array comparison as "regardless of order", but both functions compare elements at the same index position, which is order-dependent.

### Changes

Updated the description and example in both JSDoc blocks to reflect the actual behaviour.

**Issue number:** closes #5162

---

### DIFF
--- a/examples/app/package.json
+++ b/examples/app/package.json
@@ -51,7 +51,7 @@
     "@types/node": "25.5.0",
     "aws-cdk": "^2.1113.0",
     "constructs": "^10.6.0",
-    "esbuild": "^0.27.5",
+    "esbuild": "^0.28.0",
     "typescript": "^6.0.2"
   }
 }
--- a/layers/package.json
+++ b/layers/package.json
@@ -44,6 +44,6 @@
   "dependencies": {
     "aws-cdk": "^2.1113.0",
     "aws-cdk-lib": "^2.244.0",
-    "esbuild": "^0.27.5"
+    "esbuild": "^0.28.0"
   }
 }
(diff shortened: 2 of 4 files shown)


Zelys-DFKH:
Thanks for merging, appreciate the quick review

PR #5173, opened 2026-04-07
chore(deps): add esbuild as explicit devDependency

### LINKED ISSUE #5172: Maintenance: Add esbuild as explicit devDependency after Vite 8 dropped it as transitive dependency
### Summary

The upgrade to Vitest (which now uses Vite 8) removed esbuild as a transitive dependency since Vite 8 switched from esbuild to Rolldown. This caused CDK's `NodejsFunction` to silently fall back to Docker-based bundling during e2e tests, resulting in `exec format error` failures due to architecture mismatches on CI runners.

### Why is this needed?

Without esbuild explicitly listed as a devDependency, the e2e tests fail with a cryptic `exec /bin/sh: exec format error` during the Docker-based bundling fallback. This blocks CI pipelines and is difficult to diagnose since the root cause (missing esbuild binary) is hidden behind the Docker fallback behavior.

Tests

### Solution

Add `esbuild` as an explicit `devDependency` in the root `package.json`.

### Acknowledgment

- [X] This request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.aws.amazon.com/powertools/typescript/latest/#tenets)

### Future readers

sdangol:
### Changes

Add `esbuild` as an explicit `devDependency` in the root `package.json`. After the Vitest upgrade moved to Vite 8, esbuild was dropped as a transitive dependency since Vite 8 switched from esbuild to Rolldown. This caused CDK's `NodejsFunction` to fall back to Docker-based bundling during e2e tests, resulting in `exec format error` failures on CI runners.

**Issue number:** closes #5172

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -140,9 +140,9 @@
       "license": "MIT"
     },
     "node_modules/@aws-cdk/asset-awscli-v1": {
-      "version": "2.2.263",
-      "resolved": "https://registry.npmjs.org/@aws-cdk/asset-awscli-v1/-/asset-awscli-v1-2.2.263.tgz",
-      "integrity": "sha512-X9JvcJhYcb7PHs8R7m4zMablO5C9PGb/hYfLnxds9h/rKJu6l7MiXE/SabCibuehxPnuO/vk+sVVJiUWrccarQ==",
+      "version": "2.2.273",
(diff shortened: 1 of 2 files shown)


svozza (approved):

dreamorosi:
I disagree that this should be a root-level package.

AWS CDK is not a root level package either and is instead used (and depended upon) in `packages/testing` and `layers` - `esbuild` should've been added as a non-dev dependency to both of these imo.

PR #5175, opened 2026-04-09
improv(parser): export InferOutput from public types entry point

### LINKED ISSUE #5174: Bug: `InferOutput` not exported from public types entry point, causing TS2883 with TypeScript 6
### Expected Behavior

Code using the `parser` middleware with `safeParse: true` should compile cleanly under TypeScript 6. All types that appear in the inferred return type of a consumer function must be reachable via a public package export so TypeScript can generate valid `.d.ts` declaration files.

### Current Behavior

Compiling with TypeScript 6 raises:

```
error TS2883: The inferred type of 'myHandler' cannot be named without a reference to
'InferOutput' from '.../node_modules/@aws-lambda-powertools/parser/lib/esm/types/parser.js'.
This is likely not portable. A type annotation is necessary.
```

`InferOutput` is used in the `parser()` middleware's type signature but is only reachable at the internal path `lib/esm/types/parser.js`. It is not re-exported from `@aws-lambda-powertools/parser/types` or any other public entry point, so TypeScript cannot include a portable reference to it in generated declaration files.

### Code snippet

```typescript
import { parser } from '@aws-lambda-powertools/parser/middleware';
import { APIGatewayProxyEventV2Schema } from '@aws-lambda-powertools/parser/schemas';

// TS2883 is raised on this exported symbol
export const myHandler = middy()
  .use(parser({ schema: APIGatewayProxyEventV2Schema, safeParse: true }))
  .handler(async (event) => {
    return { statusCode: 200 };
  });
```

### Steps to Reproduce

1. Create a TypeScript project with `typescript@^6.0.0`
2. Install `@aws-lambda-powertools/parser`, `@middy/core`, and `zod`
3. Add the code snippet above to an exported file (declaration emit must be on, or use `tsc --noEmit`)
4. Run `tsc`
5. Observe TS2883

### Possible Solution

Export `InferOutput` from `packages/parser/src/types/index.ts` so it is available at the public `@aws-lambda-powertools/parser/types` entry point. This is a one-line, type-only change with no runtime impact — TypeScript will then automatically use the public path in generated `.d.ts` files and the error resolves without any changes required on the consumer side.

I am opening a PR with this fix alongside this issue.

latest

24.x

### Packaging format used

npm

### Execution logs

```Shell

```

yashar-new10:
## Summary
Exports `InferOutput` from the public `@aws-lambda-powertools/parser/types` entry point.

### Changes

TypeScript enforces that all types appearing in a function's inferred return type must be nameable via public package exports (see [TS2742](https://github.com/microsoft/TypeScript/issues/47663) / TS2883 — the latter is a variant that also reports the source module path). This check becomes harder to avoid in TypeScript 6 because the compiler is stricter about declaration emit portability.

When using the `parser` middleware with `safeParse: true`, the return type of the wrapping function includes `InferOutput<TSchema>`. Because `InferOutput` was only exported from the internal path `lib/esm/types/parser.js` and not from any public entry point, TypeScript 6 could not generate a valid declaration file and raised:

```
error TS2883: The inferred type of 'X' cannot be named without a reference to
'InferOutput' from '...node_modules/@aws-lambda-powertools/parser/lib/esm/types/parser.js'.
This is likely not portable. A type annotation is necessary.
```

Consumers hitting this error were forced to add explicit type casts or annotations as a workaround.

### Fix

Add `InferOutput` to the existing re-exports in `packages/parser/src/types/index.ts`. This is a one-line, type-only change with no runtime impact. Once `InferOutput` is available at `@aws-lambda-powertools/parser/types`, TypeScript can reference it in generated `.d.ts` files automatically and the error is resolved without any changes required on the consumer side.

### Testing

Verified locally by:
1. Manually adding `InferOutput` to the built `lib/esm/types/index.d.ts`
2. Confirming that TS2883 is resolved without any type annotations or casts in consumer code
3. Confirming that removing the manual edit brings the error back

**Issue number:** closes #5174

---

### DIFF
--- a/packages/parser/src/types/index.ts
+++ b/packages/parser/src/types/index.ts
@@ -1,4 +1,5 @@
 export type {
+  InferOutput,
   ParsedResult,
   ParsedResultError,
   ParsedResultSuccess,


svozza (approved):

svozza:
Thank you for the fix!

PR #5186, opened 2026-04-14
fix(logger): correctly serialize shared object references in JSON replacer

### LINKED ISSUE #5167: Bug: extraInput shows array of null values when destructing array of objects
### Expected Behavior

When destructuring an array of objects to a field in the log's extraInput, I expect that the values will be shown

### Current Behavior

After destructing an array of objects, an array of null values is shown

Image

### Code snippet

```typescript
const x1 = [{ value: "1" }];
const x2 = [{ value: "2" }];
log.info("### fails", {
  x1,
  destructed: [...x1],
});
log.info("### fails2", {
  destructed1: [...x1],
  destructed2: [...x1],
});

log.info("### no destruction", {
  x1,
  x2,
});
````

### Steps to Reproduce

1. run the code above

### Possible Solution

_No response_

latest

24.x

### Packaging format used

npm

### Execution logs

```Shell
{
    "level": "INFO",
    "message": "### fails",
    "timestamp": "2026-04-07T10:53:53.763Z",
    "service": "service_fixStationEntitiesHandler",
    "sampling_rate": 0,
    "xray_trace_id": "1-69d4e240-0aea710630d9117d4d529cd4",
    "x1": [
        {
            "value": "1"
        }
    ],
    "destructed": [
        null
    ]
}

{
    "level": "INFO",
    "message": "### fails2",
    "timestamp": "2026-04-07T10:53:53.768Z",
    "service": "service_fixStationEntitiesHandler",
    "sampling_rate": 0,
    "xray_trace_id": "1-69d4e240-0aea710630d9117d4d529cd4",
    "destructed1": [
        {
            "value": "1"
        }
    ],
    "destructed2": [
        null
    ]
}
```

svozza:
## Summary

The Logger's `getJsonReplacer()` used a `WeakSet` to detect circular references during JSON serialization, which incorrectly treated duplicate object references as circular. This caused shared objects (e.g. spread array elements) to serialize as `null` in arrays or be omitted from objects.

### Changes

- Replace `WeakSet` with ancestor-chain tracking in `getJsonReplacer()` — only objects on the current path from root to the node being serialized are tracked, allowing the same object to appear in multiple branches while still detecting true circular references
- Switch from arrow function to regular function in the replacer so `JSON.stringify` can bind `this` to the parent object, enabling ancestor-chain pruning
- Add two test cases reproducing the exact scenarios from the bug report (spread array alongside original, multiple spreads of the same array)

**Issue number:** closes #5167

---


packages/logger/src/Logger.ts
```diff
@@ -761,25 +761,29 @@ class Logger extends Utility implements LoggerInterface {
    * @see {@link ConstructorOptions.jsonReplacerFn}
    */
   protected getJsonReplacer(): (key: string, value: unknown) => void {
-    const references = new WeakSet();
+    const ancestors: unknown[] = [];
+    const jsonReplacerFn = this.#jsonReplacerFn;
+    const logFormatter = this.getLogFormatter();
 
-    return (key, value) => {
+    return function (this: unknown, key: string, value: unknown) {
```
dreamorosi, line 768: Now that we have changed the signature of the function, does it break the [BYO JSON replacer feature](https://docs.aws.amazon.com/powertools/typescript/latest/features/logger/#extend-json-replacer-function)?

svozza: Ah good catch, this function is the return value, I was thinking of it more as an internal detail but of course this is actually the public API.

svozza: So, this is not actually a breaking change. We are using the TypeScript [feature](https://www.typescriptlang.org/docs/handbook/2/functions.html#declaring-this-in-a-function) that allows you to specify `this` in a function:

> ...[T]here are a lot of cases where you need more control over what object `this` represents. The JavaScript specification states that you cannot have a parameter called `this`, and so TypeScript uses that syntax space to let you declare the type for this in the function body.
>
>...
>
> This pattern is common with callback-style APIs, where another object typically controls when your function is called. Note that you need to use function and not arrow functions to get this behavior:

The `this` parameter  gets erased by the compiler at build time: note that you can see the actual return type in the type definition of the `getJsonReplacer` method hasn't changed and there are no compiler errors.

The old implementation returned an arrow function (so `this` inside was the Logger), the new one returns a regular function so `JSON.stringify` binds `this` to the parent object. That change allows us to use `this` as a way to identify when an object at the top of the ancestors list is the same reference as `this`: that means the parent and child are the same object and thus we have a circular reference.

dreamorosi: Ok that makes sense - thanks for explaining, I forgot it gets erased.

For the purpose of the docs, should be do any change to suggest passing the `this` reference or writing a full function vs an arrow one, to avoid `this` binding implications?

svozza: No, I think we're fine. This is an implementation detail that doesn't get exposed to consumers.

dreamorosi (approved):

PR #5227, opened 2026-05-07
fix(logger): flush log buffer on critical() when flushOnErrorLog is e…

### LINKED ISSUE #5210: Bug: critical() does not flush the log buffer when flushOnErrorLog is enabled
## Expected Behavior

When `flushOnErrorLog` is enabled in the log buffer options, calling `logger.critical()` should flush the buffer before emitting the log — the same behavior as `logger.error()`. CRITICAL is more severe than ERROR, so it should at minimum have the same flush-on-error semantics.

## Current Behavior

`error()` checks `flushOnErrorLog` and flushes the buffer (Logger.ts:411-413), but `critical()` (Logger.ts:387-392) does not. This means buffered diagnostic logs (debug, info, etc.) are flushed when an ERROR is logged but silently lost when a CRITICAL is logged.

## Code snippet

```typescript

const logger = new Logger({
  logLevel: 'ERROR',
  logBufferOptions: { enabled: true, flushOnErrorLog: true },
});

logger.debug('diagnostic info that should be preserved');
logger.critical('system failure');
// Expected: the debug log is flushed to output before the critical log
// Actual: the debug log is silently discarded
```

## Steps to Reproduce

1. Create a Logger with `logBufferOptions: { enabled: true, flushOnErrorLog: true }`
2. Log a debug message: `logger.debug('diagnostic info')`
3. Log a critical message: `logger.critical('system failure')`
4. Observe that the buffered debug log was never output

## Possible Solution

The most direct fix is to add the same flush check to `critical()`. However, a more robust approach would be to centralise the flush-on-error logic inside `processLogItem()`, which all log level methods already funnel through. `processLogItem` could check whether the incoming log level meets a severity threshold (e.g. `>= LogLevelThreshold.ERROR`) and flush before processing. This would make the individual methods simple one-liners and prevent the issue from recurring if new severity levels are added in the future.

2.33.0

22.x

## Packaging format used

npm

---

vishwakt:
### Changes

Centralizes the flush-on-error logic from `error()` into `processLogItem()` so that any log level `>= ERROR` (including `CRITICAL`) will automatically flush the buffer when `flushOnErrorLog` is enabled.

Previously, only `error()` checked `flushOnErrorLog` and flushed the buffer. `critical()`, despite being more severe, did not flush — causing buffered diagnostic logs to be silently discarded.

#### What changed

- **`Logger.ts`**: Removed the buffer flush check from `error()` and moved it into `processLogItem()`, which all log level methods funnel through. The condition checks `logLevel >= LogLevelThreshold.ERROR && logLevel < LogLevelThreshold.SILENT` to cover both `ERROR` and `CRITICAL` while excluding `SILENT`.
- **`logBuffer.test.ts`**: Added two unit tests:
  - Verifies `critical()` flushes the buffer when `flushOnErrorLog` is enabled (default)
  - Verifies `critical()` does **not** flush the buffer when `flushOnErrorLog` is explicitly disabled

**Issue number:** Closes #5210

---

### DIFF
--- a/packages/logger/tests/unit/logBuffer.test.ts
+++ b/packages/logger/tests/unit/logBuffer.test.ts
@@ -271,8 +271,8 @@ describe('Buffer logs', () => {
     logger.critical('This is a critical message');
 
     // Assess
-    expect(console.debug).toBeCalledTimes(1);
-    expect(console.error).toBeCalledTimes(1); // critical uses console.error
+    expect(console.debug).toHaveBeenCalledTimes(1);
+    expect(console.error).toHaveBeenCalledTimes(1); // critical uses console.error
   });
 
   it('does not flush on critical logs when flushOnErrorLog is disabled', () => {
@@ -287,8 +287,8 @@ describe('Buffer logs', () => {
     logger.critical('This is a critical message');
 
     // Assess
-    expect(console.debug).toBeCalledTimes(0);
-    expect(console.error).toBeCalledTimes(1); // critical uses console.error
+    expect(console.debug).toHaveBeenCalledTimes(0);
+    expect(console.error).toHaveBeenCalledTimes(1); // critical uses console.error
   });
 
   it('passes down the same buffer config to child loggers', () => {
--- a/packages/logger/tests/unit/logBuffer.test.ts
+++ b/packages/logger/tests/unit/logBuffer.test.ts
@@ -259,6 +259,38 @@ describe('Buffer logs', () => {
     expect(console.error).toBeCalledTimes(1);
   });
 
+  it('it flushes the buffer when a critical log is logged', () => {
+    // Prepare
+    const logger = new Logger({
+      logLevel: LogLevel.ERROR,
+      logBufferOptions: { enabled: true },
+    });
+
+    // Act
+    logger.debug('This is a log message');
+    logger.critical('This is a critical message');
+
+    // Assess
+    expect(console.debug).toBeCalledTimes(1);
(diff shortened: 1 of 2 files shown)


packages/logger/tests/unit/logBuffer.test.ts
```diff
@@ -259,6 +259,38 @@ describe('Buffer logs', () => {
...
+      logLevel: LogLevel.ERROR,
+      logBufferOptions: { enabled: true },
+    });
+
+    // Act
+    logger.debug('This is a log message');
+    logger.critical('This is a critical message');
+
+    // Assess
+    expect(console.debug).toBeCalledTimes(1);
```
svozza: I know this is following the existing pattern but `toBeCalledTimes` has been deprecated in Vitest so let's use `toHaveBeenCalledTimes` here. Don't worry about the existing tests, I will create a separate issue but let's ensure new tests use the new method.

svozza:
One minor comment but other than that LGTM!

svozza (changes requested):

packages/logger/tests/unit/logBuffer.test.ts
vishwakt: Thanks! Updated the new tests to use toHaveBeenCalledTimes

svozza (approved):

packages/logger/tests/unit/logBuffer.test.ts
svozza: Nice, merging now!

PR #5228, opened 2026-05-07
fix(metrics): account for all dimensions in setDefaultDimensions() li…

### LINKED ISSUE #5206: Bug: setDefaultDimensions() ignores regular dimensions in limit check
## Expected Behavior

`setDefaultDimensions()` should account for all dimensions (default + regular + dimension sets) when enforcing the maximum dimension count, ensuring the total never exceeds CloudWatch's limit.

## Current Behavior

`setDefaultDimensions()` at `Metrics.ts:796-813` only checks the count of existing *default* dimensions against the limit, completely ignoring regular dimensions added via `addDimension()`:

```ts
const currentCount = Object.keys(currentDefaultDimensions).length;
const newSetCount = Object.keys(newDimensions).length;
  throw new RangeError(...);
}
```

You can exceed CloudWatch's limit by adding 25 regular dimensions, then calling `setDefaultDimensions()` with 20 new keys — the check only sees `1 (service) + 20 = 21 < 29` and allows it, even though the total is now 46.

CloudWatch will silently drop the metric.

## Code snippet

```typescript

const metrics = new Metrics({ namespace: 'test' });

// Add 25 regular dimensions (plus 'service' default = 26 total)
for (let i = 0; i < 25; i++) {
  metrics.addDimension(`regular-${i}`, 'test');
}

// ❌ Should throw but doesn't — only checks default count (1) + new (20) = 21 < 29
const newDefaults: Record<string, string> = {};
for (let i = 0; i < 20; i++) {
  newDefaults[`new-default-${i}`] = 'test';
}
metrics.setDefaultDimensions(newDefaults);
// No error — but total dimensions is now 46, well above CloudWatch's limit
```

## Steps to Reproduce

1. Create a `Metrics` instance (1 default `service` dimension)
2. Add 25 regular dimensions via `addDimension()` (total = 26)
3. Call `setDefaultDimensions()` with 20 new keys
4. Observe no error, even though total dimensions = 46

## Possible Solution

Count only the genuinely new keys that don't overlap with existing defaults, then check against the full dimension count:

```ts
const newKeysCount = Object.keys(newDimensions)
  .filter((key) => !Object.hasOwn(currentDefaultDimensions, key)).length;
const totalAfterMerge = this.#dimensionsStore.getDimensionCount() + newKeysCount;
if (totalAfterMerge >= MAX_DIMENSION_COUNT) {
  throw new RangeError(...);
}
```

2.33.0

22.x

## Packaging format used

npm

---

vishwakt:
### Changes

Fixes the dimension limit check in `setDefaultDimensions()` to account for **all** dimension types (default + regular + dimension sets), not just existing defaults.

Previously, `setDefaultDimensions()` only checked `Object.keys(currentDefaultDimensions).length + Object.keys(newDimensions).length`, completely ignoring regular dimensions added via `addDimension()` and dimension sets from `addDimensions()`. This allowed silently exceeding CloudWatch's 29-dimension limit, causing metrics to be dropped.

#### What changed

- **`Metrics.ts`**: Replaced the narrow default-only count with `getDimensionCount()` (which correctly sums all dimension types). Also filters for genuinely new keys to avoid double-counting when overriding existing defaults.
- **`dimensions.test.ts`**: Added two tests:
  - Verifies `setDefaultDimensions()` throws when regular dimensions already fill the limit
  - Verifies overriding existing default keys doesn't trigger a false limit violation

**Issue number:** Closes #5206

---

### DIFF
--- a/packages/metrics/tests/unit/dimensions.test.ts
+++ b/packages/metrics/tests/unit/dimensions.test.ts
@@ -361,6 +361,42 @@ describe('Working with dimensions', () => {
     );
   });
 
+  it('throws when setDefaultDimensions would exceed the limit with existing regular dimensions', () => {
+    // Prepare
+    const metrics = new Metrics({
+      singleMetric: true,
+    });
+
+    // Act
+    for (let i = 1; i < MAX_DIMENSION_COUNT - 1; i++) {
+      metrics.addDimension(`regular-${i}`, 'test');
+    }
+
+    // Assess
+    expect(() =>
+      metrics.setDefaultDimensions({ 'new-default': 'test' })
+    ).toThrow(
+      `The number of metric dimensions must be lower than ${MAX_DIMENSION_COUNT}`
+    );
+  });
+
+  it('allows overriding existing default dimension keys without triggering the limit', () => {
+    // Prepare
+    const metrics = new Metrics({
+      singleMetric: true,
+    });
+
+    // Act
+    for (let i = 1; i < MAX_DIMENSION_COUNT - 1; i++) {
+      metrics.setDefaultDimensions({ [`dimension-${i}`]: 'test' });
+    }
+
+    // Assess
+    expect(() =>
+      metrics.setDefaultDimensions({ 'dimension-1': 'updated' })
+    ).not.toThrow();
+  });
+
   it('throws when adding dimension sets would exceed the limit', () => {
     // Prepare
     const metrics = new Metrics({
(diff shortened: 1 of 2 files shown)


packages/metrics/tests/unit/dimensions.test.ts
```diff
@@ -361,6 +361,44 @@ describe('Working with dimensions', () => {
...
   });
 
+  it('throws when setDefaultDimensions would exceed the limit with existing regular dimensions', () => {
+    // Prepare
+    const metrics = new Metrics({
+      singleMetric: true,
+    });
+
+    // Act
+    // Add regular dimensions until near the limit (service is already 1 default)
```
svozza: Let's remove these all these comments and just stick with //Act, //Preoare etc

vishwakt: Updated

svozza (approved):

svozza:
Thank you @vishwakt, merged!

PR #5229, opened 2026-05-08
fix(metrics): enforce dimension limits per-array

### LINKED ISSUE #5205: Bug: setDefaultDimensions() double-counts overlapping keys, rejecting valid updates near the limit
## Expected Behavior

Calling `setDefaultDimensions()` with keys that already exist in the current default dimensions should be allowed, since it's updating values — not increasing the total dimension count. The limit check should be based on the final merged count, not the naive sum of current + new.

## Current Behavior

`setDefaultDimensions()` at `Metrics.ts:796-813` sums the current default dimension count and the new dimension count without deducting overlapping keys:

```ts
const currentCount = Object.keys(currentDefaultDimensions).length;
const newSetCount = Object.keys(newDimensions).length;
  throw new RangeError(...);
}
```

If you have 21 default dimensions (including `service`) and call `setDefaultDimensions()` with the same 20 keys to update their values, it calculates `21 + 20 = 41 >= 29` and throws — even though the final count would still be 21.

## Code snippet

```typescript

const defaultDimensions: Record<string, string> = {};
for (let i = 0; i < 20; i++) {
  defaultDimensions[`dim-${i}`] = 'value';
}

const metrics = new Metrics({
  namespace: 'test',
  defaultDimensions,
});

// ❌ Throws RangeError even though we're just updating existing keys
const updatedDimensions: Record<string, string> = {};
for (let i = 0; i < 20; i++) {
  updatedDimensions[`dim-${i}`] = 'new-value';
}
metrics.setDefaultDimensions(updatedDimensions);
// RangeError: The number of metric dimensions must be lower than 29
```

## Steps to Reproduce

1. Create a `Metrics` instance with 20 default dimensions (+ `service` = 21 total)
2. Call `setDefaultDimensions()` with the same 20 keys but different values
3. Observe `RangeError` even though the final merged count would still be 21

## Possible Solution

Calculate the final merged count instead of naively summing:

```ts
const merged = { ...currentDefaultDimensions, ...newDimensions };
if (Object.keys(merged).length >= MAX_DIMENSION_COUNT) {
  throw new RangeError(...);
}
```

2.33.0

22.x

## Packaging format used

npm

---

vishwakt:
### Changes

This PR fundamentally fixes how dimension limits (`MAX_DIMENSION_COUNT`) are enforced across the `Metrics` class. It resolves the issue where valid dimension updates were incorrectly rejected near the limit due to overlapping keys being double-counted.

**What is this PR solving?**
Previously, limit checks relied on `this.#dimensionsStore.getDimensionCount()`, which naively summed the sizes of `dimensions`, `defaultDimensions`, and **all** `dimensionSets`. This approach was flawed for a few reasons:
1. **CloudWatch EMF Constraints:** CloudWatch publishes each `dimensionSet` as a completely separate array of dimensions. The CloudWatch limit of 30 dimensions applies to the maximum size of *any individual array*, not the sum of all dimensions defined across all sets. 
2. **Double Counting:** Using simple arithmetic (`currentCount + newCount`) meant that updating an existing key would incorrectly inflate the dimension count, causing a `RangeError` (issue #5205).
3. **Off-by-one Inconsistencies:** `addDimension` allowed 29 dimensions, whereas `addDimensions` and `setDefaultDimensions` only allowed 28 due to a `>=` vs `>` discrepancy.

**Fix Details:**
- **Replaced naive sums with projected array sizes:** We now use Javascript `Set` to calculate the exact maximum size of the specific dimension arrays that will actually be published to CloudWatch EMF.
- **Natively Handles Updates:** Because `Set` automatically deduplicates overlapping keys, updating existing dimensions inherently evaluates to a size of `0` new dimensions—fixing #5205 automatically.
- **Fixed off-by-one bug:** The limit checks now uniformly use `projectedSize > MAX_DIMENSION_COUNT` so exactly 29 custom dimensions are permitted across all methods (accommodating the 30th reserved `service` dimension).
- **Added comprehensive edge-case tests:** Added tests proving that overlapping keys at the limit threshold behave correctly, and added a regression test proving that adding multiple valid dimension sets no longer incorrectly hits the global dimension limit.

**Issue number:** closes #5205

---

### DIFF
--- a/packages/metrics/src/Metrics.ts
+++ b/packages/metrics/src/Metrics.ts
@@ -247,13 +247,18 @@ class Metrics extends Utility implements MetricsInterface {
       );
       return this;
     }
-    if (MAX_DIMENSION_COUNT <= this.#dimensionsStore.getDimensionCount()) {
+    const dimensions = this.#dimensionsStore.getDimensions();
+    const defaultDimensions = this.#dimensionsStore.getDefaultDimensions();
+    const projectedSize = new Set([
+      ...Object.keys(defaultDimensions),
+      ...Object.keys(dimensions),
+      name,
+    ]).size;
+    if (projectedSize > MAX_DIMENSION_COUNT) {
       throw new RangeError(
         `The number of metric dimensions must be lower than ${MAX_DIMENSION_COUNT}`
       );
     }
-    const dimensions = this.#dimensionsStore.getDimensions();
-    const defaultDimensions = this.#dimensionsStore.getDefaultDimensions();
     if (
       Object.hasOwn(dimensions, name) ||
       Object.hasOwn(defaultDimensions, name)
@@ -280,9 +285,12 @@ class Metrics extends Utility implements MetricsInterface {
    */
   public addDimensions(dimensions: Dimensions): this {
     const newDimensions = this.#sanitizeDimensions(dimensions);
-    const currentCount = this.#dimensionsStore.getDimensionCount();
-    const newSetCount = Object.keys(newDimensions).length;
-    if (currentCount + newSetCount >= MAX_DIMENSION_COUNT) {
+    const defaultDimensions = this.#dimensionsStore.getDefaultDimensions();
+    const projectedSize = new Set([
+      ...Object.keys(defaultDimensions),
+      ...Object.keys(newDimensions),
+    ]).size;
+    if (projectedSize > MAX_DIMENSION_COUNT) {
       throw new RangeError(
         `The number of metric dimensions must be lower than ${MAX_DIMENSION_COUNT}`
       );
@@ -798,13 +806,29 @@ class Metrics extends Utility implements MetricsInterface {
     const newDimensions = this.#sanitizeDimensions(dimensions);
     const currentDefaultDimensions =
       this.#dimensionsStore.getDefaultDimensions();
-    const newKeysCount = Object.keys(newDimensions).filter(
-      (key) => !Object.hasOwn(currentDefaultDimensions, key)
-    ).length;
-    if (
-      this.#dimensionsStore.getDimensionCount() + newKeysCount >=
-      MAX_DIMENSION_COUNT
-    ) {
+    const currentDimensions = this.#dimensionsStore.getDimensions();
+    const dimensionSets = this.#dimensionsStore.getDimensionSets();
+
+    const combinedDefaultKeys = [
+      ...Object.keys(currentDefaultDimensions),
+      ...Object.keys(newDimensions),
+    ];
+    let maxProjectedSize = new Set([
+      ...combinedDefaultKeys,
+      ...Object.keys(currentDimensions),
+    ]).size;
+
+    for (const dimensionSet of dimensionSets) {
+      const setSize = new Set([
+        ...combinedDefaultKeys,
+        ...Object.keys(dimensionSet),
+      ]).size;
+      if (setSize > maxProjectedSize) {
+        maxProjectedSize = setSize;
+      }
+    }
+
+    if (maxProjectedSize > MAX_DIMENSION_COUNT) {
       throw new RangeError(
         `The number of metric dimensions must be lower than ${MAX_DIMENSION_COUNT}`
       );
--- a/packages/metrics/src/Metrics.ts
+++ b/packages/metrics/src/Metrics.ts
@@ -249,11 +249,15 @@ class Metrics extends Utility implements MetricsInterface {
     }
     const dimensions = this.#dimensionsStore.getDimensions();
     const defaultDimensions = this.#dimensionsStore.getDefaultDimensions();
+    // addDimension only mutates the regular dimensions array, so we don't need to project against dimensionSets
+    // (each set is an independent published array).
     const projectedSize = new Set([
       ...Object.keys(defaultDimensions),
       ...Object.keys(dimensions),
       name,
     ]).size;
+    // MAX_DIMENSION_COUNT is 29 (EMF 30 dimension cap - 1 reserved for service).
+    // Thus exactly 29 custom dimensions are allowed, and we throw only if strictly greater.
     if (projectedSize > MAX_DIMENSION_COUNT) {
       throw new RangeError(
         `The number of metric dimensions must be lower than ${MAX_DIMENSION_COUNT}`
@@ -286,6 +290,7 @@ class Metrics extends Utility implements MetricsInterface {
   public addDimensions(dimensions: Dimensions): this {
     const newDimensions = this.#sanitizeDimensions(dimensions);
     const defaultDimensions = this.#dimensionsStore.getDefaultDimensions();
+    // addDimensions creates a new independent dimensionSet array, so we only project against defaultDimensions.
     const projectedSize = new Set([
       ...Object.keys(defaultDimensions),
       ...Object.keys(newDimensions),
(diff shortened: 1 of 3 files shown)


packages/metrics/tests/unit/dimensions.test.ts
```diff
@@ -407,20 +463,16 @@ describe('Working with dimensions', () => {
     });
 
     // Act
+    const newDimensionSet: Record<string, string> = {};
     // We start with 2 dimensions because the default dimension & service name are already added
-    for (let i = 2; i < MAX_DIMENSION_COUNT; i++) {
-      metrics.addDimension(`dimension-${i}`, 'test');
+    // We need 28 more to exceed the limit of 29 (2 + 28 = 30 > 29)
```
svozza: I don't think we need this comment

packages/metrics/src/Metrics.ts
```diff
@@ -280,9 +285,12 @@ class Metrics extends Utility implements MetricsInterface {
...
     const newDimensions = this.#sanitizeDimensions(dimensions);
-    const currentCount = this.#dimensionsStore.getDimensionCount();
-    const newSetCount = Object.keys(newDimensions).length;
-    if (currentCount + newSetCount >= MAX_DIMENSION_COUNT) {
+    const defaultDimensions = this.#dimensionsStore.getDefaultDimensions();
+    const projectedSize = new Set([
+      ...Object.keys(defaultDimensions),
+      ...Object.keys(newDimensions),
+    ]).size;
+    if (projectedSize > MAX_DIMENSION_COUNT) {
```
svozza, line 298: How come we're chsnging this from `>=` to `>`?

packages/metrics/src/Metrics.ts
```diff
@@ -247,13 +247,18 @@ class Metrics extends Utility implements MetricsInterface {
...
-    if (MAX_DIMENSION_COUNT <= this.#dimensionsStore.getDimensionCount()) {
+    const dimensions = this.#dimensionsStore.getDimensions();
+    const defaultDimensions = this.#dimensionsStore.getDefaultDimensions();
+    const projectedSize = new Set([
+      ...Object.keys(defaultDimensions),
+      ...Object.keys(dimensions),
+      name,
+    ]).size;
+    if (projectedSize > MAX_DIMENSION_COUNT) {
       throw new RangeError(
```
dreamorosi, line 262: Worth a short comment here noting that `addDimension` only mutates the regular-dimensions array, so we don't need to project against `dimensionSets` (each set is an independent published array). Same applies to `addDimensions` below. Without this context, a future reviewer is likely to ask "shouldn't we also check all sets?".

packages/metrics/src/Metrics.ts
```diff
@@ -798,13 +806,29 @@ class Metrics extends Utility implements MetricsInterface {
...
+      const setSize = new Set([
+        ...combinedDefaultKeys,
+        ...Object.keys(dimensionSet),
+      ]).size;
+      if (setSize > maxProjectedSize) {
+        maxProjectedSize = setSize;
+      }
+    }
+
+    if (maxProjectedSize > MAX_DIMENSION_COUNT) {
```
dreamorosi, line 839: Minor: when `currentDimensions` is empty the main array isn't actually emitted (`Metrics.ts:729` gates on `length > 0`), so the initial `maxProjectedSize` includes a phantom array. It's harmless because the per-set loop will dominate when relevant, but a comment or a small guard would make the intent clearer.

packages/metrics/tests/unit/dimensions.test.ts
```diff
@@ -397,6 +418,41 @@ describe('Working with dimensions', () => {
...
+      metrics.addDimensions({
+        'dimension-1': 'updated',
+        'dimension-2': 'updated',
+      })
+    ).not.toThrow();
+  });
+
   it('throws when adding dimension sets would exceed the limit', () => {
     // Prepare
     const metrics = new Metrics({
```
dreamorosi, line 458: Nit: the test title says "via addDimensions" but the body uses `setDefaultDimensions` to populate, then calls `addDimensions` to override. The assertion is valid, but the title could be clearer — e.g. "allows addDimensions to override existing default dimension keys without triggering the limit".

dreamorosi (commented):
Validated the approach against the [EMF specification](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format_Specification.html#CloudWatch_Embedded_Metric_Format_Specification_structure_dimensionset): the spec defines the 30-dimension cap *per `DimensionSet`*, not as an aggregate across all sets, and the outer `Dimensions` array has no upper bound in the schema. The new `Set`-based projected-size check correctly mirrors what `serializeMetrics` actually emits (each set merged with `defaultDimensions` as its own array, see `Metrics.ts:733-738`), which makes the validation match runtime behaviour. The off-by-one unification on `projectedSize > MAX_DIMENSION_COUNT` is also a nice cleanup. Left a few minor observations inline — none are blockers.

packages/metrics/tests/unit/dimensions.test.ts
vishwakt: Done. I've removed this block comment as the math is straightforward and self-explanatory in the context of the test.

packages/metrics/src/Metrics.ts
vishwakt: I updated it to `>` to align with the 30-dimension CloudWatch EMF limit. Since 1 slot is inherently reserved for the `service` dimension, we can support exactly 29 custom dimensions. Throwing on `projectedSize > MAX_DIMENSION_COUNT` (where `MAX_DIMENSION_COUNT = 29`) correctly allows exactly 29 custom dimensions, whereas the previous `>=` check artificially capped us at 28.

Also added a comment to make this clear for future readers

packages/metrics/src/Metrics.ts
vishwakt: I've added a short explanatory comment right above the `projectedSize` logic in both `addDimension` and `addDimensions`.

packages/metrics/src/Metrics.ts
vishwakt: Good catch. I added a guard to check `currentDimensionsKeys.length > 0` so that `maxProjectedSize` accurately reflects what's emitted at runtime (since the main array isn't emitted if it's empty, per `Metrics.ts:729`)

Adding this guard also exposed a tiny bug where `combinedDefaultKeys` wasn't naturally deduplicated if the main array was skipped, so I made sure to wrap it in a `new Set(combinedDefaultKeys).size` as a fallback to ensure it's fully bulletproof

packages/metrics/tests/unit/dimensions.test.ts
vishwakt: Done. I've updated the test title

svozza (dismissed):

svozza:
@vishwakt looks like we're getting a CI failure due to a drop in code coverage. My guess is probably because the new guard you added doesn't have a test.

vishwakt:
@svozza the drop in coverage was actually a direct result of the recent refactor

the `getDimensionCount()` method in `DimensionsStore` became dead code which i've now removed. I've also added another test case for a boundary condition

svozza (approved):

svozza:
Thank you for your patience @vishwakt, this looks good now. Approved!

PR #5240, opened 2026-05-11
fix(metrics): detect key collisions between dimensions, metrics, and metadata

### LINKED ISSUE #5208: Bug: No collision detection between dimension keys, metric names, and metadata keys in EMF output
## Expected Behavior

The library should prevent collisions between dimension keys, metric names, and metadata keys since they all share the same top-level namespace in the EMF JSON. Per the [EMF spec](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format_Specification.html), dimension targets MUST be strings and metric targets MUST be numbers — they cannot coexist under the same key.

## Current Behavior

In `serializeMetrics()` at `Metrics.ts:747-769`, the EMF output is constructed by spreading objects in this order:

```ts
return {
  _aws: { ... },
  ...defaultDimensions,     // 1st
  ...dimensions,            // 2nd
  ...dimensionSetsFlat,     // 3rd
  ...metricValues,          // 4th — overwrites dimensions
  ...this.#metadataStore.getAll(),  // 5th — overwrites metrics
};
```

Since these keys share the same namespace, collisions can silently corrupt the output:
- A metric named `environment` overwrites a dimension `environment: 'prod'` with a number — CloudWatch silently loses the dimension.
- Metadata with key `request_count` overwrites a metric `request_count: 42` with a string — CloudWatch fails to parse the metric.

## Code snippet

```typescript

const metrics = new Metrics({ namespace: 'test' });

// Problem 1: metric overwrites dimension
metrics.addDimension('environment', 'prod');
metrics.addMetric('environment', MetricUnit.Count, 1);
const output1 = metrics.serializeMetrics();
console.log(output1.environment);
// ❌ Outputs: 1 (number) — the dimension string 'prod' was silently overwritten

// Problem 2: metadata overwrites metric
const metrics2 = new Metrics({ namespace: 'test' });
metrics2.addMetric('request_count', MetricUnit.Count, 42);
metrics2.addMetadata('request_count', 'not-a-number');
const output2 = metrics2.serializeMetrics();
console.log(output2.request_count);
// ❌ Outputs: 'not-a-number' (string) — the numeric metric value 42 was overwritten
```

## Steps to Reproduce

**Metric overwrites dimension:**
1. Create a `Metrics` instance
2. Add a dimension `environment: 'prod'`
3. Add a metric named `environment`
4. Call `serializeMetrics()` — `output.environment` is `1`, not `'prod'`

**Metadata overwrites metric:**
1. Create a `Metrics` instance
2. Add a metric `request_count` with value `42`
3. Add metadata with key `request_count`
4. Call `serializeMetrics()` — `output.request_count` is `'not-a-number'`, not `42`

## Possible Solution

Maintain a `Set` of reserved keys and check against it in `storeMetric()` and `addMetadata()`:

```ts
// In storeMetric():
const dimensionKeys = new Set([
  ...Object.keys(this.#dimensionsStore.getDimensions()),
  ...Object.keys(this.#dimensionsStore.getDefaultDimensions()),
]);
if (dimensionKeys.has(name)) {
  throw new Error(`Metric name "${name}" conflicts with an existing dimension key`);
}

// In addMetadata():
if (this.#metricsStore.getMetric(key) !== undefined) {
  throw new Error(`Metadata key "${key}" conflicts with an existing metric name`);
}
```

2.33.0

22.x

## Packaging format used

npm

---

Zelys-DFKH:
## Summary

Closes #5208. \`serializeMetrics()\` builds EMF output by spreading multiple key sources in a fixed order: metadata → defaultDimensions → dimensions → dimensionSets → metricValues. When a key appears in more than one source, the later value wins silently. Two of those overwrites break the EMF schema:

- A metric name matching a string key (dimension or metadata) replaces the string value with a number, so CloudWatch drops the dimension or fails to parse the metric.
- A string key matching a metric name replaces a number with a string, causing the same failure in the opposite call order.

The bug is call-order-dependent. A guard at the call site of \`addMetric()\` or \`addDimension()\` can only catch one direction; the fix belongs at \`serializeMetrics()\`, where all sources are present before the spread.

#### What changed

- **\`Metrics.ts\`**: Added \`#checkKeyCollisions()\`, called from \`serializeMetrics()\` before the spread. Seeds string keys in spread-precedence order (metadata lowest, dimension set highest), warns when a key appears in two string sources, throws when a metric name matches any string key.
- **\`types/Metrics.ts\`** and **\`types/index.ts\`**: Added \`EmfKeySource\` union type for source labels in error and warning messages.
- **\`dimensions.test.ts\`**: Eight new tests: five for metric/dimension name collisions (regular dimension, default dimension, built-in \`service\` dimension, via \`addDimensions\`, and reverse call order) and three for dimension-on-dimension overwrites that should warn.
- **\`metadata.test.ts\`**: Six new tests: two for metadata/metric name collisions in both call orderings and four for metadata-on-dimension overwrites that should warn.

**Issue number:** Closes #5208

---


svozza:
Thanks for tackling this! Two pieces of feedback:

**1. Only throw on type changes**

Same-type collisions (dimension ↔ metadata, both strings) overwrite silently but don't corrupt the EMF schema — CloudWatch only rejects when a key flips between string and number. Throwing on same-type clashes is stricter than the spec requires and inconsistent with `addDimension`, which already handles dimension-on-dimension overwrites with a warn. I'd keep the throw for type changes only and leave same-type collisions on the existing warn path.

**2. Centralise the check in `serializeMetrics()`**

The current guards in `storeMetric()` and `addMetadata()` only catch two of the four type-change directions. They miss:
- `addDimension`/`addDimensions`/`setDefaultDimensions` after `addMetric` (num → str)
- `addMetric` after `addMetadata` (str → num)

Reverse the call order in either of the existing tests and the corruption comes back. Rather than adding guards to every mutator, `serializeMetrics()` already has all the sources in scope right before the spread that causes the overwrite — one pass there catches every combination regardless of call order:

```ts
// inside serializeMetrics(), just before the return
type Source = 'dimension' | 'default dimension' | 'dimension set' | 'metadata';
const stringKeys = new Map<string, Source>();
for (const k of Object.keys(defaultDimensions)) stringKeys.set(k, 'default dimension');
for (const k of Object.keys(dimensions)) stringKeys.set(k, 'dimension');
for (const set of dimensionSets) for (const k of Object.keys(set)) stringKeys.set(k, 'dimension set');
for (const k of Object.keys(this.#metadataStore.getAll())) stringKeys.set(k, 'metadata');

for (const name of Object.keys(metricValues)) {
  const source = stringKeys.get(name);
  if (source) {
    throw new Error(
      `EMF key collision on "${name}": registered as both a metric (number) and a ${source} (string)`
    );
  }
}
```

svozza:
Hi @Zelys-DFKH, do you need any help with this PR? Anything you need me to clarify from my comment?

Zelys-DFKH:
Sorry for the slow response, I missed the notification. Both points were right.

Pushed a second commit that removes the guards from `storeMetric()` and `addMetadata()` and moves the check into `serializeMetrics()`, where all four sources (default dimensions, dimensions, dimension sets, metadata) are in scope before the spread. The throw is scoped to number-vs-string only, so same-type string overwrites stay on the existing warn path. Tests cover both call orderings for the metadata case and the reverse-order dimension case.

Thanks for the ping.

svozza:
No worries! Couple of comments, the way the code is now, we only warn on dimesnions overwriting each other but we should warn whenever any keys are overwritten. Secondly, there are merge conflicts in the test file that need to be resolved. Also, the `serializeMetrics` method has become quite long, you can see there is now a warning about the cyclomatic complexity so we may want to pull this new validation logic into a private method.

packages/metrics/src/Metrics.ts
```diff
@@ -792,6 +778,46 @@ class Metrics extends Utility implements MetricsInterface {
...
   }
 
+  #checkKeyCollisions(
+    defaultDimensions: Dimensions,
+    dimensions: Dimensions,
+    dimensionSets: Dimensions[],
+    metricValues: Record<string, number | number[]>,
+    metadata: Record<string, string>
+  ): void {
+    type Source =
```
svozza: Let's not inline types, there's a types module for this.

packages/metrics/src/Metrics.ts
```diff
@@ -792,6 +778,46 @@ class Metrics extends Utility implements MetricsInterface {
...
+    for (const k of Object.keys(defaultDimensions))
+      seenKeys.set(k, 'default dimension');
+    for (const k of Object.keys(dimensions)) seenKeys.set(k, 'dimension');
+    for (const set of dimensionSets)
+      for (const k of Object.keys(set)) seenKeys.set(k, 'dimension set');
+
+    for (const k of Object.keys(metadata)) {
+      const src = seenKeys.get(k);
+      if (src !== undefined) {
+        this.#logger.warn(
```
svozza, line 810: Is there a way we could make it so dimension takes precedence?

Zelys-DFKH:
Thanks for the thorough review — pushed b2a61faf.

1. `setKey` now warns whenever any source overwrites another (dimension over default dimension, dimension set over dimension or default dimension), not just when metadata overlaps. The metric throw is unchanged.

2. Synced `dimensions.test.ts` with upstream's #5229 (loop bounds, `toThrow` vs `toThrowError`, the three new upstream tests) and added regression tests for the new warn paths.

3. The repeated `get/check/set` blocks are now a single `setKey` helper, which absorbed `warnOverwrite`. `serializeMetrics` is unchanged. The complexity was all in `#checkKeyCollisions`, which is now shorter.

Let me know if anything else needs a look!

svozza:
Left some comments, also looks like we still have merge conflicts

Zelys-DFKH:
Thanks for the careful review @svozza. Pushed 3e92ae89 with all three points addressed.

1. Moved `EmfKeySource` into `packages/metrics/src/types/Metrics.ts`, re-exported from `types/index.ts` and imported in `Metrics.ts`.

2. Dimensions take precedence over metadata. While I was in there, I noticed the spread order in `serializeMetrics` had metadata last, which was silently overriding dimensions in the output even though the warning said the dimension would win. Flipped the spread (metadata first, dimensions last) and reordered the seed in `#checkKeyCollisions` to match. Warning message and payload now agree.

3. Cleared the merge conflict by moving my collision tests into a new `packages/metrics/tests/unit/keyCollisions.test.ts`. `dimensions.test.ts` is now byte-identical to upstream/main, so GitHub's 3-way merge resolves cleanly. Also dropped `DimensionsStore.getDimensionCount()` since #5229 stopped using it.

All checks green.

packages/metrics/tests/unit/keyCollisions.test.ts
```diff
@@ -0,0 +1,138 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { Metrics, MetricUnit } from '../../src/index.js';
+
+describe('EMF key collision detection', () => {
+  const ENVIRONMENT_VARIABLES = process.env;
```
svozza: Let's use the built in Vitest environment variable functions here: https://vitest.dev/api/vi.html#vi-stubenv.

packages/metrics/tests/unit/keyCollisions.test.ts
```diff
@@ -0,0 +1,138 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
```
svozza: Am I missing something, these tests look incredibile similar to the ones in `metadata.test.ts`?

packages/metrics/tests/unit/keyCollisions.test.ts
Zelys-DFKH: Deleted `keyCollisions.test.ts`. The 8 tests are now in `dimensions.test.ts`, which uses the same env setup.

packages/metrics/tests/unit/keyCollisions.test.ts
Zelys-DFKH: The tests weren't duplicating `metadata.test.ts` (those cover metadata/metric collisions, these cover dimension/metric), but having a separate file for it was the real issue. Moved all 8 into `dimensions.test.ts` and deleted `keyCollisions.test.ts`. Dimension/metric and dimension/dimension collision tests live there now; `metadata.test.ts` owns the metadata side.

svozza:
Thank you for your work on this @Zelys-DFKH, everything looks good now. Merging!

svozza (approved):

Zelys-DFKH:
Thanks @svozza. The centralized-check shape was definitely a better design than the per-mutator guards, and the spread-order catch in R3 saved a follow-up PR.

PR #5255, opened 2026-05-14
fix(event-handler): normalize trailing slashes in HTTP Router prefix joining

### LINKED ISSUE #5252: Bug: event-handler HTTP Router prefix joining drops trailing slash and produces double slashes
### Expected Behavior

When constructing `new Router({ prefix: '/api' })` and registering a root route with `app.get('/', handler)`, the route should be reachable at the prefix path (`/api` and/or `/api/`). When the prefix accidentally ends with `/`, joining a child path should not produce a double slash.

### Current Behavior

`resolvePrefixedPath` (`packages/event-handler/src/http/utils.ts:334-346`) ends with:

```ts
return `${prefix}${path}`.replace(/\/$/, '') as Path;
```

Two concrete failures:

1. `resolvePrefixedPath('/', '/api')` returns `'/api'` — the trailing slash is stripped. A handler registered at `/` under `/api` is unreachable from `/api/`; the request resolves to 404.
2. `resolvePrefixedPath('/users', '/api/')` returns `'/api//users'` — a literal double slash. Incoming requests to `/api/users` don't match the registered route id.

### Code snippet

```typescript

const app = new Router({ prefix: '/api' });
app.get('/', () => ({ root: true }));

// Request to /api/ → 404 Not Found (expected: 200 with { root: true })

const app2 = new Router({ prefix: '/api/' });
app2.get('/users', () => ({ users: [] }));

// Internally registers GET:/api//users; request to /api/users → 404
```

### Steps to Reproduce

1. Construct a `Router` with `prefix: '/api'` and register a handler at `'/'`.
2. Resolve a request for `/api/`.
3. Observe `statusCode: 404`.
4. Repeat with `prefix: '/api/'` and `path: '/users'`; observe the registered route id contains `//`.
5. Or run the failing test: `npx vitest --run tests/unit/http/bugHunt.test.ts -t "Bug 5"` from `packages/event-handler/`.

### Possible Solution

_No response_

latest

22.x

### Packaging format used

npm

### Execution logs

_(not applicable — bug is reproducible from a unit test)_

svozza:
## Summary

`Router` registered route ids did not handle trailing slashes consistently: a root route under a prefix (e.g. `app.get('/', …)` with `prefix: '/api'`) was unreachable from `/api/`, and a prefix that accidentally ended with `/` produced double slashes in the route id (e.g. `'/api//users'`), breaking matches.

### Changes

- `resolvePrefixedPath` now strips trailing slashes from the prefix before joining, so `'/api'` and `'/api/'` produce the same result and no `//` appears in the route id.
- `Router.resolve` normalizes incoming request paths the same way (preserving `/` itself), so a route registered as `/` under prefix `/api` matches both `/api` and `/api/`.
- Added unit cases for `resolvePrefixedPath` covering prefix-with-trailing-slash inputs, and integration tests in `basic-routing.test.ts` for both bug scenarios from the issue.
- Updated the `resolvePrefixedPath` JSDoc to document the normalization preconditions.

**Issue number:** closes #5252

---

### DIFF
--- a/packages/event-handler/src/http/utils.ts
+++ b/packages/event-handler/src/http/utils.ts
@@ -339,15 +339,24 @@ export const composeMiddleware = (middleware: Middleware[]): Middleware => {
  * @param path - The path to resolve
  * @param prefix - The prefix to prepend to the path; trailing slashes are ignored
  */
+// Linear-time trailing-slash strip. Avoids `replace(/\/+$/, '')` to keep
+// behavior linear on attacker-controlled request paths.
+export const stripTrailingSlashes = (input: string): string => {
+  let end = input.length;
+  while (end > 0 && input[end - 1] === '/') end--;
+  return end === input.length ? input : input.slice(0, end);
+};
+
 export const resolvePrefixedPath = (path: Path, prefix?: Path): Path => {
   if (!prefix) return path;
-  const prefixStr = getPathString(prefix).replace(/\/+$/, '');
+  const prefixStr = stripTrailingSlashes(getPathString(prefix));
   if (isRegExp(prefix) || isRegExp(path)) {
     const pathStr = getPathString(path);
     const sep = pathStr.startsWith('/') ? '' : '/';
     return new RegExp(`${prefixStr}${sep}${pathStr}`);
   }
-  return `${prefixStr}${path}`.replace(/\/$/, '') as Path;
+  const joined = `${prefixStr}${path}`;
+  return (joined.endsWith('/') ? joined.slice(0, -1) : joined) as Path;
 };
 
 export const HttpResponseStream =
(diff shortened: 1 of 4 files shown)


packages/event-handler/src/http/utils.ts
```diff
@@ -322,27 +322,41 @@ export const composeMiddleware = (middleware: Middleware[]): Middleware => {
   };
 };
 
+// Linear-time trailing-slash strip. Avoids `replace(/\/+$/, '')` to keep
```
dreamorosi: Multi line comments should have a different format

dreamorosi (dismissed):

packages/event-handler/src/http/utils.ts
svozza: Fixed!

dreamorosi (approved):

PR #5260, opened 2026-05-15
fix(event-handler): allow PATCH request body in ALB converter

### LINKED ISSUE #5258: Bug: event-handler HTTP ALB converter silently drops PATCH request bodies
### Expected Behavior

A `PATCH` request body received via an Application Load Balancer event should be readable from `req` inside a route handler (e.g. `await req.json()`).

### Current Behavior

`albEventToWebRequest` in `packages/event-handler/src/http/converters.ts:180-184` discards bodies for both `GET` and `PATCH`:

```ts
// ALB events represent GET and PATCH request bodies as empty strings
const body =
  httpMethod === HttpVerbs.GET || httpMethod === HttpVerbs.PATCH
    ? null
    : createBody(event.body ?? null, event.isBase64Encoded);
```

The comment is incorrect (verified against a live ALB → Lambda integration): per [RFC 5789](https://datatracker.ietf.org/doc/html/rfc5789), `PATCH` is body-bearing and routinely carries bodies through ALB. Handlers calling `await req.json()` for `PATCH` on ALB silently receive `null`.

The [AWS ALB Lambda docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#receive-event-from-load-balancer) describe a generic event payload with a `body` field and don't carve out PATCH as special.

### Code snippet

```typescript

const app = new Router();
app.patch('/items/:id', async ({ req }) => {
  const data = await req.json(); // null on ALB ❌
  return { updated: data };
});

```

### Steps to Reproduce

1. Register a `PATCH` handler that consumes the request body.
2. Invoke with an ALB event containing a non-empty `body`.
3. Observe the body inside the handler is `null`.

### Possible Solution

Drop `PATCH` from the bodyless list. Per [RFC 9110 §9.3.1](https://datatracker.ietf.org/doc/html/rfc9110#section-9.3.1) and [§9.3.2](https://datatracker.ietf.org/doc/html/rfc9110#section-9.3.2), only `GET` and `HEAD` payloads have no defined semantics:

```ts
const body =
  httpMethod === HttpVerbs.GET || httpMethod === HttpVerbs.HEAD
    ? null
    : createBody(event.body ?? null, event.isBase64Encoded);
```

latest

22.x

### Packaging format used

npm

enaktes9-hub:
## Summary

Fixes #5258 — the ALB event-to-web-request converter was silently dropping `PATCH` request bodies.

## Problem

In `packages/event-handler/src/http/converters.ts`, the `albEventToWebRequest` function treated `PATCH` as a body-less HTTP method alongside `GET`:

```ts
const body =
  httpMethod === HttpVerbs.GET || httpMethod === HttpVerbs.PATCH
    ? null
    : createBody(event.body ?? null, event.isBase64Encoded);
```

Per [RFC 5789](https://datatracker.ietf.org/doc/html/rfc5789) and [RFC 9110 §9.3](https://datatracker.ietf.org/doc/html/rfc9110#section-9.3), only `GET` and `HEAD` have no defined body semantics. `PATCH` routinely carries request bodies, and AWS ALB correctly forwards them (verified against live ALB → Lambda integration). The comment was also incorrect — ALB does forward PATCH bodies.

## Changes

- Remove `PATCH` from the body-less list in `albEventToWebRequest`
- Add `HEAD` (the actually body-less method alongside `GET`, per RFC 9110)
- Update the comment to accurately describe the behavior

## Verification

- `PATCH` handlers using `await req.json()` will now receive the request body
- `HEAD` requests (which have no body semantics per RFC 9110) continue to get `null`
- No other HTTP methods are affected
- All existing tests should continue to pass

Closes #5258

### DIFF
--- a/packages/event-handler/src/http/converters.ts
+++ b/packages/event-handler/src/http/converters.ts
@@ -182,9 +182,9 @@ const albEventToWebRequest = (event: ALBEvent): Request => {
   const url = new URL(path, `${protocol}://${hostname}/`);
   populateV1QueryParams(url, event);
 
-  // ALB events represent GET and PATCH request bodies as empty strings
+  // ALB events represent GET and HEAD request bodies as empty strings
   const body =
-    httpMethod === HttpVerbs.GET || httpMethod === HttpVerbs.PATCH
+    httpMethod === HttpVerbs.GET || httpMethod === HttpVerbs.HEAD
       ? null
       : createBody(event.body ?? null, event.isBase64Encoded);
 


svozza:
This has been implemented in #5259.

PR #5261, opened 2026-05-15
fix(event-handler): allow ALB PATCH requests to preserve body payloads

### LINKED ISSUE #5258: Bug: event-handler HTTP ALB converter silently drops PATCH request bodies

vishwakt:
### Changes

The ALB HTTP event converter (`albEventToWebRequest`) was incorrectly dropping the request body for `PATCH` requests by returning `null`. This violates RFC 5789, where `PATCH` requests are body-bearing and expected to carry a payload.

This PR fixes the issue by updating the condition in `albEventToWebRequest` to drop request bodies only for `GET` and `HEAD` requests (per RFC 9110), ensuring `PATCH` bodies are preserved.

- Swapped `HttpVerbs.PATCH` for `HttpVerbs.HEAD` in `packages/event-handler/src/http/converters.ts`.
- Updated `createAlbBody` in test `helpers.ts` to reflect the fix.
- Added tests verifying that ALB `PATCH` bodies are accurately parsed and ALB `HEAD` bodies are dropped.

**Issue number:** closes #5258

---

### DIFF
--- a/packages/event-handler/src/http/constants.ts
+++ b/packages/event-handler/src/http/constants.ts
@@ -116,6 +116,26 @@ const COMPRESSION_ENCODING_TYPES = {
   ANY: '*',
 } as const;
 
+/**
+ * Known headers that are defined as a comma-separated list of values (1#element).
+ * These headers can be safely split by commas to populate multiValueHeaders.
+ */
+const MULTI_VALUE_HEADERS_ALLOWLIST = new Set([
+  'accept',
+  'accept-encoding',
+  'accept-language',
+  'cache-control',
+  'vary',
+  'connection',
+  'allow',
+  'x-forwarded-for',
+  'te',
+  'expect',
+  'transfer-encoding',
+  'content-encoding',
+  'content-language',
+]);
+
 const HttpStatusText: Record<number, string> = {
   // 2xx Success
   200: 'OK',
@@ -192,6 +212,7 @@ export {
   HttpStatusCodes,
   HttpStatusText,
   HttpVerbs,
+  MULTI_VALUE_HEADERS_ALLOWLIST,
   PARAM_PATTERN,
   SAFE_CHARS,
   UNSAFE_CHARS,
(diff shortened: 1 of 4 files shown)


svozza:
This has been implemented in #5259.

PR #5267, opened 2026-05-15
fix(event-handler): allow cors preflight without request headers

### LINKED ISSUE #5263: Bug: event-handler HTTP cors() middleware rejects valid CORS preflight without Access-Control-Request-Headers
### Expected Behavior

Per the [Fetch Standard §4.8 \"CORS-preflight fetch\"](https://fetch.spec.whatwg.org/#cors-preflight-fetch), `Access-Control-Request-Headers` is only appended to a preflight when the request has CORS-unsafe headers (\"If `headers` is not empty, then ... Append (`Access-Control-Request-Headers`, value) to preflight's header list\"). A preflight with a valid `Access-Control-Request-Method` and no `Access-Control-Request-Headers` (the common case for `application/json` POSTs from same-origin-tooled UIs, etc.) should return `204 No Content` with the appropriate CORS headers.

### Current Behavior

`isValidPreflightRequest` short-circuits to `undefined` when `Access-Control-Request-Headers` is missing because of how the `?.split(...).every(...)` chain is composed:

```ts
return (
  accessControlRequestMethod &&
  allowedMethods.includes(accessControlRequestMethod) &&
  accessControlRequestHeaders?.split(',').every(...)   // undefined when header absent
);
```

The middleware then falls through to `await next()`. Because no `OPTIONS` route is registered for the path, the request resolves to `404 Not Found` instead of a CORS preflight `204`.

### Code snippet

```typescript
import { cors } from '@aws-lambda-powertools/event-handler/http/middleware';

const app = new Router();
app.use(cors({ origin: 'https://example.com' }));
app.get('/test', () => ({ ok: true }));

```

### Steps to Reproduce

1. Wire up `cors({ origin: 'https://example.com' })` and a `GET /test` route as above.
2. Issue a preflight request: method `OPTIONS`, headers `Origin: https://example.com` and `Access-Control-Request-Method: GET` (note: no `Access-Control-Request-Headers`).
3. Observe `statusCode: 404` instead of `204`.

### Possible Solution

Treat a missing `Access-Control-Request-Headers` as \"no headers requested\" → still valid:

```ts
return (
  accessControlRequestMethod &&
  allowedMethods.includes(accessControlRequestMethod) &&
  (!accessControlRequestHeaders ||
    accessControlRequestHeaders
      .split(',')
      .every((h) => allowedHeaders.includes(h.trim())))
);
```

latest

22.x

### Packaging format used

npm

vishwakt:
### Changes

The `cors` middleware incorrectly rejects valid CORS `OPTIONS` preflight requests if the `Access-Control-Request-Headers` header is missing. Per the Fetch Standard §4.8, `Access-Control-Request-Headers` is only required when the request includes CORS-unsafe headers. Preflight requests containing only `Access-Control-Request-Method` should be treated as valid and succeed with a `204 No Content`.

This PR fixes the issue by updating `isValidPreflightRequest` to explicitly accept requests where `Access-Control-Request-Headers` is missing or undefined.

- Allowed `accessControlRequestHeaders` to be undefined/empty in the validation logic inside `packages/event-handler/src/http/middleware/cors.ts`.
- Added a unit test covering valid preflight requests without request headers in `packages/event-handler/tests/unit/http/middleware/cors.test.ts`.

**Issue number:** closes #5263

---

### DIFF
--- a/packages/event-handler/tests/unit/http/middleware/cors.test.ts
+++ b/packages/event-handler/tests/unit/http/middleware/cors.test.ts
@@ -203,4 +203,40 @@ describe('CORS Middleware', () => {
       corsConfig.maxAge.toString()
     );
   });
+
+  it('handles OPTIONS preflight requests without Access-Control-Request-Headers', async () => {
+    // Prepare
+    const app = new Router();
+    const corsConfig = {
+      origin,
+      allowMethods: ['GET', 'POST'],
+      allowHeaders: ['Authorization', 'Content-Type'],
+      maxAge: 3600,
+    };
+    app.use(cors(corsConfig));
+
+    // Act
+    const result = await app.resolve(
+      createTestEvent('/test', 'OPTIONS', {
+        Origin: origin,
+        'Access-Control-Request-Method': 'GET',
+      }),
+      context
+    );
+
+    // Assess
+    expect(result.statusCode).toBe(204);
+    expect(result.headers?.['access-control-allow-origin']).toEqual(
+      corsConfig.origin
+    );
+    expect(result.multiValueHeaders?.['access-control-allow-methods']).toEqual(
+      corsConfig.allowMethods
+    );
+    expect(result.multiValueHeaders?.['access-control-allow-headers']).toEqual(
+      corsConfig.allowHeaders.map((header) => header.toLowerCase())
+    );
+    expect(result.headers?.['access-control-max-age']).toEqual(
+      corsConfig.maxAge.toString()
+    );
+  });
 });
--- a/packages/event-handler/tests/unit/http/middleware/cors.test.ts
+++ b/packages/event-handler/tests/unit/http/middleware/cors.test.ts
@@ -167,7 +167,16 @@ describe('CORS Middleware', () => {
     expect(result.headers?.['access-control-allow-origin']).toBeUndefined();
   });
 
-  it('handles OPTIONS preflight requests', async () => {
+  it.each([
+    [
+      'handles OPTIONS preflight requests',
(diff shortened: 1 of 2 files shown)


packages/event-handler/tests/unit/http/middleware/cors.test.ts
```diff
@@ -167,7 +167,16 @@ describe('CORS Middleware', () => {
     expect(result.headers?.['access-control-allow-origin']).toBeUndefined();
   });
 
-  it('handles OPTIONS preflight requests', async () => {
+  it.each([
```
svozza, line 170: Nit, I prefer to use the object syntax for `it.each` rather than tuples.

svozza (dismissed):

vishwakt:
@svozza I've updated the test to use the object syntax for `it.each` as suggested.

Just a heads up on the latest push: when switching to the object syntax, TypeScript struggled to infer the type of the spread headers (since one test case omitted the `Access-Control-Request-Headers` key, creating a union type). To fix the TS error without introducing unconventional explicit type arguments to `it.each`, I just updated the test objects to define the full `headers` payload directly. Should be good to go now!

svozza:
Hmm, looks like we're getting type errors in the tests, I'm ok to put it back to a tuple rather than have to jump through hoops to get a test to compile.

vishwakt:
I've switched it back to the tuple syntax as you preferred, and just casted the empty object as `Record<string, string>` to keep the compiler happy

svozza (approved):

svozza:
Good stuff. Merged!

PR #5295, opened 2026-05-26
fix(commons): clear LRUCache pointers on eviction of last item

### LINKED ISSUE #5290: Bug: LRUCache retains stale entry when maxSize=0 across multiple add() calls
### Expected Behavior

When an `LRUCache` is instantiated with `maxSize: 0`, the cache should never hold any entries. Calling `add()` any number of times should leave `size()` at `0`, and `get()` for any key should return `undefined`.

### Current Behavior

After the **first** `add()`, `size()` is `0` (correct), but after a **second** `add()`, the cache permanently retains one stale entry — `size()` returns `1` and the most recently added value is reachable via `get()`. Subsequent `add()` calls keep cycling but the size stays at `1` forever, violating the `maxSize: 0` invariant.

Root cause is in `shift()`: when the only item being evicted has no `NEWER` neighbour, the function deletes the item from the underlying `Map` but does *not* clear `mostRecentlyUsed` / `leastRecentlyUsed`. The next `add()` then links the new item onto these stale pointers, and the next `shift()` calls `map.delete()` on the already-removed key — a no-op — leaving the new item resident.

### Code snippet

```typescript
import { LRUCache } from '@aws-lambda-powertools/commons/utils/lru-cache';

const cache = new LRUCache<string, number>({ maxSize: 0 });

cache.add('a', 1);
console.log(cache.size()); // 0  (correct)

cache.add('b', 2);
console.log(cache.size()); // 1  (BUG — expected 0)
console.log(cache.get('b')); // 2 (BUG — expected undefined)

cache.add('c', 3);
console.log(cache.size()); // 1  (still leaking)
```

### Steps to Reproduce

1. Install `@aws-lambda-powertools/commons@2.33.1` (or check out `main`).
2. Create an `LRUCache` with `{ maxSize: 0 }`.
3. Call `cache.add(...)` at least twice with distinct keys.
4. Observe that `cache.size()` returns `1` instead of `0`.

### Possible Solution

In `LRUCache.shift()` (`packages/commons/src/LRUCache.ts:192-206`), when the evicted item has no `NEWER` neighbour, clear both pointers:

```typescript
private shift(): void {
  const item = this.leastRecentlyUsed!;

  if (item[NEWER]) {
    this.leastRecentlyUsed = item[NEWER];
    this.leastRecentlyUsed[OLDER] = undefined;
  } else {
    // Evicting the only item — clear both ends of the list
    this.leastRecentlyUsed = undefined;
    this.mostRecentlyUsed = undefined;
  }

  this.map.delete(item.key);
  item[NEWER] = undefined;
  item[OLDER] = undefined;
}
```

Alternatively, short-circuit `add()` when `maxSize === 0` so the linked-list manipulation is skipped entirely.

`2.33.1`

`22.x`

### Packaging format used

`npm`

### Execution logs

_N/A — pure logic bug, reproducible in unit tests._

vishwakt:
### Changes

When LRUCache is configured with maxSize 0, adding any item triggers immediate eviction. When the evicted item has no newer neighbor, shift() was not clearing the leastRecentlyUsed and mostRecentlyUsed pointers. Subsequent additions would incorrectly link onto these stale pointers, causing the cache to permanently retain one stale entry.

This change sets both pointers to undefined in shift() when the evicted item is the only one in the cache.

**Issue number:** closes #5290

---

### DIFF
--- a/packages/commons/src/LRUCache.ts
+++ b/packages/commons/src/LRUCache.ts
@@ -197,6 +197,9 @@ class LRUCache<K, V> {
     if (item[NEWER]) {
       this.leastRecentlyUsed = item[NEWER];
       this.leastRecentlyUsed[OLDER] = undefined;
+    } else {
+      this.leastRecentlyUsed = undefined;
+      this.mostRecentlyUsed = undefined;
     }
 
     // Remove the item from the map
--- a/packages/commons/tests/unit/LRUCache.test.ts
+++ b/packages/commons/tests/unit/LRUCache.test.ts
@@ -52,9 +52,14 @@ describe('Class: LRUMap', () => {
 
       // Act
       cache.add('a', 1);
+      cache.add('b', 2);
+      cache.add('c', 3);
 
       // Assess
       expect(cache.size()).toBe(0);
+      expect(cache.get('a')).toBeUndefined();
+      expect(cache.get('b')).toBeUndefined();
+      expect(cache.get('c')).toBeUndefined();
     });
   });
 


svozza (approved):

svozza:
Merged!

PR #5310, opened 2026-06-02
fix(event-handler): avoid runtime import of metrics package in HTTP metrics middleware

### LINKED ISSUE #5309: Bug: event-handler HTTP metrics middleware forces a runtime import of Metrics
## Expected Behavior

Using only the `@aws-lambda-powertools/event-handler` HTTP `Router` (with the `cors` middleware) should not require `@aws-lambda-powertools/metrics` to be installed or present at runtime. A user who never opts into the metrics middleware should be able to bundle their handler (esbuild, ESM) and run it without `@aws-lambda-powertools/metrics` resolvable.

## Current Behavior

`@aws-lambda-powertools/event-handler@2.33.1`'s HTTP metrics middleware file (`packages/event-handler/src/http/middleware/metrics.ts`) does a VALUE import: `import { MetricUnit } from '@aws-lambda-powertools/metrics';`. Because `MetricUnit` is a runtime value (not a type), esbuild retains this import in the bundle even when the consuming code only imports the `Router` and `cors` middleware and never the metrics middleware. When `@aws-lambda-powertools/metrics` is not a dependency (it is only an optional/transitive of event-handler) and there is no Lambda layer providing it, the bundled function crashes at init on AWS Lambda (Node.js 24 ESM) with:

`Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@aws-lambda-powertools/metrics' imported from /var/task/index.mjs`

resulting in API Gateway returning HTTP 502.

Note for contrast: the sibling tracer middleware (`http/middleware/tracer.ts`) uses `import type { Tracer }` (type-only, erased at compile time) and does NOT cause this problem. Only the metrics middleware's value import of `MetricUnit` leaks into the bundle.

## Code snippet

```typescript
import { getStringFromEnv } from '@aws-lambda-powertools/commons/utils/env';
import { cors } from '@aws-lambda-powertools/event-handler/http/middleware';

const app = new Router();
app.use(cors());
app.get('/hello', async () => ({ ok: true }));

export const handler = async (event: unknown, context: unknown) =>
  app.resolve(event, context);
```

Bundled with esbuild (`format: 'esm'`, `bundle: true`), the output retains `import { MetricUnit } from "@aws-lambda-powertools/metrics";` even though the metrics middleware is never imported by this code. With `@aws-lambda-powertools/metrics` NOT installed, the bundle fails to resolve at runtime.

## Steps to Reproduce

1. New project that depends on `@aws-lambda-powertools/event-handler` and `@aws-lambda-powertools/commons` (2.33.1), WITHOUT `@aws-lambda-powertools/metrics` installed and without a Powertools Lambda layer.
2. Write the handler in the code snippet above (Router + cors only; no metrics middleware).
3. Bundle with esbuild as ESM (`format: 'esm'`, `bundle: true`, `platform: 'node'`).
4. Inspect the bundle output: it contains a bare `import { MetricUnit } from "@aws-lambda-powertools/metrics"` statement.
5. Deploy to AWS Lambda (Node.js 24) and invoke (e.g. behind API Gateway). The function fails at init with `ERR_MODULE_NOT_FOUND` for `@aws-lambda-powertools/metrics`; API Gateway returns 502.

## Possible Solution

Avoid leaking a runtime dependency on the metrics package from code paths that don't use metrics. Options the maintainers may consider: (a) make the metrics middleware import `MetricUnit` lazily / only when the middleware is actually constructed, (b) move the `MetricUnit` value import to a type-only import or a local constant within the metrics middleware so it isn't pulled into the bundle unless the middleware is used, or (c) ensure the metrics middleware module is only reachable when explicitly imported (subpath isolation) so tree-shaking/bundlers don't retain the metrics import for Router-only consumers. (As a consumer-side workaround we aliased `@aws-lambda-powertools/metrics` to a tiny MetricUnit-only stub at bundle time, but the library ideally shouldn't force this.)

`2.33.1`

24.x

## Packaging format used

npm

## Execution logs

```Shell
ERROR Init Error Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@aws-lambda-powertools/metrics' imported from /var/task/index.mjs
    at ... (node:internal/modules/esm/resolve)
    code: 'ERR_MODULE_NOT_FOUND'
Node.js v24.14.1
INIT_REPORT Init Duration: 223.12 ms Phase: invoke Status: error Error Type: Runtime.Unknown
```

vishwakt:
### Changes

The HTTP metrics middleware (`packages/event-handler/src/http/middleware/metrics.ts`) did a value import of `MetricUnit` from `@aws-lambda-powertools/metrics`:

```ts
import { MetricUnit } from '@aws-lambda-powertools/metrics';
```

Because `MetricUnit` is a runtime value (not a type), bundlers such as esbuild retain this import even when the consuming code only uses the `Router` and other middleware (e.g. `cors`) and never opts into the metrics middleware. The middleware barrel (`http/middleware/index.ts`) re-exports `metrics` alongside `cors`, and the package does not declare `sideEffects: false`, so a `Router + cors`-only bundle keeps `metrics.js` and its `MetricUnit` import.

Since `@aws-lambda-powertools/metrics` is an optional peer dependency, bundles that do not install it (and have no Powertools Lambda layer) crash at init on AWS Lambda with `ERR_MODULE_NOT_FOUND`, returning HTTP 502 behind API Gateway.

This change inlines the metric unit string literals (`'Milliseconds'`, `'Count'`) instead of importing the `MetricUnit` value. The literals are type-safe because `addMetric`'s `unit` parameter is typed as the union of those exact string literals. This matches the type-only approach already used by the sibling tracer middleware (`http/middleware/tracer.ts`), so no runtime dependency on the metrics package leaks into Router-only bundles.

A regression test asserts that the middleware source carries no value (non-`type`) import from `@aws-lambda-powertools/metrics`. Verified that reintroducing the value import makes the test fail. Also verified manually by bundling a `Router + cors`-only handler with esbuild (`format: 'esm'`, `bundle: true`, `platform: 'node'`): after the change the output contains no reference to `@aws-lambda-powertools/metrics`.

**Issue number:** closes #5309

---

### DIFF
--- a/packages/event-handler/tests/unit/http/middleware/metrics.test.ts
+++ b/packages/event-handler/tests/unit/http/middleware/metrics.test.ts
@@ -422,15 +422,42 @@ describe('Metrics Middleware', () => {
     const source = readFileSync(
       new URL('../../../../src/http/middleware/metrics.ts', import.meta.url),
       'utf-8'
-    )
-      // Strip comments so documentation examples (e.g. JSDoc `@example`
-      // snippets that show a value import) don't trigger a false positive.
-      .replace(/\/\*[\s\S]*?\*\//g, '')
-      .replace(/\/\/.*$/gm, '');
-    const valueImportFromMetrics =
-      /import\s+(?!type\b)[^;]*from\s+['"]@aws-lambda-powertools\/metrics['"]/;
+    );
 
-    // Act & Assess
-    expect(source).not.toMatch(valueImportFromMetrics);
+    // Strip comments with a linear character scan (no regex) so documentation
+    // examples that show a value import don't cause a false positive, while
+    // avoiding ReDoS-prone backtracking patterns.
+    const stripComments = (code: string): string => {
+      let result = '';
+      let index = 0;
+      while (index < code.length) {
+        const marker = code.slice(index, index + 2);
+        if (marker === '/*') {
+          const end = code.indexOf('*/', index + 2);
+          index = end === -1 ? code.length : end + 2;
+        } else if (marker === '//') {
+          const end = code.indexOf('\n', index + 2);
+          index = end === -1 ? code.length : end;
+        } else {
+          result += code[index];
+          index += 1;
+        }
+      }
+      return result;
+    };
+
+    // Act
+    const importsMetricsAsValue = stripComments(source)
+      .split(';')
+      .map((statement) => statement.trim())
+      .some(
+        (statement) =>
+          statement.startsWith('import') &&
+          !statement.startsWith('import type') &&
+          statement.includes('@aws-lambda-powertools/metrics')
+      );
+
+    // Assess
+    expect(importsMetricsAsValue).toBe(false);
   });
 });
--- a/packages/event-handler/tests/unit/http/middleware/metrics.test.ts
+++ b/packages/event-handler/tests/unit/http/middleware/metrics.test.ts
@@ -1,4 +1,3 @@
-import { readFileSync } from 'node:fs';
 import { Metrics } from '@aws-lambda-powertools/metrics';
 import context from '@aws-lambda-powertools/testing-utils/context';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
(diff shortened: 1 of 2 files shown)


packages/event-handler/tests/unit/http/middleware/metrics.test.ts
```diff
@@ -410,4 +411,35 @@ describe('Metrics Middleware', () => {
       })
     );
   });
+
+  it('does not import any runtime value from the metrics package', () => {
```
dreamorosi: I'd remove this test - I get where it's coming from, but it's just verbose and doesn't add much - plus it's very brittle.

dreamorosi (changes requested):
Thank you for the quick PR - left a comment to remove the test case added in this PR.

packages/event-handler/tests/unit/http/middleware/metrics.test.ts
vishwakt: fair call out. i've removed the test

dreamorosi (approved):

dreamorosi (approved):

svozza (approved):

PR #5336, opened 2026-06-12
build(internal): add test compilation check to package builds

### LINKED ISSUE #5073: Maintenance: Add test compilation check to package builds
### Summary

Package builds (e.g. `npm run build`) only compile source code under `src/`. Test files under `tests/` are never type-checked as part of the build, meaning type errors in tests can go unnoticed until runtime.

Every package already has a `tests/tsconfig.json` configured with `noEmit: true` that includes both source and test files — it just isn't invoked during the build.

### Why is this needed?

Type errors in test files can silently accumulate and only surface at runtime. Including test compilation in the build ensures type errors are caught early and consistently.

All packages

### Solution

For each package, make two changes:

**1. Remove redundant source inclusion from `tests/tsconfig.json`**

Source files imported by tests are already resolved via module resolution, so explicitly including `../src/**/*` just duplicates work that the main build already does.

For packages that include `../../testing/src/setupEnv.ts` (e.g. event-handler, logger, parameters):

```diff
-"include": ["../../testing/src/setupEnv.ts", "../src/**/*", "./**/*"]
+"include": ["../../testing/src/setupEnv.ts", "./**/*"]
```

For all other packages:

```diff
-"include": ["../src/**/*", "./**/*"]
+"include": ["./**/*"]
```

**2. Add test type-checking to the build in `package.json`**

The test type-check runs in parallel with the ESM and CJS builds (note the single `&`).

```diff
  "scripts": {
    ...
+   "build:tests": "tsc --noEmit -p tests/tsconfig.json",
-   "build": "npm run build:esm & npm run build:cjs",
+   "build": "npm run build:esm & npm run build:cjs & npm run build:tests",
    ...
  }
```

Apply to all packages.

**3. Create missing `tests/tsconfig.json` for `jmespath`**

The `jmespath` package is the only one missing this file. Create `packages/jmespath/tests/tsconfig.json`:

```json
{
  "extends": "../tsconfig.json",
  "compilerOptions": {
    "rootDir": "../",
    "noEmit": true
  },
  "include": ["./**/*"]
}
```

### Acknowledgment

### Future readers

dreamorosi:
### Changes

Package builds only compile source code under `src/`; test files under `tests/` were never type-checked, so type errors in tests could go unnoticed until runtime. This PR adds a dedicated test type-checking step so those errors are caught in CI.

- **`build:tests` script** added to each package (`tsc --noEmit -p tests/tsconfig.json`).
- **`build:tests` is intentionally NOT wired into the regular `build`** — day-to-day builds stay fast and never emit test files into the published bundle.
- **CI runs the type-check as a dedicated step** (`Type check tests`) in the `code-quality` matrix job, after dependencies are installed and the whole workspace is built (so every package's `lib/` exists for cross-package import resolution).
- **Created the missing `tests/tsconfig.json` for `jmespath`** (the only package without one).
- **Fixed pre-existing test type errors** surfaced by the new check:
  - `commons` — `deepMerge.property.test.ts` (invalid `depthSize` on `fc.array` constraints → `size`) and `deepMerge.test.ts` (typed the result of `deepMerge({}, source)`).
  - `metrics` — `basicFeatures.decorator.test.functionCode.ts` (made the decorated handler `async` so it satisfies `HandlerMethodDecorator`).

### Notes for reviewers

- The original proposal suggested wiring `build:tests` into `build` in parallel (`& npm run build:tests`). That is unsafe here: the test type-check resolves cross-package imports against each dependency's built `lib/`, but the workspace build order builds some dependencies (e.g. `testing`) after their consumers, and the parallel `&` also races a package's own emit build. Running it as a separate CI step after the full build avoids the race entirely.
- The proposal also suggested removing `../src/**/*` from `tests/tsconfig.json`. That cannot be done here: the configs inherit `composite: true`, which makes TypeScript enforce TS6307 ("Projects must list all files or use an `include` pattern"). The source glob is kept (including in the new `jmespath` file).
- Maintainer caveat (tests must not end up in the published npm bundle) holds: the emitting builds use `rootDir: ./src` + `include: ["./src/**/*"]`, `build:tests` uses `--noEmit`, and it is not part of the regular `build`. Verified no test files land in any `lib/` output.
- Verified locally: emit-only `npm run build` passes, and `npm run build:tests -w <pkg>` passes for every package after a full build.

**Issue number:** closes #5073

---


svozza:
Can we add the check to the pre-commit hook so we don't have to wait until the CI fails if we forget to run it locally?

dreamorosi:
> Can we add the check to the pre-commit hook so we don't have to wait until the CI fails if we forget to run it locally?

Good point - added, but to `pre-push` in line with when we run the other tests.

svozza (approved):

PR #5338, opened 2026-06-12
feat(event-handler): add HTTP request/response data to tracer subsegment

### LINKED ISSUE #1179: Feature request: HTTP request/response annotations
### Use case

I have a nextjs CDK construct with a simple APIGW lambda wrapper around nextjs. I want to annotate my traces with HTTP request and response metadata, e.g. "URL" and "method" and "status code"

I would like to use it here: https://github.com/jetbridge/cdk-nextjs/blob/main/assets/lambda/NextJsHandler.ts#L45

I'd like to see the status code, path, method, etc in xray:
image

### Solution/User Experience

Perhaps something like `traceRequestResponseCycle` in https://docs.aws.amazon.com/xray-sdk-for-nodejs/latest/reference/AWSXRay.html

### Alternative solutions

```Markdown
Discussed on discord: https://discord.com/channels/1006478942305263677/1006527385409179678/1042309163818168380
```

### Acknowledgment

- [X] Should this be considered in other Lambda Powertools languages? i.e. [Python](https://github.com/awslabs/aws-lambda-powertools-python/), [Java](https://github.com/awslabs/aws-lambda-powertools-java/)

dreamorosi:
### Changes

This PR populates the `http` field of the X-Ray subsegment created by the Event Handler `tracer` middleware with request/response data, implementing **Option 1 (child subsegment)** from the design discussion in #1179.

For each HTTP route, the subsegment now carries an `http` object mirroring the X-Ray SDK's `IncomingRequestData` shape:

- `http.request`: `method`, `url`, and (when available) `user_agent`, `client_ip`, `x_forwarded_for`
- `http.response`: `status`, and `content_length` when the response has a `Content-Length` header

The data is attached to the **route subsegment** (not the immutable Lambda function segment), which preserves the lineage between the function segment and the request while avoiding the misleading "self-calling" trace map nodes that the sibling/standalone-segment approach would introduce (per the discussion with @leandrodamascena).

Example trace data for a `GET /` route:

```json
"http": {
  "request": {
    "url": "https://<api-id>.execute-api.<region>.amazonaws.com/",
    "method": "GET",
    "user_agent": "curl/8.7.1",
    "client_ip": "<ip>",
    "x_forwarded_for": true
  },
  "response": { "status": 200, "content_length": 27 }
}
```

on the CloudWatch console, the segments look like this - notice the **Request & Response** fields being populated on the right side of the screen:

Screenshot 2026-06-12 at 12 52 22

Notes for reviewers:

- Because the data is on a subsegment rather than the top-most segment, the CloudWatch Traces UI does not populate the top-level **HTTP Method** / **URL Address** fields. The data is available in the "Raw data" and single-trace views. This is documented in the middleware JSDoc.
- `content_length` is read from the `Content-Length` header only (no extra body read). It is therefore only present when the handler sets the header or the `compress` middleware is registered as an inner middleware relative to `tracer`. This is documented in the JSDoc as well.
- `client_ip` extraction follows the same per-event-source logic already used by the `metrics` middleware (API Gateway v1/v2 request context, with `X-Forwarded-For` fallback for ALB/other).

**Issue number:** closes #1179

---


packages/event-handler/src/http/middleware/tracer.ts
```diff
@@ -1,6 +1,124 @@
...
+
+/**
+ * Extracts the client IP address from the request context.
+ *
+ * For API Gateway events the source IP is taken from the request context, while
+ * for ALB (or any other source) it falls back to the `X-Forwarded-For` header.
+ *
+ * @param reqCtx - The request context for the current request
+ */
+const getClientIp = (reqCtx: RequestContext): string | undefined => {
```
svozza, line 28: Are we repeating this function from the Metrics middleware to keep it self contained?

dreamorosi: Yea, I wanted to keep the diff scoped to this feature and we don't have a place under middlewares where we can share this type of utils - but I'm happy to do it as a follow up PR that touches both.

svozza: Yeah, I think we can have utils or common folder for middleware but agree doesn't need to be this PR

svozza (approved):

PR #5343, opened 2026-06-12
feat(parameters): add throwOnMissing option to fail fast on missing values

### LINKED ISSUE #4086: Feature request: NonNullable return type for parameters
### Use case

Fetching a non-existing parameter usually results in an error from sdk (not sure about every provider secrets/appconfig etc). Not receiving a param usually block an execution flow. If developers need to, they can account for such errors and do some actions on catch.

However, since powertools parameters utility uses Map for caching, it is by default typed to be T | undefined, even though it's the utility that is responsible for adding the parameter to the map, and it can guarantee if something has indeed been added to the Map. Hence, is it possible to return NonNullable<T> in the get method of the params util? If it's not a desirable maybe with throwOnUndefined option?

This would greatly reduce boilerplate of checking if the returned param is not undefined or throwing an error otherwise.

### Solution/User Experience

Return NonNullable<T> for get method of the params utilities.

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR adds an opt-in `throwOnMissing` option to the Parameters utility's `get` functions, addressing a known papercut where callers must manually guard against `undefined` after fetching a parameter that may not exist.

When `throwOnMissing: true` is set:

- **Runtime**: a missing value throws a new `ParameterNotFoundError` instead of returning `undefined` (or, for SSM/Secrets, instead of the generic `GetParameterError`).
- **Types**: the return type is narrowed to exclude `undefined`, so no post-fetch nullish check is needed.

```ts
const param = await getParameter('/my/parameter', { throwOnMissing: true });
//    ^? string   (no longer string | undefined)
```

**Key design points:**

- `ParameterNotFoundError` **extends `GetParameterError`**, so it is `instanceof`-discriminable for callers who want to handle a genuinely missing value separately from other retrieval failures, while remaining backward-compatible with existing `catch (e instanceof GetParameterError)` blocks.
- "Missing" means the store returned no value (`null`/`undefined`), evaluated on the raw value **before** any transform. Falsy-but-present values (`""`, `0`, `false`, JSON `null`) are considered present and do **not** throw.
- The change is **purely additive**. The SSM and Secrets providers signal a missing parameter by *throwing* a not-found error from the SDK (rather than returning nullish); they translate that into `ParameterNotFoundError` **only when `throwOnMissing` is set**, leaving the long-standing default behavior (throwing `GetParameterError`) untouched.
- Applies to `getParameter`, `getSecret`, `getAppConfig`, and the provider `get()` methods. `getMultiple` is unaffected (it returns a `Record`, never `undefined`).

Includes unit tests (runtime behavior across providers, including the preserved default), type tests (return-type narrowing), and documentation with a compiling snippet.

> Validated on a real deployed Lambda function against SSM (present value, missing value with and without the flag, and a falsy-but-present value).

**Issue number:** closes #4086

---


packages/parameters/tests/unit/BaseProvider.test.ts
```diff
@@ -103,6 +104,63 @@ describe('Class: BaseProvider', () => {
       );
     });
 
+    it('returns undefined when the parameter is not found and throwOnMissing is not set', async () => {
+      // Prepare
+      const provider = new TestProvider();
+      vi.spyOn(provider, '_get').mockResolvedValue(
+        undefined as unknown as string
```
svozza: Nit: any way to avoid this casting?

packages/parameters/tests/unit/BaseProvider.test.ts
```diff
@@ -103,6 +104,63 @@ describe('Class: BaseProvider', () => {
...
+        .get('my-parameter', { throwOnMissing: true })
+        .catch((err) => err);
+
+      // Assess
+      expect(error).toBeInstanceOf(ParameterNotFoundError);
+      expect(error).toBeInstanceOf(GetParameterError);
+      expect(error.cause).toBeUndefined();
+    });
+
+    it('does not throw when the parameter resolves to a falsy but present value and throwOnMissing is set', async () => {
```
svozza, line 146: Should we use `it.each` to test other falsy values or is that overkill?

dreamorosi: I think for this case it's not needed - the reason why this test is here is because we have a special handling for when the SDK returns `undefined` or `null`, and during the planning I decided to leave alone falsy values. This test is just there as a proof of that, but we don't need to test all of them.

svozza (dismissed):

dreamorosi:
@svozza sorry I had to fix some type issues that I noticed after you approved. If you approve it, feel free to merge right away

svozza (approved):

svozza:
Merging now!

PR #5344, opened 2026-06-12
feat(signer): add SigV4 request signing utility

### LINKED ISSUE #2146: Feature request: module to Sign Requests
### Use case

It would be nice to have a module in the AWS Powertools that would sign requests made to AWS Resources via the SigV4 method. Currently doing it yourself is a tedious process and needing to understand the underlying mechanism.

This is also discussed in the Powertools for Python repositor 
- https://github.com/aws-powertools/powertools-lambda-python/issues/2713
- https://github.com/aws-powertools/powertools-lambda-python/issues/2493
- https://github.com/aws-powertools/powertools-lambda-python/pull/2435

### Solution/User Experience

Same as for the python version,  I am proposing adding a new class that will take in the parameters to create this signed request for the end user to simplify it since it is not a very straightforward method and should be made easier.

Currently it differs based on which request library you are using, for example with Axios there is an [intercepter](https://github.com/jamesmbourne/aws4-axios), for got there is a [got4aws](https://github.com/SamVerschueren/got4aws) you could use or in cases where I'm at currently I've implemented our own hook in a request library.

Having an supported module by powertools that does this would be preferred. 

### Alternative solutions

_No response_

### Acknowledgment

### Future readers

dreamorosi:
### Changes

This PR introduces a new `@aws-lambda-powertools/signer` package that signs HTTP requests to AWS services using the [AWS Signature Version 4 (SigV4)](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv4_signing.html) signing process, so customers can call IAM-authenticated endpoints (API Gateway, Lambda function URLs, AppSync) from within their Lambda functions.

The design intentionally splits **signing** from **sending**:

- **`SigV4Signer`** (`/sigv4`) — signs web-standard `Request` objects and returns a signed `Request`. Performs no network I/O. Reads credentials and region from the Lambda runtime environment variables by default (no extra dependencies, no cold-start cost from the credential provider chain).
- **`createSignedFetcher`** (`/fetch`) — a drop-in signed `fetch` that consumes a signer instance and signs each request before sending it. Composes with libraries that accept a custom `fetch`.
- Signing without fetching is first-class: the signed `Request` exposes `url`, `method`, and `headers` for use with any HTTP client (axios, got, interceptors, generated SDK clients).

Design decisions worth highlighting for review:

- A structural **`Signer`** interface is defined, and the fetcher depends on it (not the concrete class), so a future `SigV4aSigner` can be added at a `/sigv4a` subpath as a non-breaking change with its heavier dependencies isolated to that entry point.
- Subpath exports from day one (`/sigv4`, `/fetch`, `/errors`, `/types`) — no root export — to keep import-level isolation and make future signer variants additive.
- Opinionated, lean credentials: defaults to the standard AWS env vars the Lambda runtime always injects (including `AWS_SESSION_TOKEN`), avoiding a dependency on `@aws-sdk/credential-provider-node`. Customers can pass explicit `credentials` (e.g. `fromNodeProviderChain()`) for non-Lambda environments.
- Only runtime dependencies are `@smithy/signature-v4` and `@smithy/types`; hashing uses `node:crypto` (no `@aws-crypto/sha256-js`). We do not depend on `@smithy/protocol-http` — the signing process accepts a plain object typed via `@smithy/types`, and a guard test protects that contract.
- Typed errors (`/errors`): base `SignerError`, plus `SignerConfigError` (missing region at construction / unresolvable credentials) and `RequestSigningError` (signing failures, including non-bufferable streaming bodies). Transport errors from the underlying `fetch` propagate untouched.

**Out of scope / deferred** (accommodated by the design): SigV4a, streaming/`UNSIGNED-PAYLOAD` bodies (would arrive as additive options or a future `signStream` method), service auto-detection from the URL, and `uriEscapePath`/payload-signing knobs.

Includes: unit tests with 100% coverage, a feature page (`docs/features/signer.md`), package README, doc snippets, navigation/index updates, and the `signer` semantic-commit scope.

**Issue number:** closes #2146

---


packages/signer/tests/unit/Sha256.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    // Prepare
+    const hash = new Sha256('secret');
+    hash.update('message');
+
+    // Act
+    const digest = await hash.digest();
+
+    // Assess
+    const expected = new Uint8Array(
+      createHmac('sha256', 'secret').update('message').digest()
```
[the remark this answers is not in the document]
dreamorosi: it literally says "secret" gosh

packages/signer/src/Sha256.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @internal
+ */
+class Sha256 implements Checksum {
+  readonly #hash: Hash | Hmac;
+
+  public constructor(secret?: SourceData) {
+    this.#hash =
+      secret === undefined
+        ? createHash('sha256')
+        : createHmac('sha256', secret as Uint8Array | string);
```
svozza: How come we need these casts?

packages/signer/tests/unit/SigV4Signer.test.ts
```diff
@@ -0,0 +1,328 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { RequestSigningError, SignerConfigError } from '../../src/errors.js';
+import { SigV4Signer } from '../../src/sigv4.js';
+
+describe('Class: SigV4Signer', () => {
+  const ENVIRONMENT_VARIABLES = process.env;
+
+  beforeEach(() => {
+    process.env = {
```
svozza: Let's use the vitest stubEnv/unstubEnv fucntions

packages/signer/README.md
```diff
@@ -0,0 +1,150 @@
+# Powertools for AWS Lambda (TypeScript) - Signer Utility <!-- omit in toc -->
+
+Powertools for AWS Lambda (TypeScript) is a developer toolkit to implement Serverless [best practices and increase developer velocity](https://docs.aws.amazon.com/powertools/typescript/latest/#features).
+
+You can use the package in both TypeScript and JavaScript code bases.
```
svozza: Is this necessary? JavaScript is valid TypeScript.

dreamorosi: Yes fair, this is a copy of other readme files that used to do this type of statement but agreed I'll remove it

packages/signer/src/Sha256.ts
dreamorosi: Good question - let me investigate

dreamorosi: Dug into this — good catch, the casts were actually hiding a real gap. Smithy's `SourceData` is `string | ArrayBuffer | ArrayBufferView`, while `node:crypto` accepts `string | NodeJS.ArrayBufferView`: a bare `ArrayBuffer` isn't assignable (and would throw at runtime), and TS's structural `ArrayBufferView` isn't assignable to Node's concrete union of typed arrays either.

In practice `@smithy/signature-v4` only ever passes strings and `Uint8Array`s, so it never bit us, but casting was papering over it. Replaced the casts with a small `toBinary` helper that wraps non-string inputs in a `Uint8Array` (no byte copy), and added unit tests for `ArrayBuffer` and offset `ArrayBufferView` inputs in 383176db.

packages/signer/tests/unit/SigV4Signer.test.ts
dreamorosi: Done in 383176db — switched to `vi.stubEnv` per test with `vi.unstubAllEnvs()` in `afterEach`, matching what we do in the other packages.

packages/signer/README.md
dreamorosi: Removed in 383176db.

svozza:
I wonder are the unit test failures related to: https://github.com/aws-powertools/powertools-lambda-typescript/pull/5351. Might need to explicitly import some types or something?

packages/signer/src/Sha256.ts
svozza: Nice!

dreamorosi:
> I wonder are the unit test failures related to: #5351. Might need to explicitly import some types or something?

Yes, you're correct - it was `BodyInit` that I should've now imported from `undici-types`.

svozza (approved):

PR #5349, opened 2026-06-12
feat(metrics): add Disposable support for automatic flushing via using

### LINKED ISSUE #5200: Feature request: Add Disposable support to Metrics for automatic flushing via `using` keyword
### Use case

The Metrics class has a fluent interface where methods like `addMetric()`, `addDimension()`, and `publishStoredMetrics()` return `this`. Currently, users who are not using the decorator or Middy middleware must either call `publishStoredMetrics()` manually or wrap their handler logic in `try/finally` to ensure metrics are flushed on error paths:

```typescript
const metrics = new Metrics({ namespace: 'serverlessAirline', serviceName: 'orders' });

  try {
    metrics.addMetric('successfulBooking', MetricUnit.Count, 1);
    // business logic
  } finally {
  }
};
```

JavaScript's [Explicit Resource Management](https://github.com/tc39/proposal-explicit-resource-management) proposal (Stage 3, shipping in Node.js 24) introduces the `using` keyword, which calls `[Symbol.dispose]()` automatically when a binding goes out of scope — including when an exception is thrown. Since the fluent methods already return `this`, making `Metrics` implement `Disposable` is a natural fit.

### Solution/User Experience

Implement `[Symbol.dispose]()` on the `Metrics` class, delegating to `publishStoredMetrics()`. This allows users to write:

```typescript
const metrics = new Metrics({ namespace: 'serverlessAirline', serviceName: 'orders' });

  using _ = metrics.addMetric('successfulBooking', MetricUnit.Count, 1);
  metrics.addDimension('region', 'us-west-2');
  // metrics are automatically flushed when scope exits, including on exceptions
};
```

The implementation is straightforward. Add the method to the `Metrics` class:

```typescript
public [Symbol.dispose](): void {
  this.publishStoredMetrics();
}
```

And the corresponding signature to `MetricsInterface`:

```typescript
[Symbol.dispose](): void;
```

No `AsyncDisposable` is needed since `publishStoredMetrics()` is synchronous. The method is safe to call on an already-flushed instance — `publishStoredMetrics()` checks `hasStoredMetrics()` and simply logs a warning if the buffer is empty, which is the same behaviour that already occurs with the decorator and middleware patterns.

On the TypeScript side, `ESNext.Disposable` needs to be added to the `lib` compiler option since the `Disposable` types have not yet been promoted to any yearly ES lib (including ES2024). This is purely a compile-time type concern — it does not change emitted code or require runtime polyfills:

```json
{
  "compilerOptions": {
    "target": "ES2023",
    "lib": ["ES2023", "ESNext.Disposable"]
  }
}
```

The `using` keyword requires Node.js 24+. The `[Symbol.dispose]` method itself is harmless on older runtimes (the symbol exists from Node 18+, it simply won't be called without the `using` syntax). Unit and e2e tests for this feature should be gated with `describe.skipIf(process.versions.node < '24')`.

### Alternative solutions

The existing decorator (`@metrics.logMetrics()`) and Middy middleware already provide automatic flushing. The `using` pattern is a lighter-weight alternative for users who prefer not to adopt those patterns.

### Acknowledgment

The `using`/`IDisposable` pattern is native to .NET and could be considered there. Python and Java do not have an equivalent language feature.

### Future readers

svozza:
## Summary

Implements the `Disposable` interface on the `Metrics` class so instances can be declared with the `using` keyword (Node.js 24+). When the binding leaves scope — including when an exception is thrown — stored metrics are flushed automatically via `publishStoredMetrics()`, offering a lightweight alternative to a manual `try/finally` block or the middleware/decorator.

### Changes

- Add `[Symbol.dispose]()` to `Metrics` (`packages/metrics/src/Metrics.ts`), delegating to `publishStoredMetrics()`, and declare it on `MetricsInterface` (`packages/metrics/src/types/Metrics.ts`) with TSDoc and examples
- Add `ESNext.Disposable` to `lib` in the root `tsconfig.json` — `Symbol.dispose` types aren't part of any yearly ES lib yet; compile-time only, no emitted-code or polyfill impact
- Add unit tests (`packages/metrics/tests/unit/dispose.test.ts`) covering flush on scope exit, flush on thrown error, empty-metrics warning, and direct disposal — skipped on Node.js < 24
- Add e2e test (`packages/metrics/tests/e2e/basicFeatures.using.test.ts`) verifying metrics are delivered to CloudWatch when flushed via `using`
- Document the new flushing method in `docs/features/metrics.md` with a new snippet (`examples/snippets/metrics/usingKeyword.ts`)

**Issue number:** closes #5200

---

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,8 +5,10 @@
     "target": "ES2024", // Node.js 22
     // `Disposable`/`Symbol.dispose` are not part of any yearly ES lib yet (not ES2024 nor
     // ES2025 as of TS 6.x), so we pull them in explicitly. Compile-time only: it doesn't
-    // change emitted code or require a runtime polyfill.
-    "lib": ["ES2024", "ESNext.Disposable"],
+    // change emitted code or require a runtime polyfill. DOM libs must be listed too:
+    // setting `lib` replaces the target's implicit default (ES2024 + DOM + DOM.Iterable),
+    // and the codebase relies on DOM types like `BodyInit` and `Response`.
+    "lib": ["ES2024", "DOM", "DOM.Iterable", "ESNext.Disposable"],
     "experimentalDecorators": true,
     "module": "NodeNext",
     "moduleResolution": "NodeNext",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,6 +3,10 @@
     "incremental": true,
     "composite": true,
     "target": "ES2024", // Node.js 22
+    // `Disposable`/`Symbol.dispose` are not part of any yearly ES lib yet (not ES2024 nor
+    // ES2025 as of TS 6.x), so we pull them in explicitly. Compile-time only: it doesn't
+    // change emitted code or require a runtime polyfill.
+    "lib": ["ES2024", "ESNext.Disposable"],
     "experimentalDecorators": true,
     "module": "NodeNext",
     "moduleResolution": "NodeNext",
--- a/docs/features/metrics.md
+++ b/docs/features/metrics.md
@@ -328,7 +328,9 @@ You can manually flush the metrics with `publishStoredMetrics` as follows:
 
 #### Using the `using` keyword
 
-If you are running on Node.js 24 or newer, the `Metrics` class implements the [`Disposable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/dispose){target="_blank"} interface. This means you can declare your instance with the [`using`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/using){target="_blank"} keyword and have your metrics flushed automatically when the binding goes out of scope, including when an exception is thrown. This is a lightweight alternative to wrapping your handler in a `try/finally` block.
+If you are running on Node.js 24 or newer, the `Metrics` class implements the [`Disposable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/dispose){target="_blank"} interface.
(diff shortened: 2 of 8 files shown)


tsconfig.json
```diff
@@ -3,6 +3,12 @@
     "incremental": true,
     "composite": true,
     "target": "ES2024", // Node.js 22
+    // `Disposable`/`Symbol.dispose` are not part of any yearly ES lib yet (not ES2024 nor
+    // ES2025 as of TS 6.x), so we pull them in explicitly. Compile-time only: it doesn't
+    // change emitted code or require a runtime polyfill. DOM libs must be listed too:
+    // setting `lib` replaces the target's implicit default (ES2024 + DOM + DOM.Iterable),
+    // and the codebase relies on DOM types like `BodyInit` and `Response`.
+    "lib": ["ES2024", "DOM", "DOM.Iterable", "ESNext.Disposable"],
```
svozza, line 11: Note for @dreamorosi: setting `lib` explicitly replaces TypeScript's target-derived default (`ES2024` + `DOM` + `DOM.Iterable`), so `DOM`/`DOM.Iterable` must be listed here to preserve the status quo — the codebase currently relies on DOM types like `BodyInit` and the looser DOM `Response` typings. Removing the DOM dependency properly (using `undici-types` instead) is tracked in #5350.

svozza:
E2E failures we unrelated rate-limits.

tsconfig.json
dreamorosi: Gotcha - we definitely need to do this asap.

dreamorosi (approved):
Thanks for this - looking forward to see if we can do other core utilities as well!

svozza:
Yes, should be a very nice DX!

PR #5351, opened 2026-06-13
chore(maintenance): remove DOM lib, use undici-types for fetch types

### LINKED ISSUE #5350: Maintenance: Remove DOM lib dependency and use undici-types for server-side fetch types
## Summary

The root `tsconfig.json` historically had no explicit `lib`, so TypeScript derived it from `target: ES2024` — which implicitly includes the full browser **DOM** and **DOM.Iterable** libs. The entire codebase has therefore always type-checked against browser types, even though we ship a server-side library running on Node.js. PR #5349 made `lib` explicit and had to include `DOM` to preserve the status quo, because dropping it surfaces real dependencies on browser types:

1. `BodyInit` is referenced in `packages/event-handler/src/http/converters.ts` (~line 526). `@types/node` exposes `fetch`, `Request`, `Response`, and `Headers` as globals (backed by `undici-types`), but **not** `BodyInit` — that name only exists globally in the DOM lib.
2. `@types/node`'s web globals are conditional on DOM's presence (detected via `globalThis.onmessage`). With DOM loaded, DOM's `Response` wins and `json()` returns `Promise<any>`; without it, undici-types' stricter `Response` applies and `json()` returns `Promise<unknown>`. Roughly 40 call sites across e2e tests and `examples/snippets` do `const data = await response.json(); data.foo`, which only compiles under the looser DOM typing.

This issue proposes removing `DOM`/`DOM.Iterable` from `lib` and migrating those dependencies to the Node-native types.

## Why is this needed?

- **Type-checking against the wrong platform.** DOM types let browser-only globals (`document`, `window`, `localStorage`, etc.) pass the type checker in code that runs exclusively on Lambda's Node.js runtime, hiding a class of bugs the compiler should catch.
- **Wrong fetch semantics.** undici-types are the truthful types for Node's actual `fetch` implementation. DOM's `Response.json(): Promise<any>` also silently disables type checking at every response-parsing call site — the `unknown` errors that appear when DOM is removed are real unchecked accesses, not noise.
- **Lib hygiene.** Now that `lib` is explicit, `DOM` in a Lambda library's tsconfig is a visible oddity that will invite questions; better to remove it deliberately than carry it indefinitely.

Governance, Event Handler, Tests

## Solution

1. Change root `tsconfig.json` `lib` from `["ES2024", "DOM", "DOM.Iterable", "ESNext.Disposable"]` to `["ES2024", "ESNext.Disposable"]`.
2. In `packages/event-handler/src/http/converters.ts`, import `BodyInit` from `undici-types` instead of relying on the DOM global, and declare `undici-types` as an explicit root devDependency (pinned to the range `@types/node` resolves) rather than relying on it being hoisted transitively via `@types/node` — placed at the root alongside `@types/node`, consistent with how the monorepo manages ambient Node typings.
3. Fix the ~40 `'data' is of type 'unknown'` sites in e2e tests and `examples/snippets` by typing the expected response shape (e.g. `await response.json() as { message: string }` or a small helper) rather than relying on `any`.
4. Verify with `npm run build:tests -ws --if-present` that all workspaces type-check, and audit for any remaining accidental DOM-global usage that the removal surfaces.

## Acknowledgment

---

svozza:
## Summary

Removes `DOM` and `DOM.Iterable` from the root `tsconfig.json` `lib` so the codebase type-checks against Node.js types instead of browser types. Node's `fetch`/`Request`/`Response`/`Headers` globals now come from `@types/node` (backed by `undici-types`), whose stricter `Response.json(): Promise<unknown>` surfaced ~50 previously unchecked response accesses in tests and snippets, which are now explicitly typed.

### Changes

- Change root `tsconfig.json` `lib` from `["ES2024", "DOM", "DOM.Iterable", "ESNext.Disposable"]` to `["ES2024", "ESNext.Disposable"]` and update the explanatory comment
- Import `BodyInit` as a type from `undici-types` in `packages/event-handler/src/http/converters.ts`, since `@types/node` does not expose it as a global
- Declare `undici-types` as an explicit root devDependency (pinned to `^7.24.6`, matching the range `@types/node` resolves) rather than relying on it being hoisted transitively via `@types/node` — placed at the root alongside `@types/node`, consistent with how the monorepo manages ambient Node typings
- Type the `response.json()` results in the event-handler e2e tests (`httpRouter.test.ts`) with the expected response shape of each route
- Cast `req.json()` to `JSONValue` in e2e/unit test route handlers so echoed request bodies satisfy `HandlerResponse`
- Type the `json()` results in `examples/snippets` (`advanced_fine_grained_responses.ts`, `customProviderVault.ts`)
- Audited the codebase for remaining DOM-global usage (`document`, `window`, `localStorage`, etc.) — none found

> [!NOTE]
> On the `(await response.json()) as { ... }` pattern: under undici-types, `Response.json()` returns `Promise<unknown>` by design — bytes off the wire have no compile-time type, and DOM's `Promise<any>` was silently disabling type checking at every one of these call sites. Some per-site type annotation is therefore unavoidable; the question is only where the trust lives. We considered a `fetchJson<T>` helper (centralizes the cast), `toMatchObject` assertions (no casts, but can't express numeric comparisons like `toBeGreaterThan`), and zod parsing (real runtime validation, but adds a dependency and the most churn). We settled on the plain inline assertion as the standard, explicit idiom — each test documents the response shape it expects right where it's used.

**Issue number:** closes #5350

---


examples/snippets/event-handler/http/advanced_fine_grained_responses.ts
```diff
@@ -24,7 +24,7 @@ app.get('/todos', async () => {
 });
 
 app.post('/todos', async ({ req }) => {
-  const body = await req.json();
+  const body = (await req.json()) as { title: string };
```
dreamorosi, line 27: I don't mind having these in the snippets, but I'm mildly concerned about the implications that these new types have on end customers.

Does this mean customers who use the library now have to do this cast every time they use this?

package.json
```diff
@@ -65,6 +65,7 @@
     "typedoc": "^0.28.19",
     "typedoc-plugin-missing-exports": "^4.1.3",
     "typescript": "^6.0.3",
+    "undici-types": "^7.24.6",
```
dreamorosi, line 68: I read the PR body and I am unsure whether we should declare this vs relying on what `@types/node` brings.

Now that we have it explicitly, does it mean we need to keep it in lockstep w the types? And what happens if they go out of sync?

What's the advantage of declaring this explicitly?

packages/event-handler/src/http/converters.ts
```diff
@@ -8,6 +8,9 @@ import type {
   APIGatewayProxyResult,
   APIGatewayProxyStructuredResultV2,
 } from 'aws-lambda';
+// `BodyInit` is not exposed as a global by @types/node, so we import it from
```
dreamorosi: I think we can remove this comment

examples/snippets/event-handler/http/advanced_fine_grained_responses.ts
svozza: Yeah, good point, there will be effects on customer. If you are doing stuff like `req.json()` in middleware or a handler like in this example, you would need to fix the types with a cast now that you are not getting `Promise<any>` back. I'm not sure that's a bad thing though, because it makes clear that you are handling untrusted data. Ultimately if you want real safety guarantees you should really be using a validation library like Zod and this problem goes away. That said, this does feel quite disruptive so I'm open to persuasion that we should leave things as they are.

package.json
svozza: I just would rather us declare our dependencies explicitly rather than relying on transitive dependencies of other packages.

packages/event-handler/src/http/converters.ts
svozza: Done!

package.json
svozza: That said, I don't feel strongly about this so happy to revert.

dreamorosi: No, I am onboard with declaring it explicitly - I am just concerned about type mismatches, but we'll cross that bridge when we get there.

examples/snippets/event-handler/http/advanced_fine_grained_responses.ts
dreamorosi: Let me take it for a spin and see how it works in practice before I insist further in either way.

svozza: Sounds good!

dreamorosi: Ok - I have looked into this and I think I agree with you on the fact that it being stricter is a feature and not a bug.

Ultimately you don't know what `await req.json()` will return so while it's true that it being `Promise<any>`  was allowing you to use it indiscriminately, it also masked the fact that you were potentially accessing something w/o knowing it's actual shape.

Now that it's `Promise<unknown>` it forces you to either acknowledge that you're typecasting, or do proper runtime validation.

I'm on board with that.

dreamorosi (approved):

PR #5352, opened 2026-06-13
feat(parser): add errorHandler option for inline parse error handling

### LINKED ISSUE #4191: Feature request: Injectable error handler for parser
### Use case

Today when there is a parsing error (except when using safeParse) the error is thrown uncaught. This could be potentially undesired, as we may want to instead return a specific response to the request - I'm thinking specifically API Gateway here but others might benefit too (I just don't have those use cases).

### Solution/User Experience

Add an error handler function to ParserOptions. In the decorator add a try/catch around the parse, and in the catch if the error handler function is present, use it and return its value otherwise rethrow the error.

This way we can handle errors JUST from the parse function, rather than some other upstream decorator catching ZodError (or whatever) from where exactly? maybe not just the parser.

### Alternative solutions

```Markdown

```

### Acknowledgment

### Future readers

Zelys-DFKH:
### Changes

Adds an optional `errorHandler` callback to `ParserOptions`, letting callers intercept parse failures and return a custom response instead of throwing.

The motivation comes from #4191: when `@parser` is stacked with other decorators that expect a fully-typed event, `safeParse: true` doesn't compose well -- the outer decorators see `ParsedResult` rather than the actual payload. An error handler keeps the event type clean while still providing a recovery path on parse failure. Thanks to @codyfrisch for walking through this use case in detail.

The callback receives the `ParseError`. If it returns a non-undefined value, that becomes the Lambda/Middy response and stops further processing. If it returns `undefined`, the original error is rethrown -- so you can handle specific failure cases without silently absorbing unexpected ones. Non-`ParseError` exceptions (e.g., from schema transforms) always propagate as-is.

Both the `@parser` class decorator and `parser()` Middy middleware support the new option.

**Issue number:** closes #4191

---


packages/parser/src/middleware/index.ts
```diff
@@ -38,10 +39,20 @@ const parser = <
...
+  const before = (request: MiddyLikeRequest): unknown => {
+    const { schema, envelope, safeParse, errorHandler } = options;
 
-    request.event = parse(request.event, envelope, schema, safeParse);
+    try {
+      request.event = parse(request.event, envelope, schema, safeParse);
+    } catch (error) {
+      if (errorHandler && error instanceof ParseError) {
+        const result = errorHandler(error);
+        if (result !== undefined) {
```
dreamorosi: The `errorHandler` type (`(error: ParseError) => unknown`) allows customers to pass an `async` function. When that happens, `result` is a `Promise`, which is never `undefined`, so the rethrow branch can never be taken. Worse: Middy.js `await`s the value returned by `before`, so an async handler that resolves to `undefined` means no early response is set and the Lambda handler runs with the raw, unparsed event - the parse failure is silently swallowed.

The same issue exists in the decorator version, where the pending promise would be returned as the handler response.

Since `parse` is synchronous, I'd constrain the callback to be synchronous at the type level and document that expectation. Either way, we need explicit test coverage for what happens when an async function is passed.

packages/parser/src/middleware/index.ts
```diff
@@ -38,10 +39,20 @@ const parser = <
...
+          return result;
```
dreamorosi: Returning a value from `before` makes Middy.js short-circuit the entire request: the handler is skipped, but so are the `after` phases of every other middleware attached to the stack - so for example Logger, Tracer, and Metrics would never run their cleanup logic. I don't think early-return from `before` is the right mechanism for this feature in the Middy.js flavor.

A more idiomatic approach would be to let the `ParseError` propagate and invoke the customer's `errorHandler` from an `onError` middleware step, setting `request.response` when it returns a value and leaving `request.error` untouched otherwise. That keeps the normal middleware lifecycle intact and composes with the `onError` handlers of other middlewares.

packages/parser/src/types/parser.ts
```diff
@@ -17,6 +18,7 @@ type ParserOptions<
   schema: TSchema;
   envelope?: TEnvelope;
   safeParse?: TSafeParse;
+  errorHandler?: (error: ParseError) => unknown;
```
dreamorosi: `safeParse: true` and `errorHandler` are contradictory: with `safeParse` the parser never throws, so the `errorHandler` is never invoked. As written, the types allow the combination and customers will reasonably expect the two to compose.

Let's make them mutually exclusive at the type level (e.g. a union of option shapes) so the invalid combination is rejected at compile time, and mention in the docs when to reach for one vs the other.

dreamorosi: The callback should also receive the original event, e.g. `(error: ParseError, event: unknown)`. For the motivating use case - building an error response - customers will often need data from the raw request such as a correlation ID, path, or headers.

Since this is public API, changing the signature after release would be a breaking change, so let's get it in now.

packages/parser/src/middleware/index.ts
```diff
@@ -38,10 +39,20 @@ const parser = <
 >(
   options: ParserOptions<TSchema, TEnvelope, TSafeParse>
 ): MiddlewareObj<ParserOutput<TSchema, TEnvelope, TSafeParse>> => {
```
dreamorosi: With this change, whatever `errorHandler` returns can become the function's response, but that's invisible to the type system - the response type parameter of `MiddlewareObj` isn't updated. For the Middy.js flavor this needs to be typed: we should thread the error handler's return type through (e.g. a generic inferred from the callback) so customers get a correctly typed response.

For the decorator flavor I'm okay keeping the return type loose.

dreamorosi (changes requested):
Hi @Zelys-DFKH, thank you for picking this up and for the well-structured PR - and apologies for the delay in getting to the review.

The direction matches what we discussed in #4191 and the implementation is easy to follow, however there are a few things we need to address before this can move forward - see the inline comments for the details and reasoning:

- an `async` `errorHandler` currently swallows parse failures and lets the handler run with the unparsed event (must fix)
- in the Middy.js flavor, short-circuiting from `before` skips every other middleware's `after` phase - I'd like the recovery to go through the `onError` phase instead
- `safeParse` and `errorHandler` should be mutually exclusive at the type level
- the callback should receive the original event alongside the error
- the Middy.js middleware's response type should reflect the error handler's return type

On top of that:

- the feature needs documentation in `docs/features/parser.md` (the `safeParse` section is the natural home for it) plus code snippets under `examples/snippets/parser`
- missing test cases: envelope + `errorHandler`, non-`ParseError` propagation, and the `safeParse` interaction
- `biome check` currently fails on formatting in `parser.middy.test.ts`

Happy to clarify or discuss any of these - thanks again for the contribution!

dreamorosi:
btw could you please leave a quick comment on #4191? We'd like to assign the issue to you for tracking, but GitHub only allows assigning users who have either commented on the issue or are collaborators — so right now we can't. Once you've commented, we'll assign it to you. Thanks!

Zelys-DFKH:
Thanks for the thorough review, @dreamorosi, and sorry for the slow turnaround. All five points are addressed below, along with the docs, tests, and the biome fix.

- errorHandler is now constrained to synchronous returns at the type level, with a runtime guard as a backstop. A type test and a runtime test cover the async case.
- Moved the Middy.js recovery into onError, exactly as you suggested: request.response is set directly (never returned), so the rest of the onError chain still runs. A regression test attaches another middleware's onError and confirms it still fires after parser recovers.
- safeParse and errorHandler are now mutually exclusive via a discriminated union; the invalid combination fails to compile, covered by a type test.
- errorHandler now receives (error, event) in both flavors.
- The Middy MiddlewareObj response type threads through the errorHandler's return type now. Kept the decorator loosely typed per your note.
- Docs live under the safeParse section in docs/features/parser.md, with snippets for both flavors under examples/snippets/parser.
- Added the three missing test cases (envelope + errorHandler, non-ParseError propagation, safeParse/errorHandler exclusion), plus a couple more that fell out of the onError change.
- Fixed the biome formatting failure.

One more thing came up while testing the onError change: since ParseError is public API, both flavors were treating any ParseError reaching them as this parser's own failure, including one a customer's handler might throw for unrelated reasons. Fixed that by scoping recovery to this parser instance's own before-phase failure (a WeakSet keyed on the request for the middleware, a narrower try/catch around just the parse() call for the decorator), with regression tests for both.

Let me know if I've missed anything, or if you'd rather take a different approach on any of these.

svozza:
There is a SonarQube issue (very simple fix), can you address it?

packages/parser/src/errorHandler.ts
```diff
@@ -0,0 +1,40 @@
...
+    typeof result === 'object' &&
+    result !== null &&
+    typeof (result as { then?: unknown }).then === 'function'
+  ) {
+    throw new TypeError(
+      'errorHandler must return synchronously; async errorHandler functions are not supported'
+    );
+  }
+
+  return result ?? NO_RECOVERY;
```
dreamorosi: The switch to `??` here means a `null` return from the `errorHandler` now also triggers a rethrow, while the docs, the JSDoc above, and the `ErrorHandler` type docs all state that only `undefined` rethrows. `null` can be a legitimate response value a customer intentionally returns, so I'd rather keep the documented contract:

```ts
return result === undefined ? NO_RECOVERY : result;
```

If SonarQube complains about the ternary again, an explicit `if (result === undefined) return NO_RECOVERY;` works too.

Could you also add a small test covering an `errorHandler` that returns `null`, so the behavior is locked in?

dreamorosi (changes requested):
Thanks for the quick and thorough turnaround @Zelys-DFKH - all the points from the previous review have been addressed exactly as discussed: the sync-only `errorHandler` constraint, the `onError`-based recovery in the Middy.js flavor, the mutually exclusive types, the event in the callback signature, the threaded response type, plus the docs, snippets, and the extra test coverage. We're getting close to merging this.

There's one small inconsistency left, introduced by the SonarQube follow-up - see the inline comment. Once that's sorted I think we're good to go.

packages/parser/src/errorHandler.ts
Zelys-DFKH: Thanks for catching this — `??` broke the null/undefined distinction. Reverted to the explicit `if (result === undefined)` check, added a null-returning test for both flavors, and left a comment explaining why so it doesn't happen again. All green locally. Ready for another look.

dreamorosi (approved):
Thank you for addressing the review comments and implementing the feature!

Issue #3988, opened 2025-05-28
Maintenance: align implementation of Bedrock Agent Function resolver


dreamorosi:
### Summary

After an internal bug bash for this upcoming feature we have noticed a few inconsistencies in the implementations and decided to converge in other areas. Because of this we need to make some light changes to the implementation.

### Why is this needed?

We need to implement these changes to ensure feature parity & alignment with other Powertools for AWS runtimes and give a consistent experience.

### Which area does this relate to?

Event Handler

### Acknowledgment

- [x] This request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.powertools.aws.dev/lambda/typescript/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Python](https://github.com/aws-powertools/powertools-lambda-python/), [Java](https://github.com/aws-powertools/powertools-lambda-java/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

### Future readers

Please react with 👍 and your use case to help us understand customer demand.

PR #5363, opened 2026-06-19
fix(event-handler): preserve original string for invalid Bedrock Agent number params

### LINKED ISSUE #5362: Bug: Bedrock Agent number/integer params lose their value when not numeric (no string fallback like Python)
### Expected Behaviour

When a Bedrock Agent sends a parameter typed as `number` or `integer` whose `value` is not a valid numeric string, `BedrockAgentFunctionResolver` should preserve the original string value and pass it to the tool function — the same way the Python runtime does.

The Python `BedrockAgentFunctionResolver` wraps the numeric coercion in a `try`/`except` and falls back to the original value ([`bedrock_agent_function.py`](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/event_handler/bedrock_agent_function.py#L212-L221)):

```python
elif param_type == "integer":
    try:
        parameters[param.name] = int(param.value)
    except (ValueError, TypeError):
        parameters[param.name] = param.value
elif param_type == "number":
    try:
        parameters[param.name] = float(param.value)
    except (ValueError, TypeError):
        parameters[param.name] = param.value
```

### Current Behaviour

In the TypeScript resolver the coercion is unconditional ([`BedrockAgentFunctionResolver.ts#L227-L230`](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/event-handler/src/bedrock-agent/BedrockAgentFunctionResolver.ts#L227-L230)):

```ts
case 'number':
case 'integer': {
  toolParams[param.name] = Number(param.value);
  break;
}
```

`Number('not-a-number')` returns `NaN`, and `NaN` serializes to `null` via `JSON.stringify`. So when a tool echoes such a parameter back, the original value is silently lost and the tool receives `NaN` instead of the string the agent sent.

### Code snippet

```ts
const app = new BedrockAgentFunctionResolver();
app.tool(async (params) => params.arg, {
  name: 'echo',
  description: 'Echoes the received parameter',
});

// Bedrock sends a `number`-typed parameter whose value is not numeric
const event = {
  messageVersion: '1.0',
  agent: { name: 'agent', id: 'id', alias: 'alias', version: '1' },
  inputText: '...',
  sessionId: '...',
  actionGroup: 'group',
  function: 'echo',
  parameters: [{ name: 'arg', type: 'number', value: 'not-a-number' }],
  sessionAttributes: {},
  promptSessionAttributes: {},
};

// result.response.functionResponse.responseBody.TEXT.body === 'null'   (value lost)
// Python runtime would return '"not-a-number"'
```

### Steps to Reproduce

1. Register a tool that returns one of its parameters.
2. Invoke the resolver with that parameter typed `number` (or `integer`) but a non-numeric `value`.
3. Observe the response body is `null` (TS) instead of the original string (Python).

### Possible Solution

Mirror the Python behavior with a minimal, branch-local fallback:

```ts
case 'number':
case 'integer': {
  const parsed = Number(param.value);
  toolParams[param.name] = Number.isNaN(parsed) ? param.value : parsed;
  break;
}
```

This keeps the existing valid-number behavior unchanged and only adds a fallback for the `NaN` case.

latest

nodejs22.x

### Packaging format used

npm

### Debugging logs

_No response_

---

Context: this is a small follow-up to the cross-runtime alignment tracked in #3988 (released in v2.21.0) — the `number`/`integer` string fallback that the Python runtime already implements was not carried over to the TypeScript resolver.

_This report was prepared with the help of an AI agent (Claude Code); a human reviewed the reproduction, the proposed fix, and the test results before filing._

kimnamu:
## Summary

First, thank you to the maintainers for Powertools for AWS Lambda — it makes building Lambda functions a pleasure, and the cross-runtime parity work is much appreciated.

This is a small follow-up to that alignment effort. The `BedrockAgentFunctionResolver` coerces `number`/`integer` parameters with `Number(param.value)` **unconditionally**. When a Bedrock Agent sends a `number`/`integer`-typed parameter whose value is not a valid numeric string, `Number(...)` returns `NaN`, which serializes to `null` via `JSON.stringify` — the original value sent by the agent is silently lost.

The Python runtime already guards against this: it wraps `int()`/`float()` in `try`/`except` and falls back to the original string ([`bedrock_agent_function.py` at `L212-L221`](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/event_handler/bedrock_agent_function.py#L212-L221)). This PR carries the same fallback over to the TypeScript resolver, so the two runtimes behave consistently.

### Changes

`packages/event-handler/src/bedrock-agent/BedrockAgentFunctionResolver.ts` — for the `number`/`integer` case, fall back to the original string when `Number(...)` yields `NaN`:

```ts
case 'number':
case 'integer': {
  const parsed = Number(param.value);
  // Fall back to the original string when the value is not a valid
  // number (e.g. `Number('abc')` is `NaN`, which serializes to `null`).
  // This matches the behavior of the Python runtime.
  toolParams[param.name] = Number.isNaN(parsed) ? param.value : parsed;
  break;
}
```

This is the minimal change required: it touches only the `NaN` branch and leaves every other path untouched.

### Before / After

Tool that echoes back a `number`-typed parameter with value `'not-a-number'`:

| Case | Before | After |
|------|--------|-------|
| `number`/`integer` param, **non-numeric** value (`'not-a-number'`) | ❌ `NaN` → response body `null` (value lost) | ✅ original string `"not-a-number"` preserved (matches Python) |
| `number`/`integer` param, **valid** value (`'42'`) | ✅ `42` | ✅ `42` (unchanged) |
| `boolean` / `string` / `array` params | ✅ unchanged | ✅ unchanged |
| Public API / method signatures / valid-number behavior | — | ✅ unchanged |

**Issue number:** Closes #5362

This continues the Bedrock Agent Function resolver alignment effort, which was previously tracked under the resolver parity work released in v2.21.0.

## Testing

Added a parametrized regression test (`it.each(['number', 'integer'])`) in `packages/event-handler/tests/unit/bedrock-agent/BedrockAgentFunctionResolver.test.ts` asserting the original string is preserved.

**Green (with fix):**

```
✓ tests/unit/bedrock-agent/BedrockAgentFunctionResolver.test.ts (24 tests)
 Test Files  1 passed (1)
      Tests  24 passed (24)
```

**Red (fix reverted via `git stash`, new test kept) — proves the test catches the bug:**

```
 FAIL  ... > preserves the original string when a 'number' parameter is not a valid number
 FAIL  ... > preserves the original string when a 'integer' parameter is not a valid number
AssertionError: expected 'null' to deeply equal '"not-a-number"'
Expected: ""not-a-number""
Received: "null"
      Tests  2 failed | 22 passed (24)
```

Local gates:

- `npx biome lint` on both changed files → clean.
- Coverage on the changed file: `Statements 100% (33/33), Branches 100% (23/23), Functions 100% (4/4), Lines 100% (33/33)` — the new branch is fully covered.

---

_This contribution was prepared with the help of an AI agent (Claude Code); a human reviewed the change, the rationale, and the test results before submitting._

### DIFF
--- a/packages/event-handler/src/bedrock-agent/BedrockAgentFunctionResolver.ts
+++ b/packages/event-handler/src/bedrock-agent/BedrockAgentFunctionResolver.ts
@@ -226,7 +226,11 @@ class BedrockAgentFunctionResolver {
         }
         case 'number':
         case 'integer': {
-          toolParams[param.name] = Number(param.value);
+          const parsed = Number(param.value);
+          // Fall back to the original string when the value is not a valid
+          // number (e.g. `Number('abc')` is `NaN`, which serializes to `null`).
+          // This matches the behavior of the Python runtime - see #3988.
+          toolParams[param.name] = Number.isNaN(parsed) ? param.value : parsed;
           break;
         }
         // this default will also catch array types but we leave them as strings
--- a/packages/event-handler/tests/unit/bedrock-agent/BedrockAgentFunctionResolver.test.ts
+++ b/packages/event-handler/tests/unit/bedrock-agent/BedrockAgentFunctionResolver.test.ts
@@ -451,6 +451,44 @@ describe('Class: BedrockAgentFunctionResolver', () => {
     );
   });
 
+  it.each([
+    { type: 'number' as const },
+    { type: 'integer' as const },
+  ])(
+    'preserves the original string when a $type parameter is not a valid number',
+    async ({ type }) => {
+      // Prepare
+      const toolFunction: ToolFunction<{ arg: number | string }> = async (
+        params,
+        _options
+      ) => params.arg;
+
+      const toolParams: Configuration = {
+        name: 'echo',
+        description: 'Echoes the received parameter',
+      };
+
+      const parameters: Parameter[] = [
+        { name: 'arg', type, value: 'not-a-number' },
+      ];
+
+      const app = new BedrockAgentFunctionResolver();
+      app.tool(toolFunction, toolParams);
+
+      // Act
+      const actual = await app.resolve(
+        createEvent(toolParams.name, parameters),
+        context
+      );
+
+      // Assess
+      expect(actual.response.function).toEqual(toolParams.name);
+      expect(actual.response.functionResponse.responseBody.TEXT.body).toEqual(
+        '"not-a-number"'
+      );
+    }
+  );
+
   it('correctly parses string parameters', async () => {
     // Prepare
     const toolFunction: ToolFunction<{ arg: string }> = async (


kimnamu:
Thanks for the automated checks 🙏 A quick note on the **Related Issue Check**: the PR is correctly linked to #5362 (`closes #5362`, and it shows under the PR's linked issues), but the check appears to require the linked issue to be triaged/labeled by a maintainer first. As an external contributor I don't have label permissions, so I've left #5362 for maintainer triage.

I opened #5362 specifically because the original parity issue (#3988) is already closed/released, so I couldn't link to it directly — #5362 captures this remaining TypeScript-side gap. Happy to adjust the linkage if you'd prefer a different issue reference. The change itself is a small, test-covered cross-runtime parity fix (matches the Python `try/except` string fallback).

dreamorosi (approved):
Thank you for the PR and for following the correct process & addressing CI checks.

kimnamu:
Thanks for the nudge from the automation 🙏 I tracked down why the **Related Issue Check** kept failing even though `closes #5362` is present and shows under the PR's linked issues.

The body also contained a bare `#3988` reference inside the code-comment illustration. The automation scans the body for every `#<number>` token and expects each to be a linked issue — so the (already-closed) `#3988` was being counted as a second, unlinked reference, which tripped the check.

I've rewritten that line so it no longer uses a bare `#<number>` token; the only remaining issue reference is the real closing link `closes #5362`. No code changes — body only.

_This follow-up was prepared with the help of an AI agent (Claude Code); a human reviewed it before posting._

kimnamu:
The Related Issue Check is green now ✅ — the earlier failure was a stale run that didn't re-evaluate after I cleaned up the issue reference. The PR body now links exactly one issue (`Closes #5362`), and the check passes on the latest commit.

This is approved and the merge state is clean — ready whenever you'd like to land it. Thanks again for the review!

_(This note was prepared with the help of an AI agent (Claude Code); I verified the check status and merge state before posting.)_

PR #5399, opened 2026-07-07
feat(parameters): add support for AppConfig Agent

### LINKED ISSUE #5312: Feature request: add support for AppConfig Agent in Parameters
### Use case

As a customer I want to use the Parameters utility to fetch configuration objects from AppConfig using the [AWS AppConfig Agent](https://docs.aws.amazon.com/appconfig/latest/userguide/appconfig-integration-lambda-extensions.html) which runs as Lambda extension inside my execution environment.

### Solution/User Experience

Similarly to what customers can do today with the getAppConfig helper[^1], customers should be able to easily fetch a configuration object using a helper function like this:

```ts
import { getConfig } from '@aws-lambda-powertools/parameters/appconfig-agent';

// Retrieve a configuration
const config = await getConfig('my-configuration', {
  environment: 'my-env',
  application: 'my-app',
  transform: 'json'
});

  console.log(config);
};
```

_**Note**: the import has two differences from the existing one: 1/ function name becomes `getConfig`, 2/ the sub-path export becomes `/parameters/appconfig-agent`._

Since caching is handled by the extension and there's no AWS SDK client to tune, we will neither implement nor offer a corresponding provider class to AppConfigProvider[^2]. This allows us to create a lean implementation that always fetches the configuration object from the extension's endpoint, transforms it, and returns it to the caller.

Similarly, since caching is handled entirely by the AppConfig Agent Lambda extension, the `getConfig` helper function **will not** support the `maxAge`[^3] and `forceFetch`[^4] options. The `transform`[^5] parameter will still be supported, together with the current formats.

Customers using the `getConfig` helper function will be able to configure both caching and pre-fetching by using two environment variables exposed by the AppConfig Agent:
- `AWS_APPCONFIG_EXTENSION_PREFETCH_LIST` - Tells the extension to fetch config during INIT (before your handler runs), reducing cold-start latency. Value is the path to your config, e.g. /applications/my_app/environments/my_env/configurations/my_config.
- `AWS_APPCONFIG_EXTENSION_POLL_TIMEOUT_MILLIS` (default 3000) - Max time the extension waits for an AppConfig response during a poll. If it times out, the cached value is used.

So for example, if you set:
```sh
AWS_APPCONFIG_EXTENSION_POLL_INTERVAL_SECONDS=120
AWS_APPCONFIG_EXTENSION_PREFETCH_LIST=/applications/myapp/environments/prod/configurations/feature-flags
```

The extension will prefetch during cold start and then check for updates at most every 2 minutes on subsequent invocations.

It's important to note that the prefetch list is purely an optimization for cold starts in that it tells the extension to start fetching before your handler runs. At runtime, your function can request any configuration by calling the local HTTP endpoint:
`GET http://localhost:2772/applications/{app}/environments/{env}/configurations/{config}`.

If the requested configuration isn't already in the cache (because it wasn't prefetched), the extension will call AppConfig on the spot, cache the result, and return it to your function. Subsequent calls for the same configuration will be served from cache (subject to the poll interval for freshness).

So the prefetch list doesn't limit what you can access — it only controls what gets pre-warmed into the cache during INIT. Any configuration your IAM permissions allow can be fetched on-demand at any point during invocation.

Prefetching is particularly interesting for Powertools customers as it unlocks a new level of optimization not available with the current AppConfigProvider-based approach, specifically allowing parallelism during cold start.

**Timeline comparison**

With prefetch (AWS_APPCONFIG_EXTENSION_PREFETCH_LIST):

```sh
|-- Extension INIT --|
|   prefetch starts → AppConfig network call  |
|                                             |
|-- Runtime INIT (in parallel) ---------------|
|   module loading, top-level code...         |
|                                             |
|-- Handler invoked --------------------------|
|   GET localhost:2772 → already cached ✓     |
```

The extension begins fetching from AppConfig concurrently with your Node.js runtime initializing (parsing modules, running top-level code). By the time your code asks for the config, it's likely already in the cache.

Fetching in top-level code (no prefetch):

```sh
|-- Extension INIT --|
|   (no prefetch)    |
|                    |
|-- Runtime INIT ----|------------------------------|
|   module loading...                               |
|   await fetch("http://localhost:2772/...") -----> | calls AppConfig (blocking)
|   ...waits...                                     |
|   config received ←                               |
|                                                   |
|-- Handler invoked --------------------------------|
```

Your top-level await hits the extension, which has nothing cached, so it makes a synchronous (from your perspective) call to AppConfig. Your INIT blocks on the full network round-trip.

**Practical impact**
- Prefetch overlaps the AppConfig network latency with your module/dependency loading — effectively "free" if your module init takes longer than the AppConfig call.
- Top-level fetch adds the AppConfig latency on top of your module loading time.
- 
For a typical Node.js function with several dependencies (SDK, ORM, etc.), prefetch can shave hundreds of milliseconds off cold starts because the network call happens while Node is busy parsing and loading your `node_modules`.

**When the difference is negligible**
- If your function has minimal dependencies (fast module load), the prefetch may not have finished by the time your code runs — in which case the extension blocks until the in-flight request completes, similar to a top-level fetch.
- On warm invocations, both behave identically (served from cache).

A note on testing: the `getConfig` helper function should be able to detect when it's not running in AWS Lambda and either become a no-op or return a fixed value. This is helpful for testing and local environments.

As part of the scoping for this issue, also check the benchmark analysis in #3754.
---

[^1]: [`getAppConfig`](https://docs.aws.amazon.com/powertools/typescript/latest/features/parameters/#fetching-app-configurations)
[^2]: [`AppConfigProvider`](https://docs.aws.amazon.com/powertools/typescript/latest/features/parameters/#appconfigprovider)
[^3]: [Adjusting cache TTL](https://docs.aws.amazon.com/powertools/typescript/latest/features/parameters/#adjusting-cache-ttl)
[^4]: [Always fetching the latest](https://docs.aws.amazon.com/powertools/typescript/latest/features/parameters/#always-fetching-the-latest)
[^5]: [Deserializing values with transform parameter](https://docs.aws.amazon.com/powertools/typescript/latest/features/parameters/#deserializing-values-with-transform-parameter)

### Alternative solutions

Since the internal implementation and mechanics of retrieving configurations is more similar to the recently released [metadata feature](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/commons/src/metadata.ts) than to the existing AWS SDK client-based approach we have decided to forego the existing BaseProvider approach used in Parameters.

While we considered using the same pattern in the name of consistency, we realized that following the same structure would've created a lot of indirection and boilerplate code which would've ended up in customers bundles.

### Acknowledgment

### Future readers

svozza:
## Summary

Adds a new `getConfig` function exported from `@aws-lambda-powertools/parameters/appconfig-agent` that retrieves configurations from the [AWS AppConfig Agent Lambda extension](https://docs.aws.amazon.com/appconfig/latest/userguide/appconfig-integration-lambda-extensions.html)'s local HTTP endpoint. Since caching, polling, and prefetching are handled by the agent, the implementation is a lean standalone function modeled on the commons `metadata` module rather than the `BaseProvider` pattern, keeping customer bundles small.

### Changes

- Add `getConfig` function in `packages/parameters/src/appconfig-agent/` that fetches from `http://localhost:2772/applications/{app}/environments/{env}/configurations/{config}` using native `fetch` with a configurable timeout (default 3000ms)
- Support `json` and `binary` transforms by reusing the existing `transformValue` utility; no `maxAge`/`forceFetch` options since caching is handled by the agent
- Return `undefined` when not running in a Lambda environment (local development/testing)
- Respect the `AWS_APPCONFIG_EXTENSION_HTTP_PORT` environment variable and fall back to `POWERTOOLS_SERVICE_NAME` when `application` is not provided
- Add `./appconfig-agent` and `./appconfig-agent/types` subpath exports with typed return values (`string`, `JSONValue`, or explicit generic)
- Add unit tests (100% coverage), type tests, and an e2e test suite that deploys the agent extension layer alongside AppConfig profiles
- Document the new function in the Parameters feature page, README, and add a doc snippet

**Issue number:** closes #5312

---


svozza:
E2E tests pass: https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/28928374948

packages/parameters/src/appconfig-agent/getConfig.ts
```diff
@@ -0,0 +1,160 @@
...
+  let value: string;
+  try {
+    const res = await fetch(
+      `http://localhost:${port}/applications/${encodeURIComponent(application)}/environments/${encodeURIComponent(options.environment)}/configurations/${encodeURIComponent(name)}`,
+      {
+        signal: AbortSignal.timeout(options.timeout ?? 3000),
+      }
+    );
+    value = await res.text();
+    if (res.status === 404) {
```
svozza, line 138: Note: we always throw if a parameter is not found. If we want to do something different and say return `undefined`, then we need to revisit the behaviour when we are not running in a lambda because currently that is what `undefined` means. 

The issue here is I can't think of a different way to represent not running in a lambda other than throwing: we can't do what we do in `metdata` by just returning an empty object because that is a valid return value for a parameter. 

Open to suggestions here: maybe allowing users to supply a default value that means they can avoid this `ParameterNotFoundError` error.  There is also the issue that we are missing the `throwOnMissing` behaviour that other parameter types have, which could confuse users.

packages/parameters/tests/e2e/appConfigAgent.test.functionCode.ts
```diff
@@ -0,0 +1,53 @@
...
+      getConfig(freeFormJsonName, { ...baseOptions, transform: 'json' }),
+      // Test 3 - get a free-form base64-encoded plain text and apply binary transformation (should return a decoded string)
+      getConfig(freeFormBase64encodedPlainText, {
+        ...baseOptions,
+        transform: 'binary',
+      }),
+      // Test 4 - get a feature flag and apply json transformation (should return an object with the evaluated flag values)
+      getConfig(featureFlagName, { ...baseOptions, transform: 'json' }),
+      // Test 5 - get a configuration that does not exist (should throw a ParameterNotFoundError)
+      getConfig('does-not-exist', baseOptions).then(
```
dreamorosi: Non-blocking style nit: we generally prefer `async`/`await` over `Promise.then` / `Promise.catch` for readability and consistency. Since this is just capturing the expected missing-config error for the e2e fixture, could we move it into a small local async helper (for example `getMissingConfigErrorName`) and call that from `Promise.all` instead of using `.then()` here?

packages/parameters/src/appconfig-agent/getConfig.ts
dreamorosi: This for me is the biggest one-way door / question mark to be honest.

On one hand, I understand why we're throwing based on your explanation above, but on the other hand I don't think it's in line with the rest of the library and that's a minus - which you also have pointed out.

What do you think about not throwing unless `throwOnMissing` is present and also returning a clearly fake object like:

```ts
{
  "testing": true
}
```

or something like this - which would have to properly document.

Conversely, what would it take for us to allow customers to specify what to return? Would having something like `POWERTOOLS_APPCONFIG_AGENT_RETURN_VALUE="{\"foo\": true}"` be an option?

packages/parameters/tests/e2e/appConfigAgent.test.functionCode.ts
svozza: Moved it into a local `getMissingConfigErrorName` async helper with a try/catch, called from `Promise.all`. Addressed in 6d4d4186.

packages/parameters/src/appconfig-agent/getConfig.ts
svozza: Went with a combination of both of your suggestions, implemented in 113fc081:

1. **404 now returns `undefined` by default**, in line with the rest of the library (DynamoDB/AppConfig provider semantics). This also means users get per-call fallback values for free via nullish coalescing — `await getConfig(...) ?? DEFAULT_FLAGS` — which we think covers the 'return a specific value when not found' use case with full type safety and no new API.
2. **`throwOnMissing: true`** opts into `ParameterNotFoundError`, same flag and semantics as the provider-based functions, including the `GetMaybeUndefined` return-type narrowing.
3. **`POWERTOOLS_APPCONFIG_AGENT_RETURN_VALUE`** for the not-in-Lambda case: rather than a hardcoded fake object (which would be a type lie for the no-transform/`binary` cases and a dangerously plausible `{}` for `json`), the env var value is treated as the *agent response body* and runs through the normal transform pipeline. Users pick a shape-correct value for their config; unset means `undefined` as before. Documented in a new 'Testing getConfig' section with a snippet.

On the fake-object idea specifically: anything the library invents can't satisfy the `string` return branches, and for `json` an empty-ish object reads as 'all flags off' with no failure signal — the env var sidesteps both since the customer supplies the value.

svozza: Small follow-up in 1560d724: a review pass flagged that the `throwOnMissing` return-type narrowing was unsound outside Lambda — the local path returned `undefined` when no `POWERTOOLS_APPCONFIG_AGENT_RETURN_VALUE` is set, even with `throwOnMissing: true`, while the type excludes `undefined`.

Fixed by making the local path honor `throwOnMissing` too: when not running in Lambda, no local value configured, and `throwOnMissing: true`, we now throw the same `ParameterNotFoundError` as the agent-404 path. This is semantically consistent rather than just type-appeasing — the not-in-Lambda environment behaves as one where no configurations exist unless injected, so 'no value + throwOnMissing' means an exception everywhere. The contract is now uniform: with `throwOnMissing: true` you always get a value or a `ParameterNotFoundError`, and the narrowed type is sound in every environment.

One implication for tests: a test exercising a `throwOnMissing: true` code path must either set `POWERTOOLS_APPCONFIG_AGENT_RETURN_VALUE` or assert the throw — which is the opted-into fail-fast behavior working as designed. Docs and unit tests updated accordingly.

dreamorosi (approved):

dreamorosi:
Thanks for working on this - and glad we were able to smooth out the non-Lambda experience.

svozza:
Yeah, the non-lambda part was trickier than I expected but I think we landed on a nice solution in the end. Thank you for helping iterate on the design!

PR #5405, opened 2026-07-07
ci: slow CDK stack-event polling to avoid CFN throttling

### LINKED ISSUE #5404: Maintenance: patch CDK toolkit stack-event polling interval to stop e2e CloudFormation throttling
### Summary

Every recent "Run e2e Tests" run loses one or more matrix jobs to CloudFormation `Throttling: Rate exceeded` (`CDK_TOOLKIT_E5500` while monitoring a stack, then a fatal `CDK_TOOLKIT_E7900` on destroy). Failures are random per run — the same code passes in sibling matrix cells — because the cause is account-level API rate exhaustion, not the code under test.

Root cause, confirmed via CloudTrail: `@aws-cdk/toolkit-lib`'s `StackActivityMonitor` polls `DescribeStackEvents` every 2 s per active stack operation and exposes no way to change that. Our matrix is 36 jobs (9 packages × 2 Node versions × 2 architectures, `max-parallel: 25`), all starting within ~2 s, each deploying several stacks concurrently. That's ~50–100 concurrent stack operations driving 25–50 requests/sec into the shared account-level CloudFormation read-API token bucket. CloudTrail for one run window shows 1,550+ `DescribeStackEvents` calls at a sustained ~8–10/sec with `ThrottlingException`s every 1–2 s at peak. The toolkit's 7-attempt retry strategy can't outlast that, so destroys exhaust retries and the suite's `afterAll` throws.

### Why is this needed?

E2E is our release gate and it's currently unreliable: recent runs consistently lose matrix cells to throttling, and each re-run costs a full matrix wall-clock. This is a threshold effect — every package added to the matrix (e.g. data-masking in #5143) adds stacks and pushes peak load further past the limit — so it gets worse with growth, not better. Mechanisms that scale with the fleet beat one-time tuning here.

Tests, Automation

### Solution

Interim fix, ready on `fix/e2e-cfn-throttling-polling-patch`: a module-load patch in `packages/testing` (`patchCdkStackEventPolling.ts`, invoked from `TestStack.ts`) wraps `StackActivityMonitor.prototype.start` and raises the polling interval from 2 s to 10 s — a 5× cut in `DescribeStackEvents` volume. This costs nothing observable: our `ioHost` already surfaces progress at 10 s intervals. If a future toolkit release moves the internal module, the patch degrades to a console warning instead of breaking e2e.

Durable fix: we filed aws/aws-cdk-cli#1709 for a public `pollingInterval` option and offered to implement it. When it ships, we swap the monkeypatch for the supported option in `TestStack.ts` and delete `patchCdkStackEventPolling.ts`.

If upstream stalls and the patch proves fragile across toolkit upgrades, the fallback is to replace toolkit-lib in `packages/testing` with a minimal deploy/destroy on `cdk-assets` + AWS SDK `DescribeStacks` waiters, which don't call `DescribeStackEvents` at all.

### Acknowledgment

### Future readers

svozza:
## Summary

Every recent "Run e2e Tests" run loses one or more matrix jobs to CloudFormation `Throttling: Rate exceeded` (`CDK_TOOLKIT_E5500`/`CDK_TOOLKIT_E7900`). `@aws-cdk/toolkit-lib`'s `StackActivityMonitor` polls `DescribeStackEvents` every 2 s per active stack operation with no public option to change it; with 36 matrix jobs deploying stacks concurrently, that saturates the account-level CloudFormation read-API rate. This PR patches the monitor at module load to poll every 10 s instead — a 5× cut in call volume with no observable cost, since our `ioHost` only surfaces progress every 10 s anyway.

This is an interim fix: we've asked for a public `pollingInterval` option upstream (aws/aws-cdk-cli#1709) and offered to implement it. Once it ships, the patch gets replaced by the supported option in `TestStack.ts`. If a future toolkit release moves the internal module, the patch degrades to a console warning rather than breaking e2e.

### Changes

- Add `packages/testing/src/patchCdkStackEventPolling.ts`: wraps `StackActivityMonitor.prototype.start` via a deep CJS require into toolkit-lib internals and raises `pollingInterval` from 2 s to 10 s; no-ops with a warning if the internal module can't be resolved
- Invoke the patch at module load in `packages/testing/src/TestStack.ts`, so every e2e suite gets it with no per-package changes

**Issue number:** closes #5404

---

### DIFF
--- a/packages/testing/src/TestStack.ts
+++ b/packages/testing/src/TestStack.ts
@@ -9,7 +9,10 @@ import {
 } from '@aws-cdk/toolkit-lib';
 import { App, Stack } from 'aws-cdk-lib';
 import { generateTestUniqueName } from './helpers.js';
+import { patchCdkStackEventPolling } from './patchCdkStackEventPolling.js';
 import type { TestStackProps } from './types.js';
+
+patchCdkStackEventPolling();
 
 const testConsole = new Console({
   stdout: process.stdout,
--- a/packages/testing/src/patchCdkStackEventPolling.ts
+++ b/packages/testing/src/patchCdkStackEventPolling.ts
@@ -1,0 +1,66 @@
+import { createRequire } from 'node:module';
+import { dirname, join } from 'node:path';
+
+/**
+ * How often the CDK toolkit polls `DescribeStackEvents` for each in-flight
+ * stack operation once patched. The default is 2s; our e2e `ioHost` only
+ * surfaces progress every 10s anyway, so nothing is lost by polling slower.
+ */
+const PATCHED_POLLING_INTERVAL_MS = 10_000;
+
+let patched = false;
+
+/**
+ * Slow down the CDK toolkit's CloudFormation stack-event polling.
+ *
+ * `@aws-cdk/toolkit-lib` polls `DescribeStackEvents` every 2s per active stack
+ * operation and exposes no way to configure this. During e2e runs the CI matrix
+ * deploys/destroys dozens of stacks concurrently, which saturates the
+ * account-level CloudFormation read API rate and randomly fails jobs with
+ * `Throttling: Rate exceeded` (CDK_TOOLKIT_E5500/E7900).
+ *
+ * Until the interval is configurable upstream, we reach into the toolkit's
+ * internals and raise it on every monitor before it starts. The patch is a
+ * no-op (with a warning) if the internal module moves in a future release.
+ */
+const patchCdkStackEventPolling = (): void => {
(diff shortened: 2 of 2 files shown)


dreamorosi (approved):

dreamorosi:
Can you please link the upstream issue/PR in a comment here, so we can keep tabs on it?

svozza:
I was going to create a separate issue to remove this workaround and link to it in there. Having multiple links to issues in a PR breaks the `closes: #<issue-number>` automation.

PR #5422, opened 2026-07-08
feat(idempotency): warn when persistence layer reconfiguration is ignored

### LINKED ISSUE #5409: Docs: document persistence layer reuse across multiple idempotent operations
### Use case

I encountered some surprising behaviour when I built a function which performed two sequential operations wrapped with makeIdempotent but with different JMESPath expressions (as both were handling different input data). In a nutshell - the second operation did not work more than once before the key expired even though we confirmed that the input data _was_ changing between invocations.

Digging into the code, it looks like this is because `makeIdempotent` actually [**mutates** the `persistenceStore`](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/idempotency/src/IdempotencyHandler.ts#L88) with the values in `config` the first time it is invoked, but there is an [explicit guard](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/idempotency/src/persistence/BasePersistenceLayer.ts#L60) preventing it from being mutated more than once (except for the `idempotencyKeyPrefix`).

A succinct example of how to reproduce this:

```ts
  tableName: 'idempotencyTableName',
});

const processUser = makeIdempotent(
  userHandler,
  {
    persistenceStore,
    config: {eventKeyJmesPath: '["userId"]'}
  }
);

const processProduct = makeIdempotent(
  productHandler,
  {
    persistenceStore,
    config: {eventKeyJmesPath: '["productId"]'}
  }
);

processUser({userId: randomUUID()}); // Succeeds always
processProduct({productId: randomUUID()}); // Succeeds once only, until the idempotency TTL expires
```

In the above code, what actually happens is the second invocation to `makeIdempotent` does not reconfigure the `persistenceStore` and the store retains the previous eventKeyJmesPath (userId). As the `product` does not contain `userId` it evaluates to `null` which produces the exact same hash value every invocation.

It feels wrong to file as a bug as it has clearly been designed this way with intent. I'd like to open a feature discussion to ask if it can be considered as something to change, or at least put more explicit information/guards around.

### Solution/User Experience

Immediate need: Explicit documentation around this caveat today and the intended usage of the persistence layer and configuration in combination. Right now I would argue the documentation implies reuse of persistence stores as they are almost always declared as `const` outside of `makeIdempotent` and injected in. I think it's a very subtle failure to run into and would benefit from more explicit changes to how it's handled: If the persistence layer threw an error instead of silently skipping the reconfiguration that would have been a better experience as we would have had immediate feedback.

Longer term, I'd personally like to see no mutation to the inputs, but at the very least the ability to reuse a persistence store across multiple calls to the idempotent operation factory, as it basically represents static unchanging data that I imagine would not typically be changed between multiple operations in the same lambda/context.

I appreciate the suggestions to throw errors or change behaviour to avoid this mutation would constitute breaking changes and would likely require a major version bump - though I cannot imagine a scenario where someone is presently reliant on this behaviour.

### Alternative solutions

```Markdown
The alternative I'm using today is just explicitly re-declaring the persistence store and making sure I keep the input configuration to that static and shared, along with the dynamo client. Explicit guidance to use it in this way could help people avoid this silent failure today if no changes are acceptable.
```

### Acknowledgment

### Future readers

dreamorosi:
## Summary

A persistence layer instance is configured only once, so when the same instance is shared across multiple idempotent operations with different idempotency configurations, all but the first configuration are silently ignored. As reported in #5409, this can lead to subtle bugs, e.g. an `eventKeyJmesPath` that doesn't match the payload of the second operation, producing the same idempotency key on every invocation.

This PR emits a console warning - once per persistence layer instance - when an already configured instance receives a configuration that materially differs from the one applied, naming the mismatched settings so customers get immediate, actionable feedback instead of silent misbehavior.

### Changes

- Add a diff-aware, warn-once check in `BasePersistenceLayer.configure()`: strict equality on the scalar settings snapshotted during configuration (`eventKeyJmesPath`, `payloadValidationJmesPath`, `throwOnNoIdempotencyKey`, `expiresAfterSeconds`, `useLocalCache`, `maxLocalCacheSize`, `hashFunction`) plus a constructor identity check for custom JMESPath functions
- Store `maxLocalCacheSize` on the instance so it can be compared (previously only passed to the LRU cache)
- Add unit tests covering: warning emitted with mismatched settings named, warn-once behavior, no warning for equivalent configurations, and custom JMESPath functions mismatch
- Fix three existing tests in `makeIdempotent.test.ts` that shared a module-level persistence store across different configurations - an instance of the exact footgun this warning detects

**Issue number:** closes #5409

---


packages/idempotency/src/persistence/BasePersistenceLayer.ts
```diff
@@ -81,6 +86,68 @@ abstract class BasePersistenceLayer implements BasePersistenceLayerInterface {
...
+
+    const mismatchedSettings: string[] = [];
+    if (config.eventKeyJmesPath !== this.eventKeyJmesPath) {
+      mismatchedSettings.push('eventKeyJmesPath');
+    }
+    if (config.payloadValidationJmesPath !== this.validationKeyJmesPath) {
+      mismatchedSettings.push('payloadValidationJmesPath');
+    }
+    if (
+      (config.throwOnNoIdempotencyKey || false) !== this.throwOnNoIdempotencyKey
```
svozza: Nit: `??` is more strictly correct here

dreamorosi: Good catch, addressed in 349bde7d.

dreamorosi: Turns out the fallback was dead code entirely: `IdempotencyConfig` always defaults `throwOnNoIdempotencyKey` to a boolean in its constructor, so `??` left an unreachable branch that tripped the 100% coverage gate. Removed the fallback in d642007e.

svozza (approved):

Issue #5434, opened 2026-07-09
Bug: Tracer middy middleware writes annotations to wrong invocation's subsegment under LMI concurrency


svozza:
### Expected Behavior

When multiple invocations are multiplexed into the same execution environment (Lambda Managed Instances with `perExecutionEnvironmentMaxConcurrency` > 1), annotations, metadata, responses, and errors captured by the `captureLambdaHandler` middy middleware should be recorded on the subsegment of the invocation that produced them.

### Current Behavior

The middy middleware manages the subsegment lifecycle through the X-Ray SDK's context primitives:

```typescript
// packages/tracer/src/middleware/middy.ts (open/close)
const segment = target.getSegment();
handlerSegment = lambdaSegment.addNewSubsegment(`## ${process.env._HANDLER}`);
target.setSegment(handlerSegment);
```

In Lambda mode, `aws-xray-sdk-core` creates **one** CLS context for the whole process at init (`namespace.enter(namespace.createContext())` in `lib/env/aws_lambda.js`) and never creates another. `setSegment()` therefore writes the active segment into a single process-wide slot shared by all concurrent invocations.

Under LMI multiplexing, invocation B's `before` hook overwrites the active segment while invocation A is still in flight. From that point, every `putAnnotation`, `putMetadata`, `addResponseAsMetadata`, and `addErrorAsMetadata` from A resolves `getSegment()` to **B's** subsegment: annotations land on the wrong trace, response payloads are attached to another invocation's (potentially another tenant's) subsegment, and error flags are misattributed.

The **decorator API is not affected** by this mechanism: `@tracer.captureLambdaHandler()` goes through `captureAsyncFunc`, which wraps the handler in `namespace.run(...)`, giving each invocation a nested CLS context — interleaved invocations through the decorator path stay isolated.

Related issues:
- #5433 — the middleware additionally stores the segments in factory-closure variables (separate bug, same file)
- #5429 — same class of bug (per-invocation state in shared storage) fixed in Logger

The underlying single-shared-CLS-context design is an upstream `aws-xray-sdk-node` issue and will be reported there, but Powertools can side-step it entirely because the SDK already ships the isolation mechanism the decorator path uses.

### Code snippet

Failing unit test — two overlapping invocations through the real middleware, real `Tracer`, and the X-Ray SDK's real Lambda-mode CLS context (no mocked provider):

```typescript
import context from '@aws-lambda-powertools/testing-utils/context';
[...]

Issue #5429, opened 2026-07-08
Bug: Logger emits wrong function_request_id under Lambda Managed Instances concurrency


svozza:
### Expected Behavior

When multiple invocations are multiplexed into the same execution environment (Lambda Managed Instances with `perExecutionEnvironmentMaxConcurrency` > 1), each log line should carry the Lambda context of the invocation that emitted it — i.e. `function_request_id` should match that invocation's `context.awsRequestId`, and the same for `tenant_id` and `cold_start`.

### Current Behavior

`Logger.addContext()` stores the Lambda context in plain instance state (`powertoolsLogData.lambdaContext`), not in the InvokeStore. Under LMI, concurrent invocations sharing an execution environment overwrite each other's context: whichever invocation called `addContext()` last wins, and every subsequent log line — from any in-flight invocation — carries that invocation's `function_request_id`, `tenant_id`, and `cold_start`.

Temporary keys added via `appendKeys()` are correctly invoke-scoped through `LogAttributesStore` (which uses the InvokeStore when `AWS_LAMBDA_MAX_CONCURRENCY` is set), so on the same log line the attributes added via `appendKeys()` are right while the context-derived fields are wrong.

Found while building the LMI e2e tests for #5092: with 30 concurrent invocations multiplexed into ~8 environments, virtually every invocation's emitted logs carried a different invocation's request id. The same handler with no concurrency matches fine.

### Code snippet

```typescript
import { Logger } from '@aws-lambda-powertools/logger';
import type { Context } from 'aws-lambda';

const logger = new Logger({ serviceName: 'lmi-repro' });

export const handler = async (event: unknown, context: Context) => {
  logger.addContext(context);
  // block long enough for a peer invocation in the same execution
  // environment to call addContext() and overwrite the shared slot
  await new Promise((resolve) => setTimeout(resolve, 2000));
  logger.info('hello'); // function_request_id is the peer's, not ours
  return context.awsRequestId;
};
```

### Steps to Reproduce

1. Deploy the function attached to a Lambda Managed Instances capacity provider with `perExecutionEnvironmentMaxConcurrency: 10` (memory ≥ 2048 MB).
2. Fire ~30 simultaneous invocations so the scheduler multiplexes invocations into shared environments.
3. Compare each invocation's `context.awsRequestId` with the `function_request_id` on the log lines it emitted.
[...]

PR #5440, opened 2026-07-09
fix(tracer): scope middy middleware segment state per invocation

### LINKED ISSUE #5433: Bug: Tracer middy middleware shares segment state across concurrent invocations
### Expected Behavior

When multiple invocations are multiplexed into the same execution environment (Lambda Managed Instances with `perExecutionEnvironmentMaxConcurrency` > 1), each invocation wrapped with the `captureLambdaHandler` middy middleware should close its **own** handler subsegment and restore its **own** facade segment when it completes.

### Current Behavior

`captureLambdaHandler()` stores the per-invocation segments in closure variables created once at handler initialization:

```typescript
// packages/tracer/src/middleware/middy.ts
const captureLambdaHandler = (target, options) => {
  let lambdaSegment: Segment;      // shared by ALL concurrent invocations
  let handlerSegment: Subsegment;  // shared by ALL concurrent invocations
  ...
```

Since the middleware factory runs once, all concurrent invocations share these two variables. When invocation B's `before` hook runs while invocation A is still in flight, B's `open()` overwrites both variables. A's `close()` then closes **B's** subsegment (which B later closes again), while A's own subsegment is never closed, and A restores B's facade segment as active.

Consequences under LMI multiplexing: lost/never-closed subsegments, double-closed subsegments, and annotations/metadata written to another invocation's (potentially another tenant's) subsegment after the clobber.

Related: #5429 fixed the same class of bug (per-invocation state in init-scoped storage) in Logger. The decorator API (`@tracer.captureLambdaHandler()`) is not affected by this specific issue because `captureAsyncFunc` scopes the subsegment to the invocation's async context.

### Code snippet

Failing unit test (interleaves two invocations through the same middy handler, mimicking LMI multiplexing):

```typescript

  it('closes each invocation subsegment and restores its own facade segment when invocations overlap', async () => {
    // Prepare
    const tracer = new Tracer();
    vi.spyOn(tracer, 'annotateColdStart').mockImplementation(() => ({}));
    vi.spyOn(tracer, 'addServiceNameAnnotation').mockImplementation(
      () => ({})
    );
    const setSegmentSpy = vi
      .spyOn(tracer.provider, 'setSegment')
      .mockImplementation(() => ({}));

    const facadeSegmentA = new Segment('facade');
    const handlerSubsegmentA = new Subsegment('## index.handler');
    vi.spyOn(facadeSegmentA, 'addNewSubsegment').mockImplementation(
      () => handlerSubsegmentA
    );
    const facadeSegmentB = new Segment('facade');
    const handlerSubsegmentB = new Subsegment('## index.handler');
    vi.spyOn(facadeSegmentB, 'addNewSubsegment').mockImplementation(
      () => handlerSubsegmentB
    );
    vi.spyOn(tracer.provider, 'getSegment')
      .mockImplementationOnce(() => facadeSegmentA)
      .mockImplementationOnce(() => facadeSegmentB);

    const gates = [Promise.withResolvers<void>(), Promise.withResolvers<void>()];
    const handler = middy(async (event: { idx: number }, _context: Context) => {
    }).use(captureLambdaHandler(tracer, { captureResponse: false }));

    // Act
    // (its `before` overwrites the middleware's closed-over segment variables),
    // then A completes, then B completes.
    const invocationA = handler({ idx: 0 }, context);
    const invocationB = handler({ idx: 1 }, context);
    gates[0].resolve();
    await invocationA;
    gates[1].resolve();
    await invocationB;

    // Assess
    // Each invocation must close its own handler subsegment...
    expect(handlerSubsegmentA.isClosed()).toBe(true);
    expect(handlerSubsegmentB.isClosed()).toBe(true);
    // ...and restore its own facade segment when it completes
    expect(setSegmentSpy).toHaveBeenNthCalledWith(3, facadeSegmentA);
    expect(setSegmentSpy).toHaveBeenNthCalledWith(4, facadeSegmentB);
  });
});
```

Fails on current `main`:

```
AssertionError: expected false to be true
  54|     expect(handlerSubsegmentA.isClosed()).toBe(true);
```

Invocation A's subsegment is never closed — A's `close()` ran against B's subsegment after B's `before` hook overwrote the shared `handlerSegment` variable.

### Steps to Reproduce

3. The first assertion fails: invocation A's subsegment is left open

On real LMI: deploy a middy-wrapped handler attached to a capacity provider with `perExecutionEnvironmentMaxConcurrency` > 1 and fire enough simultaneous invocations to force multiplexing; subsegments are lost or double-closed and annotations land on the wrong invocation's subsegment.

### Possible Solution

Store the two segments per-request instead of in the factory closure — e.g. in `request.internal` (the middleware already writes a cleanup function there) so each middy request carries its own `lambdaSegment`/`handlerSegment`, or in the InvokeStore keyed by invocation, following the pattern used for Logger in #5429.

Note: even with this fixed, restoring the facade segment via `setSegment()` writes to the X-Ray SDK's single shared CLS context in Lambda mode, which has its own cross-invocation issues upstream (`aws-xray-sdk-node`'s Lambda facade is a module-level singleton). Fixing the closure state is still necessary and is entirely within Powertools' control.

latest

24.x

### Packaging format used

npm

### Execution logs

_N/A — see failing unit test output above._

dreamorosi:
## Summary

The `captureLambdaHandler` middy middleware stored the per-invocation facade segment and handler subsegment in factory-closure variables created once at handler initialization. When multiple invocations are multiplexed into the same execution environment (Lambda Managed Instances with `perExecutionEnvironmentMaxConcurrency` > 1), those variables are shared by all in-flight invocations: one invocation's `after`/`onError` hook would close another invocation's subsegment and restore the wrong facade segment.

This PR moves the per-invocation state to `request.internal`, so each invocation closes the subsegment it opened and restores the facade segment it captured, regardless of interleaving or completion order.

> [!note]
> This PR was originally scoped to also address #5434 (wrong-segment attribution of annotations/metadata under LMI concurrency) via per-invocation CLS contexts. Real-Lambda testing showed that approach cannot work from within a middy middleware: the context entered in a `before` hook does not propagate to the handler because middy's `runRequest` continuations are created before the hook runs, and `cls-hooked` restores their captured (root) context. See [this comment on #5434](https://github.com/aws-powertools/powertools-lambda-typescript/issues/5434) for the full analysis - that part requires per-invocation context isolation in `aws-xray-sdk-node` and is tracked separately.

### Changes

- `packages/tracer/src/middleware/middy.ts`: store `lambdaSegment` and `handlerSegment` in `request.internal` under `TRACER_KEY`-prefixed keys instead of factory-closure variables; `close()` reads them from the request so cleanup always targets the invocation that opened them
- `packages/tracer/tests/unit/middy.concurrency.test.ts`: new regression tests running two interleaved invocations through the middleware, asserting each invocation closes its own subsegment exactly once and restores its own facade segment - including a reverse-completion-order variant

**Issue number:** closes #5433

---

### DIFF
--- a/packages/tracer/src/middleware/middy.ts
+++ b/packages/tracer/src/middleware/middy.ts
@@ -8,17 +8,18 @@ import type { Tracer } from '../Tracer.js';
 import type { CaptureLambdaHandlerOptions } from '../types/Tracer.js';
 
 /**
- * Keys used to store the per-invocation segments and CLS context in the
- * middy `request.internal` object.
+ * Keys used to store the per-invocation segments in the middy
+ * `request.internal` object.
  *
  * Storing this state per-request rather than in factory-closure variables is
  * required for correctness when multiple invocations are multiplexed into the
  * same execution environment (e.g. Lambda Managed Instances with
- * `perExecutionEnvironmentMaxConcurrency` > 1).
+ * `perExecutionEnvironmentMaxConcurrency` > 1): closure variables are shared
+ * by all in-flight invocations, so one invocation's `after`/`onError` hook
+ * would close and restore another invocation's segments.
  */
 const lambdaSegmentKey = `${TRACER_KEY}.lambdaSegment`;
 const handlerSegmentKey = `${TRACER_KEY}.handlerSegment`;
-const clsContextKey = `${TRACER_KEY}.clsContext`;
 
 /**
  * A middy middleware automating capture of metadata and annotations on segments or subsegments for a Lambda Handler.
@@ -63,35 +64,11 @@ const captureLambdaHandler = (
     };
   };
 
-  /**
-   * Open a new handler subsegment inside a fresh CLS context dedicated to
-   * this invocation.
-   *
-   * In Lambda mode the X-Ray SDK enters a single process-wide CLS context at
-   * initialization, so `setSegment()` alone writes the active segment into a
-   * slot shared by all concurrent invocations. To keep interleaved
-   * invocations isolated, we create and enter a new CLS context first: the
-   * context tracking in `cls-hooked` is based on `async_hooks` resource
-   * creation, so — as long as this function runs synchronously within the
-   * `before` hook — every async resource created downstream (the handler and
-   * the `after`/`onError` hooks) inherits the new context, while other
-   * in-flight invocations keep operating on their own. This mirrors what
-   * `captureAsyncFunc` does in the decorator path via `namespace.run()`.
-   *
-   * The created context prototypally inherits the facade segment from the
-   * SDK's root context, and `setSegment()` then shadows it with the handler
-   * subsegment for this invocation only.
-   *
-   * @param request - The request object
-   */
   const open = (request: MiddyLikeRequest): void => {
     const segment = target.getSegment();
     if (segment === undefined) {
       return;
     }
-    const namespace = target.provider.getNamespace();
-    const clsContext = namespace.createContext();
-    namespace.enter(clsContext);
     // If segment is defined, then it is a Segment as this middleware is only used for Lambda Handlers
     const lambdaSegment = segment as Segment;
     const handlerSegment = lambdaSegment.addNewSubsegment(
@@ -102,16 +79,9 @@ const captureLambdaHandler = (
       ...request.internal,
       [lambdaSegmentKey]: lambdaSegment,
       [handlerSegmentKey]: handlerSegment,
-      [clsContextKey]: clsContext,
     };
   };
 
-  /**
-   * Close the handler subsegment for this invocation, restore the facade
-   * segment, and exit the invocation's CLS context.
-   *
-   * @param request - The request object
-   */
   const close = (request: MiddyLikeRequest): void => {
     const handlerSegment = request.internal[handlerSegmentKey] as
       | Subsegment
@@ -132,10 +102,6 @@ const captureLambdaHandler = (
       );
     }
     target.setSegment(lambdaSegment);
-    const clsContext = request.internal[clsContextKey];
-    if (clsContext !== undefined) {
-      target.provider.getNamespace().exit(clsContext);
-    }
   };
 
   const before = (request: MiddyLikeRequest) => {
--- a/packages/tracer/src/middleware/middy.ts
+++ b/packages/tracer/src/middleware/middy.ts
@@ -8,6 +8,19 @@ import type { Tracer } from '../Tracer.js';
 import type { CaptureLambdaHandlerOptions } from '../types/Tracer.js';
 
 /**
+ * Keys used to store the per-invocation segments and CLS context in the
+ * middy `request.internal` object.
+ *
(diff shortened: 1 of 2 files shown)


dreamorosi:
Closing in favor of #5444, which was opened by the contributor assigned to #5433 and implements the same approach (per-invocation segment state in `request.internal` under `TRACER_KEY`-prefixed keys).

Worth carrying over from this branch: the approach was validated on a deployed Lambda function (Node.js 22, X-Ray active tracing) - annotations, metadata, and outbound-call subsegments were all correctly placed across sequential invocations - so #5444's mechanism is de-risked beyond unit tests. I've left suggestions on #5444 for a couple of extra test assertions from this branch.

The wider investigation into #5434 (why per-invocation CLS context isolation cannot be done from within the middleware) is documented in https://github.com/aws-powertools/powertools-lambda-typescript/issues/5434#issuecomment-4927700583 and tracked upstream in aws/aws-xray-sdk-node#760 and middyjs/middy#1661.

svozza:
Following up here to close the loop: #5446 re-adds the #5434 half that was descoped from this PR, using a different isolation mechanism that doesn't hit the propagation wall documented above.

The key insight from your real-Lambda diagnostics was that a middy hook cannot *establish* an async context that reaches the handler — `runRequest`'s continuations are created before the hook body executes, so they restore the pre-hook context on resume. #5446 sidesteps that constraint entirely: the middleware never creates a context. Under LMI, the runtime itself wraps the whole invocation (all hooks + handler) in an `AsyncLocalStorage` context via the Invoke Store before any user code runs, so by the time middy creates its continuations the invocation-scoped context is already active and every continuation snapshots the right one by construction. `ProviderService.getSegment()`/`setSegment()` read/write the active segment in that store when `AWS_LAMBDA_MAX_CONCURRENCY` is set (falling back to the SDK's CLS context otherwise) — the same dual-path pattern Logger shipped in #5429.

Since everything that attributes data to a segment (`putAnnotation`/`putMetadata`, response/error capture) resolves through those two methods, wrong-segment attribution is covered without touching cls-hooked. The boundary you identified remains real, just narrower: SDK-patched clients (`captureAWSv3Client`, `captureHTTPsGlobal`) resolve their parent through the SDK's own `contextUtils.resolveSegment()` and stay subject to the upstream issue until the InvokeStore-scoped context lands in `aws-xray-sdk-node`.

On your testing caveat — that transpiled unit tests gave false confidence for the CLS approach — this was validated on a real LMI deployment in both directions, using the barrier/saturation harness from the Logger LMI PoC: with the fix, 30 concurrent invocations across genuinely multiplexed environments each annotated exactly their own subsegment; with only `ProviderService.ts` reverted to `main`, the same harness failed immediately with one subsegment carrying two invocations' annotations. The unit suite also includes a scenario derived from your propagation finding (an async `before` middleware registered ahead of the tracer) — it fails under the per-invocation CLS approach and passes with the store approach, so the constraint you documented stays regression-guarded even at unit level. Full results on #5446.

PR #5444, opened 2026-07-09
fix(tracer): scope middy middleware segments per invocation

### LINKED ISSUE #5433: Bug: Tracer middy middleware shares segment state across concurrent invocations

vishwakt:
### Changes

The `captureLambdaHandler` middy middleware stored the facade segment and handler subsegment in variables scoped to the middleware factory, which runs once per handler. Under Lambda Managed Instances concurrency, overlapping invocations share these variables: a later invocation's `before` hook overwrites them, so an earlier invocation closes the wrong subsegment, leaves its own subsegment open, and restores another invocation's facade segment.

Following the direction agreed in the issue, both segments are now stored in `request.internal`, so each middy request carries its own segments through `after`, `onError`, and the early-return cleanup path (`cleanupMiddlewares` already passes `request` to the stored cleanup function).

Added a unit test that interleaves two invocations through the same handler and asserts each closes its own subsegment and restores its own facade segment. It fails on `main` and passes with this change.

**Issue number:** closes #5433

---

### DIFF
--- a/packages/tracer/src/middleware/middy.ts
+++ b/packages/tracer/src/middleware/middy.ts
@@ -38,8 +38,11 @@ const captureLambdaHandler = (
   target: Tracer,
   options?: CaptureLambdaHandlerOptions
 ): MiddlewareLikeObj => {
-  let lambdaSegment: Segment;
-  let handlerSegment: Subsegment;
+  // The segments are stored in `request.internal` rather than in variables scoped
+  // to this factory so that each invocation operates on its own segments when
+  // multiple invocations run concurrently in the same execution environment.
+  const lambdaSegmentKey = `${TRACER_KEY}.lambdaSegment`;
+  const handlerSegmentKey = `${TRACER_KEY}.handlerSegment`;
 
   /**
    * Set the cleanup function to be called in case other middlewares return early.
@@ -53,21 +56,29 @@ const captureLambdaHandler = (
     };
   };
 
-  const open = (): void => {
+  const open = (request: MiddyLikeRequest): void => {
     const segment = target.getSegment();
     if (segment === undefined) {
       return;
     }
     // If segment is defined, then it is a Segment as this middleware is only used for Lambda Handlers
-    lambdaSegment = segment as Segment;
-    handlerSegment = lambdaSegment.addNewSubsegment(
+    const lambdaSegment = segment as Segment;
+    const handlerSegment = lambdaSegment.addNewSubsegment(
       `## ${process.env._HANDLER}`
     );
+    request.internal[lambdaSegmentKey] = lambdaSegment;
+    request.internal[handlerSegmentKey] = handlerSegment;
     target.setSegment(handlerSegment);
   };
 
-  const close = (): void => {
-    if (handlerSegment === undefined || lambdaSegment === null) {
+  const close = (request: MiddyLikeRequest): void => {
+    const lambdaSegment = request.internal[lambdaSegmentKey] as
+      | Segment
+      | undefined;
+    const handlerSegment = request.internal[handlerSegmentKey] as
+      | Subsegment
+      | undefined;
+    if (handlerSegment === undefined || lambdaSegment === undefined) {
       return;
     }
     try {
@@ -84,7 +95,7 @@ const captureLambdaHandler = (
 
   const before = (request: MiddyLikeRequest) => {
     if (target.isTracingEnabled()) {
-      open();
+      open(request);
       setCleanupFunction(request);
       target.annotateColdStart();
       target.addServiceNameAnnotation();
@@ -96,14 +107,14 @@ const captureLambdaHandler = (
       if (options?.captureResponse ?? true) {
         target.addResponseAsMetadata(request.response, process.env._HANDLER);
       }
-      close();
+      close(request);
     }
   };
 
   const onError = (request: MiddyLikeRequest) => {
     if (target.isTracingEnabled()) {
       target.addErrorAsMetadata(request.error as Error);
-      close();
+      close(request);
     }
   };
 
--- a/packages/tracer/src/middleware/middy.ts
+++ b/packages/tracer/src/middleware/middy.ts
@@ -66,8 +66,11 @@ const captureLambdaHandler = (
     const handlerSegment = lambdaSegment.addNewSubsegment(
       `## ${process.env._HANDLER}`
     );
-    request.internal[lambdaSegmentKey] = lambdaSegment;
-    request.internal[handlerSegmentKey] = handlerSegment;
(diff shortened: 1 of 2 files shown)


dreamorosi (commented):
Thanks for picking this up - the approach is correct, and it's additionally de-risked beyond unit tests: while investigating #5434 I validated this exact mechanism (per-invocation segment state in `request.internal` under `TRACER_KEY`-prefixed keys) on a deployed Lambda function with X-Ray active tracing, and annotations, metadata, and outbound-call subsegments were all correctly attributed across invocations.

Two small suggestions for the new test, based on regressions the current assertions can't catch:

1. **Reverse completion order**: add a variant where invocation B enters second but finishes *first* (`gates[1].resolve()` before `gates[0]`). With the old closure-variable bug, completion order determines *which* wrong subsegment gets closed - covering both orders locks down the behavior regardless of interleaving.
2. **Close called exactly once**: `isClosed()` can't distinguish "closed by its own invocation" from "closed twice by two invocations" (the old bug closed one subsegment twice and left the other open, but a partial regression could still pass an `isClosed()` check). Spying on `close` and asserting `toHaveBeenCalledTimes(1)` per subsegment makes the test airtight.

Cosmetic nit, take it or leave it: `open()` assigns `request.internal[key] = ...` directly while the existing `setCleanupFunction` uses spread-assignment - might be worth aligning on one style within the file.

vishwakt:
@dreamorosi 

Applied all three suggestions including the nit.

The concurrency test now runs both completion orders via `it.each`, asserts `close` is called exactly once per subsegment, and `open()` uses the same spread style as `setCleanupFunction`. Verified both test variants fail against the pre-fix implementation.

svozza (approved):

svozza:
Nice work @vishwakt , I'm able to build on this now to fix #5434 .

Issue #5092, opened 2026-03-14
Maintenance: Add E2E tests for Lambda Managed Instances


svozza:
### Summary

Now that CloudFormation supports provisioning Lambda Managed Instances, we should add e2e tests fot his configuration. Specifically, the logger, metrics, and batch utilities have specific code paths that are only invoked in LMI environments and we should test these.

### Why is this needed?

We support this platform but we have no automated e2e tests to confirm it works.

### Solution

Update our CDK test to create LMI environments:

```ts
import * as ec2 from 'aws-cdk-lib/aws-ec2';

const vpc = new ec2.Vpc(this, 'MyVpc');
const securityGroup = new ec2.SecurityGroup(this, 'SecurityGroup', { vpc });

const capacityProvider = new lambda.CapacityProvider(this, 'MyCapacityProvider', {
  subnets: vpc.privateSubnets,
  securityGroups: [securityGroup],
  scalingOptions: lambda.ScalingOptions.manual([
    lambda.TargetTrackingScalingPolicy.cpuUtilization(70),
  ]),
});

const fn = new lambda.Function(this, 'MyFunction', {
  // Runtime must be equal to or newer than NODEJS_LATEST
  runtime: lambda.Runtime.NODEJS_LATEST,
  handler: 'index.handler',
  code: lambda.Code.fromAsset(path.join(__dirname, 'lambda-handler')),
});

// Associate the function with the capacity provider
capacityProvider.addFunction(fn, {
  perExecutionEnvironmentMaxConcurrency: 10,
  executionEnvironmentMemoryGiBPerVCpu: 4,
});
```

If possible we should try to run some of the concurrency tests in this particular:

https://github.com/aws-powertools/powertools-lambda-typescript/tree/main/packages/logger/tests/unit/concurrency
https://github.com/aws-powertools/powertools-lambda-typescript/tree/main/packages/metrics/tests/unit/concurrency
https://github.com/aws-powertools/powertools-lambda-typescript/tree/main/packages/batch/tests/unit/concurrency

### Acknowledgment

- [x] This request meets [Powertools for AWS Lambda (TypeScript) Tenets](https://docs.aws.amazon.com/powertools/typescript/latest/#tenets)
[...]

PR #5446, opened 2026-07-09
fix(tracer): isolate active segment per invocation under LMI concurrency

### LINKED ISSUE #5434: Bug: Tracer middy middleware writes annotations to wrong invocation's subsegment under LMI concurrency

svozza:
## Summary

In Lambda mode the X-Ray SDK keeps the active segment in a single process-wide CLS context shared by all concurrent invocations. When multiple invocations are multiplexed into the same execution environment (Lambda Managed Instances with `perExecutionEnvironmentMaxConcurrency` > 1), one invocation's `setSegment()` clobbers the slot for every other in-flight invocation, so `putAnnotation`, `putMetadata`, response capture, and error capture resolve to whichever subsegment was set last — annotations land on the wrong trace and response payloads attach to another invocation's (potentially another tenant's) subsegment.

This PR isolates the active segment per invocation by routing `ProviderService.getSegment()`/`setSegment()` through the Lambda Invoke Store when `AWS_LAMBDA_MAX_CONCURRENCY` is set — the same pattern already used by Logger (#5429), Metrics, and Batch. Unlike the per-invocation CLS context approach explored and invalidated in #5440 (a context entered in a middy `before` hook never reaches the handler because middy's continuations are created before the hook runs), the middleware never establishes an async context here — the LMI runtime does, wrapping the entire invocation in its own `AsyncLocalStorage` context before any user code runs, so isolation doesn't depend on hook timing or middleware registration order. Builds on the per-request segment state introduced in #5444.

Design note: no dedicated `*Store` class as in Logger/Metrics/Batch — `ProviderService` already is the tracer's segment-state seam and the non-LMI fallback is the SDK's CLS context rather than instance state, so a `SegmentStore` would be a pure pass-through layer.

Scope boundary: anything resolving its parent through the SDK's own context (`captureAWSv3Client`-patched clients, direct `AWSXRay.getSegment()` calls) still reads the shared slot and remains subject to aws/aws-xray-sdk-node#760.

### Changes

- `packages/tracer/src/provider/ProviderService.ts`: `getSegment()`/`setSegment()` gain a dual path gated on `shouldUseInvokeStore()` — under `AWS_LAMBDA_MAX_CONCURRENCY` the active segment is read/written in the Invoke Store under a private `Symbol` (throwing `InvokeStore is not available` if the store is missing, matching Logger); otherwise behavior is unchanged. `getSegment()` falls through to the SDK context when the invocation hasn't set a segment yet so the facade segment stays reachable
- `packages/tracer/tests/unit/middy.concurrency.test.ts`: regression tests running interleaved invocations through the real middleware, real `Tracer`, and the SDK's real Lambda-mode CLS context, each wrapped in `invokeStore.run()` as on LMI — covering in-order and reverse-order completion, error attribution, an async `before` middleware registered ahead of the tracer, and the store-unavailable error path
- `packages/tracer/package.json` / `package-lock.json`: add `@aws/lambda-invoke-store` 0.3.0 (same pin as logger/metrics/batch)

**Issue number:** closes #5434

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9126,6 +9126,7 @@
       "license": "MIT-0",
       "dependencies": {
         "@aws-lambda-powertools/commons": "2.33.1",
+        "@aws/lambda-invoke-store": "0.3.0",
         "aws-xray-sdk-core": "^3.12.0"
       },
       "devDependencies": {
--- a/packages/tracer/package.json
+++ b/packages/tracer/package.json
@@ -89,6 +89,7 @@
   },
   "dependencies": {
     "@aws-lambda-powertools/commons": "2.33.1",
+    "@aws/lambda-invoke-store": "0.3.0",
     "aws-xray-sdk-core": "^3.12.0"
   },
   "keywords": [
--- a/packages/tracer/src/provider/ProviderService.ts
+++ b/packages/tracer/src/provider/ProviderService.ts
@@ -26,8 +26,13 @@ const {
 import { subscribe } from 'node:diagnostics_channel';
 import http from 'node:http';
 import https from 'node:https';
+import type { InvokeStoreBase } from '@aws/lambda-invoke-store';
+import '@aws/lambda-invoke-store';
 import { addUserAgentMiddleware } from '@aws-lambda-powertools/commons';
-import { getXRayTraceIdFromEnv } from '@aws-lambda-powertools/commons/utils/env';
+import {
+  getXRayTraceIdFromEnv,
+  shouldUseInvokeStore,
+} from '@aws-lambda-powertools/commons/utils/env';
 import type { DiagnosticsChannel } from 'undici-types';
 import {
   findHeaderAndDecode,
@@ -47,6 +52,28 @@ import {
  * they should work as expected. However, support for these methods is not guaranteed.
  */
 class ProviderService implements ProviderServiceInterface {
+  /**
+   * Key used to store the active segment in the Lambda Invoke Store when
+   * multiple invocations may be multiplexed into the same execution
+   * environment (Lambda Managed Instances with
+   * `perExecutionEnvironmentMaxConcurrency` > 1).
+   *
+   * In Lambda mode the X-Ray SDK keeps the active segment in a single
+   * process-wide CLS context shared by all concurrent invocations, so
+   * writing to it from one invocation clobbers the others. The Invoke Store
+   * is scoped to the invocation by the runtime, so it provides the isolation
+   * the SDK context can't.
+   */
+  readonly #segmentKey = Symbol('powertools.tracer.segment');
+
+  #getInvokeStore(): InvokeStoreBase {
+    const store = globalThis.awslambda?.InvokeStore;
+    if (store === undefined) {
+      throw new Error('InvokeStore is not available');
+    }
+    return store;
+  }
+
   /**
    * @deprecated Use {@link captureAWSv3Client} instead.
    */
@@ -94,6 +121,18 @@ class ProviderService implements ProviderServiceInterface {
   }
 
   public getSegment(): Segment | Subsegment | undefined {
+    if (shouldUseInvokeStore()) {
+      // The facade segment isn't stored per-invocation: it's created by the
+      // X-Ray SDK at init and lives in its process-wide context, so when the
+      // current invocation hasn't set an active segment yet we fall through
+      // to the SDK
+      const segment = this.#getInvokeStore().get<Segment | Subsegment>(
+        this.#segmentKey
+      );
+      if (segment !== undefined) {
+        return segment;
+      }
+    }
     return getSegment();
   }
 
@@ -271,6 +310,13 @@ class ProviderService implements ProviderServiceInterface {
   }
 
   public setSegment(segment: Segment | Subsegment): void {
+    if (shouldUseInvokeStore()) {
+      // Writing to the SDK context here would clobber the shared slot for
+      // every other in-flight invocation, so the active segment goes in the
+      // invocation-scoped store instead
+      this.#getInvokeStore().set(this.#segmentKey, segment);
+      return;
+    }
     setSegment(segment);
   }
 }
--- a/packages/tracer/tests/unit/middy.concurrency.test.ts
+++ b/packages/tracer/tests/unit/middy.concurrency.test.ts
@@ -1,0 +1,222 @@
+import { InvokeStore } from '@aws/lambda-invoke-store';
+import context from '@aws-lambda-powertools/testing-utils/context';
+import middy from '@middy/core';
+import type { Context } from 'aws-lambda';
+import { Segment, Subsegment } from 'aws-xray-sdk-core';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { Tracer } from '../../src/index.js';
+import { captureLambdaHandler } from '../../src/middleware/middy.js';
+
+// Must run before aws-xray-sdk-core is imported so the SDK initializes in
+// Lambda mode with its real CLS context, which is what these tests exercise:
+// in Lambda mode the SDK enters a single process-wide CLS context at init,
+// shared by all concurrent invocations. `AWS_LAMBDA_MAX_CONCURRENCY` must be
+// set before the InvokeStore instance is created so it uses per-invocation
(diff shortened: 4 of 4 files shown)


svozza:
**Validated on real Lambda Managed Instances, in both directions.**

Given the finding in #5440 that transpiled unit tests can give false confidence for async-context changes, I ran this fix on a deployed function: an LMI e2e suite modeled on the Logger LMI PoC (#5092 infra) — ephemeral capacity provider with a minimum-size fleet (12 vCPU, single AZ), `perExecutionEnvironmentMaxConcurrency: 10`, and 30 simultaneous invocations to force genuine multiplexing. The handler blocks on a module-scoped barrier until a peer invocation is in flight in the same execution environment, then writes a per-invocation annotation — deliberately inside the window where a peer's `before` hook has already overwritten the SDK's shared slot. Attribution is asserted from in-process ground truth (the SDK's `annotations` object on every handler subsegment the middleware creates, captured via a prototype patch), so the X-Ray delivery path and the context-resolution machinery under test play no part in the measurement. All invocations confirmed `AWS_LAMBDA_INITIALIZATION_TYPE=lambda-managed-instances` and at least one genuine overlap per run.

- **With this fix:** all 30 invocations' annotations landed on exactly the subsegment their own `before` hook opened — no foreign annotations, no duplicates, none lost.
- **Negative control (same harness, only `ProviderService.ts` reverted to `main`):** failed immediately on the isolation invariant — a single handler subsegment carrying **two different invocations' annotations**, the exact #5434 clobber signature.

Same infrastructure, same invocation pattern, one variable. The suite depends on LMI test infrastructure (ephemeral capacity provider construct, LMI function props) that isn't on `main` yet, so it's not part of this PR's diff; we'll be adding LMI e2e tests covering this as a follow-up once that infrastructure lands.

dreamorosi (approved):

PR #5462, opened 2026-07-13
feat(ci): fold GovCloud/China layer publishing into Make Release

### LINKED ISSUE #5451: Maintenance: Fold GovCloud/China layer publishing and ARN doc update into Make Release
### Summary

Today, three release steps that could run automatically as one pipeline are instead split into separate manual triggers:

1. `layers_partitions.yml` ("Layer Deployment (Partitions)") publishes the Lambda layer to the GovCloud and China partitions via a manually-triggered `workflow_dispatch` run per partition, after `Make Release` completes.
2. The "update layer ARN on documentation" PR (opened by `update_layer_arn_docs` in `publish_layer.yml`) is created right after the *commercial* Prod deploy, before GovCloud/China are published — so it has to guess those regions' version numbers ahead of time.
3. A maintainer has to review/approve that PR, then wait for the GovCloud/China workflows to finish, before finally merging it.

This split originated when the layer version had to be manually noted from the commercial deployment output and passed along by hand to the next steps. That's no longer true: `reusable_deploy_layer_stack.yml` already outputs the deployed `layer-version` from the `us-east-1` Prod deploy job, and `layers_partitions.yml`'s own `download` job already resolves the latest commercial layer version automatically via `aws lambda list-layer-versions`, with no human input. Additionally, none of the GovCloud/China GitHub Environments (`GovCloud Gamma`, `GovCloud Prod`, `China Gamma`, `China Prod`) have required-reviewer protection rules, so the separate manual trigger isn't enforcing an approval checkpoint today — it's just leftover structure.

Because the ARN doc PR is generated before GovCloud/China deploy, it also causes a real correctness gap: it pre-emptively bumps the GovCloud/China ARN version numbers based on the assumption that those regions will get the same version as commercial, which isn't guaranteed (e.g. paused regions).

### Why is this needed?

Removes multiple manual steps (and the wait time between triggering them) from every release, and fixes a source of incorrect documentation by generating a single ARN-update PR only after all partitions (commercial + GovCloud + China) have actually deployed, rather than guessing ahead of time.

Automation, Governance

### Solution

- Add GovCloud and China deploy jobs to `publish_layer.yml` (or `make-release.yml`), gated on `needs: deploy-prod`, so they run automatically right after the commercial Prod layer deploy completes — no separate `workflow_dispatch` trigger needed.
- Move the `update_layer_arn_docs` job out of its current position (right after commercial Prod deploy) to run only after commercial + GovCloud + China have all completed, so a single ARN-update PR reflects the true, final state of every partition.
- Fully automate China the same as GovCloud and all commercial regions — no additional approval gate beyond the existing `Release` environment's required-reviewer check on npm publish.
- End state: one ARN-update PR per release, opened only once every partition has actually deployed, instead of today's three separate manual triggers plus an early, partially-guessed PR.

### Acknowledgment

svozza:
## Summary

Today, GovCloud/China layer publishing requires a separate manual `workflow_dispatch` of `Layer Deployment (Partitions)` after every release, and the ARN-doc PR is opened right after the *commercial* Prod deploy — before GovCloud/China even run — so it has to guess those regions' version numbers ahead of time. This folds both partitions directly into `publish_layer.yml` so they deploy automatically right after commercial Prod finishes, using the exact version `deploy-prod` just published (no more guessing via `list-layer-versions`), and only opens the ARN-doc PR once all three (commercial, GovCloud, China) Prod deployments are complete.

This was designed and rehearsed end-to-end on an internal fork before porting it here, including two full dispatches of `make-release.yml` that confirmed the job graph runs automatically (commercial beta → commercial prod → GovCloud/China gamma+prod in parallel → single consolidated ARN-update PR). The actual AWS CLI and npm-publish calls were mocked/stubbed for the rehearsal — only the job wiring, `needs:` graph, and control flow were verified live; the AWS-facing steps themselves are unchanged from their existing implementations.

### Changes

- `publish_layer.yml`: added `download-layer` (pulls the layer version `deploy-prod` just deployed) plus `deploy-govcloud-gamma`/`deploy-govcloud-prod` and `deploy-china-gamma`/`deploy-china-prod` jobs, both partitions deploying in parallel after commercial Prod; `update_layer_arn_docs` now waits on all three partitions' Prod deployments before opening its PR
- `make-release.yml`: updated release-process comments to reflect the new step and dependency
- `layers_partitions.yml`: updated comments to note it's now a standalone manual recovery tool rather than part of the automated release path (kept as-is otherwise, per the issue)
- `docs/maintainers.md`: updated the release checklist and release-process Gantt chart to remove the now-obsolete manual GovCloud/China trigger steps and reflect the new parallel deployment step

**Issue number:** closes #5451

---


dreamorosi (approved):
Diff looks good as far as I can tell by reading it, but I guess we'll know once we make the next release.

Do we have an idea of the failure modes? I think knowing them would help us understand if we need to setup a sample environment to run these or if it's fine to test during the release because it's recoverable.

svozza:
The most likely failure mode is that the newly added workflow fails to execute because of some wriing or permission issue. In that case we'd have to fall back to running the [old workflows](https://github.com/aws-powertools/powertools-lambda-typescript/actions/workflows/layers_partitions.yml). Another thing that could happen is that the push to beta works but for some reason the verification fails so the push to prod doesn't happen. Again, we would have to fall back to the old workflows (although only run it for the prod environment).

PR #5465, opened 2026-07-13
chore(tests): add LMI e2e suite and run-scoped shared capacity provider

### LINKED ISSUE #5518: Maintenance: Add Logger E2E tests for Lambda Managed Instances
### Summary

Add an end-to-end test validating Logger's `InvokeStore`-backed log attribute isolation on Lambda Managed Instances, where concurrent invocations are multiplexed into shared execution environments.

This includes the supporting testing infrastructure: a capacity provider construct, shared run-scoped capacity provider stacks, and workflow jobs. Part of #5092 — this issue tracks the Logger portion; metrics and batch are tracked separately by the parent. Implemented by PR #5465.

### Why is this needed?

Logger's `InvokeStore` concurrency code paths are currently exercised only by unit tests with mocked Lambda Managed Instances environments. Nothing proves that log attributes remain isolated between concurrent invocations on the real platform. This work validates that behavior against Lambda Managed Instances, following the prototype findings discussed in #5092.

- Automation
- Tests
- Logger

### Solution

Add a Logger LMI end-to-end suite that forces concurrent invocations to share execution environments and verifies that invocation-scoped log attributes do not leak between requests. Add reusable capacity provider infrastructure, shared run-scoped capacity provider stacks, and workflow jobs needed to provision and run the suite.

### Acknowledgment

### Future readers

svozza:
## Summary

Adds an end-to-end test that validates Logger's `InvokeStore`-backed log-attribute isolation on Lambda Managed Instances (LMI), where multiple invocations run concurrently in the same execution environment. Also introduces the testing infrastructure to support it, including a run-scoped shared capacity provider so LMI suites don't each provision their own VPC + capacity provider.

### Changes

- Add a Logger LMI e2e suite (`packages/logger/tests/e2e/lmi.test.ts`) that saturates a deliberately small fleet, forces concurrent invocations to multiplex into shared execution environments, and asserts each invocation's appended keys stay isolated
- Capture the handler's own log lines by teeing `process.stdout.write` (the production write path) and returning them in the response payload, making log collection deterministic — no CloudWatch polling or ingestion latency
- Filter captured lines by the `InvokeStore`-scoped `invocationKey`, and additionally assert on `function_request_id` as a regression check now that the Lambda context is scoped per invocation under LMI (#5430)
- Add `TestLmiCapacityProvider` (dual-stack IPv6-only VPC, egress-only IGW, CloudWatch Logs interface endpoint — no NAT) and extend `TestNodejsFunction` to attach a function to a capacity provider by construct or by ARN
- Add `packages/testing/src/lmi/cli.ts` with deploy/destroy commands that manage a run-scoped `LmiShared-<runId>-<arch>` stack and print the capacity provider ARN, so a workflow run can provision one shared capacity provider per architecture up front and pass it to each suite via `LMI_CAPACITY_PROVIDER_ARN` (suites fall back to an ephemeral per-suite provider when unset)
- Run the LMI suite unconditionally on every e2e dispatch (removes the `RUN_LMI_TESTS` gate and workflow input)

**Issue number:** closes #5518

---


packages/testing/src/lmi/cli.ts
```diff
@@ -0,0 +1,112 @@
...
+  const { app } = buildApp(stackName);
+  const cli = new Toolkit({ color: false });
+  const { cx, outputFilePath } = await makeAssembly(cli, app, stackName);
+
+  if (action === 'deploy') {
+    await cli.deploy(cx, {
+      stacks: { strategy: StackSelectionStrategy.ALL_STACKS },
+      outputsFile: outputFilePath,
+    });
+    const outputs = JSON.parse(await readFile(outputFilePath, 'utf-8'))[
```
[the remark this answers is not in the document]
svozza: Fixed in 482b3cf0. The `--run-id` value is now validated at the boundary (`assertValidRunId`) against `^[A-Za-z0-9-]+$` — the same character set CloudFormation already requires for stack names — before it's used in either the stack name or the `join(tmpdir(), ...)` assembly path. A crafted value containing `../` (or slashes/spaces) is rejected, so it can no longer escape `tmpdir()`.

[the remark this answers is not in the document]
svozza: Added a sink-side guard in b814b127. The scanner's taint analysis doesn't credit the run-id regex as sanitisation, and its own guidance is to "validate the constructed path before accessing the file system" — so `makeAssembly` now `resolve()`s the output directory and throws unless it stays within `os.tmpdir()`, before anything is written or read back. The run-id validation stays as the input-side guard, giving defence in depth on both ends of the flow.

packages/testing/src/lmi/cli.ts
dreamorosi, line 1: I am not sure I understand why we need a bespoke CLI and we don't use CDK directly to deploy the stack.

I understand from the JSDoc below (and we've discussed this) that we need to deploy a single capacity provider stack independent of tests but I wonder if we can still keep this as part of the vitest lifecycle and specifically the setup/teardown parts.

I'm concerned about this type of bespoke scripts as they usually become very brittle / outdated over time and often times they're dependent on a specific way to run things.

However I am also perhaps missing part of the picture because I don't see how you actually run this CLI (the PR has no updates to the `.github/workflows/*` files nor an example on how to call it.

svozza: So that CLI was actually a leftover from when I had a different idea of how these tests were going to run but you're right, there's no need for it in the current architecture. I've removed the CLI code and we just execute it like standard CDK now. 

I've also split the LMI end to end tests using a naming convention so we can use Vitest globs. I've also made the LMI infra be explicitly spun up by the GitHub workflow rather than implicitly provisioning through Vitest hooks. Combining these two approaches means that the non-LMI tests can start executing while all the LMI infra is being provisioned.

Something to note: the shared stacks are torn down at the end of the run, so "re-run failed jobs" only works for the non-LMI jobs, a re-run LMI cell will fail at the ARN resolve step ("stack does not exist"), and recovering an LMI failure needs "re-run all jobs" so the setup job redeploys the stacks.

packages/testing/src/lmi/deploySharedCapacityProvider.ts
```diff
@@ -0,0 +1,34 @@
...
+ * ```yaml
+ * - run: npm run lmi:deploy -w packages/testing
+ * ```
+ * The ARNs are deliberately not exposed as job outputs: they contain the AWS
+ * account id, which CI masks, and GitHub silently drops job outputs that
+ * contain masked values. The stack names are deterministic, so test jobs
+ * resolve the ARN from the stack outputs instead (see the e2e workflow).
+ */
+const main = async (): Promise<void> => {
+  await Promise.all(
```
dreamorosi, line 20: **Blocker:** Running two `TestStack.deploy()` calls concurrently in one process also runs two `fromAssemblyBuilder` synths concurrently. toolkit-lib explicitly does not support this unless `clobberEnv: false`: each synth temporarily replaces the global `process.env` with an immutable proxy via `temporarilyWriteEnv`, and interleaved disposal can leave the process with the wrong/stale env (including leaked `CDK_OUTDIR`/`CDK_CONTEXT_JSON`). This works today only because each `App` is constructed before the synth window, so it is correct by accident. Could we either pass `clobberEnv: false` in `TestStack.#synthAssembly()` (safe here because the builder ignores the injected env), or deploy the architectures sequentially?

packages/testing/src/lmi/destroySharedCapacityProvider.ts
```diff
@@ -0,0 +1,38 @@
...
+ * ```yaml
+ * - run: npm run lmi:destroy -w packages/testing
+ * ```
+ * The stack names are deterministic for a given run id and architecture, so
+ * the teardown job reconstructs the same stacks the setup job deployed. Each
+ * destroy failure is reported but does not prevent the other architecture's
+ * teardown from being attempted.
+ */
+const main = async (): Promise<void> => {
+  const results = await Promise.allSettled(
```
dreamorosi, line 20: The same concurrent-synth concern described in `deploySharedCapacityProvider.ts` applies here.

.github/workflows/run-e2e-tests.yml
```diff
@@ -58,4 +101,102 @@ jobs:
...
+        run: npm run test:e2e:lmi -w ${{ matrix.package }}
+
+  # Tears down the shared capacity provider stacks as soon as the LMI suites
+  # finish: the non-LMI matrix doesn't use them, so it doesn't hold them up.
+  # Retries: non-LMI job failures can be recovered with "re-run failed jobs"
+  # (those cells never touch the shared stacks). LMI job failures need
+  # "re-run all jobs": this teardown destroys the stacks at the end of each
+  # attempt, so a re-run LMI cell fails fast at ARN resolution ("stack does
+  # not exist") until the setup job re-runs and redeploys them.
+  teardown-lmi-capacity-providers:
```
dreamorosi, line 182: **Blocker:** If an LMI cell is cancelled or times out before its `afterAll` runs, the per-suite function stack survives with its function still attached to the shared capacity provider. This teardown then tries to delete the provider stack while associations/ENIs still exist—the classic `DELETE_FAILED` path—leaking an EC2-backed fleet plus VPC with no sweeper to catch it. Could we add at least a retry/second pass in the destroy script, and/or scheduled cleanup for `LmiShared-*` stacks and stacks tagged `Service: Powertools-for-AWS-e2e-tests`?

.github/workflows/run-e2e-tests.yml
```diff
@@ -7,6 +7,49 @@ permissions:
   contents: read
 
 jobs:
+  # Deploys one run-scoped shared LMI capacity provider stack per architecture.
+  # EC2-backed capacity providers are the slowest resources in the LMI e2e
+  # suites and are subject to account vCPU quotas, so the suites share one per
+  # architecture instead of provisioning their own (see
+  # packages/testing/src/lmi/sharedCapacityProviderStack.ts).
+  setup-lmi-capacity-providers:
```
dreamorosi, line 15: Related to the teardown concern: none of the three new jobs sets `timeout-minutes`, so the 360-minute default lets a hung deploy/cell keep two EC2 fleets alive for up to six hours before teardown even starts. The suite hooks themselves cap at 20 minutes each, so a roughly 60-minute job timeout seems easy to justify.

.github/workflows/run-e2e-tests.yml
```diff
@@ -58,4 +101,102 @@ jobs:
...
+      # dropped by GitHub). `>> $GITHUB_ENV` is safe: masking only redacts
+      # logs, not environment values.
+      - name: Resolve shared LMI capacity provider ARN
+        env:
+          ARCH: ${{ matrix.arch }}
+        run: |
+          arn=$(aws cloudformation describe-stacks \
+            --stack-name "LmiShared-${GITHUB_RUN_ID}-${ARCH//_/-}" \
+            --query "Stacks[0].Outputs[?OutputKey=='CapacityProviderArn'].OutputValue" \
+            --output text)
```
dreamorosi, line 155: **Should-fix:** `--output text` exits 0 with an empty string when the JMESPath filter matches nothing, and prints `None` when outputs are null. An empty value is silently treated as “no shared provider” by the test, so all four cells fall back to provisioning their own VPC and capacity provider, defeating the shared-provider design and pressuring the VPC quota without warning. Please fail unless the value is non-empty and not `None`, for example: `[[ -n "$arn" && "$arn" != "None" ]] || exit 1`.

packages/testing/src/resources/TestNodejsFunction.ts
```diff
@@ -71,6 +80,72 @@ class TestNodejsFunction extends NodejsFunction {
...
+    if (typeof capacityProvider === 'string') {
+      this.#attachToCapacityProviderArn(capacityProvider, scaling);
+    } else {
+      const {
+        perExecutionEnvironmentMaxConcurrency,
+        executionEnvironmentMemoryGiBPerVCpu,
+        minExecutionEnvironments,
+        maxExecutionEnvironments,
+      } = scaling;
+      capacityProvider.addFunction(this, {
```
dreamorosi, line 99: **Should-fix:** The ARN path explicitly sets `publishToLatestPublished = true` at line 131, while this construct path relies on CDK's `@default - True` even though the underlying CFN property documents no default. Passing `publishToLatestPublished: true` here would make both attach paths emit identical templates and remove dependence on an undocumented service default. This is also the path used by every local run and the CI fallback.

packages/logger/package.json
```diff
@@ -73,7 +74,10 @@
   "types": "./lib/cjs/index.d.ts",
   "main": "./lib/cjs/index.js",
   "devDependencies": {
-    "@aws-lambda-powertools/testing-utils": "file:../testing"
+    "@aws-lambda-powertools/testing-utils": "file:../testing",
+    "@aws-sdk/client-cloudwatch-logs": "^3.1079.0",
```
dreamorosi: **Should-fix:** This dependency is not imported anywhere in the package and looks left over from the CloudWatch-polling approach, so it can be dropped. Conversely, `@aws-sdk/client-lambda` is imported by `tests/e2e/lmi.test.ts` but is not declared here (it currently resolves via workspace hoisting), so that dependency should be declared instead.

packages/logger/tests/e2e/lmi.test.ts
```diff
@@ -0,0 +1,218 @@
...
+// other output, and when it fails these markers are the only way to tell
+// which phase died.
+const testConsole = new Console({
+  stdout: process.stdout,
+  stderr: process.stderr,
+});
+
+describe('Logger E2E - Lambda Managed Instances', () => {
+  // The LMI scheduler scales out to fresh execution environments until the
+  // capacity provider's fleet is saturated (8 environments with a 12 vCPU
```
dreamorosi: **Should-fix (comment accuracy):** The real CI run (29740823390) reported `30/30 responses; 12 multiplexed across 24 execution environments` in all four cells, so the “fleet saturates at 8 environments” model here—and in the similar file-header comment—does not match observed behavior, likely because both runtime cells now share each per-architecture provider. The assertion only needs one overlap, so the test held, but could we describe the actual mechanics so future readers do not tune counts against the wrong model?

packages/testing/src/invokeTestFunction.ts
```diff
@@ -11,6 +11,7 @@ const lambdaClient = new LambdaClient({});
 const invokeFunctionOnce = async ({
   functionName,
   payload = {},
+  includeTailLogs = true,
```
dreamorosi: **Should-fix:** This new option has no callers; the LMI suite hand-rolls its own `LambdaClient`/`InvokeCommand` instead. Could we either wire the suite through `invokeFunction` or drop the option? Also, with `includeTailLogs: false`, callers currently receive silently empty `TestInvocationLogs` rather than an error.

dreamorosi (commented):
These are findings from a deep review pass, grouped as 2 blockers plus several smaller fixes. Happy to discuss any of them.

dreamorosi (changes requested):
Apologies for the long wait - I just finally found some time to review this now.

Some review comments that I think are valid, let me know if you disagree and/or I misunderstood anything.

dreamorosi:
The infrastructure-leak concern raised in review—cancelled cells leaving functions attached to the shared capacity provider and causing `DELETE_FAILED` during teardown—will be backstopped by the scheduled sweeper tracked in #5519. Within this PR, it would be good to keep the scope to job `timeout-minutes` and a retry/second pass in the destroy script. Before merge, a fresh full e2e workflow run against the current head SHA would also be good, since the last green run (29740823390) predates the newest commits and the merge from main.

packages/testing/src/invokeTestFunction.ts
svozza: So this was initially included because you cannot invoke an LMI function with the `TAIL` option. Initiall we were using these invoke function helpers and then checking the CloudWatch logs. Then we changed to capturing the log output by intercepting stdout and just returning a function payload with the logs to be validated in the response. If we ever decided to write tests that use these helpers then we will need this functionality. I'm happy to remove it on the YAGNI principle but I do think we should leave a comment so this information doesn't get lost.

.github/workflows/run-e2e-tests.yml
svozza: Agreed this shouldn't rely on a best-effort cron. The teardown job is already `if: always()` and `needs: run-e2e-tests-lmi`, so it *does* run when a cell is cancelled or times out — the gap is only that it deletes the providers without first clearing any orphaned function stacks, and those orphans' ENIs are exactly what make the provider delete fail.

Plan: make teardown authoritative with an ordered, retried sequence:

1. **Sweep first** — `ListStacks` filtered to stacks whose name carries **both** an LMI marker **and** the current `GITHUB_RUN_ID`, and delete any found. We'll name the LMI function stacks accordingly (the shared provider already uses the `LmiShared-` prefix; the function stacks will get a matching `Lmi` marker plus the run id), so the filter matches *exactly this run's LMI function stacks* — never non-LMI stacks and never another run's. On the happy path this is a no-op, since each suite's `afterAll` already deleted its own stack; it only does work when a cell died before teardown. This must precede step 2 because the orphaned function stack pins the ENIs on the shared provider.
2. **Delete the provider stacks** (`LmiShared-<runId>-<arch>`) — now guaranteed detached, so no `DELETE_FAILED`.

Both steps get a retry pass for transient CFN throttling. That makes teardown deterministic for every normal / cancelled / timed-out cell.

The remaining cases teardown can't cover — the whole workflow being cancelled so `teardown-lmi` never starts, env-protection blocks, `DISABLE_TEARDOWN`, and `LmiShared-local-*` stacks from dev machines — are tracked separately as the scheduled sweeper in #5519.

dreamorosi:
The temporary orphan-cleanup experiment succeeded in [the latest e2e run](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/31110436322): all four LMI cells passed, teardown found and deleted all four deliberately orphaned function stacks, and both shared capacity provider stacks were then deleted successfully. The workflow was red overall because four non-LMI matrix cells failed, but the targeted LMI teardown path completed successfully.

Before this is merge-ready, please:

- revert the temporary `f0faefd8` commit that sets `DISABLE_TEARDOWN`
- bring the branch up to date with `main` and resolve the current conflicts in `package-lock.json` and `packages/testing/package.json`
- run the final checks against the resulting head

I have resolved the review threads whose requested changes are now present. I am leaving the requested-changes review in place until the temporary commit and conflicts are cleared.

svozza:
Yeah, we're not finished yet, I think the no clobber change is causing unrelated tests to fail. I'm going to be doing multiple CI runs to diagnose this so we can't merge until that's done anyway.

svozza:
Thanks @dreamorosi — since you've already resolved the threads from the commits, just closing the loop with everything that landed:

**Blockers**
- **Concurrent-synth `process.env` race** — `clobberEnv: false` in `TestStack.#synthAssembly()` (covers both the deploy and destroy scripts).
- **Teardown leak on cancelled/timed-out cells** — LMI function stacks now carry an `Lmi-<runId>` marker; the teardown job sweeps leftover function stacks and waits for them to delete **before** the provider stacks (so ENIs are detached and the provider delete can't `DELETE_FAILED`), with a retry pass. Residual cases (workflow-level cancellation, `DISABLE_TEARDOWN`, `LmiShared-local-*`) tracked in #5519.

**Should-fixes**
- `timeout-minutes: 60` on all three LMI jobs.
- ARN resolution now fails loudly on empty/`None` instead of silently falling back to per-suite provisioning.
- `publishToLatestPublished` set explicitly on the construct attach path so both paths emit identical templates.
- Logger deps: dropped unused `@aws-sdk/client-cloudwatch-logs`, declared `@aws-sdk/client-lambda`.
- Multiplexing comment in `lmi.test.ts` rewritten to match observed behaviour (barrier guarantees the overlap; env counts vary and aren't asserted).
- Removed the unused `includeTailLogs` option, with a comment on `invokeFunctionOnce` noting `Tail` is unsupported on capacity-provider functions.

**Validation**
- Teardown sweep proven end-to-end in real CI (a run with teardown skipped on every LMI cell left 4 orphans; the sweep found and deleted all 4, then both providers deleted cleanly with the real teardown role) and reproduced locally against real capacity-provider infra.
- Happy-path LMI run: setup + all 4 LMI cells + teardown green, zero throttling.

~**⚠️ Not ready to merge yet**
While validating I hit non-LMI cells failing with `NoStack: CloudFormationStack object does not hold a stack`. It's **not** from this PR — it reproduces on a clean `main` (run [31155239957](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/31155239957), 6/40 cells) and is independent of the `clobberEnv` change (confirmed by a controlled run with it reverted). Filed as #5537. I'd like to **hold this PR until #5537 is resolved**, otherwise the branch's CI stays red on unrelated flakes and we can't get a clean signal that the LMI work is green.~ Edit: issue in CDK has been fixed upstream.

svozza:
Merge blocker cleared. #5537 (the pre-existing `NoStack` flake) is fixed upstream in `@aws-cdk/toolkit-lib` 1.39.0 (aws/aws-cdk-cli#1848); this branch is rebased on latest `main` and bumps the pin to `^1.39.0`.

Full e2e matrix on the updated branch — [run 32657685591](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/32657685591) — is **46/46 green, zero `NoStack`** (setup + all 4 LMI cells + teardown included). This is the first fully-green run, so the LMI work is now verifiable on a clean signal. Ready for review/merge.

dreamorosi:
@svozza - ack, let's resolve the merge conflicts and do another run, once green tag me and I'll approve it, so we can finally merge

svozza:
@dreamorosi rebase completed

dreamorosi (approved):
Thanks for the long and complex work on this, appreciate it!

svozza:
Yeah, glad we got it overr the line. Was a real gap in the tests and we now have a nice foundation for the the other utilities

Issue #3705, opened 2025-03-10
Feature request: Clear log buffer before starting next one


zirkelc:
Hi,

I went through the code of the new log buffering feature and noticed a potential memory issue. The way it is currently implemented, both `clearBuffer()` and `flushBuffer()` will reset the log items only for the current X-Ray Trace ID:

`clearBuffer`
https://github.com/aws-powertools/powertools-lambda-typescript/blob/cab716df528b0f6ce9082dbc089cd5c234ab1264/packages/logger/src/Logger.ts#L1386-L1392

`flushBuffer`
https://github.com/aws-powertools/powertools-lambda-typescript/blob/cab716df528b0f6ce9082dbc089cd5c234ab1264/packages/logger/src/Logger.ts#L1379

However, that means if the `Logger` is created outside of the handler and is not explicitly cleared, or flushed through an error, the buffer will keep logs from previous invocations with different X-Ray Trace IDs. For example, a Lambda instance that runs every few minutes will be re-used and eventually fill up the buffer with log items.

Since the buffer is implemented as a `Map`, I think calling `this.#buffer?.clear()` instead of `this.#buffer?.delete(traceId)` should have the same effect without the potential memory issue.

dreamorosi:
Hi @zirkelc, this design was intentional however now that you mention it I think we should've called this out in the documentation more clearly.

The `logger.clearBuffer()` method is called automatically after each invocation/request whenever you use the `injectLambdaContext()` Middy.js middleware or class method decorator. Because of this, there's no danger of the buffer growing in size.

The only case where this can happen is if you're not using either of the two above, and instead you're using the logger "manually". In that case you're responsible for either flushing it or clearing it manually, i.e.

```ts
import { Logger } from '@aws-lambda-powertools/logger';

const logger = new Logger({
  logBufferOptions: { enabled: true }
});

export const logger = async () => {
  logger.debug('a buffered log');

  try {

  } finally {
    logger.clearBuffer();
  }
}
```

We decided to go in this direction because we don't know yet how our customers will use this feature and while we officially support AWS Lambda as only compute environment, if customers use the logger elsewhere and especially in environments that support concurrent requests, clearing the entire Map would result in data loss.

As I mentioned, I believe we don't call out the fact that you need to clear the buffer when not using any of the two `injectLambdaContext()`, I'll open a PR to add it to the docs.

dreamorosi:
Retrieving comments

[...]

PR #5543, opened 2026-08-08
fix(logger): scope log buffer per invocation under LMI concurrency

### LINKED ISSUE #5541: Bug: Logger log buffer wipes other in-flight invocations' buffered logs under Lambda Managed Instances concurrency
### Expected Behavior

When multiple invocations are multiplexed into the same execution environment (Lambda Managed Instances with `perExecutionEnvironmentMaxConcurrency` > 1), each invocation using the log buffer should keep its own buffered logs until it flushes or clears them. An invocation that errors should be able to flush the debug logs it buffered, regardless of what other in-flight invocations did in the meantime.

### Current Behavior

`Logger.bufferLogItem()` clears the entire buffer whenever it sees a trace id for the first time:

```typescript
// packages/logger/src/Logger.ts
protected bufferLogItem(xrayTraceId: string, log: LogItem, logLevel: number): void {
  log.prepareForPrint();
  // This is the first time we see this traceId, so we need to clear the buffer
  // from previous requests. This is ok because in AWS Lambda, the same sandbox
  // environment can only ever be used by one request at a time.
  if (this.#buffer?.has(xrayTraceId) === false) {
    this.#buffer?.clear();
  }
  ...
}
```

The comment states the exact assumption that LMI multiplexing breaks. The buffer is a single `CircularMap` on the shared `Logger` instance, keyed by trace id, and trace id resolution is already InvokeStore-aware, so under LMI each concurrent invocation gets its own distinct key. That is precisely what makes the `clear()` fire: invocation B's first buffered log wipes every other in-flight invocation's entries, not just stale ones.

Consequences: invocation A buffers debug logs, invocation B starts concurrently and buffers its first log, A's entries are deleted. When A later errors and `flushBuffer()` runs, its buffer is gone and nothing is emitted, not even the eviction warning, because `hasEvictedLog` was never set. The debug context for the exact failure the buffer exists to explain is silently lost. The wipe also ping-pongs: A's next buffered log wipes B's entries in return.

Related: #5429 fixed the same class of bug (per-invocation state in init-scoped storage) for log attributes via `LogAttributesStore`, but the buffer does not go through that store.

### Code snippet

Failing unit test (put in `packages/logger/tests/unit/concurrency/`):

```typescript
import { InvokeStore } from '@aws/lambda-invoke-store';
import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
import { Logger } from '../../../src/index.js';

const XRAY_TRACE_ID_KEY = Symbol.for('_AWS_LAMBDA_X_RAY_TRACE_ID');

describe('Logger buffer under concurrent invocations', () => {
  beforeEach(() => {
    InvokeStore._testing?.reset();
    vi.stubEnv('POWERTOOLS_DEV', 'true');
    vi.stubEnv('AWS_LAMBDA_MAX_CONCURRENCY', '10');
    vi.clearAllMocks();
  });

  afterEach(() => {
    vi.unstubAllEnvs();
  });

  it('keeps each invocation buffered logs when a concurrent invocation starts buffering', async () => {
    // Prepare
    const logger = new Logger({
      logLevel: 'INFO',
      logBufferOptions: { enabled: true },
    });
    const store = await InvokeStore.getInstanceAsync();
    const aBuffered = Promise.withResolvers<void>();
    const bBuffered = Promise.withResolvers<void>();

    // Act
    // Invocation A buffers a debug log, yields (simulating I/O), invocation B
    // starts concurrently and buffers its first debug log under a different
    // trace id, then A errors and flushes its buffer
    const invocationA = store.run(
      { [XRAY_TRACE_ID_KEY]: '1-aaaaaaaa-111111111111111111111111' },
      async () => {
        logger.debug('A buffered debug log');
        aBuffered.resolve();
        await bBuffered.promise;
        logger.flushBuffer();
      }
    );
    const invocationB = (async () => {
      await aBuffered.promise;
      await store.run(
        { [XRAY_TRACE_ID_KEY]: '1-bbbbbbbb-222222222222222222222222' },
        async () => {
          logger.debug('B buffered debug log');
        }
      );
      bBuffered.resolve();
    })();
    await Promise.all([invocationA, invocationB]);

    // Assess
    expect(console.debug).toHaveLogged(
      expect.objectContaining({ message: 'A buffered debug log' })
    );
  });
});
```

Fails on current main: A's `flushBuffer()` emits nothing because B's first buffered log cleared the whole buffer. Removing the `clear()` makes it pass.

### Steps to Reproduce

1. Add the test above and run `npx vitest run tests/unit/concurrency` in `packages/logger`
2. The assertion fails: invocation A's buffered log was wiped by invocation B

On real LMI: enable log buffering, attach the function to a capacity provider with `perExecutionEnvironmentMaxConcurrency` > 1, fire simultaneous invocations, and make one of them throw. Its buffered debug logs are not flushed.

### Possible Solution

`CircularMap` already supports multiple trace keys concurrently and `flushBuffer()`/`clearBuffer()` already delete only the current trace's entry, so the `clear()` looks removable. The remaining question is stale-entry cleanup for invocations that never flush (the reason the wipe exists): options include relying on the buffer's existing byte-budget eviction, or scoping buffered entries through the InvokeStore like #5429 did for attributes so entries die with their invocation context.

latest

24.x

### Packaging format used

npm

### Execution logs

```
N/A, see failing unit test above.
```

vishwakt:
### Changes

The log buffer cleared all entries whenever a log was buffered for a trace id it had not seen before, based on the assumption that only one request can be in flight per execution environment (the comment in `bufferLogItem()` stated this explicitly). Under Lambda Managed Instances concurrency that assumption no longer holds: an invocation starting to buffer wiped every other in-flight invocation's buffered logs, so a failing invocation had nothing left to flush and its debug context was silently lost.

Following the direction agreed in the issue, the buffer is now scoped through the InvokeStore when concurrency is enabled (`AWS_LAMBDA_MAX_CONCURRENCY` set): a new private `#getBuffer()` accessor returns an invocation-scoped `CircularMap` stored in the InvokeStore, so each invocation's buffered logs live and die with its invocation context and no cross-invocation cleanup is needed. The sequential path is unchanged: it keeps the instance-level buffer and its new-trace `clear()` behavior, preserving the stale-buffer safeguard introduced in #3705.

`bufferLogItem()`, `flushBuffer()`, and `clearBuffer()` all resolve the buffer through the accessor. The `clear()` now only runs on the sequential path.

Added regression tests in `tests/unit/concurrency/logBuffer.test.ts`:
- interleaved invocations keep and flush their own buffered logs (fails on main, where invocation B's first buffered log wipes invocation A's buffer)
- each invocation flushes independently while the other is mid-flight
- `clearBuffer()` throws when concurrency is enabled but the InvokeStore is unavailable, matching the behavior of the attribute store

All existing logger unit tests pass unchanged (226 tests), confirming sequential behavior is preserved, with 100% coverage on `Logger.ts`.

**Issue number:** closes #5541

---

### DIFF
--- a/packages/logger/src/Logger.ts
+++ b/packages/logger/src/Logger.ts
@@ -16,6 +16,7 @@ import {
   getStringFromEnv,
   getXRayTraceIdFromEnv,
   isDevMode,
+  shouldUseInvokeStore,
 } from '@aws-lambda-powertools/commons/utils/env';
 import type { Callback, Context, Handler } from 'aws-lambda';
 import {
@@ -214,9 +215,17 @@ class Logger extends Utility implements LoggerInterface {
   };
 
   /**
-   * Contains buffered logs, grouped by `_X_AMZN_TRACE_ID`, each group with a max size of `maxBufferBytesSize`
+   * Contains buffered logs, grouped by `_X_AMZN_TRACE_ID`, each group with a max size of `maxBufferBytesSize`.
+   *
+   * Used when invocations run sequentially; under concurrency the buffer is scoped
+   * to each invocation via the InvokeStore, see {@link Logger.#getBuffer | `#getBuffer()`}.
    */
   #buffer?: CircularMap<string>;
+
+  /**
+   * Key used to store the buffer in the InvokeStore when concurrency is enabled.
+   */
+  readonly #bufferKey = Symbol('powertools.logger.buffer');
 
   /**
    * Search function for the correlation ID.
@@ -1419,19 +1428,54 @@ class Logger extends Utility implements LoggerInterface {
    * @param log - Log to be buffered
    * @param logLevel - The level of log to be buffered
    */
+  /**
+   * Get the buffer holding this invocation's logs.
+   *
+   * When invocations run concurrently in the same execution environment
+   * (Lambda Managed Instances), each invocation gets its own buffer stored in
+   * the InvokeStore, so buffered logs die with their invocation context.
+   * Otherwise the instance-level buffer shared across sequential invocations
+   * is used.
+   */
+  #getBuffer(): CircularMap<string> | undefined {
+    if (this.#bufferConfig.enabled === false) {
+      return undefined;
+    }
+    if (!shouldUseInvokeStore()) {
+      return this.#buffer;
+    }
+
+    if (globalThis.awslambda?.InvokeStore === undefined) {
+      throw new Error('InvokeStore is not available');
+    }
+
+    const store = globalThis.awslambda.InvokeStore;
+    let buffer = store.get(this.#bufferKey) as CircularMap<string> | undefined;
+    if (buffer == null) {
+      buffer = new CircularMap({
+        maxBytesSize: this.#bufferConfig.maxBytes,
+      });
+      store.set(this.#bufferKey, buffer);
+    }
+    return buffer;
+  }
+
   protected bufferLogItem(
     xrayTraceId: string,
     log: LogItem,
     logLevel: number
   ): void {
     log.prepareForPrint();
-    // This is the first time we see this traceId, so we need to clear the buffer
-    // from previous requests. This is ok because in AWS Lambda, the same sandbox
-    // environment can only ever be used by one request at a time.
-    if (this.#buffer?.has(xrayTraceId) === false) {
-      this.#buffer?.clear();
+    const buffer = this.#getBuffer();
+    // When invocations run sequentially, seeing a new traceId means the
+    // previous request is done, so its leftover entries are cleared to avoid
+    // retaining stale logs. Under concurrency the buffer is scoped to the
+    // invocation via the InvokeStore, so no cleanup is needed and clearing
+    // would wipe other in-flight invocations' logs.
+    if (!shouldUseInvokeStore() && buffer?.has(xrayTraceId) === false) {
+      buffer?.clear();
     }
-    this.#buffer?.setItem(
+    buffer?.setItem(
       xrayTraceId,
       JSON.stringify(
         log.getAttributes(),
@@ -1454,7 +1498,8 @@ class Logger extends Utility implements LoggerInterface {
       return;
     }
 
-    const buffer = this.#buffer?.get(traceId);
+    const requestBuffer = this.#getBuffer();
+    const buffer = requestBuffer?.get(traceId);
     if (buffer === undefined) {
       return;
     }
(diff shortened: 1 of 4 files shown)


packages/logger/src/Logger.ts
```diff
@@ -1419,19 +1428,54 @@ class Logger extends Utility implements LoggerInterface {
    * @param log - Log to be buffered
    * @param logLevel - The level of log to be buffered
    */
+  /**
```
dreamorosi: Could you move the `Add a log to the buffer` TypeDoc block so it sits immediately above `bufferLogItem()`? Inserting `#getBuffer()` here detached the block and its `@param` tags from the method they document.

dreamorosi (changes requested):
Thanks for the contribution! I’ve left only one small inline comment. The implementation otherwise looks good and matches the intended behavior. Happy to merge once that comment is addressed.

svozza:
I would prefer if we followed the pattern of isolating all interactions with the async context in the [`LogAttributesStore` ](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/logger/src/LogAttributesStore.ts)abstraction. This is the pattern we have used for all other shared state that neds to be scoped across invocations.

vishwakt:
Thanks both. Moved all the buffer's InvokeStore access into a dedicated `LogBufferStore` that mirrors `LogAttributesStore`, so `Logger` no longer touches the async context directly. That also reunited the `bufferLogItem` TypeDoc block with its method. Behavior and tests are unchanged.

Note: the log-level scoping from #5544 (already merged) still accesses `InvokeStore` directly in `Logger`. Happy to follow up with a small PR moving that into a store too, for consistency, if you'd like.

dreamorosi:
> Note: the log-level scoping from #5544 (already merged) still accesses `InvokeStore` directly in `Logger`. Happy to follow up with a small PR moving that into a store too, for consistency, if you'd like.

Yes please, I missed that during my review - please open a second PR, you can reuse the existing issue you already had.

vishwakt:
Opened #5548 for the log-level follow-up @dreamorosi asked about — it moves the `InvokeStore` access from #5544 into a dedicated `LogLevelStore`, matching the `LogBufferStore` here.

Heads-up: both this PR and #5548 touch the same import block and adjacent regions of `Logger.ts`, so whichever merges second will have a small conflict to resolve (the `shouldUseInvokeStore` import plus the store field/method neighborhood). I'm happy to rebase whichever one you'd like to land first.

dreamorosi (approved):
Thanks for the PR and thanks for addressing the review comments.

PR #5548, opened 2026-08-10
refactor(logger): isolate log level InvokeStore access in a store

### LINKED ISSUE #5542: Bug: Logger debug sampling decision mutates shared log level across concurrent invocations
### Expected Behavior

With `sampleRateValue` configured, each invocation's sampling decision should apply to that invocation only: an invocation that was sampled in should emit debug logs for its entire duration, and one that was sampled out should not.

### Current Behavior

`refreshSampleRateCalculation()` runs once per invocation (from both the `injectLambdaContext` middy middleware and the decorator) and writes its random sampling decision into `this.logLevel`, a plain field on the `Logger` instance that is shared by all concurrent invocations:

```typescript
// packages/logger/src/Logger.ts
public refreshSampleRateCalculation(): void {
  if (this.#debugLogSampling.refreshedTimes === 0) {
    this.#debugLogSampling.refreshedTimes++;
    return;
  }
  if (this.#shouldEnableDebugSampling() && this.logLevel > LogLevelThreshold.TRACE) {
    this.setLogLevel('DEBUG');  // writes shared this.logLevel
  } else {
    this.setLogLevel(this.getLogLevelNameFromNumber(this.#initialLogLevel));
  }
}
```

`this.logLevel` gates every log call. Under LMI multiplexing, whichever invocation's refresh ran last wins for all in-flight invocations: a sampled-in invocation has its debug logs silently dropped after a concurrent sampled-out invocation resets the level, and a sampled-out invocation emits debug logs it was never sampled for (inflating log volume and potentially leaking verbose data for unsampled requests). The `#debugLogSampling.refreshedTimes` cold-start guard is shared the same way, so if the first two invocations of a fresh environment overlap, only one is treated as the cold start.

`LogAttributesStore` (#5429) covers attributes, context, and correlation id, but not the effective log level.

### Code snippet

Failing unit test (put in `packages/logger/tests/unit/concurrency/`; `randomInt` is mocked to make the sampling rolls deterministic):

```typescript
import { randomInt } from 'node:crypto';
import { InvokeStore } from '@aws/lambda-invoke-store';
import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
import { Logger } from '../../../src/index.js';

vi.mock('node:crypto', async (importOriginal) => {
  const mod = await importOriginal<typeof import('node:crypto')>();
  return { ...mod, randomInt: vi.fn(mod.randomInt) };
});

const XRAY_TRACE_ID_KEY = Symbol.for('_AWS_LAMBDA_X_RAY_TRACE_ID');

describe('Logger debug sampling under concurrent invocations', () => {
  beforeEach(() => {
    InvokeStore._testing?.reset();
    vi.stubEnv('POWERTOOLS_DEV', 'true');
    vi.stubEnv('AWS_LAMBDA_MAX_CONCURRENCY', '10');
    vi.clearAllMocks();
  });

  afterEach(() => {
    vi.unstubAllEnvs();
  });

  it('applies the debug sampling decision only to the invocation that was sampled', async () => {
    // Prepare
    // Rolls: constructor -> not sampled (INFO); invocation A's refresh ->
    // sampled in (DEBUG); invocation B's refresh -> not sampled (INFO)
    vi.mocked(randomInt)
      .mockReturnValueOnce(99 as never)
      .mockReturnValueOnce(0 as never)
      .mockReturnValueOnce(99 as never);
    const store = await InvokeStore.getInstanceAsync();
    const logger = new Logger({ logLevel: 'INFO', sampleRateValue: 0.5 });
    // Cold-start invocation: the first refresh keeps the constructor decision
    logger.refreshSampleRateCalculation();
    const aStarted = Promise.withResolvers<void>();
    const bRefreshed = Promise.withResolvers<void>();

    // Act
    // Warm invocation A is sampled in by its refresh, emits a debug log,
    // yields, invocation B's refresh re-rolls and is NOT sampled, then A
    // emits another debug log
    const invocationA = store.run(
      { [XRAY_TRACE_ID_KEY]: '1-aaaaaaaa-111111111111111111111111' },
      async () => {
        logger.refreshSampleRateCalculation();
        logger.debug('A sampled debug log 1');
        aStarted.resolve();
        await bRefreshed.promise;
        logger.debug('A sampled debug log 2');
      }
    );
    const invocationB = (async () => {
      await aStarted.promise;
      await store.run(
        { [XRAY_TRACE_ID_KEY]: '1-bbbbbbbb-222222222222222222222222' },
        async () => {
          logger.refreshSampleRateCalculation();
        }
      );
      bRefreshed.resolve();
    })();
    await Promise.all([invocationA, invocationB]);

    // Assess
    expect(console.debug).toHaveLogged(
      expect.objectContaining({ message: 'A sampled debug log 1' })
    );
    expect(console.debug).toHaveLogged(
      expect.objectContaining({ message: 'A sampled debug log 2' })
    );
  });
});
```

Fails on current main: the first assertion passes (A was sampled in and emitted its first debug log), the second fails (A's second debug log was dropped after B's refresh reset the shared level). With B's concurrent refresh removed, both pass.

### Steps to Reproduce

1. Add the test above and run `npx vitest run tests/unit/concurrency` in `packages/logger`
2. The second assertion fails

On real LMI: configure `POWERTOOLS_LOGGER_SAMPLE_RATE`, run with `perExecutionEnvironmentMaxConcurrency` > 1, and compare sampled invocations' expected vs actual debug output; debug coverage becomes non-deterministic under overlap.

### Possible Solution

Store the effective per-invocation log level in invoke-scoped storage (the same InvokeStore-backed pattern as `LogAttributesStore` from #5429) with the instance field as fallback outside a context, so a refresh only affects the invocation that rolled it. The `refreshedTimes` cold-start guard needs the same treatment or an InvokeStore-aware equivalent.

latest

24.x

### Packaging format used

npm

### Execution logs

```
N/A, see failing unit test above.
```

vishwakt:
### Changes

Follow-up to the review discussion on #5543: the per-invocation log level scoping (added in #5542 / #5544) accessed the `InvokeStore` directly from within `Logger`. Per @svozza's feedback, rather than a dedicated store this folds the log level into the existing `LogInvocationStore` (introduced for the buffer in #5543), keeping `LogAttributesStore` separate.

- `LogInvocationStore` now also owns the effective log level: `getLogLevel()`/`setLogLevel()` resolve the invocation-scoped level when a context is active, and `getBaseLogLevel()`/`setBaseLogLevel()` operate on the base level (set at initialization and for Advanced Logging Controls, which apply environment-wide).
- Because the log level is always needed (buffering is optional), the store is now created eagerly and its buffer is enabled via `configureBuffer()` when log buffering is turned on.
- `Logger` delegates every log-level read and write to the store and no longer references the `InvokeStore` directly for either the level or the buffer.

Pure refactor, no behavior change: the existing concurrency, sampling, and buffer suites pass unchanged (231 logger tests), with 100% coverage on `Logger.ts` and `LogInvocationStore.ts`.

This branch merges `main` (which has #5543) and resolves the conflict, so no rebase is needed.

**Issue number:** fixes #5542

---


svozza:
A store just for the log level seems overkill, I wonder can we put the buffer and log level in the same store as they are quite related. The log attributes stuff makes sense to be logically separated though I think we could make an argument that there could just be one store for all async context related stuff. That however would be a bigger change here that could be. done later as none of these stores are public facing.

vishwakt:
Good point @svozza, I'll consolidate. Rather than a dedicated LogLevelStore, I'll move the log-level scoping into the same store as the buffer from [#5543](https://github.com/aws-powertools/powertools-lambda-typescript/pull/5543), keeping LogAttributesStore separate which matches your "buffer and level are related" grouping. I'd suggest letting [#5543](https://github.com/aws-powertools/powertools-lambda-typescript/pull/5543) land first (it's the bug fix), then reworking this PR on top to fold the level into that store (renamed to cover both), which also avoids the churn of introducing a store here that then gets absorbed. The single-store-for-all-async-context idea makes sense as a later, separate change. What do you think?

svozza:
Agreed on mergeing https://github.com/aws-powertools/powertools-lambda-typescript/pull/5543 first and also that we need to rework that PR too so it has a more generic store name. And yes, store consolidation is out of scope here too.

vishwakt:
@svozza renamed the store to LogInvocationStore in [#5543](https://github.com/aws-powertools/powertools-lambda-typescript/pull/5543) (kept it holding just the buffer for now). Once that lands I'll rework this PR to fold the log level into it and drop the standalone LogLevelStore.

dreamorosi:
@vishwakt the other PR is merged, please rebase & resolve conflicts when ready.

vishwakt:
@svozza the store was renamed to `LogInvocationStore` in #5543, which is now merged. I'll rework this PR to fold the log level into it and drop the standalone `LogLevelStore`.

vishwakt:
@svozza @dreamorosi , this one's ready for review

dreamorosi:
@vishwakt I think your agent might've mistakenly overwritten the PR body, ask it to restore its format please.

We'll review later today!

svozza:
Do we need a concurrency test that specifically targets log levels? There's quite a bit of conditional logic here that has the added complication of being dependent on the absence/presence of the invoke store. I want to make sure we're covered here.

vishwakt:
@svozza Log-level scoping under concurrency is covered in [tests/unit/concurrency/sampling.test.ts](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/logger/tests/unit/concurrency/sampling.test.ts), which exercises exactly the InvokeStore presence/absence branches you flagged:

- throws when reading/setting the level with the InvokeStore unavailable (concurrency on, store missing),
- sets the base level when called outside an invocation context (the fallback path),
- applies the sampling decision only to the sampled invocation across interleaved invocations,
- keeps an invocation-scoped level from leaking into later invocations.

I've also added a test that interleaves two concurrent invocations each calling setLogLevel and asserts neither clobbers the other's effective level. Also, I confirmed it fails if the level is stored as a shared instance field and passes with the per-invocation store, so it's guarding the actual behavior.

svozza:
Is there a reason why we didn't use the `sequence` utility from the testing package? The other concurrency tests use it and it looks like we're reimplantimg it here but I maybe I'm missing something.

vishwakt:
No good reason, I just hadn't reached for it. Switched to sequence in the latest commit.

the sampling tests still use explicit store.run() because they need specific X-Ray trace IDs in the context, and sequence runs invocations with an empty context. Happy to extend sequence to accept a per-invocation context object if you'd like those converted too.

svozza:
> Happy to extend sequence to accept a per-invocation context object if you'd like those converted too.

Oh good shout! Yeah, let's do this, it's a useful addition. If the context is not provided, we can fall back to an empty object.

svozza (approved):

svozza:
Thank you for your work on this @vishwakt, the PR is in a great state now. Merging!

PR #5567, opened 2026-08-18
ci(maintenance): add aggregate unit-test check for branch protection

### LINKED ISSUE #5569: Maintenance: unit tests and dependency review do not gate merges on main
### Summary

Branch protection on `main` currently requires `Acknowledgment Check`, `Related Issue Check`, `Semantic PR Title Check` and `CodeQL`. It does not require the unit test workflow or `dependency-review`, so a pull request whose tests are red can still be merged — nothing stops it mechanically, only reviewer attention.

Making the existing contexts required isn't a straight fix. `run-unit-tests` fans out to 30 jobs (two Node.js versions across 13 workspaces, plus four more jobs), and required checks are listed by name, so pinning all 30 would silently change meaning every time a workspace or Node.js version is added or renamed.

### Why is this needed?

The protection rule is the only thing that holds when attention lapses — a busy review day, an automated pull request, or a maintainer merging on a phone. Right now the most valuable signal we produce, the test suite, isn't part of it, so "the checks passed" is a convention rather than a guarantee.

It is also a prerequisite for #5491. Auto-merging even the most boring Dependabot updates means merging without a human reading the checks, and that is only safe once the test suite is part of the protection rule instead of advisory. The workflow for that is in #5568, which must not merge until the contexts described here are required.

Automation

### Solution

Add one aggregate job to `reusable-run-linting-check-and-unit-tests.yml` that `needs` the five existing jobs and fails if any did not succeed, then require that single context plus `dependency-review`:

- `run-unit-tests / unit-tests-complete`
- `dependency-review`

Two details are easy to get wrong and worth stating:

- The aggregate job needs `if: always()`, and must treat `skipped` as a failure alongside `failure` and `cancelled`. A skipped job reports its check run as `skipped`, and GitHub counts a skipped required check as **satisfied** — so an aggregate job that skipped when a dependency failed would green-light a broken pull request
- `pr-run-linting-check-and-unit-tests.yml` triggers on `[opened, synchronize]` only. Without `reopened`, a reopened pull request never re-reports and would sit blocked forever on a context that can't arrive

`dependency-review.yml` needs no change: bare `pull_request` trigger, no `paths:` filter, so it always reports.

A context has to report at least once before it can be added to a protection rule, so the workflow change has to merge before the settings change.

### Acknowledgment

### Future readers

svozza:
## Summary

Unit tests and `dependency-review` don't gate merges on `main` today, so a pull request can merge with failing tests. This adds a single aggregate context that branch protection can require, and makes sure it always reports.

Requiring the 30 individual `code-quality` matrix contexts instead would be brittle: adding a Node.js version or a workspace silently changes which checks are required.

### Changes

- Add a `unit-tests-complete` job to `reusable-run-linting-check-and-unit-tests.yml` that `needs` all five existing jobs and fails if any of them did not succeed
- Treat `skipped` as a failure alongside `failure` and `cancelled`, and run the job with `if: always()`. Both are load-bearing: a skipped job reports its check run as `skipped`, and GitHub counts a skipped required check as *satisfied*, so an aggregate job that skipped when a dependency failed would green-light a broken pull request
- Add `reopened` to `pr-run-linting-check-and-unit-tests.yml` triggers, so a reopened pull request re-reports instead of stranding the required context

`dependency-review.yml` needs no change: it triggers on bare `pull_request` and has no `paths:` filter, so it always reports.

Follow-up for a repo admin once this merges: add `run-unit-tests / unit-tests-complete` and `dependency-review` to the required status checks for `main`. A context must report at least once before it can be required, which is why this is a separate step.

**Issue number:** closes #5569

---

### DIFF
--- a/.github/workflows/pr-run-linting-check-and-unit-tests.yml
+++ b/.github/workflows/pr-run-linting-check-and-unit-tests.yml
@@ -2,7 +2,7 @@ name: On PR code update
 
 on:
   pull_request:
-    types: [opened, synchronize]
+    types: [opened, synchronize, reopened]
 
 permissions:
   contents: read
--- a/.github/workflows/reusable-run-linting-check-and-unit-tests.yml
+++ b/.github/workflows/reusable-run-linting-check-and-unit-tests.yml
@@ -147,3 +147,24 @@ jobs:
       - *setup_dependencies
       - name: Run linting
         run: npm run lint:markdown
+  unit-tests-complete:
+    # Single stable context (`run-unit-tests / unit-tests-complete`) that branch protection can
+    # require, so that adding a Node.js version or a workspace to the `code-quality` matrix
+    # doesn't silently change which checks are required.
+    #
+    # `if: always()` is required so this job still runs when a dependency fails - otherwise it
+    # would be skipped, and a skipped required check counts as satisfied. For the same reason
+    # `skipped` is treated as a failure below: none of the jobs above are conditional, so a skip
+    # means something upstream went wrong.
+    needs: [code-quality, check-examples, check-layer-publisher, check-docs-snippets, check-docs]
+    if: always()
+    runs-on: ubuntu-latest
+    steps:
+      - name: Verify all unit test jobs succeeded
+        env:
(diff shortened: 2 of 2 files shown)


dreamorosi (approved):
Good idea - never thought of this

PR #5570, opened 2026-08-19
fix(event-handler): handle +json content types and bodyless responses

### LINKED ISSUE #5566: Bug: `http` event handler validation does not recognise `+json` structured syntax suffix media types
### Expected Behavior

The `validate` middleware in the `http` event handler should parse a request or response body as JSON whenever the `Content-Type` uses the `+json` structured syntax suffix defined in [RFC 6839 §3.1](https://datatracker.ietf.org/doc/html/rfc6839#section-3.1), not only when it is exactly `application/json`.

Given a schema of `z.object({ name: z.string() })`, a request body of `{"name":"John"}` sent as `application/vnd.api+json` should validate successfully — exactly as it does when sent as `application/json`.

### Current Behavior

`extractBody` in https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/event-handler/src/http/middleware/validation.ts#L188-L216 gates JSON parsing on a substring match:

```ts
const contentType = source.headers.get('content-type');

if (contentType?.includes('application/json')) {
  try {
    return await cloned.json();
  } catch {
    // throws RequestValidationError / ResponseValidationError ("Invalid JSON body")
  }
}

return await cloned.text();
```

So, the body falls through to `cloned.text()` and the schema receives the raw JSON **string** instead of the parsed object. Validation then fails with a type mismatch.

This affects both directions, because `extractBody` is shared by request and response validation:

| Direction | `Content-Type` | Result |
| --- | --- | --- |
| Request body from client | `application/vnd.api+json` | **422** `RequestValidationError` — `Invalid input: expected object, received string` |
| Response body from handler | `application/vnd.api+json` | **500** `ResponseValidationError` — `Invalid input: expected object, received string` |

The response case breaks a path the package explicitly supports. `handlerResultToWebResponse` keeps a user-supplied `Content-Type` (https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/event-handler/src/http/converters.ts#L490), and a test asserts it: `preserves user-provided Content-Type when body is a JSON object` (https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/event-handler/tests/unit/http/converters.test.ts#L1346). So a handler may return `application/vnd.api+json` — and validation then rejects the body the router itself just serialized, with a 500.

### Validation is the only affected layer

Drop the `validation` option from the same two routes and both work. The router handles these media types end to end; enabling validation is what breaks them. [`validation.ts:192`](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/event-handler/src/http/middleware/validation.ts#L192) is the only content-type gate in the request/response path.

### Code snippet

These tests assert the current behavior, bug included. The first two show validation turning a `+json` body into a 422/500; the last two run the same routes without validation. 

```typescript
import type { APIGatewayProxyEvent, Context } from 'aws-lambda';
import { describe, expect, it } from 'vitest';
import { z } from 'zod';

const createTestEvent = (options: {
  path: string;
  httpMethod: string;
  headers?: Record<string, string>;
}): APIGatewayProxyEvent => ({
  path: options.path,
  httpMethod: options.httpMethod,
  headers: options.headers ?? {},
  body: null,
  multiValueHeaders: {},
  isBase64Encoded: false,
  pathParameters: null,
  queryStringParameters: null,
  multiValueQueryStringParameters: null,
  stageVariables: null,
  requestContext: {
    httpMethod: options.httpMethod,
    path: options.path,
    domainName: 'localhost',
  } as APIGatewayProxyEvent['requestContext'],
  resource: '',
});

const headers = {
  'content-type': 'application/vnd.api+json',
};

describe('vnd+json media types with validation (bug)', () => {
  it('rejects a request body sent as application/vnd.api+json with 422', async () => {
    // Prepare
    const app = new Router();
    app.post('/users', () => ({ statusCode: 201, body: 'ok' }), {
      validation: { req: { body: z.object({ name: z.string() }) } },
    });

    const event = createTestEvent({
      path: '/users',
      httpMethod: 'POST',
      headers,
    });
    event.body = JSON.stringify({ name: 'John' });

    // Act
    const result = await app.resolve(event, {} as Context);

    // Assess
    expect(result.statusCode).not.toBe(201);
    expect(result.statusCode).toBe(422);
  });

  it('rejects a response body returned as application/vnd.api+json with 500', async () => {
    // Prepare
    const app = new Router();
    app.get(
      '/users',
      () => ({
        statusCode: 200,
        body: { id: '1', name: 'John' },
        headers,
      }),
      {
        validation: {
          res: { body: z.object({ id: z.string(), name: z.string() }) },
        },
      }
    );

    // Act
    const result = await app.resolve(
      createTestEvent({ path: '/users', httpMethod: 'GET', headers: {} }),
      {} as Context
    );

    // Assess
    expect(result.statusCode).not.toBe(200);
    expect(result.statusCode).toBe(500);
  });
});

describe('vnd+json media types without validation', () => {
  it('reads a request body sent as application/vnd.api+json', async () => {
    // Prepare
    const app = new Router();
    app.post('/users', async ({ req }) => {
      const body = (await req.json()) as { name: string };
      return { statusCode: 201, body: { name: body.name } };
    });

    const event = createTestEvent({
      path: '/users',
      httpMethod: 'POST',
      headers,
    });
    event.body = JSON.stringify({ name: 'John' });

    // Act
    const result = await app.resolve(event, {} as Context);

    // Assess
    expect(result.statusCode).toBe(201);
    expect(JSON.parse(result.body)).toEqual({ name: 'John' });
  });

  it('returns a response body as application/vnd.api+json', async () => {
    // Prepare
    const app = new Router();
    app.get('/users', () => ({
      statusCode: 200,
      body: { id: '1', name: 'John' },
      headers,
    }));

    // Act
    const result = await app.resolve(
      createTestEvent({ path: '/users', httpMethod: 'GET', headers }),
      {} as Context
    );

    // Assess
    expect(result.headers?.['content-type']).toBe('application/vnd.api+json');
    expect(JSON.parse(result.body)).toEqual({ id: '1', name: 'John' });
  });
});

```

Output:
```
 ✓ vnd+json media types with validation (bug) (2)
     ✓ rejects a request body sent as application/vnd.api+json with 422 17ms
     ✓ rejects a response body returned as application/vnd.api+json with 500 1ms
 ✓ vnd+json media types without validation (2)
   ✓ reads a request body sent as application/vnd.api+json 1ms
   ✓ returns a response body as application/vnd.api+json 0ms

 Test Files  1 passed (1)
        Tests  4 passed (4)
```

That every one passes is the point: the router carries `application/vnd.api+json` end to end (201 and 200 in the control pair), and only the validated routes reject it (422 and 500).

### Steps to Reproduce

1. Save the snippet above as `vnd-json.test.ts` in a project with `@aws-lambda-powertools/event-handler@2.34.0`, `zod`, and `vitest` installed.
2. Run `npx vitest run vnd-json.test.ts` — all four tests pass, i.e. the validated routes return 422 and 500 while the same routes without validation return 201 and 200.
3. Change `headers` to `application/json`: the first two tests now fail (201 and 200), confirming the media type is the only variable.

## Is this deliberate?

- It mirrors upstream. Powertools for AWS Lambda (Python) has the same check in [`aws_lambda_powertools/event_handler/middlewares/openapi_validation.py`](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/event_handler/middlewares/openapi_validation.py#L140): `content_type.startswith(APPLICATION_JSON_CONTENT_TYPE)`, which likewise rejects `application/vnd.api+json`. The TypeScript port appears to have inherited the limitation for parity.
- However, nothing in the docs, tests, or the introducing PR (#4736) shows `+json` types being considered and rejected, and the same package *explicitly tests* that a handler may return `application/vnd.api+json` (`converters.test.ts`), which response validation then breaks with a 500.

So this looks like a gap in the implementation. The same gap likely warrants an upstream issue on the Python repo.

### Possible Solution

Create a utility function like this:

```typescript
export const isJsonContentType = (
  contentType: string | null | undefined
): boolean => {
  if (!contentType) return false;
  const type = contentType.split(';')[0].trim().toLowerCase();
  return type === 'application/json' || type.endsWith('+json');
};
```

And use it in `event-handler/src/http/middleware/validation.ts`. `packages/event-handler/src/http/middleware/tracer.ts:159` has the same blind spot, we can use the function there too. 
Worth fixing alongside, or splitting into its own issue if preferred. 

Parameters keep working (`application/json; charset=utf-8`), which is what `includes` was tolerating, and `application/vnd.api+json`, `application/problem+json; charset=utf-8`, and `model/gltf+json` now match too — per [RFC 6839 §3.1](https://datatracker.ietf.org/doc/html/rfc6839#section-3.1) the `+json` suffix can appear on any top-level type, so the check isn't limited to `application/*`.

The doc note at `docs/features/event-handler/http.md:207` should be updated in the same change, since it currently documents the JSON-only behavior.

I'm happy to open a PR for this.

2.34.0

24.x

### Packaging format used

npm

### Execution logs

```Shell

```

arnabrahman:
## Summary
Body deserialization in the HTTP event handler only recognized `application/json`, so responses using RFC 6839 structured syntax suffixes (`application/vnd.api+json`, `application/problem+json`) were treated as plain strings and failed schema validation. This adds a shared `isJsonContentType` helper used by both the validation and tracer middleware, and hardens the tracer against bodyless or malformed-JSON responses.
### Changes
- Add `isJsonContentType` to `packages/event-handler/src/http/utils.ts`, matching `application/json` and any `+json` suffix, ignoring parameters and casing
- Use it in `validation.ts` `extractBody` in place of the `contentType?.includes('application/json')` check
- Use it in the tracer middleware's response-metadata capture in place of the strict `=== 'application/json'` equality check
- Skip response metadata capture when the response has no body, avoiding a spurious parse attempt
- Wrap `res.clone().json()` in the tracer with a try/catch that logs a warning instead of throwing when the body isn't valid JSON
- Update the body validation note in `docs/features/event-handler/http.md` to document `+json` support
- Add unit tests for `isJsonContentType`, the tracer's bodyless/invalid-JSON paths, and `+json` response validation 

**Issue number:** closes #5566 

---

### DIFF
--- a/docs/features/event-handler/http.md
+++ b/docs/features/event-handler/http.md
@@ -204,7 +204,8 @@ If you prefer to use the decorator syntax, you can use the same methods on a cla
 Event Handler supports built-in request and response validation using [Standard Schema](https://standardschema.dev){target="_blank"}, a common specification supported by popular TypeScript validation libraries including [Zod](https://zod.dev){target="_blank"}, [Valibot](https://valibot.dev){target="_blank"}, and [ArkType](https://arktype.io){target="_blank"}.
 
 !!! note "Body validation is limited to JSON-serializable values"
-    Standard Schema validators operate on JavaScript values, not raw bytes. For body validation, Event Handler deserializes the body before passing it to your schema: `application/json` content is parsed as an object, all other content types are passed as a plain string. Binary data, streams, and other non-serializable response types cannot be validated.
+    Standard Schema validators operate on JavaScript values, not raw bytes. For body validation, Event Handler deserializes the body before passing it to your schema: `application/json` content, and any content type with a `+json` structured syntax suffix (e.g. `application/vnd.api+json`, `application/problem+json`), is parsed as an object.
+    All other content types are passed as a plain string. Binary data, streams, and other non-serializable response types cannot be validated.
 
 #### Validating requests
 
--- a/packages/event-handler/src/http/middleware/tracer.ts
+++ b/packages/event-handler/src/http/middleware/tracer.ts
@@ -6,6 +6,7 @@ import type {
   SegmentHttpData,
   TracerOptions,
 } from '../../types/http.js';
+import { isJsonContentType } from '../utils.js';
 import { getClientIp } from './commons.js';
 import type { compress } from './compress.js';
 
@@ -156,7 +157,7 @@ const tracer = (tracer: Tracer, options?: TracerOptions): Middleware => {
 
       if (
         captureResponse &&
-        reqCtx.res.headers.get('Content-Type') === 'application/json'
+        isJsonContentType(reqCtx.res.headers.get('Content-Type'))
       ) {
         const responseBody = await reqCtx.res.clone().json();
         tracer.addResponseAsMetadata(responseBody, segmentName);
--- a/packages/event-handler/src/http/middleware/tracer.ts
+++ b/packages/event-handler/src/http/middleware/tracer.ts
@@ -159,8 +159,16 @@ const tracer = (tracer: Tracer, options?: TracerOptions): Middleware => {
         captureResponse &&
         isJsonContentType(reqCtx.res.headers.get('Content-Type'))
       ) {
-        const responseBody = await reqCtx.res.clone().json();
-        tracer.addResponseAsMetadata(responseBody, segmentName);
+        try {
+          const responseBody = await reqCtx.res.clone().json();
+          tracer.addResponseAsMetadata(responseBody, segmentName);
+        } catch (error) {
+          logger.warn(
+            'Failed to parse response body as JSON for segment %s. Response metadata will not be captured.',
+            segmentName,
+            error
+          );
+        }
       }
     } catch (err) {
       tracer.addErrorAsMetadata(err as Error);
--- a/packages/event-handler/src/http/middleware/tracer.ts
+++ b/packages/event-handler/src/http/middleware/tracer.ts
@@ -157,6 +157,7 @@ const tracer = (tracer: Tracer, options?: TracerOptions): Middleware => {
 
       if (
         captureResponse &&
+        reqCtx.res.body &&
         isJsonContentType(reqCtx.res.headers.get('Content-Type'))
       ) {
         try {
--- a/packages/event-handler/src/http/middleware/validation.ts
+++ b/packages/event-handler/src/http/middleware/validation.ts
@@ -12,6 +12,7 @@ import type {
   ValidationConfig,
 } from '../../types/http.js';
 import { RequestValidationError, ResponseValidationError } from '../errors.js';
+import { isJsonContentType } from '../utils.js';
 
 /**
  * Creates a validation middleware from the provided validation configuration.
@@ -189,7 +190,7 @@ async function extractBody(source: Request | Response): Promise<unknown> {
   const cloned = source.clone();
   const contentType = source.headers.get('content-type');
 
-  if (contentType?.includes('application/json')) {
+  if (isJsonContentType(contentType)) {
     try {
       return await cloned.json();
     } catch {
--- a/packages/event-handler/src/http/utils.ts
+++ b/packages/event-handler/src/http/utils.ts
@@ -438,10 +438,13 @@ export const getBase64EncodingFromHeaders = (headers: Headers): boolean => {
 };
 
 /**
- * Checks whether a `Content-Type` value is `application/json` or uses the RFC 6839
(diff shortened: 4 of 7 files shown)


svozza (approved):

svozza:
Great work as always @arnabrahman. Merged!

arnabrahman:
@svozza when can we expect this fix to be released? This is an ongoing issue at my work.

PR #4898, opened 2025-12-29
chore(ci): harden workflow permissions for security compliance

dreamorosi:
### Changes

Harden GitHub Actions workflow permissions to satisfy both OSSF Scorecard and SonarCloud security requirements. This addresses 6 open code scanning alerts related to token permissions.

**Strategy applied:**
- Set `permissions: {}` at workflow level (disables all permissions by default)
- Declare specific permissions at each job level (only grants what each job needs)

**Files updated:**
- `reusable-run-linting-check-and-unit-tests.yml` - added `contents: read` to 5 jobs
- `ossf_scorecard.yml` - added `contents: read` to `analysis` job
- `quality_check.yml` - added `contents: read` to 5 jobs
- `publish_layer.yml` - added `contents: read` to `build-layer` job
- `layers_partitions.yml` - added `id-token: write, contents: read` to deploy jobs
- `layers_partitions_deploy.yml` - added top-level `permissions: {}`

**Issue number:** closes #4897

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #5577, opened 2026-08-23
fix(layers): bundle all publishable utilities in the Lambda layer

### LINKED ISSUE #5576: Maintenance: Lambda layer does not bundle all published utilities (data-masking, signer, kafka, validation)
### Summary

The published Lambda layer bundles only 10 of the 14 public `@aws-lambda-powertools/*` utilities. The hardcoded `utilities` array in `layers/src/layer-publisher-stack.ts` (lines 107-118) omits four utilities that are published to npm:

- `@aws-lambda-powertools/data-masking` (added #5143)
- `@aws-lambda-powertools/signer` (added #5344)
- `@aws-lambda-powertools/kafka` (added #4058)
- `@aws-lambda-powertools/validation` (added #3592)

`validation` and `kafka` predate the last update to the layer list (#4956, which added `event-handler`); `data-masking` and `signer` came afterward. None were ever appended. `@aws-lambda-powertools/testing-utils` is `private: true` and is correctly excluded.

### Why is this needed?

Customers using the layer cannot `import` these utilities without installing them separately — the modules are simply absent from `/opt/nodejs/node_modules`, so a function relying on the layer alone fails at runtime with a module-not-found error. Beyond fixing the four missing packages, the hardcoded list is a recurring drift hazard: it has silently fallen behind the workspace four times already.

Other (Lambda layer publishing / `layers/`)

### Solution

Rather than adding the four names to the hardcoded list (which will drift again the next time a utility is added), **derive the layer contents from the workspace**: scan `packages/*/package.json` and bundle every non-private `@aws-lambda-powertools/*` package. This bundles the four missing utilities today and any future utility automatically.

Touch points:

- `layers/src/layer-publisher-stack.ts` — replace the hardcoded `utilities` array with a scan of `packages/`, filtering `private: true`. The published-install path (`<name>@<version>`), the `buildFromLocal` build/pack path (`-w packages/<dir>`), and the `npm pack` tarball naming all need to key off the derived package name/dir rather than an assumed short name.
- `layers/tests/e2e/layerPublisher.class.test.functionCode.ts` (lines 89-99) — the version-parity check has the same hardcoded list (also missing `jmespath`). Make it iterate the layer's `@aws-lambda-powertools/` directory so it verifies whatever actually shipped.

Considerations to review during implementation:

- **`validation`** pulls in `ajv@^8` → increases layer size (`jmespath`, its other dep, is already bundled).
- **`data-masking`** declares `@aws-crypto/client-node` as an *optional peer*, so it won't be installed by default. Decide whether to ship a crypto backend in the layer or leave it to the customer (consistent with the "we don't ship AWS SDK clients" policy from #5512).
- **`kafka`** optional peers are arktype/valibot/zod; `zod` is already installed in the layer, the others are optional and not bundled.
- **`signer`** adds `@smithy/*` packages (small).
- Auto-derivation means any newly-added public package ships in the layer without an explicit opt-in — acceptable given all 14 current public packages are intended to ship, but worth noting.

svozza:
## Summary

The Lambda layer bundled a hardcoded list of 10 packages, silently omitting the
published `data-masking`, `signer`, `kafka`, and `validation` utilities.
Customers relying on the layer alone hit module-not-found errors for those
features. This derives the bundled list from the workspace so the four missing
utilities are included and any future utility is bundled automatically, without
the list drifting out of sync.

### Changes

- `layers/src/layer-publisher-stack.ts`: replace the hardcoded `utilities` array with `getLayerUtilities()`, which scans `packages/*/package.json` and keeps every non-private `@aws-lambda-powertools/*` package; both the published-install and `buildFromLocal` paths now key off the derived package name / workspace dir. Adds `packTarballPrefix()` and an empty-list guard.
- `layers/tests/e2e/layerPublisher.class.test.functionCode.ts`: import and exercise the four newly-bundled utilities (`DataMasking.erase`, `validate`, `kafkaConsumer`, `SigV4Signer`) so the E2E proves they load from the layer; read the layer's `@aws-lambda-powertools/` directory for the version-parity check instead of a hardcoded list.
- `layers/tests/unit/layer-publisher.test.ts`: assert the locally-bundled layer contains every derived utility and excludes the private `testing-utils`.
- `layers/tests/unit/getLayerUtilities.test.ts` (new): pure tests for the helpers, including asserting the real workspace resolves to exactly the 14 public packages.

### Testing

- Unit + pure-function tests pass; `biome ci` and `tsc --noEmit` clean.
- Local bundle inspection: all 14 packages, 24 MB; no `@aws-sdk/*` client shipped (only the types-only `@aws-sdk/types`), preserving the #5512 invariant; `@aws-crypto/client-node` correctly absent (optional peer).
- Real E2E against a sandbox account (eu-west-1): `layers/tests/e2e/layerPublisher.test.ts` **8/8 passed** — published a live layer, deployed and invoked CJS **and** ESM functions on it exercising all four new utilities, then auto-torn-down with no leftover stacks.
- Not covered locally (handled by the release pipeline): production multi-region publish and SSM/ARN wiring.

**Issue number:** closes #5576

---


dreamorosi (approved):
There's an alert about uncontrolled paths (was there prior to this PR) which I assume will be addressed in #4898, if that's the case feel free to merge this

svozza:
Yeah, the issue is pre-exiting, just got triggered because we touched the string with the shell command. There are several other places in the script with the same issue, which the fix for https://github.com/aws-powertools/powertools-lambda-typescript/issues/5579 should handle too.

PR #5595, opened 2026-08-29
chore(maintenance): move dev tooling to devDependencies in private workspaces

### LINKED ISSUE #5593: Maintenance: move dev tooling out of production dependencies in private workspaces
### Summary

A few of our private workspaces list build/test tools (`esbuild`, `typescript`, `@types/*`) under `dependencies` instead of `devDependencies`. Because Dependabot decides whether an update is a "development" or "production" dependency by looking at where the package is declared across the repo, these misplacements cause Dependabot to label the updates as production. That in turn excludes them from the new dev-dependency auto-merge workflow, so a maintainer has to review and merge them by hand even though they're routine tooling bumps.

We found this while checking why a couple of Dependabot PRs weren't auto-merging (#5591 for `@types/node`, and the `esbuild`/tooling group). The auto-merge workflow was working correctly — the dependency declarations were the problem.

### Why is this needed?

The whole point of the auto-merge workflow is to stop maintainers from hand-merging routine dev-dependency bumps. Every tool stuck in the wrong section quietly defeats that for its updates. Fixing the declarations lets these bumps flow through auto-merge like the rest.

All three affected workspaces are private and never published to npm, so moving these packages to `devDependencies` has no effect on anything we ship — there's no risk here, just cleanup.

Automation, Governance

### Solution

Correct the dependency declarations, using the root `package-lock.json` as the source of truth for what's declared where:

**`examples/app`**
- `@types/node` — listed in both `dependencies` and `devDependencies`; remove from `dependencies`.
- `@types/aws-lambda` — same, listed in both; remove from `dependencies`.
- `typescript` — same, listed in both; remove from `dependencies`.
- `esbuild` — only in `dependencies`; move to `devDependencies`.

**`layers`**
- `esbuild` — only in `dependencies`; move to `devDependencies`.

**`packages/testing`** (`@aws-lambda-powertools/testing-utils`, private)
- `esbuild` — only in `dependencies`; move to `devDependencies`.

After the change, run `npm install` so `package-lock.json` is regenerated, and confirm the build/test still pass for these workspaces.

Note: the audit only turned up dev tooling in *private* workspaces. No published package (`packages/*`) declares a dev tool as a production dependency, so there's nothing to fix on the shipping side.

### Acknowledgment

### Future readers

svozza:
## Summary

A couple of private workspaces declared build/test tooling (`esbuild`, `typescript`, `@types/*`) under `dependencies` instead of `devDependencies`. Dependabot classifies updates as production or development based on where a package is declared across the repo, so these misplacements caused routine tooling bumps to be labelled as production dependencies and excluded from the dev-dependency auto-merge workflow. Both workspaces are private and never published to npm, so the change has no effect on anything we ship.

### Changes

- `examples/app`: remove `@types/node`, `@types/aws-lambda`, and `typescript` from `dependencies` (they were already duplicated in `devDependencies`), and move `esbuild` to `devDependencies`
- `layers`: move `esbuild` to `devDependencies`
- Update `package-lock.json` so the workspace declarations and `dev` flags reflect the moves; all resolved versions stay at main's pins

`packages/testing` keeps `esbuild` as a production dependency — it backs the `NodejsFunction` L3 construct the package provides, so it's production as far as this package is concerned (see review discussion).

**Issue number:** closes #5593

---

### DIFF
--- a/packages/testing/package.json
+++ b/packages/testing/package.json
@@ -105,11 +105,11 @@
     "@aws/lambda-invoke-store": "0.3.0",
     "@smithy/util-utf8": "^4.5.2",
     "aws-cdk-lib": "^2.266.0",
+    "esbuild": "^0.28.2",
     "promise-retry": "^2.0.1"
   },
   "devDependencies": {
     "@types/promise-retry": "^1.1.6",
-    "aws-sdk-client-mock-vitest": "^7.1.0",
-    "esbuild": "^0.28.2"
+    "aws-sdk-client-mock-vitest": "^7.1.0"
   }
 }
--- a/packages/testing/package.json
+++ b/packages/testing/package.json
@@ -105,11 +105,11 @@
     "@aws/lambda-invoke-store": "0.3.0",
     "@smithy/util-utf8": "^4.5.2",
     "aws-cdk-lib": "^2.266.0",
-    "esbuild": "^0.28.2",
     "promise-retry": "^2.0.1"
   },
   "devDependencies": {
     "@types/promise-retry": "^1.1.6",
-    "aws-sdk-client-mock-vitest": "^7.1.0"
+    "aws-sdk-client-mock-vitest": "^7.1.0",
+    "esbuild": "^0.28.2"
   }
 }
--- a/examples/app/package.json
+++ b/examples/app/package.json
@@ -33,6 +33,7 @@
     "@types/node": "26.3.0",
     "aws-cdk-lib": "^2.266.0",
     "constructs": "^10.8.1",
+    "esbuild": "^0.28.2",
     "source-map-support": "^0.5.21",
     "typescript": "^6.0.3",
     "vitest": "^4.1.11"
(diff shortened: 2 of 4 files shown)


packages/testing/package.json
```diff
@@ -105,11 +105,11 @@
     "@aws/lambda-invoke-store": "0.3.0",
     "@smithy/util-utf8": "^4.5.2",
     "aws-cdk-lib": "^2.266.0",
-    "esbuild": "^0.28.2",
```
dreamorosi, line 108: Considering the main purpose of this package is to provide CDK and it’s the parent `NodejsFunction` L3 construct site, one could argue that `esbuild` - as far as this package is concerned - is a production dependency (just like `aws-cdk-lib`).

I understand why we’re moving it, but I still wanted to explain why it was there and it was on purpose.

svozza: I was actually a bit unsure of this one tbh. I think you're right though, I'd rather it reflect reality so changing back.

svozza: Good point — reverted in fbbd53be, `esbuild` stays as a production dependency of `testing-utils` since it backs the `NodejsFunction` construct this package provides.

dreamorosi (approved):

PR #5599, opened 2026-08-31
chore(tests): add scheduled sweeper for leaked e2e infrastructure

### LINKED ISSUE #5519: Maintenance: Add a scheduled sweeper for leaked e2e infrastructure
### Summary

Add a scheduled GitHub Actions workflow that removes leaked e2e infrastructure from the test account.

This is a follow-up to #5465, which introduces run-scoped shared Lambda Managed Instances (LMI) capacity provider stacks named `LmiShared-<runId>-<arch>`. These stacks include EC2 fleets with a minimum capacity of 12 vCPUs and no scale-to-zero, a VPC, and a CloudWatch Logs interface endpoint. Unlike existing serverless e2e leftovers, leaked LMI infrastructure incurs hourly charges.

This issue tracks only the scheduled sweeper. Complementary mitigations, including job `timeout-minutes` and a destroy retry/second pass, are being handled during the review of #5465.

### Why is this needed?

A cancelled or timed-out LMI test cell can leave its per-suite function stack running, with the function attached to the shared capacity provider by ARN. Because this relationship crosses stacks and is invisible to CloudFormation, the `if: always()` teardown job can encounter `DELETE_FAILED` when deleting the provider stack. The current destroy script does not retry, and no later process revisits leaked resources in the account.

Other leak paths include:

- The teardown job being blocked by runner availability.
- Environment protection rules preventing teardown from starting.
- Runs using `DISABLE_TEARDOWN`.
- `LmiShared-local-*` stacks created from developer machines.

A recurring cleanup mechanism is needed as a final safety net to limit cost and prevent abandoned e2e resources from accumulating.

This is a follow-up to #5465 and is related to #5092 and #5518.

- Automation
- Tests

### Solution

Add a cron-based GitHub Actions workflow that runs nightly or every few hours and can also be invoked manually. It should assume the existing E2E OIDC role and require no standing AWS infrastructure of its own.

In `eu-west-1`, the workflow should:

1. List CloudFormation stacks that are older than an age threshold of approximately three hours and either:
   - Match `LmiShared-*`, including `LmiShared-local-*`; or
   - Have the tag `Service: Powertools-for-AWS-e2e-tests`, which all e2e stacks receive through `TestStack`.
2. Delete resources in dependency order:
   - Delete function stacks first so their capacity-provider associations are detached.
   - Delete the corresponding shared capacity provider stacks afterward.
3. Revisit and retry stacks in `DELETE_FAILED`.
4. If deletion still fails after two attempts, file an issue or otherwise notify maintainers with the affected stack names and failure details.

The sweeper implementation can live in `packages/testing` and reuse the toolkit-lib machinery introduced by #5465.

### Acknowledgment

### Future readers

dreamorosi:
## Summary

Adds a nightly GitHub Actions workflow that removes leaked e2e CloudFormation stacks from the test account, as a final safety net for the hourly-billed LMI capacity provider infrastructure introduced by #5465. Stacks become eligible 12 hours after creation when they carry the `Service=Powertools-for-AWS-e2e-tests` tag or an `LmiShared-*` name.

### Changes

- Add `sweepStaleStacks` to `packages/testing` (new `e2e:sweep` script): paginated discovery via `DescribeStacks` (root stacks only, 12-hour age gate, exact tag or `LmiShared-` prefix), all-or-nothing discovery, two dependency-ordered deletion passes (function stacks before shared capacity providers), concurrency capped at 5, 20-minute per-stack waits by stack id, no force deletion
- Enforce an internal 50-minute time budget so the report is always emitted before the 60-minute job cap; leftover stacks are reported unresolved
- Skip and report stacks with operations in progress; `REVIEW_IN_PROGRESS` is terminal and holds no resources, so those are deleted
- Emit a sanitized single-line JSON report (account ids and ARNs redacted) on stdout, exit nonzero when stacks remain unresolved
- Add `sweep-stale-e2e-stacks.yml`: nightly cron plus manual dispatch defaulting to dry run (discovery only, job summary, no issue); reuses the existing `e2e-tests` OIDC environment; workflow concurrency group prevents overlapping sweeps
- Add `report_e2e_sweep.js`: on failed real runs, creates or comments on a single incident issue deduplicated by exact title (no labels managed, no auto-close); issue bodies are truncated below GitHub's size limit; the same formatter renders the job summary
- Apply the e2e service tag to caller-supplied `TestStack` stacks so the tag-based selector also covers the layer publisher stack; drop the now-redundant manual tag in `sharedCapacityProviderStack`
- Add 49 mocked unit tests covering pagination, selection, age and status handling, dependency ordering, bounded concurrency, retries, time budget, dry run, and report sanitization

**Issue number:** closes #5519

---

### DIFF
--- a/.github/scripts/report_e2e_sweep.js
+++ b/.github/scripts/report_e2e_sweep.js
@@ -1,0 +1,122 @@
+const INCIDENT_LABEL = 'e2e-sweeper';
+const AUTOMATION_LABEL = 'automation';
+const INCIDENT_TITLE = 'E2E stale stack cleanup failing';
+
+const ensureLabel = async ({ github, owner, repo, name, color }) => {
+  try {
+    await github.rest.issues.getLabel({ owner, repo, name });
+  } catch (error) {
+    if (error.status !== 404) throw error;
+    await github.rest.issues.createLabel({ owner, repo, name, color });
+  }
+};
+
+const findIncident = async ({ github, owner, repo }) => {
+  const { data: issues } = await github.rest.issues.listForRepo({
+    owner,
+    repo,
+    state: 'open',
+    labels: INCIDENT_LABEL,
+    per_page: 100,
+  });
+
+  return issues.find(
+    (issue) =>
+      issue.title === INCIDENT_TITLE && issue.pull_request === undefined
+  );
+};
+
+const formatSection = (title, items, format) => {
+  if (items.length === 0) return `### ${title}\n\n_None_`;
+  return `### ${title}\n\n${items.map((item) => `- ${format(item)}`).join('\n')}`;
+};
+
+const formatReport = ({ report, runUrl }) =>
+  [
+    `Automated E2E stale stack cleanup reported a problem at ${report.timestamp}.`,
+    formatSection(
+      'Candidates',
+      report.candidates,
+      (item) =>
+        `\`${item.stackName}\` — ${item.status}; ${item.ageHours} hours old; reason: ${item.reason}`
+    ),
+    formatSection('Deleted', report.deleted, (name) => `\`${name}\``),
+    formatSection(
+      'Skipped in progress',
+      report.skippedInProgress,
+      (item) => `\`${item.stackName}\` — ${item.status}`
+    ),
+    formatSection(
+      'Unresolved',
+      report.unresolved,
+      (item) => `\`${item.stackName}\` — ${item.status}; reason: ${item.reason}`
+    ),
+    `Discovery failed: **${report.discoveryFailed ? 'yes' : 'no'}**`,
+    `[View workflow run](${runUrl})`,
+  ].join('\n\n');
+
+module.exports = async ({ github, context, core, report }) => {
+  const { owner, repo } = context.repo;
+  const runUrl = `${context.serverUrl}/${owner}/${repo}/actions/runs/${context.runId}`;
+  const isOk =
+    report.unresolved.length === 0 &&
+    report.skippedInProgress.length === 0 &&
+    report.discoveryFailed === false;
+
+  await ensureLabel({
+    github,
+    owner,
+    repo,
+    name: INCIDENT_LABEL,
+    color: 'b60205',
+  });
+  await ensureLabel({
+    github,
+    owner,
+    repo,
+    name: AUTOMATION_LABEL,
+    color: '0e8a16',
+  });
+
+  const incident = await findIncident({ github, owner, repo });
+
+  if (!isOk) {
+    const body = formatReport({ report, runUrl });
+    if (incident === undefined) {
+      const { data: issue } = await github.rest.issues.create({
+        owner,
+        repo,
+        title: INCIDENT_TITLE,
+        body,
+        labels: [INCIDENT_LABEL, AUTOMATION_LABEL],
+      });
+      core.info(`Created incident issue #${issue.number}`);
+      return;
+    }
+
+    await github.rest.issues.createComment({
+      owner,
+      repo,
+      issue_number: incident.number,
+      body,
+    });
+    core.info(`Updated incident issue #${incident.number}`);
+    return;
+  }
+
+  if (incident !== undefined) {
+    await github.rest.issues.createComment({
+      owner,
+      repo,
+      issue_number: incident.number,
+      body: `The E2E test account is clean. [View workflow run](${runUrl}).`,
+    });
+    await github.rest.issues.update({
+      owner,
+      repo,
+      issue_number: incident.number,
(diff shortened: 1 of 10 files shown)


.github/scripts/report_e2e_sweep.js
```diff
@@ -0,0 +1,100 @@
...
+  await github.rest.issues.createComment({
+    owner,
+    repo,
+    issue_number: incident.number,
+    body,
+  });
+  core.info(`Updated incident issue #${incident.number}`);
+};
+
+module.exports = reportE2eSweep;
```
svozza, line 99: How come we're not using ESM exports?

packages/testing/src/e2e/sweepStaleStacks.ts
```diff
@@ -0,0 +1,825 @@
...
+ * @param now - the instant stack ages are measured against, in epoch milliseconds
+ * @internal
+ */
+const discoverStaleStacks = async (
+  client: CloudFormationClient,
+  now: number
+): Promise<StaleStackCandidate[]> => {
+  const candidates: StaleStackCandidate[] = [];
+  let nextToken: string | undefined;
+  do {
```
svozza: Could we use `paginateDescribeStacks` and a `for..await` loop instead of handrolling this? https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-cloudformation/Function/paginateDescribeStacks/

packages/testing/src/e2e/sweepStaleStacks.ts
```diff
@@ -0,0 +1,825 @@
...
+      `Dry run: would delete ${deletable.length} stack(s): ${deletable.map(({ stackName }) => stackName).join(', ')}`
+    );
+    report.ok = true;
+
+    return report;
+  }
+
+  const deleted = new Set<string>();
+  const unresolved = new Map<string, SweepReport['unresolved'][number]>();
+  const markUnresolved = (
```
svozza: Is this function really necessary? Inlining`unresolved.set(stackName, { stackName, status, reason });` is clear enough imo.

packages/testing/tests/unit/sweepStaleStacks.test.ts
```diff
@@ -0,0 +1,985 @@
...
+    if (command instanceof DeleteStackCommand) {
+      deleteInputs.push(command.input);
+      events.push(`delete:${stackNameOf(command.input.StackName as string)}`);
+      return {};
+    }
+    throw new Error('Unexpected command');
+  });
+
+  return {
+    client: { send } as unknown as CloudFormationClient,
```
svozza: Is there a way to avoid this cast?

.github/scripts/report_e2e_sweep.js
dreamorosi: The scripts in `.github/scripts` are CommonJS by construction: the directory has its own `package.json` with `"type": "commonjs"`, and `actions/github-script` loads these files with a CJS `require()` inside the runner, so an ESM export would not be loadable there. Same pattern as `post_release.js`.

packages/testing/src/e2e/sweepStaleStacks.ts
dreamorosi: Good call — switched to `paginateDescribeStacks` with `for await...of` in 28036d33. Error propagation is unchanged, so discovery still aborts all-or-nothing.

packages/testing/src/e2e/sweepStaleStacks.ts
dreamorosi: Agreed, inlined at all four call sites in 28036d33.

packages/testing/tests/unit/sweepStaleStacks.test.ts
dreamorosi: Removed in 28036d33 by switching the harness to `aws-sdk-client-mock` (already backing our `aws-sdk-client-mock-vitest` devDep), so tests pass a real `CloudFormationClient` instance. This also turned out to be required for the paginator change: `@smithy/core` rejects clients that are not real instances.

.github/scripts/report_e2e_sweep.js
svozza: Ah I see. We should probably have a separate issue to make them ESM, espeically because especially for scripts we may need top level await at some point.

svozza (approved):

.github/scripts/report_e2e_sweep.js
dreamorosi: Yes, last time I checked almost 2y ago the GH action didn't play nice with it, but if it's doable now let's do it

PR #5630, opened 2026-09-04
fix(data-masking): honor throwOnMissingField consistently and resolve paths before masking

### LINKED ISSUE #5626: Bug: `maskingRules` in `erase` does not honour `throwOnMissingField` when a rule matches no field
### Expected Behavior

`DataMasking` is constructed with `throwOnMissingField`, defaulting to `true`. When a path expression passed to `erase` matches nothing in the data, the call throws `DataMaskingFieldNotFoundError`, or logs a warning when the option is `false`. This is the behaviour for paths listed in `fields`, and a path listed as a key in `maskingRules` should behave the same way.

### Current Behavior

A `maskingRules` key that matches no field is silently skipped. No error is thrown and no warning is logged regardless of `throwOnMissingField`, and `erase` returns the data with that rule not applied. The `fields` path in the same method does throw or warn.

### Code snippet

```typescript
import { DataMasking } from '@aws-lambda-powertools/data-masking';

const masker = new DataMasking(); // throwOnMissingField defaults to true
const data = { ssn: '123-45-6789' };

masker.erase(data, { fields: ['snn'] });
// throws DataMaskingFieldNotFoundError: Field not found: 'snn'

masker.erase(data, { maskingRules: { snn: { dynamicMask: true } } });
// returns { ssn: '123-45-6789' }, no error, no warning
```

### Steps to Reproduce

1. Construct `DataMasking` with default options
2. Call `erase` with a `maskingRules` key that does not match any path in the data
3. Observe that the call returns normally and nothing is logged

### Possible Solution

In `#applyMaskingRules`, when `#resolveFieldPaths` returns no paths for a rule key, apply the same throw-or-warn check that `#eraseFields` already performs. Extracting that check into a shared private method keeps the two paths in step. Add tests alongside the existing `throwOnMissingField` cases in `erase.test.ts`. The docs page does not currently mention `throwOnMissingField`, so a short note there would help too.

latest

24.x

### Packaging format used

npm

svozza:
## Summary

A `maskingRules` key that matched nothing was silently skipped while the same path in `fields` threw, and `encrypt` and `decrypt` ignored `throwOnMissingField` entirely. Fixing that exposed ordering problems in how paths were applied, so path resolution and target selection are now done up front for all three operations. Scope grew beyond the issue as filed; each addition came from reviewing the initial fix and is listed on the issue.

### Changes

- `erase`, `encrypt` and `decrypt` all resolve every path expression through one private method that throws or warns per `throwOnMissingField`; `encrypt` and `decrypt` resolve before calling the provider so a missing field cannot leave provider calls in flight
- Targets are collected into a first-wins map keyed by path, and any target beneath another target is dropped, so a rule on a parent sees the original children and overlapping expressions apply a rule once
- A wildcard over an empty collection is a match with no paths rather than a missing field; only an absent non-wildcard segment counts as missing
- Path resolution uses `Object.hasOwn`, so present keys holding `undefined` resolve; reserved keys, empty expressions and non-index keys on arrays never resolve, and the guards in `getAtPath` and `setAtPath` are removed as unreachable
- `encrypt` and `decrypt` leave `undefined` values untouched, matching how a masking rule passes them through
- `applyRuleToLeaves` reuses `resolveWildcardEntries` instead of duplicating it
- Docs gain a "Missing fields" section and the `throwOnMissingField` JSDoc names all three operations
- Package `vitest.config.ts` loads the shared `setupEnv.ts`, and the tests use its console mock instead of local spies

**Issue number:** closes #5626

---

### DIFF
--- a/packages/data-masking/src/DataMasking.ts
+++ b/packages/data-masking/src/DataMasking.ts
@@ -121,10 +121,7 @@ export class DataMasking {
   ): Set<string> {
     const touched = new Set<string>();
     for (const [field, rule] of Object.entries(rules)) {
-      for (const path of this.#resolveFieldPaths(
-        copy as Record<string, unknown>,
-        field
-      )) {
+      for (const path of this.#resolvePathsOrReportMissing(copy, field)) {
         setAtPath(copy, path, maskLeaf(getAtPath(copy, path), rule));
         touched.add(pathKey(path));
       }
@@ -140,19 +137,7 @@ export class DataMasking {
     skip?: Set<string>
   ): void {
     for (const field of fields) {
-      const paths = this.#resolveFieldPaths(
-        copy as Record<string, unknown>,
-        field
-      );
-      if (paths.length === 0) {
-        if (this.#throwOnMissingField) {
-          throw new DataMaskingFieldNotFoundError(
-            `Field not found: '${field}'`
-          );
-        }
-        console.warn(`Field not found: '${field}'`);
-      }
-      for (const path of paths) {
+      for (const path of this.#resolvePathsOrReportMissing(copy, field)) {
         if (skip?.has(pathKey(path))) continue;
         // A plain field erase replaces any value with the mask; a rule stringifies
         // the leaf first (null/undefined pass through), matching the Python utility.
@@ -274,6 +259,25 @@ export class DataMasking {
     }
 
     await Promise.all(operations);
+  }
+
+  /**
+   * Resolve a path expression, throwing or warning per `throwOnMissingField`
+   * when it matches nothing in the data.
+   */
+  #resolvePathsOrReportMissing<T>(copy: T, field: string): string[][] {
+    const paths = this.#resolveFieldPaths(
+      copy as Record<string, unknown>,
+      field
+    );
+    if (paths.length === 0) {
+      if (this.#throwOnMissingField) {
+        throw new DataMaskingFieldNotFoundError(`Field not found: '${field}'`);
+      }
+      console.warn(`Field not found: '${field}'`);
+    }
+
+    return paths;
   }
 
   #resolveFieldPaths(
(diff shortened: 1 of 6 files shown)


packages/data-masking/src/DataMasking.ts
```diff
@@ -258,62 +222,135 @@ export class DataMasking {
...
+const withoutSubsumed = (targets: Map<string, MaskTarget>): MaskTarget[] => {
+  const kept: MaskTarget[] = [];
+  for (const target of targets.values()) {
+    if (!hasTargetedAncestor(targets, target.path)) kept.push(target);
+  }
+
+  return kept;
+};
+
+const hasTargetedAncestor = (
```
dreamorosi, line 373: Could you point the agent against the new coding standards doc, I think there's a bit to improve here in terms of comments

dreamorosi:
> Your PR title does not follow the [Conventional Commits](https://www.conventionalcommits.org/) rules for this repository.
> 
> **Problem:** The scope `data-masking` is not allowed for this repository.
> 
> Expected format: `<type>[optional scope][optional !]: <description>`
> 
> **Allowed types:** `feat`, `fix`, `docs`, `style`, `refactor`, `perf`, `test`, `build`, `ci`, `chore`, `revert`, `improv`
> 
> **Allowed scopes:** `tracer`, `logger`, `metrics`, `parameters`, `idempotency`, `commons`, `parser`, `jmespath`, `event-handler`, `validation`, `batch`, `layers`, `ci`, `docs`, `tests`, `internal`, `maintenance`, `deps`, `deps-dev`, `roadmap`, `kafka`, `signer`
> 
> Please update your PR title and the check will re-run automatically.

I'll fix this

packages/data-masking/src/DataMasking.ts
svozza: Yeah, sorry, ended up having to get really deep into the implementation here and let this slip.

dreamorosi (approved):

PR #5663, opened 2026-09-07
chore(tests): add Metrics e2e suite for Lambda Managed Instances

### LINKED ISSUE #5612: Maintenance: Add Metrics E2E tests for Lambda Managed Instances
### Summary

Add an end-to-end test validating Metrics' `InvokeStore`-backed metric isolation on Lambda Managed Instances, where concurrent invocations are multiplexed into shared execution environments.

Part of #5092. This issue tracks the Metrics portion; Logger is tracked in #5518.

### Why is this needed?

Metrics' `InvokeStore` concurrency code paths are currently exercised only by unit tests with mocked Lambda Managed Instances environments. Nothing proves that metric data and metadata remain isolated between concurrent invocations on the real platform.

- Tests
- Metrics

### Solution

Add a Metrics LMI end-to-end suite that saturates a deliberately small fleet, forcing concurrent invocations to share execution environments. Each invocation should emit uniquely identifiable metrics and metadata, then verify that its state does not leak into another invocation.

Follow the Logger test pattern from #5518: tee `process.stdout.write` inside the handler to capture the EMF payloads written through the production output path, and return the captured output in the Lambda response. Filter the returned payloads using an invocation-specific key and assert that each contains only its own metrics and metadata. This provides deterministic assertions without tailing CloudWatch Logs or waiting for log ingestion.

Reuse the shared capacity provider infrastructure and workflow support introduced by #5465.

### Acknowledgment

### Future readers

svozza:
## Summary

Adds an end-to-end suite proving that Metrics keeps metric values and metadata isolated between invocations multiplexed into the same execution environment on Lambda Managed Instances, following the Logger suite from #5465. The stdout tee and peer barrier the two handlers would otherwise duplicate move into testing-utils so a third LMI suite does not copy them again.

### Changes

- Add `packages/metrics/tests/e2e/lmi.test.ts` and its handler. 30 concurrent invocations are held open on a module-scoped barrier so the scheduler multiplexes at least one pair. Each invocation stores an `invocationKey` metadata entry and an `LmiIsolation` metric with a value unique to it before blocking, publishes after, and returns the EMF payloads carrying its own key. The test asserts LMI initialization type and max concurrency, at least one genuine overlap, and for every invocation exactly one payload holding only its own metric definition and its own value against the expected namespace. Dimensions stay at the default `service` so each run adds a single metric to the throwaway namespace.
- Add `captureJsonStdout()` and `createPeerBarrier()` to testing-utils under a new `lmi/handler` subpath, bundled into the function by esbuild, with unit tests. The Logger LMI handler is rewritten on top of them; its assertions and response shape are unchanged.
- Add the `test:e2e:lmi` script to metrics and exclude `lmi.*` from its regular e2e scripts, so `list_workspaces.js` routes the suite to the LMI job with no workflow change. `@aws-sdk/client-lambda` becomes a metrics dev dependency because the testing-utils invoke helper requires Tail logs, which LMI rejects.
- Notes for reviewers: verified on [run 34137445915](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/34137445915). All eight LMI cells passed, four per package across Node 22 and 24 on arm64 and x86_64, each with 12 to 14 of 30 invocations observing a peer, so the isolation assertion ran against real overlap in every cell. The strict EMF assertion was checked against the library's serialized output before the run.

**Issue number:** closes #5612

---


dreamorosi (approved):
Thanks for adding these

PR #2938, opened 2024-08-16
test(batch): migrate unit tests to Vitest

dreamorosi:
### Changes

> Please provide a summary of what's being changed

This PR migrates the unit tests for the Batch Processing feature to use Vitest.

As part of this work, I have also:
- Removed any Jest-specific configurations
- Extracted the tests in CI as part of a new and improved matrix strategy test
- Made coverage reporting opt-in during local tests for fast iteration
- Changed the `pre-push` git hook to run tests only on files changed since branching out from `main`

> Please add the issue number below, if no issue is present the PR might get blocked and not be reviewed

**Issue number:** closes #2937 

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Disclaimer**: We value your time and bandwidth. As such, any pull requests created on non-triaged issues might not be successful.

PR #5665, opened 2026-09-07
chore(ci): align cached-node-modules pins and drop a stale header comment

### LINKED ISSUE #5642: Maintenance: align `cached-node-modules` pins and remove a stale migration comment in CI
### Summary

The `aws-powertools/actions/.github/actions/cached-node-modules` composite action is pinned at two different commits across the workflows: thirteen uses point at a January 2025 commit (`29979bc5…`) whose input is `nodeVersion`, and six point at `3b5b8e2e…` (`v1.5.0`, August 2025) whose input is `node-version`. Each caller passes the spelling its pin expects, so nothing is broken, but the older pin carries no version comment and Dependabot therefore never proposes bumping it.

Separately, `reusable-run-linting-check-and-unit-tests.yml` still carries the note "Since #2938, we are in the process of improving our test suite, so we are gradually extracting the tests for each package … to their own job". That extraction finished long ago; every package already runs in its own matrix cell.

### Why is this needed?

Two pins of one action means two behaviours to reason about and a bump that has to be made twice. The missing version comment silently exempts thirteen call sites from Dependabot. The stale comment misleads anyone reading the workflow about what is still in flight.

Automation

### Solution

- Pin every use to the `v1.5.0` commit with a `# v1.5.0` comment, and rename the `nodeVersion` inputs to `node-version`. Affected: the reusable unit-test workflow, `quality_check.yml`, `publish_layer.yml`, `make-version.yml`, the layer and SSM workflows, and any other caller.
- Delete the "Since #2938" paragraph from the reusable workflow header.

### Acknowledgment

### Future readers

svozza:
## Summary

Every workflow now pins the `aws-powertools/actions` `cached-node-modules` composite action to the same commit, with a version comment so Dependabot tracks all of them. Previously thirteen call sites used an unannotated January 2025 commit that Dependabot ignored, and six used `v1.5.0`. The reusable unit-test workflow also loses a header paragraph describing a test-suite migration that finished long ago.

### Changes

- Pin all uses of `cached-node-modules` to the `v1.6.0` commit (`7898dbd7`) with a `# v1.6.0` comment. This release includes aws-powertools/actions#265, which fixes the action so `node-version` is actually part of the cache key again; in `v1.5.0` through `v1.5.2` the input was silently ignored.
- Rename the remaining `nodeVersion` input in `reusable-run-linting-check-and-unit-tests.yml` to `node-version`, matching the action's current interface and every other caller.
- Remove the "Since #2938, we are in the process of improving our test suite" paragraph from the reusable workflow header.

**Issue number:** closes #5642

---


.github/workflows/bootstrap_region.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         with:
           node-version: 24
       - name: Setup dependencies
-        uses: aws-powertools/actions/.github/actions/cached-node-modules@29979bc5339bf54f76a11ac36ff67701986bb0f0
+        uses: aws-powertools/actions/.github/actions/cached-node-modules@4bf64a4072489399fbf39b78f558eeeed6767189 # v1.5.2
```
dreamorosi:
```suggestion
        uses: aws-powertools/actions/.github/actions/cached-node-modules@7898dbd71be8083748aa52f07ce5201bb36078ac # v1.6.0
```

dreamorosi: Made a release with the cache key fix you posted, above the hash and version no

dreamorosi (approved):

PR #5676, opened 2026-09-09
fix(idempotency): delete only acquired records in the Middy onError hook

### LINKED ISSUE #5675: Bug: `makeHandlerIdempotent` deletes idempotency records it never acquired
### Expected Behavior

When the Middy middleware rejects a request in its `before` hook, because another invocation with the same idempotency key is still in progress (`IdempotencyAlreadyInProgressError`) or because the payload fails validation (`IdempotencyValidationError`), the existing record in the persistence store should be left untouched. Only the invocation that acquired the in-progress record should be able to delete it, and only when its own handler fails.

This is how the `makeIdempotent` wrapper and the `@idempotent` decorator already behave: they only delete the record when the wrapped function itself throws, after the record was acquired.

### Current Behavior

The middleware's `onError` hook runs whenever any earlier step throws, including its own `before` hook. It always calls `handleMiddyOnError()`, which deletes the record for the current payload unconditionally. So a rejected duplicate deletes the lock held by the still-running original invocation, and a rejected payload mismatch deletes the completed record of the original request.

Reproduced on real Lambda + DynamoDB (eu-west-1, Node.js 22.x) with an 8-second handler and three identical invocations:

| Step | Result | Table state after |
|---|---|---|
| A invoked | acquires lock, handler starts | one `INPROGRESS` record |
| B invoked at +2s | rejected with `IdempotencyAlreadyInProgressError` (correct) | **empty** |
| C invoked at +3s | acquires a new lock, handler runs **again** | `INPROGRESS` |
| A and C complete | both return `charged` | one `COMPLETED` record |

The handler executed twice. The identical timeline against `makeIdempotent` executed once and rejected both B and C.

Second variant (payload validation): after `{ orderId: "order-1", amount: 10 }` completes, a request with `amount: 20` is correctly rejected with `IdempotencyValidationError`, but the completed record is deleted. A subsequent retry of the original `amount: 10` request executes the handler a second time.

### Code snippet

```typescript
import { makeHandlerIdempotent } from '@aws-lambda-powertools/idempotency/middleware';

  tableName: process.env.IDEMPOTENCY_TABLE_NAME,
});

export const handler = middy(async (event: { orderId: string }) => {
  console.log(`CHARGE started ${event.orderId}`);
  await new Promise((r) => setTimeout(r, 8000)); // slow downstream call
  console.log(`CHARGE finished ${event.orderId}`);
  return `charged ${event.orderId}`;
}).use(makeHandlerIdempotent({ persistenceStore }));
```

### Steps to Reproduce

1. Deploy the function above with a DynamoDB idempotency table (default `id` partition key).
2. Invoke it with `{ "orderId": "order-1" }` and do not wait for the response (A).
3. ~2s later, invoke it again with the same payload (B). It fails with `IdempotencyAlreadyInProgressError`, as expected.
4. Scan the table: the `INPROGRESS` record written by A is gone.
5. ~1s later, invoke it a third time with the same payload (C).
6. Observe `CHARGE started order-1` logged twice, once by A and once by C.

For the second variant, configure `eventKeyJmesPath: 'orderId'` and `payloadValidationJmesPath: 'amount'`, complete a request with `amount: 10`, send one with `amount: 20` (rejected), then retry `amount: 10`. The handler runs again and the table no longer holds the original completed record between steps.

### Possible Solution

Track on the invocation's `IdempotencyHandler` whether this request actually acquired the record (i.e. `before` reached the point where `saveInProgress` succeeded and the handler was about to run). `handleMiddyOnError()` should only delete when that flag is set and the record has not yet been completed. Errors thrown by `before` itself (`IdempotencyAlreadyInProgressError`, `IdempotencyValidationError`, `IdempotencyInconsistentStateError`, persistence errors) must not trigger a delete.

Optionally, `deleteRecord` in the persistence layers could carry an ownership condition so a record replaced after acquisition is not removed by the original owner either.

2.35.0

22.x

### Packaging format used

npm

### Execution logs

```shell
[t+ 58401ms] --- handlerMiddy: request A
[t+ 60470ms] before B -> [{"id":"...-handlerMiddyFn#UXEpus/fKOYTGkfk4qBbpw==","status":"INPROGRESS","in_progress_expiration":1788953147671,"expiration":1788956718}]
[t+ 61157ms] B done: charged=false error=ERROR Invoke Error {"errorType":"IdempotencyAlreadyInProgressError",...}
[t+ 61166ms] after B  -> []
[t+ 67264ms] A done: charged=true
[t+ 70204ms] C done: charged=true
[t+ 70210ms] end      -> [{"id":"...-handlerMiddyFn#UXEpus/fKOYTGkfk4qBbpw==","status":"COMPLETED","data":"charged order-handlerMiddy",...}]
[t+ 70210ms] charge executed 2x

[t+ 70215ms] --- handlerWrapper (makeIdempotent, same timeline): request A
[t+ 72923ms] B done: charged=false error=ERROR Invoke Error {"errorType":"IdempotencyAlreadyInProgressError",...}
[t+ 72931ms] after B  -> [{"id":"...-handlerWrapperFn#cd0gva/MNrlfgOYd4C7oAA==","status":"INPROGRESS",...}]
[t+ 73983ms] C done: charged=false error=ERROR Invoke Error {"errorType":"IdempotencyAlreadyInProgressError",...}
[t+ 78945ms] A done: charged=true
[t+ 78955ms] charge executed 1x
```

svozza:
## Summary

The `makeHandlerIdempotent` middleware's `onError` hook runs whenever any earlier step throws, including its own `before` hook. It deleted the idempotency record unconditionally, so a rejected duplicate (`IdempotencyAlreadyInProgressError`) removed the lock of the still-running original invocation, and a rejected payload mismatch (`IdempotencyValidationError`) removed the original request's completed record. Either way a subsequent retry executed the handler again. Reproduced on Lambda + DynamoDB; the `makeIdempotent` wrapper is unaffected because its `try/catch` only wraps the function call.

The middleware now tracks whether the current request acquired an unfinished in-progress record and deletes only in that case.

### Changes

- `before` marks the request as owning the record when `handleMiddyBefore` did not replay a stored response, detected via the replay cleanup callback. Both per-request flags are reset at the start of `before`, since Middy reuses a user-supplied `internal` object across warm invocations; this also closes the same stale-flag leak in the existing `.skip` flag.
- `after` releases ownership before saving the result, so a `saveSuccess` failure no longer deletes the record, in line with the wrapper.
- `onError` returns early unless the request owns the record.
- Docs: "Handling exceptions" states when the middleware keeps the record and when it still deletes it.
- Tests: six new cases in `makeIdempotent.test.ts` covering the concurrent duplicate, payload mismatch, undefined stored response, later `after` failure, shared `internal` storage, and a later `before` failure that must still delete; the `saveSuccess` failure case now asserts no delete.

### Design note

Ownership is tracked in the middleware via `request.internal`, next to the existing `.skip` flag, rather than as a field on `IdempotencyHandler`. A core-class flag would be a smaller diff (set after `saveInProgress` resolves, cleared on `saveSuccess`, checked in `handleMiddyOnError`), but the problem is specific to Middy's lifecycle: `onError` fires for failures in `before`, which the wrapper and decorator paths never encounter. Keeping the state in the middleware leaves `IdempotencyHandler` untouched and keeps all per-request Middy state in one place.

**Issue number:** closes #5675

---

### DIFF
--- a/packages/idempotency/src/middleware/makeHandlerIdempotent.ts
+++ b/packages/idempotency/src/middleware/makeHandlerIdempotent.ts
@@ -60,6 +60,30 @@ const setIdempotencySkipFlag = (request: MiddyLikeRequest): void => {
 
 /**
  * @internal
+ * Records whether this request acquired an in-progress record it has not yet completed.
+ *
+ * @param request - The Middy request object
+ * @param owned - Whether the request owns an unfinished in-progress record
+ */
+const setIdempotencyRecordOwnership = (
+  request: MiddyLikeRequest,
+  owned: boolean
+): void => {
+  request.internal[`${IDEMPOTENCY_KEY}.ownsRecord`] = owned;
+};
+
+/**
+ * @internal
+ * Checks whether this request owns an unfinished in-progress record.
+ *
+ * @param request - The Middy request object
+ */
+const ownsIdempotencyRecord = (request: MiddyLikeRequest): boolean => {
+  return request.internal[`${IDEMPOTENCY_KEY}.ownsRecord`] === true;
+};
+
+/**
+ * @internal
  * Utility function to get the idempotency key from the request internal storage
  * and determine if the request should skip the idempotency middleware
  *
@@ -109,10 +133,12 @@ const makeHandlerIdempotent = (
    *
    * If idempotency is enabled and the idempotency key is present in the payload,
    * we then run the idempotency operations. These are handled in {@link IdempotencyHandler.handleMiddyBefore}.
+   * A stored response is returned early; otherwise this request now owns the
+   * in-progress record and `onError` is allowed to delete it.
    *
    * @param request - The Middy request object
    */
-  const before = (request: MiddyLikeRequest): unknown => {
+  const before = async (request: MiddyLikeRequest): Promise<unknown> => {
     const idempotencyConfig = options.config ?? new IdempotencyConfig({});
     const persistenceStore = options.persistenceStore;
     const keyPrefix = options.keyPrefix;
@@ -132,6 +158,9 @@ const makeHandlerIdempotent = (
       functionPayloadToBeHashed: undefined,
     });
     setIdempotencyHandlerInRequestInternal(request, idempotencyHandler);
+    // `request.internal` can be shared across invocations, so reset per-request state
+    request.internal[`${IDEMPOTENCY_KEY}.skip`] = false;
+    setIdempotencyRecordOwnership(request, false);
 
     // set the payload to be hashed
     idempotencyHandler.setFunctionPayloadToBeHashed(request.event as JSONValue);
@@ -145,7 +174,20 @@ const makeHandlerIdempotent = (
 
     idempotencyConfig.registerLambdaContext(request.context);
 
-    return idempotencyHandler.handleMiddyBefore(request, cleanupMiddlewares);
+    // the cleanup callback only runs when a stored response is replayed
+    let replayed = false;
+    const result = await idempotencyHandler.handleMiddyBefore(
+      request,
+      async (request) => {
+        replayed = true;
+        await cleanupMiddlewares(request);
+      }
+    );
+    if (!replayed) {
+      setIdempotencyRecordOwnership(request, true);
+    }
+
+    return result;
   };
 
   /**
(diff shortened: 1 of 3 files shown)


packages/idempotency/src/middleware/makeHandlerIdempotent.ts
```diff
@@ -58,6 +58,30 @@ const setIdempotencySkipFlag = (request: MiddyLikeRequest): void => {
...
+ * Records whether this request acquired an in-progress record it has not yet completed.
+ *
+ * @param request - The Middy request object
+ * @param owned - Whether the request owns an unfinished in-progress record
+ */
+const setIdempotencyRecordOwnership = (
+  request: MiddyLikeRequest,
+  owned: boolean
+): void => {
+  request.internal[`${IDEMPOTENCY_KEY}.ownsRecord`] = owned;
```
dreamorosi, line 72: Neat

dreamorosi (approved):

PR #5683, opened 2026-09-09
fix(event-handler): ignore body on GET/HEAD requests when converting to Web Request

### LINKED ISSUE #5673: Bug: GET or HEAD request with a body crashes the HTTP event handler invocation
### Expected Behavior

A `GET` or `HEAD` request whose event carries a non-null `body` should still be routed. The body should be ignored when building the Web `Request`, since the Fetch spec forbids a body on those methods, and the handler should run normally. The ALB converter already behaves this way.

### Current Behavior

For API Gateway REST (v1) and HTTP API (v2) events, `proxyEventV1ToWebRequest` and `proxyEventV2ToWebRequest` pass `event.body` to `new Request()` unconditionally. Undici throws:

```
TypeError: Request with GET/HEAD method cannot have body.
```

This is not an `InvalidHttpMethodError`, so the `catch` in `Router.#resolve` rethrows it and `app.resolve()` rejects. The invocation fails and API Gateway returns a 502 to the client instead of a 4xx or the handler's response. No route, middleware, or error handler runs, so the failure is also invisible to the metrics and tracer middleware.

API Gateway forwards request bodies on `GET` for both REST and HTTP APIs, and clients that send them are not unusual: Elasticsearch-style search APIs, some HTTP client libraries, and anything that treats `GET` with a JSON body as valid. A client can therefore take down any route with a single request.

`albEventToWebRequest` already guards against this because ALB represents `GET` and `HEAD` bodies as empty strings, which would trigger the same error; the v1 and v2 converters never received the same guard.

### Code snippet

```typescript

const app = new Router();

app.get('/search', () => ({ ok: true }));

  app.resolve(event, context);
```

Invoked with a v1 event for `GET /search` where `body` is `'{"q":"x"}'`, `handler` rejects with the `TypeError` above.

Failing unit test (runs for both v1 and v2 events using the existing `createTestEvent` and `createTestEventV2` helpers):

```typescript
it.each([
  { version: 'V1', createEvent: createTestEvent },
  { version: 'V2', createEvent: createTestEventV2 },
])(
  'does not crash when a $version GET request carries a body',
  async ({ createEvent }) => {
    // Prepare
    const app = new Router();
    app.get('/search', () => ({ ok: true }));
    const event = createEvent('/search', 'GET');
    event.body = '{"q":"x"}';

    // Act

    // Assess
    expect(result.statusCode).toBe(HttpStatusCodes.OK);
  }
);
```

### Steps to Reproduce

1. Create a `Router` and register any `GET` route.
2. Invoke `app.resolve()` with an API Gateway v1 or v2 event for that route whose `body` is a non-null string.
3. Observe the promise rejects with `TypeError: Request with GET/HEAD method cannot have body.`
4. Behind API Gateway, observe a 502 response for the same request.

### Possible Solution

In `packages/event-handler/src/http/converters.ts`, apply the same rule `albEventToWebRequest` uses (`converters.ts:186-190`) to `proxyEventV1ToWebRequest` (`converters.ts:125`) and `proxyEventV2ToWebRequest` (`converters.ts:161`): when the method is `GET` or `HEAD`, pass `null` as the body. A small shared helper would keep the three converters in step.

Dropping the body rather than returning 400 matches the ALB converter and the behaviour of other Lambda adapters, and avoids turning a client quirk into an error for handlers that never read the body. RFC 9110 permits rejecting such requests but does not require it, and the smuggling risk that motivates rejection does not apply once API Gateway has already parsed the message. The body remains available on `reqCtx.event.body` for handlers that want it.

2.35.0

24.x

### Packaging format used

npm

### Execution logs

```
TypeError: Request with GET/HEAD method cannot have body.
    at proxyEventV1ToWebRequest (src/http/converters.ts:125:10)
    at proxyEventToWebRequest (src/http/converters.ts:223:10)
    at Router.#resolve (src/http/Router.ts:312:13)
    at Router.resolve (src/http/Router.ts:454:31)
```

sabaighani:
### Changes

API Gateway v1/v2 events can include a body on GET or HEAD requests, even
though that's unusual. The V1 and V2 event converters
(`proxyEventV1ToWebRequest` and `proxyEventV2ToWebRequest`) passed the body
straight through to the Web `Request` constructor. Since a Web `Request`
with method GET/HEAD cannot carry a body, this throws an uncaught
`TypeError`, crashing the handler invocation and returning a 502 to the
client.

The ALB converter already guarded against this case. This PR moves that
guard into the shared `createBody` helper (used by all three converters) so
GET/HEAD requests always get a `null` body, matching the ALB converter's
existing behavior.

Added regression tests covering GET and HEAD requests with a body for both
the API Gateway v1 and v2 converters.

**Issue number:** closes #5673

---

### DIFF
--- a/packages/event-handler/src/http/converters.ts
+++ b/packages/event-handler/src/http/converters.ts
@@ -13,6 +13,7 @@ import type {
   ExtendedAPIGatewayProxyResult,
   ExtendedAPIGatewayProxyResultBody,
   HandlerResponse,
+  HttpMethod,
   HttpStatusCode,
   ResponseType,
   ResponseTypeMap,
@@ -37,6 +38,19 @@ import {
 } from './utils.js';
 
 /**
+ * Reads and normalises the HTTP method of an event, throwing when it is not one we route.
+ *
+ * @param rawMethod - The method as it appears on the event
+ */
+const toHttpMethod = (rawMethod: string): HttpMethod => {
+  const method = rawMethod.toUpperCase();
+  if (!isHttpMethod(method)) {
+    throw new InvalidHttpMethodError(method);
+  }
+  return method;
+};
+
+/**
  * Creates a request body from API Gateway event body, handling base64 decoding if needed.
  *
  * GET and HEAD requests are not allowed to carry a body when constructing a
@@ -49,7 +63,7 @@ import {
 const createBody = (
   body: string | null,
   isBase64Encoded: boolean,
-  httpMethod: string
+  httpMethod: HttpMethod
 ) => {
   if (httpMethod === HttpVerbs.GET || httpMethod === HttpVerbs.HEAD) {
     return null;
@@ -121,8 +135,9 @@ const populateV1QueryParams = (
  * @returns A Web API Request object
  */
 const proxyEventV1ToWebRequest = (event: APIGatewayProxyEvent): Request => {
-  const { httpMethod, path } = event;
+  const { path } = event;
   const { domainName } = event.requestContext;
+  const method = toHttpMethod(event.httpMethod);
 
   const headers = new Headers();
   populateV1Headers(headers, event);
@@ -134,9 +149,9 @@ const proxyEventV1ToWebRequest = (event: APIGatewayProxyEvent): Request => {
   populateV1QueryParams(url, event);
 
   return new Request(url.toString(), {
-    method: httpMethod,
+    method,
     headers,
-    body: createBody(event.body, event.isBase64Encoded, httpMethod),
+    body: createBody(event.body, event.isBase64Encoded, method),
   });
 };
 
@@ -148,10 +163,8 @@ const proxyEventV1ToWebRequest = (event: APIGatewayProxyEvent): Request => {
  */
 const proxyEventV2ToWebRequest = (event: APIGatewayProxyEventV2): Request => {
   const { rawPath, rawQueryString } = event;
-  const {
-    http: { method },
-    domainName,
-  } = event.requestContext;
+  const { domainName } = event.requestContext;
+  const method = toHttpMethod(event.requestContext.http.method);
 
   const headers = new Headers();
   for (const [name, value] of Object.entries(event.headers)) {
@@ -183,7 +196,8 @@ const proxyEventV2ToWebRequest = (event: APIGatewayProxyEventV2): Request => {
  * @returns A Web API Request object
  */
 const albEventToWebRequest = (event: ALBEvent): Request => {
-  const { httpMethod, path } = event;
+  const { path } = event;
+  const method = toHttpMethod(event.httpMethod);
 
   const headers = new Headers();
   populateV1Headers(headers, event);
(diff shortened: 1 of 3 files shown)


svozza (commented):
Thanks for picking this up, the shared guard in `createBody` is the right place for it.

One change I'd like before merging. The dispatcher validates the method on an uppercased copy and then throws it away, and each converter re-reads the raw method from the event:

```ts
const method = event.httpMethod.toUpperCase();
if (!isHttpMethod(method)) throw new InvalidHttpMethodError(method);
return proxyEventV1ToWebRequest(event);   // reads event.httpMethod again, raw
```

That means a lowercase `get` passes validation but misses the new guard, and `Request` throws the same `TypeError` this PR is fixing. `patch` is worse: the Fetch spec only normalises six methods and `PATCH` isn't one of them, so it reaches the router as `patch` and the route lookup misses. Neither happens with real API Gateway or ALB events today, but since you're already changing `createBody`'s signature this is the moment to close it.

Suggestion: normalise at the point the method is read, so the value that's validated is the value that's used.

```ts
/**
 * Reads and normalises the HTTP method of an event, throwing when it is not one we route.
 *
 * @param rawMethod - The method as it appears on the event
 */
const toHttpMethod = (rawMethod: string): HttpMethod => {
  const method = rawMethod.toUpperCase();
  if (!isHttpMethod(method)) {
    throw new InvalidHttpMethodError(method);
  }
  return method;
};
```

Each converter derives its method once from its own event and uses it for both the `Request` constructor and `createBody`:

```ts
const proxyEventV1ToWebRequest = (event: APIGatewayProxyEvent): Request => {
  const method = toHttpMethod(event.httpMethod);
  // ...
  return new Request(url.toString(), {
    method,
    headers,
    body: createBody(event.body, event.isBase64Encoded, method),
  });
};
```

Same for V2 (`event.requestContext.http.method`) and ALB. `createBody` can then take `method: HttpMethod` instead of `string`, and the dispatcher loses its two validation blocks and just dispatches on event type. The `InvalidHttpMethodError` is still thrown before any `Request` is built, so the existing 405 handling in `Router` is unaffected and the `CONNECT`/`TRACE` tests should pass as they are.

There's still some repetition left across the three converters after this, but that's beyond the scope of this PR; we're planning a follow-up to tidy that up separately.

Two small test asks:

- A converter-level case with a lowercase method, asserting `get` yields a `GET` request with a null body and `patch` yields `PATCH`, since those are the two behaviours that change.
- One `it.each` over V1 and V2 in `tests/unit/http/Router/basic-routing.test.ts` that resolves a GET with a body through `app.resolve()` and asserts a 200, so the fix is locked in end to end rather than only at the converter.

sabaighani:
Good catch, thanks! Moved the normalization into a shared `toHttpMethod` helper called once per converter (V1, V2, ALB), so the same normalized value now feeds both the `Request` constructor and `createBody`. The dispatcher lost its duplicate validation blocks, and `createBody` now takes `HttpMethod` instead of `string`.

Added both test cases:
- Converter-level tests for a lowercase `get` (body correctly ignored) and lowercase `patch` (normalized to `PATCH`) on both the V1 and V2 converters.
- An `it.each` over V1 and V2 in `basic-routing.test.ts` that resolves a GET request carrying a body through `app.resolve()` end to end and asserts a 200.

svozza (approved):

PR #5695, opened 2026-09-11
chore(tests): centralise TypeScript test configuration

### LINKED ISSUE #5694: Maintenance: Centralise test TypeScript configuration
### Summary

Introduce a shared test-only TypeScript configuration and use it across package test projects, reducing duplicated settings and separating test type-checking from production build configuration.

### Why is this needed?

Package test configs currently extend production configs, inheriting settings such as `composite: true`, despite using `noEmit` and running tests through Vitest.

Each package repeats test-specific settings and file inclusion rules. Shared test declarations are also included inconsistently across utilities. This makes maintaining and extending the shared test environment unnecessarily repetitive.

Tests, Automation

### Solution

- Introduce a root `tsconfig.test.json` that extends the root `tsconfig.json` and contains only intentional test-specific overrides. Common compiler options—such as `strict`, `target`, `module`, and `moduleResolution`—must remain defined in the root config so changes automatically propagate to test projects.
- Have package test configs extend the shared test config and retain only necessary package-specific overrides.
- Centralise the TypeScript inclusion of `packages/testing/src/setupEnv.ts` and its custom matcher declarations in the root `tsconfig.test.json`. Remove repeated `../../testing/src/setupEnv.ts` entries from package test configs, and make those declarations available consistently to every package test project.
- Centralise other common test settings and file inclusion rules. Consider `${configDir}` for package-relative source and test paths. Ensure package-specific `include` overrides do not accidentally drop the shared setup declarations; TypeScript replaces inherited `include` arrays rather than merging them.
- Audit whether test projects need `composite`; disable it if build scripts, project references, and CI do not require it.
- Preserve useful incremental caching with separate test build metadata.
- Verify production builds, test type-checking, Vitest type tests, and unit suites. Confirm that package test projects inherit common compiler options from the root config without duplicating them. Confirm that custom matcher assertions type-check in every utility without a per-package `setupEnv.ts` include or import.
- Compare clean and incremental type-check timings; performance improvements should be measured rather than assumed.

Proposed file layout:

```text
/
├── tsconfig.json                  # Common compiler settings
├── tsconfig.cjs.json              # CommonJS build settings
├── tsconfig.test.json             # Shared test-only configuration
│
└── packages/
    ├── logger/
    │   ├── tsconfig.json          # Production build configuration
    │   └── tests/
    │       └── tsconfig.json      # Extends root tsconfig.test.json
    │
    ├── metrics/
    │   ├── tsconfig.json
    │   └── tests/
    │       └── tsconfig.json      # Extends root tsconfig.test.json
    │
    └── …other utilities…
```

Configuration inheritance:

```text
tsconfig.json                         # Common compiler options
  └── tsconfig.test.json               # Test-specific overrides
        └── packages/*/tests/tsconfig.json
                                      # Package-specific exceptions
```

Proposed root `tsconfig.test.json` (validate the exact compiler overrides during implementation):

```json
{
  "extends": "./tsconfig.json",
  "compilerOptions": {
    "noEmit": true,
    "composite": false,
    "declaration": false,
    "declarationMap": false
  },
  "files": [
    "./packages/testing/src/setupEnv.ts"
  ],
  "include": [
    "${configDir}/../src/**/*",
    "${configDir}/**/*"
  ]
}
```

- `files` includes the shared setup entry point, resolved relative to the root test config.
- `${configDir}` resolves package source and test paths relative to the consuming package's test config.
- With `composite: false`, TypeScript follows imports from `setupEnv.ts` without requiring each imported helper to be explicitly listed.
- A package can override `include` without losing the inherited shared `files` entry, provided it does not also override `files`.
- This config makes the setup and matcher declarations available for type-checking. Vitest still needs its runtime `setupFiles` registration to execute the setup.

Most package test configs would reduce to:

```json
{
  "extends": "../../../tsconfig.test.json"
}
```

Packages with additional requirements would retain their overrides.

Related discussion: #5693.

### Acknowledgment

### Future readers

svozza:
## Summary

Centralise test compiler settings and shared setup declarations so all 16 test projects inherit one configuration. Test type-checking now uses its own incremental caches instead of reusing production build caches.

### Changes

- Add `tsconfig.test.json`, extending the root compiler configuration with test-only overrides, shared `setupEnv.ts` inclusion, and `${configDir}` paths for package sources and tests.
- Disable composite/declaration output for test projects while retaining incremental checking in each workspace's `.tsbuildinfo/tests.json`.
- Reduce package test configs to the shared base while retaining Kafka's JSON-module override.
- Preserve the layer tests' existing compiler behaviour: `allowImportingTsExtensions` and `resolveJsonModule` were previously inherited from `layers/tsconfig.json`. They are now explicit in `layers/tests/tsconfig.json` because it extends the shared test base. Its existing `include` list, including `layers/package.json`, is unchanged.
- Remove the matcher type test's explicit setup import to verify inherited declarations, and document the configuration conventions.

Validation passed: clean ESM/CJS rebuilds, workspace and layer type-checks, all Vitest type suites, 3,718 unit tests with 100% coverage, and the layer-packaging unit test. Comparison of 48 resolved configurations confirmed no existing input files were dropped; compiler probes verified matcher typing and root-option inheritance across all 16 test projects. All 1,812 production output/cache hashes remained unchanged.

Type-check timings were effectively unchanged: 52.2 → 51.7 seconds cold and 41.1 → 41.1 seconds cached in a single before/after sample.

**Issue number:** closes #5694

---

### DIFF
--- a/CODING_STANDARDS.md
+++ b/CODING_STANDARDS.md
@@ -53,6 +53,12 @@ Run from the repo root with `-w <workspace>`, or from the package directory:
 - `npm run lint` to check; `npm run lint:fix` to auto-fix (review its changes).
 - `npm run build:tests` to type-check source and tests without emitting — CI compiles them separately from running them. `npm run build` additionally compiles the CommonJS target.
 
+Test configs in `packages/*/tests/tsconfig.json` extend the root `tsconfig.test.json`, which inherits common compiler options from `tsconfig.json`. Keep only package-specific exceptions in the package test configs.
+
+The shared test config includes `packages/testing/src/setupEnv.ts` through `files` so custom matcher types are available even when a package overrides `include`; `${configDir}` resolves source and test paths relative to each package's test config.
+
+Vitest runs tests; TypeScript only checks them. Test projects disable `composite` and declaration output, retain incremental checking, and store their cache in each workspace's `.tsbuildinfo/tests.json`, separately from production build caches. Vitest's runtime `setupFiles` registration remains in its own configuration.
+
 ## Unit tests
 
 Tests use `vitest` and live in each package's `tests/unit` directory. Run with `npm run test:unit -w packages/<name>` (or `npm run test:unit` from the package directory). Write unit tests only — end-to-end tests happen when the user asks for them.
--- a/layers/tests/tsconfig.json
+++ b/layers/tests/tsconfig.json
@@ -1,8 +1,8 @@
 {
-  "extends": "../tsconfig.json",
+  "extends": "../../tsconfig.test.json",
   "compilerOptions": {
-    "rootDir": "../",
-    "noEmit": true
+    "allowImportingTsExtensions": true,
+    "resolveJsonModule": true
   },
   "include": ["../src/**/*", "../package.json", "./**/*"]
 }
--- a/packages/batch/tests/tsconfig.json
+++ b/packages/batch/tests/tsconfig.json
@@ -1,8 +1,3 @@
 {
-  "extends": "../tsconfig.json",
-  "compilerOptions": {
-    "rootDir": "../",
-    "noEmit": true
-  },
-  "include": ["../src/**/*", "./**/*"]
+  "extends": "../../../tsconfig.test.json"
 }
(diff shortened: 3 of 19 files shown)


dreamorosi (approved):
Nice one!

PR #5717, opened 2026-09-14
fix(idempotency): complete and delete the record acquired at operation start

### LINKED ISSUE #5715: Bug: Idempotency completes or deletes the wrong record when the wrapped function mutates its input
### Expected Behavior

The record acquired at the start of an idempotent operation should be the record that is completed on success or deleted on failure. The key and validation hash should be fixed when the operation acquires its record, regardless of what the wrapped function does to its arguments afterwards.

### Current Behavior

`IdempotencyHandler` keeps a reference to the payload it hashes, and `BasePersistenceLayer` recomputes the idempotency key and validation hash from that payload on every call. If the wrapped function mutates its input, for example by normalising fields or adding a `processed` flag, `saveInProgress()` hashes the original payload and `saveSuccess()` or `deleteRecord()` hash the mutated one, so they target a different key.

Consequences with the default whole-event hash:

- On success, the completed result is written under a key no caller will ever look up. The original record stays `INPROGRESS`. A retry with the same event throws `IdempotencyAlreadyInProgressError` until the in-progress deadline passes, and once the record expires the side effect can run again.
- On failure, cleanup deletes the wrong key and the original in-progress record is left behind until it expires.
- With payload validation enabled, mutating a validated field changes the hash stored at completion, so an identical retry can fail validation.

This applies to `makeIdempotent`, the decorator, and the Middy middleware, since all three pass the event object by reference. Mutating the event in a handler is common enough that the framework should not depend on callers avoiding it.

### Code snippet

```typescript
import assert from 'node:assert/strict';
import {
  IdempotencyConfig,
  makeIdempotent,

  tableName: process.env.IDEMPOTENCY_TABLE_NAME ?? 'idempotency',
});

const processOrder = makeIdempotent(
  async (order: { id: string; amount: number; normalized?: boolean }) => {
    order.normalized = true; // mutates the object used for hashing
    return { processed: order.id };
  },
  {
    persistenceStore,
    config: new IdempotencyConfig({ expiresAfterSeconds: 3600 }),
  }
);

const first = await processOrder({ id: 'order-1', amount: 10 });

// Same event again should replay the stored result.
// Currently throws IdempotencyAlreadyInProgressError: the completed result was
// saved under the hash of { id, amount, normalized: true }, while the record
// for { id, amount } is still INPROGRESS.
const second = await processOrder({ id: 'order-1', amount: 10 });
assert.deepEqual(second, first);
```

### Steps to Reproduce

1. Wrap a function with `makeIdempotent` using the default configuration, so the whole event is hashed.
2. Inside the function, add or change a property on the event object.
3. Invoke it once, then inspect the table: one record is `INPROGRESS` under the hash of the original event, and a `COMPLETED` record exists under the hash of the mutated event.
4. Invoke it again with an identical event. It throws `IdempotencyAlreadyInProgressError` instead of returning the stored result.
5. As a control, have the function clone the event before mutating it. Both invocations return the same result and only one record exists.

### Possible Solution

Bind record identity to the operation rather than recomputing it on every persistence call. Two approaches:

- Compute the idempotency key and validation hash once in `IdempotencyHandler` when the record is acquired, and pass them to the subsequent `saveSuccess()` and `deleteRecord()` calls. This is the same direction as #5707, which pinned the key prefix per operation, and would let the prefix re-application added there be replaced by an explicit key.
- Alternatively, snapshot the payload before invoking the wrapped function, for example with `structuredClone`, and hash the snapshot throughout. Simpler, but it copies every event and does not cover the shared-prefix case.

The first is recommended.

2.35.0

22.x

### Packaging format used

npm

vishwakt:
### Changes

`IdempotencyHandler` keeps a reference to the payload it hashes, and `BasePersistenceLayer` recomputed the idempotency key and validation hash from that payload on every call. If the wrapped function mutates its input (normalising a field, adding a `processed` flag), `saveInProgress()` hashes the original payload while `saveSuccess()` and `deleteRecord()` hash the mutated one, so they target a different key. The in-progress record stays `INPROGRESS`, an identical retry throws `IdempotencyAlreadyInProgressError` until it expires and can then re-run the side effect, failure cleanup deletes the wrong key, and with payload validation enabled an identical retry can fail validation. This applies to `makeIdempotent`, the decorator, and the Middy middleware, since all three pass the event by reference.

This binds record identity to the operation rather than the payload, the first approach suggested in the issue:

- `BasePersistenceLayer.getRecordIdentity()` resolves the idempotency key and payload hash for a payload. `IdempotencyHandler` calls it before attempting acquisition, so the identity is available both when a record is acquired and when the function runs without acquiring one (durable replay on an existing `INPROGRESS` record).
- `saveInProgress()`, `saveSuccess()`, and `deleteRecord()` accept an optional trailing options object, `{ identity }`, and use the identity instead of hashing `data` again. The options shape leaves room for future settings without more positional arguments. Without it they behave exactly as before, so direct callers and custom persistence layers keep working. Public signatures are otherwise unchanged: `saveInProgress()` still returns `Promise<void>`, and existing overrides of all three methods still compile. An override of `saveInProgress()` that forwards only the original arguments still lands on the right record, because the payload has not been mutated at that point. An override of `saveSuccess(data, result)` or `deleteRecord(data)` that forwards only the original arguments discards the identity and can still hash a mutated payload, so the docs now recommend accepting and forwarding the whole options object unchanged in public-method overrides, and using the protected `_putRecord` / `_updateRecord` / `_deleteRecord` hooks for custom storage, which need no changes. `BasePersistenceLayerInterface` is updated to match, and the `IdempotencyRecordIdentity` and `PersistenceOperationOptions` types are exported.

Only the post-acquisition calls need the identity: `getRecord()` and `processExistingRecord()` run in the replay branch before the wrapped function executes, so the payload cannot have mutated yet.

Regression tests assert that completion (`_updateRecord`) and cleanup (`_deleteRecord`) carry the same `idempotencyKey` as acquisition (`_putRecord`) when the wrapped function mutates its input, that the stored `payloadHash` matches when a validated field is mutated, that the same holds during a durable replay where `_putRecord` rejects with `IdempotencyItemAlreadyExistsError`, and that a subclass overriding `saveInProgress()` as `Promise<void>` compiles and completes the right record. The mutation and replay tests fail on `main` with a different key or hash, and the override test fails to compile against the first revision of this PR. Unit tests for `getRecordIdentity()` and for the identity argument on each of the three methods are included. The idempotency unit suite passes (151 tests) with the 100% coverage gate.

### Note on #5708

This is the direction the issue describes as replacing the prefix re-application in #5708: with the fully resolved key captured before acquisition, `saveSuccess()` and `deleteRecord()` no longer depend on `idempotencyKeyPrefix` being re-applied before each call. Both PRs touch `IdempotencyHandler`, so whichever lands second needs a small rebase. Happy to rebase either way, or to fold the `configure()` prefix reset from #5708 in here if that is preferred.

**Issue number:** closes #5715

---


packages/idempotency/src/persistence/BasePersistenceLayer.ts
```diff
@@ -233,11 +238,12 @@ abstract class BasePersistenceLayer implements BasePersistenceLayerInterface {
    *
    * @param data - the data payload that will be hashed to create the hash portion of the idempotency key
    * @param remainingTimeInMillis - the remaining time left in the lambda execution context
+   * @returns the acquired record, whose key and payload hash identify the operation for {@link BasePersistenceLayer.saveSuccess | `saveSuccess()`} and {@link BasePersistenceLayer.deleteRecord | `deleteRecord()`}
    */
   public async saveInProgress(
     data: JSONValue,
     remainingTimeInMillis?: number
-  ): Promise<void> {
+  ): Promise<IdempotencyRecord> {
```
svozza: Changing `saveInProgress()` from `Promise<void>` to `Promise<IdempotencyRecord>` breaks existing subclasses, even those that simply delegate to the base implementation:

```typescript
abstract class CustomPersistenceLayer extends BasePersistenceLayer {
  public override async saveInProgress(
    data: JSONValue,
    remainingTimeInMillis?: number
  ): Promise<void> {
    await super.saveInProgress(data, remainingTimeInMillis);
  }
}
```

This is valid against the current API. With this PR, TypeScript reports **TS2416** because the override returns `Promise<void>`, which cannot satisfy the base class’s new `Promise<IdempotencyRecord>` contract.

Ignoring a returned value at a call site is compatible; overriding a method with its previous return type is not. Existing subclasses would need changes just to compile.

JavaScript or already-compiled subclasses can also discard the new return value, as this example does. The handler then receives `undefined` instead of an identity and falls back to hashing the mutated payload.

Let's preserve the existing public return contract and carry operation identity through an internal mechanism that continues to honor existing overrides.

packages/idempotency/src/IdempotencyHandler.ts
```diff
@@ -376,7 +385,7 @@ export class IdempotencyHandler<Func extends AnyFunction> {
       result: undefined,
     };
     try {
-      await this.#persistenceStore.saveInProgress(
+      this.#acquiredRecord = await this.#persistenceStore.saveInProgress(
```
svozza: The durable replay path still executes the wrapped function without a captured identity. When `saveInProgress()` rejects with `IdempotencyItemAlreadyExistsError`, the `isReplay` branch below returns immediately, leaving `#acquiredRecord` undefined. `saveSuccess()` and `deleteRecord()` then fall back to hashing `data` again after the function has mutated it.

I reproduced both cases using a pre-existing `INPROGRESS` record and a context with `durableExecutionMode: 'ReplayMode'`:

- Successful execution that adds a field to the payload writes the completed result under a different key; the original record remains `INPROGRESS`.
- Execution that mutates the payload and throws deletes the wrong key; the original record remains.

The ordinary execution-and-retry control passes, so this is specifically a gap in the replay-conflict path. Let's retain the operation identity before attempting acquisition so this path can use it too, and add regressions for replay completion and failure cleanup.

packages/idempotency/src/persistence/BasePersistenceLayer.ts
vishwakt: Good catch, I only checked compatibility at call sites and never considered overrides. Fixed it.

packages/idempotency/src/IdempotencyHandler.ts
vishwakt: The handler now resolves the identity before attempting acquisition, so the replay path uses it too. Added the two replay regressions (completion and failure cleanup with a mutated payload).

Both fail if the identity is captured only after a successful acquisition.

packages/idempotency/src/persistence/BasePersistenceLayer.ts
svozza: Thanks, both original findings are addressed: `saveInProgress()` retains its `Promise<void>` contract, and identity is captured before acquisition so durable replay is covered.

One clarification on the compatibility wording: existing overrides still compile, but overrides of `saveSuccess(data, result)` or `deleteRecord(data)` that forward only the original arguments discard the optional identity and can still hash a mutated payload. Let's narrow the description accordingly.

Let's also update the existing **“Bring your own persistent store”** section in `docs/features/idempotency.md` with this guidance:

- When overriding `saveInProgress`, `saveSuccess`, or `deleteRecord`, accept and forward the optional identity argument to `super`. This preserves the base class's record-identity guarantees and avoids depending on the caller leaving its payload unchanged.
- For custom storage implementations, prefer the protected `_getRecord`, `_putRecord`, `_updateRecord`, and `_deleteRecord` extension points. Subclasses using these hooks need no changes.

Forwarding the identity should be the recommended default for public-method overrides, even when the current application does not mutate its inputs.

vishwakt: You're right, I overgeneralized from the saveInProgress() case, where the payload cannot have mutated yet. 

Iv'e narrowed the description to say only that override stays safe when it drops the argument, and that saveSuccess() / deleteRecord() overrides need to forward the identity. Added a tip to the "Bring your own persistent store" section with both points: prefer the protected hooks, and when overriding the public methods accept and forward identity by default.

svozza: One final API suggestion: let's make the new argument an optional options object, with `identity` as a property, rather than passing the identity directly:

```typescript
await store.saveSuccess(data, result, { identity });
await store.deleteRecord(data, { identity });
await store.saveInProgress(data, remainingTime, { identity });
```

This lets us add optional settings later without adding more positional arguments. Since the new parameter hasn't shipped yet, now is a good time to establish that shape.

The subclassing guidance should then recommend accepting and forwarding the **whole options object unchanged** to `super`, so overrides preserve both the identity and any future options.

vishwakt: All three methods now take an optional trailing options object with identity as a property, and PersistenceOperationOptions is exported. The docs tip now says to forward the whole options object unchanged to super, and I added a test for an override that does exactly that.

packages/idempotency/tests/unit/makeIdempotent.test.ts
```diff
@@ -840,4 +856,199 @@ describe('Function: makeIdempotent', () => {
...
+        order.normalized = true;
+        return { processed: order.id };
+      },
+      { persistenceStore }
+    );
+    const durableContext = {
+      step: vi.fn(),
+      lambdaContext: context,
+      durableExecutionMode: 'ReplayMode',
+    } as unknown as DurableContext;
```
svozza: Can we avoid these sort of casts? There's another one below.

packages/idempotency/src/persistence/BasePersistenceLayer.ts
svozza: Nice, we're in a good place now. I had one small other comment but after that we're good to go!

packages/idempotency/tests/unit/makeIdempotent.test.ts
vishwakt: Removed it @svozza 

The three isReplay passthrough tests further up still have the same cast from before this PR. Happy to convert those too if you want it in here, or leave them for a separate cleanup.

svozza: Nice. Let's do them while we're in the area and then we're good to go.

vishwakt: updated @svozza

svozza (approved):

svozza:
Great work @vishwakt, this turned out to be trickier than expect!

PR #5725, opened 2026-09-17
fix(event-handler): honor Accept-Encoding q-values in compression

### LINKED ISSUE #5691: Bug: compress() ignores Accept-Encoding qvalues and skips compression when identity is listed
`compress()` should honour content-coding tokens: skip compression on `Accept-Encoding: gzip;q=0`, and compress on `Accept-Encoding: gzip, identity`. Both are currently inverted.

`shouldCompress` in `packages/event-handler/src/http/middleware/compress.ts` tests the raw header with `String.prototype.includes`, so `'gzip;q=0'.includes('gzip')` is `true` and the `q=0` exclusion is ignored, while `'gzip, identity'.includes('identity')` is `true` and suppresses compression the client explicitly accepted. Per RFC 9110 a `q=0` coding is not acceptable.

To reproduce, send either header to any route on a Router using `compress()` and read `content-encoding` on the response:

```ts
app.use(compress());
// Accept-Encoding: gzip;q=0        -> content-encoding: gzip   (should be absent)
// Accept-Encoding: gzip, identity  -> content-encoding: absent (should be gzip)
```

One fix is to parse the header into tokens and honour qvalues, reusing the existing encoding constants. Glad to open a PR if you want it done that way.

Version 2.35.0, npm, Node 22.x. Header parsing only, so runtime-independent.

arnabrahman:
## Summary

The HTTP compression middleware matched `Accept-Encoding` with substring checks, so `gzip;q=0` was still compressed, any mention of `identity` (even `identity;q=0`) disabled compression, and `GZIP` was not recognized. This PR parses the header as a list of codings with RFC 9110 quality values and compresses only when the configured encoding is at least as preferred as `identity`.

### Changes

- Parse encoding names as exact, case-insensitive tokens and support `x-gzip` as an alias for `gzip` ([RFC 9110 §8.4.1.3](https://www.rfc-editor.org/rfc/rfc9110.html#section-8.4.1.3)).
- Parse q-values as finite numeric preferences, clamped to `[0, 1]`; a missing `q` defaults to `1`.
- Treat empty or non-numeric values (e.g. `q=`, `q=abc`, `q=1foo`, `q=Infinity`) as `q=0`, so an unreadable preference never counts as consent to compress.
- If the header names the encoding explicitly, use its q-value instead of `*`, even when it's `q=0` (e.g. `gzip;q=0`, `*` is not compressed, `*;q=0`, gzip is).
- Compare the configured encoding with `identity`: use the higher q-value and prefer compression when their qualities tie (e.g. `gzip, identity`).
- Behavior changes for existing clients: `gzip;q=0` and unparseable q-values are no longer compressed; `gzip, identity`, `identity;q=0, gzip`, and `GZIP` are now compressed.
- Preserve existing behavior for absent and empty `Accept-Encoding` headers.
- Add regression coverage for casing, whitespace, aliases, malformed and boundary q-values, wildcard precedence, identity preference, and configured `deflate` behavior.
- Assert successful response status in table-driven negotiation tests so an internal error cannot satisfy a no-compression expectation.

<details>
<summary><strong>q-value parsing approach — maintainer input requested</strong></summary>

I was initially undecided between strict [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-quality-values) parsing and accepting a broader range of numeric values, so I compared how other HTTP stacks handle q-values:

| Implementation | q-value parsing | Invalid or out-of-range q-value |
| --- | --- | --- |
| [Hono](https://github.com/honojs/hono/blob/main/src/utils/accept.ts) | `Number()`, clamped to `[0, 1]`, `q`/`Q` | Unparseable → `1` (literal `NaN` → `0`) |
| [Express `compression`](https://github.com/expressjs/compression/blob/master/index.js) ([`negotiator@0.6.4`](https://github.com/jshttp/negotiator/blob/0.6.4/lib/encoding.js)) | `parseFloat()`, not clamped, lowercase `q` only | `NaN` → coding dropped; `q=5` stays `5`; `0.5abc` → `0.5` |
| [nginx](https://github.com/nginx/nginx/blob/master/src/http/ngx_http_core_module.c) (`ngx_http_gzip_quantity`) | Strict: `0` or `1`, optional `.` and up to 3 digits, `q`/`Q` | Treated as `q=0`, so no gzip |
| [ASP.NET Core](https://github.com/dotnet/aspnetcore/blob/main/src/Http/Headers/src/HeaderUtilities.cs) (`TryParseQualityDouble`) | Strict RFC 9110 grammar, `q`/`Q`, whitespace allowed around `=` | Coding dropped from the list |
| [tower-http](https://github.com/tower-rs/tower-http/blob/main/tower-http/src/content_encoding.rs) (`QValue::parse`) | Strict RFC 9110 grammar, `q`/`Q`, no whitespace around `=` | Coding dropped from the list |
| **This PR** | `Number()`, clamped to `[0, 1]`, `q`/`Q`, whitespace allowed | Unparseable → `0` |

The comparison shows a split: the JavaScript frameworks are lenient, while nginx, ASP.NET Core, and tower-http are strict. I chose the lenient approach because this is a TypeScript/JavaScript library and it accepts headers that users may already see work with Express or Hono, including `.5`, `1.0000`, and `q = 0.5`. Unreadable values still behave as `q=0`, so they do not enable compression.

This is a compatibility choice rather than a strong preference. Strict RFC 9110 parsing would also be reasonable and would be a small, localized change to `getQuality` and its tests. I’d appreciate input on which approach the project should prefer.
</details>

**Issue number:** closes #5691

---


svozza:
The fix looks good. I ran the event-handler unit suite with coverage, type-checking, and lint; all passed.

I’m comfortable with lenient decimal parsing, but please reject non-decimal forms such as `q=0x1`. We can retain support for `.5` and extra decimal places without accepting JavaScript’s other numeric formats. Please add a regression test showing that a hexadecimal q-value disables compression.

Could you also add these two cases to the negotiation table? Both should compress:

- `*;q=0, gzip` — an explicit encoding overrides the wildcard exclusion.
- `identity;q=0, gzip` — excluding identity still allows gzip.

Both are mentioned in the PR description, so direct regression coverage would be useful.

svozza:
LGTM!

svozza (approved):

